PR #4, opened 2018-12-30
batch KG uses gpytorch interface for fantasy models

Balandat:


LiuYang-1997:
please，what is fantasy model? what is fantasy points?

Balandat:
Let's have this discussion here: https://github.com/pytorch/botorch/discussions/963

PR #151, opened 2019-05-04
Use explicit torch.autograd.grad insted of .backward

Balandat:
Being explicit in computing the gradients that we want seems to be the right thing to do.

### DIFF
--- a/botorch/acquisition/objective.py
+++ b/botorch/acquisition/objective.py
@@ -16,14 +16,7 @@ from torch.nn import Module
 
 
 class MCAcquisitionObjective(Module, ABC):
-    r"""Abstract base class for MC-based objectives.
-
-    Example:
-        >>> # This method is usually not called directly, but via the objective's
-        >>> # `__call__` method:
-        >>> samples = sampler(posterior)
-        >>> outcome = mc_obj(samples)
-    """
+    r"""Abstract base class for MC-based objectives."""
 
     def __init__(self) -> None:
         super().__init__()
@@ -39,6 +32,13 @@ class MCAcquisitionObjective(Module, ABC):
         Returns:
             Tensor: A `sample_shape x batch_shape x q`-dim Tensor of objective
             values (assuming maximization).
+
+        This method is usually not called directly, but via the objectives
+
+        Example:
+            >>> # `__call__` method:
+            >>> samples = sampler(posterior)
+            >>> outcome = mc_obj(samples)
         """
         pass  # pragma: no cover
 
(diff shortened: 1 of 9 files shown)


eytan:
Any context on this change? It does make the code more complicated / less idiomatic.

Balandat:
Mostly b/c it gets rid of some gotchas. 

Using `backward` means you always have to be aware where you're calling it, what you're calling it on, whether you've properly required gradients for the variables that you're differentiating w.r.t. You also need to remember to zero out your gradients if you reuse the tensors. Using `grad` is less gotcha-esque, and also more functional which makes it easier to understand what's going on b/c all the relevant code is in one place (rather than distributed into a `backward` call, an accessing of the `.grad` property of the relevant tensors, and zeroing out gradients). 

gotcha 1:
```
X = torch.rand(3)
f(X).backward()
X.grad # None
```
```
X = torch.rand(3)
torch.autograd.grad(f(X), X)  # raises RuntimeError: One of the differentiated Tensors does not require grad
```

gotcha2:
```
X = torch.rand(3, requires_grad=True)

Y = X ** 2
Y.sum().backward()
X.grad  # tensor([1.4779, 0.6099, 1.1670])

Y = X ** 2
Y.sum().backward()
X.grad # tensor([2.9558, 1.2199, 2.3339])
```

```
X = torch.rand(3, requires_grad=True)

Y = X ** 2
torch.autograd.grad(Y.sum(), X)[0]  # tensor([1.4779, 0.6099, 1.1670])

Y = X ** 2
torch.autograd.grad(Y.sum(), X)[0]  # tensor([1.4779, 0.6099, 1.1670])
```

botorch/acquisition/objective.py
```diff
@@ -63,7 +63,7 @@ class LinearMCObjective(MCAcquisitionObjective):
     `mc_obj(samples) = sum_{i} weights[i] * samples[..., i]`
 
     Example:
-        >>> # example for a two outcomes
+        Example for a model with two outcomes:
```
sdaulton: Sphinx won't correctly parse the following lines as a code block. Here are few options that will be correctly parsed: add a blank line after L66 before the code block, make this line a comment (starting with `>>> #`), or indent the code block to the next level.

botorch/gen.py
```diff
@@ -101,9 +101,8 @@ def f(x):
         )
         X_fix = fix_features(X=X, fixed_features=fixed_features)
         loss = -acquisition_function(X_fix).sum()
-        loss.backward()
+        gradf = _arrayify(torch.autograd.grad(loss, X)[0].contiguous().view(-1))
```
sdaulton: Mind adding a comment for posterity, since `backward` is more common.

PR #156, opened 2019-05-09
Changing baseUrl to '/' allows building locally

ericnakagawa:
The previous setting of the full path in the baseUrl resulted in an error when running the site locally. This fix will allow it to run locally without issue.

## Motivation

I wanted to run the site locally but ran into an error.

Yes

## Test Plan

Run locally... visit localhost:3000, note error.
Apply patch, run locally, visit localhost:3000 and see the page load correctly.

### DIFF
--- a/website/siteConfig.js
+++ b/website/siteConfig.js
@@ -11,7 +11,7 @@
 // site configuration options.
 
 // Define this so it can be easily modified in scripts (to host elsewhere)
-const baseUrl = 'https://botorch.org/';
+const baseUrl = '/';
 
 // List of projects/orgs using your project for the users page.
 const users = [];


Balandat:
so the added complication here is that we auto-deploy the website from CircleCI so the baseUrl needs to be set to `https://botorch.org/` for the website to work. We could hack our way around this by sed-ing the baseUrl in the build script, but that seems less than ideal. Can't Docusaurus be smart about this to always use `/` when running the site locally?

JoelMarcey:
I created the same PR in #158. I think we should make the change either from this PR or the #158 PR because it is definitely correct. And we can look into the CircleCI issue you are having. We also deploy Docusaurus itself via CircleCI and we have the `baseUrl` set to `/` with no issues.

Balandat:
Closing this out in favor of #158

PR #159, opened 2019-05-16
Fix link to Ax docs

pietern:


Balandat:
This change doesn't seem necessary, the link works fine under both Chrome and Firefox. Closing this.

PR #183, opened 2019-06-25
Fit models in SumMarginalLogLikelihood sequentially by default

Balandat:
Currently, `fit_gpytorch_model` solves a single joint optimization problem to fit ModelListGP (and, more generally, any ModuleList model that has a SumMarginalLogLikelihood).

If there are only a few models then this is just fine. If there are a lot of models, each with a lot of parameters, then the dimension of the optimization problem can become quite large, and the optimizer may have a hard time optimizing the joint problem (as it's not exploiting the indenpendence structure).

This PR changes the default behavior to fit the models sequentially, solving multiple simpler optimization problems instead a single harder one. This can be overriden by passing in `sequential=False` to `fit_gpytorch_model`.

Test Plan:
Unit tests

### DIFF
--- a/botorch/fit.py
+++ b/botorch/fit.py
@@ -9,6 +9,7 @@ Utilities for model fitting.
 from typing import Any, Callable
 
 from gpytorch.mlls.marginal_log_likelihood import MarginalLogLikelihood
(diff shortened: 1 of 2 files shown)


Balandat:
cc @jacobrgardner

Issue #203, opened 2019-07-09
Change sampler behavior if resample=False for different dtype/device 


Balandat:
Currently, the behavior of the samplers is to re-draw base samples when the `device`/`dtype` of the posterior disagrees with that of the base samples, even if `resample=False`: https://github.com/pytorch/botorch/blob/master/botorch/sampling/samplers.py#L202-L208

This seems somewhat inconsistent with the meaning of resample. Instead, we should just move the `base_samples` tensor to the `device`/`dtype` of the posterior.

Balandat:
Addressed by #204

PR #204, opened 2019-07-09
Changes to resampling behavior in MCSamplers

Balandat:
Makes the following changes in case `resample=False`:
1. if `collapse_batch_dims=True` and the requested shape is different from the shape of the base samples, only resample if the last (q and o) dimensions are different, otherwise broadcast the base samples to the requested batch shape (will just be a different number of ones)
2. if `dtype` or `device` of the posterior are different from that of the base samples don't resample if not triggered otherwise - instead automatically move the base samples to the correct device/dtype if apppropriate


Balandat:
This addresses #203

PR #210, opened 2019-07-18
Scalarization for posteriors, ScalarizedObjective

Balandat:
In some cases we may be interested in some affine function of the modeled outputs. By linearity, the posterior of that function will be a univariate normal (assuming `q=1`). 

This adds some functionality for achieving this for all analytic acquisition functions.

### DIFF
--- a/botorch/posteriors/gpytorch.py
+++ b/botorch/posteriors/gpytorch.py
@@ -13,6 +13,7 @@ import torch
 from gpytorch.distributions import MultitaskMultivariateNormal, MultivariateNormal
 from torch import Tensor
 
+from ..exceptions.errors import UnsupportedError
 from .posterior import Posterior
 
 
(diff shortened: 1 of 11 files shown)


botorch/posteriors/gpytorch.py
```diff
@@ -99,3 +100,41 @@ def variance(self) -> Tensor:
         if not self._is_mt:
             variance = variance.unsqueeze(-1)
         return variance
+
+
+def affine_transform(
+    posterior: GPyTorchPosterior, weights: Tensor, offset: float = 0.0
```
eytan: When would we want to use a non-zero offset?

eytan:
@Balandat, AFAIK these types of objectives are almost always referred to as 'scalarizations' or 'weighted sums' that are 'linear combinations' of objectives.  See e.g., https://arxiv.org/pdf/1402.0590.pdf.   Perhaps we can consider naming this `ScalarizedAcquisitionObjective` or `LinearAcquisitionObjective` instead?

Balandat:
@eytan The reason I call if affine here is because it is, in fact, not linear (b/c of the offset term). `ScalarizedAcquisitionObjective` makes a lot of sense, I can change the name to that.

botorch/posteriors/gpytorch.py
Balandat: We could use this for KG to pass in the (negative) current best, which could make the KG implementation somewhat more elegant.

Balandat: @danielrjiang @VilockLi

PR #222, opened 2019-08-01
Use bool dtype for masks

Balandat:
Summary: This is the new default behavior for masks in pytorch: https://github.com/pytorch/pytorch/pull/21113

Differential Revision: D16610024


Balandat:
ah ok that won't work though... d'oh

PR #228, opened 2019-08-09
Version 0.1.3

Balandat:
Add changelog and bump version to 0.1.3.


Balandat:
This will require a new gpytorch release & putting that in as a requirement

PR #238, opened 2019-08-14
require output dim in botorch models

sdaulton:
Summary: see title. Botorch internally squeezes the output dim as necessary before interacting with gpytorch.ExactGP code.

Differential Revision: D16633960


sdaulton:
FYI @jacobrgardner and @gpleiss, this PR makes it so that botorch models require that training (and fantasy) targets have an explicit output dimension. This convention is particularly helpful, when using `get_fantasy_model`, where `inputs` is `batch_shape x n x d` and `targets` is `batch_shape' x n x (m)`, where `m` is the number of outputs and `m` is an optional dimension. In this case, the `inputs` and `targets` may have a different number of batch dimensions, making it hard to infer whether `targets` has an explicit output dimension or not. Inferring the output dimension of `targets` is important because if `m` is an explicit dim and `m=1`, then the `m` dimension must be squeezed before calling `ExactGP.get_fantasy_model`. Furthermore when using botorch's BatchedMultiOutputModels, the `targets` needs to be permuted to make `m` a batch dimension  if `m>1`. I just wanted to flag this as it may be something you want to consider upstreaming in the future.

PR #324, opened 2019-11-26
Add subset_output functionality to (some) models

Balandat:
Summary:
In some cases we want to be able to subset models along the output dimension.
For instance, if we fit a multi-output model with a number of metrics, we may want to optimize an acquisition function with an objective that only involves a subset of the
outputs. By subsetting the model prior to that, we can save a lot of compute.

This diff adds a `subset_output` function to the model API. Calling this on a model with a list of indices will return a new model object that is restricted to the desired outputs.

For some models (e.g `AffineDeterministicModel` or `ModelListGP`) the implementation is trivial. For others it's a little more involved but doable.The main challenge is with things like passing in generic covariance modules - we really don't have any way of knowing what dimensions of the respective buffers and parameters we need to subset / rescale in this case.

Differential Revision: D18668985

### DIFF
--- a/botorch/models/deterministic.py
+++ b/botorch/models/deterministic.py
@@ -73,6 +73,21 @@ class GenericDeterministicModel(DeterministicModel):
         self._f = f
         self._num_outputs = num_outputs
 
+    def subset_output(self, idcs: List[int]) -> "GenericDeterministicModel":
+        r"""Subset the model along the output dimension.
+
+        Args:
+            idcs: The output indices to subset the model to.
+
+        Returns:
+            The current model, subset to the specified output indices.
+        """
+
+        def f_subset(X: Tensor) -> Tensor:
+            return self._f(X)[..., idcs]
+
+        return self.__class__(f=f_subset)
+
     def forward(self, X: Tensor) -> Tensor:
         r"""Compute the (deterministic) model output at X.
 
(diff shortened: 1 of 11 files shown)


jacobrgardner:
@Balandat Just for your visibility, I have plans to introduce a `__getitem__` method to `ExactGP` and `ApproximateGP` with functionality similar to what I did for `Kernel` and was recently added to `MultivariateNormal`, so that calls like `model[[4:7]]` or `model[[0, 5, 7]]` return subsets of batch GPs. It'll be pretty experimental at first though, since obviously some types of modules (like NN feature extractors) won't support it.

Balandat:
Thanks for the heads up, I had seem part of that discussion.

> so that calls like model[4:7] or model[[0, 5, 7]] return subsets of batch GPs

Is this going to return a fully separate (i.e. deep-copied subset) as my PR does here? This is quite essential to the usage I envision here b/c the resulting model should be transparently usable for anything downstream. The typical use case here is that a full model is fit to all outcomes of interest (typically in Ax), but then for acquisition function optimization we want to subset only to those outcomes present in an objective in order to save a ton of unnecessary compute overhead. So we create an ephemeral sub-set model just for computing / optimizing the acquisition function.

Either way, this may simplify some of my implementation here, so I'm looking forward to seeing some details.

jacobrgardner:
Yeah the plan would be to return a deep copied object. Currently this is how `Kernel.__getitem__` functions: it deepcopies the current kernel, subindexes any hyperparameters with the relevant index on the new kernel, and then subindexes any sub kernels recursively.

PR #357, opened 2020-01-21
Multi-Step lookahead acquisition function

Balandat:
Summary: Add generic multi-step lookahead acquisition function. This allows using arbitrary acquisition functions as the value function in a KG-like multi-step look-ahead scheme. This is an experimental feature.

Differential Revision: D19368587

### DIFF
--- a/botorch/acquisition/multi_step_lookahead.py
+++ b/botorch/acquisition/multi_step_lookahead.py
@@ -1,0 +1,241 @@
+#!/usr/bin/env python3
+# Copyright (c) Facebook, Inc. and its affiliates.
+#
+# This source code is licensed under the MIT license found in the
+# LICENSE file in the root directory of this source tree.
+
+import warnings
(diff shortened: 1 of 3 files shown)


shalijiang:
does this include multi-step EI? I guess we would have to implement a wrapper of `ExpectedImprovement` to extract `best_f` from fantasy_model as an argument of the EI value function?

Balandat:
Yeah using a wrapper is a possibility. It would be interesting to see how much different in practice the result is compared to one where `best_f` is just taken as best observed before fantasizing. Alternatively, you could use `qNoisyExpectedImprovement` which should already do the right thing.

shalijiang:
I wonder what would be the best initialization strategy for optimizing it

Balandat:
Good question. A method similar to the KG initialization strategy that starts mixes the standard EI solution and with some random points is probably a decent start.

PR #392, opened 2020-03-12
Better stopping criteria for stochastic optimization

Balandat:
The current `ConvergenceCriterion` assumes that the optimization is using a descent method. That's not true for stochastic optimization, so using this in `gen_candidates_torch` will very often prematurely terminate optimization. This diff adds a basic version of an exponential moving average smoothing of the losses to avoid this.

Since convergence is generally kind of fuzzy in stochastic optimization, this also renames `ConvergenceCriterion` to `StoppingCriterion`. It introduces an abstract base class, and moves this functionality to the `botorch.optim.stopping` submodule.

### DIFF
--- a/botorch/gen.py
+++ b/botorch/gen.py
@@ -23,12 +23,8 @@ from .optim.parameter_constraints import (
     make_scipy_bounds,
     make_scipy_linear_constraints,
 )
-from .optim.utils import (
-    ConvergenceCriterion,
-    _filter_kwargs,
-    columnwise_clamp,
-    fix_features,
-)
+from .optim.stopping import ExpMAStoppingCriterion
+from .optim.utils import _filter_kwargs, columnwise_clamp, fix_features
 
 
 def gen_candidates_scipy(
@@ -199,11 +195,11 @@ def gen_candidates_torch(
     param_trajectory: Dict[str, List[Tensor]] = {"candidates": []}
     loss_trajectory: List[float] = []
     i = 0
-    converged = False
-    convergence_criterion = ConvergenceCriterion(
-        **_filter_kwargs(ConvergenceCriterion, **options)
+    stop = False
+    stopping_criterion = ExpMAStoppingCriterion(
+        **_filter_kwargs(ExpMAStoppingCriterion, **options)
     )
-    while not converged:
+    while not stop:
         i += 1
         loss = -acquisition_function(candidates).sum()
         if verbose:
@@ -221,7 +217,7 @@ def gen_candidates_torch(
             clamped_candidates, lower_bounds, upper_bounds
         )
         candidates = fix_features(clamped_candidates, fixed_features)
-        converged = convergence_criterion.evaluate(fvals=loss.detach())
+        stop = stopping_criterion.evaluate(fvals=loss.detach())
     clamped_candidates = columnwise_clamp(
         X=candidates, lower=lower_bounds, upper=upper_bounds, raise_on_violation=True
     )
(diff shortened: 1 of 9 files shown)


cisprague:
How does this work when optimising over mini-batches?

Balandat:
> How does this work when optimising over mini-batches?

You mean fitting the model across mini-batches of data? The basic fitting tools we have in BoTorch aren't really designed to do mini-batch optimization; this is b/c the default BoTorch models are typically fit on all data. In order to do this for minibatches of data, presumably you'll want to use something like variational GPs? See the [gpytorch tutorial](https://github.com/cornellius-gp/gpytorch/blob/master/examples/04_Variational_and_Approximate_GPs/SVGP_Regression_CUDA.ipynb) on how to do this. 

In order to use the resulting model for BO, you'll need to give it the `Model` API. This should be as simple as subclassing your model from both the botorch `GPyTorchModel` and the gpytorch `ApproximateGP` (this is take from cell [4] from the tutorial linked above): 

```
from botorch.models.gpytorch import GPyTorchModel
from gpytorch.models import ApproximateGP

class GPModel(ApproximateGP, GPyTorchModel):
    ...
```

cisprague:
> > How does this work when optimising over mini-batches?
> 
> You mean fitting the model across mini-batches of data? The basic fitting tools we have in BoTorch aren't really designed to do mini-batch optimization; this is b/c the default BoTorch models are typically fit on all data. In order to do this for minibatches of data, presumably you'll want to use something like variational GPs? See the [gpytorch tutorial](https://github.com/cornellius-gp/gpytorch/blob/master/examples/04_Variational_and_Approximate_GPs/SVGP_Regression_CUDA.ipynb) on how to do this.
> 
> In order to use the resulting model for BO, you'll need to give it the `Model` API. This should be as simple as subclassing your model from both the botorch `GPyTorchModel` and the gpytorch `ApproximateGP` (this is take from cell [4] from the tutorial linked above):
> 
> ```
> from botorch.models.gpytorch import GPyTorchModel
> from gpytorch.models import ApproximateGP
> 
> class GPModel(ApproximateGP, GPyTorchModel):
>     ...
> ```

I'm doing as you suggested above. But, in my case, I am randomly sampling a batch of training data at every optimisation iteration (effectively mini-batch descent). I'm using `ExpMAStoppingCriterion` from BoTorch and it's working sufficiently. But, it would be cool if there was a way to stop when some notion overfitting is detected. Is there something I could do to achieve this? Thanks!

Balandat:
> it would be cool if there was a way to stop when some notion overfitting is detected. Is there something I could do to achieve this? Thanks!

I guess it depends what your notion of overfitting is. GPs should be reasonably robust when using the MLL. If you want something more explicit, you could set aside some holdout set and monitor the prediction error / coverage of the posterior covariances on that - you'd expect this to decrease as the model fit improves, and then to degrade as you overfit.

Is this the kind of thing you were thinking of? I'd have to think about how to do that within `fit_gpytorch_torch`.

cisprague:
> I guess it depends what your notion of overfitting is. GPs should be reasonably robust when using the MLL. If you want something more explicit, you could set aside some holdout set and monitor the prediction error / coverage of the posterior covariances on that - you'd expect this to decrease as the model fit improves, and then to degrade as you overfit.
> Is this the kind of thing you were thinking of? I'd have to think about how to do that within `fit_gpytorch_torch`.

Yup, exactly, just like regular ML with NNs. But, is there some other more principled Bayesian way to assess this notion, since we're working with GPs?

Ultimately, I just want to make sure that I'm getting accurate estimates of the GP's posterior after each sampling, whilst simultaneously ensuring that I'm not overfitting — these seem like competing objectives.
I get the feeling that overfitting early on (when the amount of samples is small) will disparage the quality of the GP posterior when future samples are added, e.g. if a local-minima is encountered early on.

eytan:
Overfitting with a vanilla exact GP won’t be an issue if you are optimizing the MLL. This will be an issue with DKL and I am pretty sure could be an issue for the approximate methods.

If you are estimating an exact GP though, why not use the default second-order methods? They very reliably obtain good model fits without any tuning of optimizer HPs. 

E

PR #410, opened 2020-03-25
Multi-step look-ahead acquisition with generic stage costs

Balandat:
Summary:
This modifies the `qMultiStepLookahead` acquisition function to allow for general "stage" or "running" value functions. This is WIP and there is a lot more to do, including

- supporting batched inputs to acquisition functions, e.g. batched values `best_f` for qEI.
- proper testing w/ full test coverage
- empirical evaluation

Differential Revision: D20657699

### DIFF
--- a/botorch/acquisition/multi_step_lookahead.py
+++ b/botorch/acquisition/multi_step_lookahead.py
@@ -1,0 +1,536 @@
+#!/usr/bin/env python3
+# Copyright (c) Facebook, Inc. and its affiliates.
+#
+# This source code is licensed under the MIT license found in the
+# LICENSE file in the root directory of this source tree.
+
+from __future__ import annotations
+
+import math
+import warnings
+from typing import Any, Callable, Dict, List, Optional, Tuple, Type
+
+import numpy as np
+import torch
+from torch import Size, Tensor
+from torch.distributions import Beta
+from torch.nn import ModuleList
+
+from ..exceptions.errors import UnsupportedError
+from ..exceptions.warnings import BotorchWarning
+from ..models.model import Model
+from ..optim.initializers import initialize_q_batch
+from ..sampling.samplers import MCSampler, SobolQMCNormalSampler
+from ..utils.transforms import match_batch_shape, t_batch_mode_transform, unnormalize
+from .acquisition import AcquisitionFunction, OneShotAcquisitionFunction
+from .analytic import AnalyticAcquisitionFunction, PosteriorMean
+from .monte_carlo import MCAcquisitionFunction
+from .objective import AcquisitionObjective, MCAcquisitionObjective, ScalarizedObjective
+
+
+TAcqfArgConstructor = Callable[[Model, Tensor], Dict[str, Any]]
+
+
+def make_best_f(model: Model, X: Tensor) -> Dict[str, Any]:
+    r"""Extract the best observed training input from the model."""
+    return {"best_f": model.train_targets.max(dim=-1).values}
+
+
+class qMultiStepLookahead(MCAcquisitionFunction, OneShotAcquisitionFunction):
+    r"""MC-based batch Multi-Step Look-Ahead (one-shot optimization)."""
+
+    def __init__(
+        self,
+        model: Model,
+        batch_sizes: List[int],
+        num_fantasies: Optional[List[int]] = None,
+        samplers: Optional[List[MCSampler]] = None,
+        valfunc_cls: Optional[List[Optional[Type[AcquisitionFunction]]]] = None,
+        valfunc_argfacs: Optional[List[Optional[TAcqfArgConstructor]]] = None,
+        objective: Optional[AcquisitionObjective] = None,
+        inner_mc_samples: Optional[List[int]] = None,
+        X_pending: Optional[Tensor] = None,
+    ) -> None:
+        r"""q-Multi-Step Look-Ahead (one-shot optimization).
+
+        Performs a `k`-step lookahead by means of repeated fantasizing.
+
+        Allows to specify the stage value functions by passing the respective class
+        objects via the `valfunc_cls` list. Optionally, `valfunc_argfacs` takes a list
+        of callables that generate additional kwargs for these constructors. By default,
+        `valfunc_cls` will be chosen as `[None, ..., None, PosteriorMean]`, which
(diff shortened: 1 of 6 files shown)


danielrjiang:
Copying some discussion here for visibility:

"This is awesome, Max!
Some points for thought below. Two scenarios that might be useful are:
(1) we want to sum up values that only depend on the x's at each stage (the case of a known cost function),
(2) we want to sum up a function of the fantasized values (costs based on the outcomes of each stage?)
For (1), should we allow for something simpler than acquisition function (any differentiable function)? I think we could just write custom acquisition functions to handle this as well.
For (2), we could grab the values from the model, but perhaps a bit messy."


"Thanks Daniel, these are great points.
(1) That makes a lot of sense. If we wanted to stay within our abstractions, we could use DeterministicModel just as we do for deterministic cost functions. That would also give us the flexibility to apply generic objectives to multiple fixed outcomes. This is likely overkill for many cases, but it would be the most generic way of setting this up I believe. The downside would be that this requires passing in ever more things into the constructor. Alternatively we could do as you suggested and just write a trivial deterministic AcquisitionFunction that simply acts on X and ignores the model.
(2) Yeah that makes sense, too. To make this work we'll likely have to just explicitly sample and then condiiton_on_observations. This would work, but it's still kind of messy and digs deeper into the API that I'd like (we are introducing additional assumptions beyond the simple fantasize API. Not sure if there is a better way though."

botorch/acquisition/multi_step_lookahead.py
```diff
@@ -213,30 +249,46 @@ def _step(
...
-        obj = terminal_value_func(X=Xs[0])
-        # shape of obj is (inner_mc_samples) x f_k x ... x f_1 x batch_shape
+    # compute stage value
+    StageValFuncClass = stage_valfunc_cls[0]
+    if StageValFuncClass is not None:
+        common_kwargs: Dict[str, Any] = {"model": model, "objective": objective}
+        if issubclass(StageValFuncClass, MCAcquisitionFunction):
+            common_kwargs["sampler"] = inner_samplers[0]
+        arg_fac = stage_valfunc_argfacs[0]
+        kwargs = arg_fac(model=model, X=Xs[0]) if arg_fac is not None else {}
```
shalijiang: so in case of ExpectedImprovement, we need to pass in best_f. Seems currently it doesn't support that? But we could easily write a wrapper to assign best_f = train_input.max()

Balandat: So the idea is that `arg_fac` would provide the required kwargs, depending on the acquisition function type. For EI we would use `make_best_f` up at the top of the module. Then `best_f` would be passed into the constructor via `**kwargs`.

shalijiang: ah, missed that part, that's nice!

botorch/acquisition/multi_step_lookahead.py
```diff
@@ -33,20 +40,23 @@ def __init__(
         model: Model,
         batch_sizes: List[int],
         num_fantasies: List[int],
-        value_function_cls: Type[AcquisitionFunction] = PosteriorMean,
-        value_function_kwargs: Optional[Dict[str, Any]] = None,
+        stage_valfunc_cls: Optional[List[Optional[Type[AcquisitionFunction]]]] = None,
+        stage_valfunc_argfacs: Optional[List[Optional[TAcqfArgConstructor]]] = None,
         objective: Optional[AcquisitionObjective] = None,
-        inner_sampler: Optional[MCSampler] = None,
+        num_inner_mc_samples: Optional[List[int]] = None,
```
shalijiang: why not pass in a list of samplers to allow the flexibility of using different samplers?

Balandat: yeah that's a possibility. I was trying to keep the interface somewhat clean to start with, but we could also support both

Balandat:
Made some further updates. Main change is that now we don't have to pre-specify the q-batch size of the initial step, but can use different ones on the same acquisition function object, which is consistent with the behavior of all other q-batch acquisition functions.

Still need to add support for Gauss-Hermite "sampler" and batch-support for `best_f` for EI. 

Here is a demo notebook for the current state:  
[demo_stage_valfunc_qMultiStep.ipynb.txt](https://github.com/pytorch/botorch/files/4418424/demo_stage_valfunc_qMultiStep.ipynb.txt)

botorch/acquisition/multi_step_lookahead.py
```diff
@@ -213,44 +247,284 @@ def _step(
...
+    kwargs = arg_fac(model=model, X=X) if arg_fac is not None else {}
+    stage_val_func = valfunc_cls(**common_kwargs, **kwargs)
+    stage_val = stage_val_func(X=X)
+    # shape of stage_val is (inner_mc_samples) x f_k x ... x f_1 x batch_shape
+    if sample_weights.numel() > 0:
+        # if there are sample weights, we need to use them
+        # sample_weights must be of shape f_k x ... x f_1 x batch_shape
+        stage_val = stage_val * sample_weights
+    # we average across all dimensions except for the batch dimension
+    stage_val = stage_val.view(-1, stage_val.size(-1)).mean(dim=0)
```
shalijiang, line 365: in case of GH sampler, .mean() should be .sum()?  \int f(x) w(x) dx = w_1 f(x_1) + w_2 f(x_2) + ...

botorch/sampling/samplers.py
```diff
@@ -249,3 +251,84 @@ def _construct_base_samples(self, posterior: Posterior, shape: torch.Size) -> No
...
+                    "Only output_dim=1 is supported by GaussHermiteSampler."
+                    f"Got output_dim={output_dim}."
+                )
+            n = shape[:-2].numel()
+            base_samples, base_weights = np.polynomial.hermite.hermgauss(n)
+            tkwargs = {"device": posterior.device, "dtype": posterior.dtype}
+            base_samples = torch.from_numpy(base_samples).to(**tkwargs)
+            base_samples = (base_samples * math.sqrt(2.0)).view(n, output_dim)
+            base_weights = torch.from_numpy(base_weights).to(**tkwargs)
+            base_samples = (base_samples / math.sqrt(math.pi)).view(-1, 1)
```
shalijiang, line 323: should be base_weights

PR #419, opened 2020-04-10
Fix sobol maxdim limitation in prune_baseline

Balandat:
Fixes https://github.com/facebook/Ax/issues/291


mrcslws:
Thanks for the fast turnaround! I can confirm, this unblocked me.

PR #498, opened 2020-07-30
`TrainingData.Yvars` should not be `Optional`

EricZLou:
Summary:
`Yvars` does not need to be declared as an `Optional` variable since it is passed directly from the `ArrayModelBridge` on the Ax side.

**Previously:**
```
class TrainingData(NamedTuple):
    Xs: List[Tensor]
    Ys: List[Tensor]
    Yvars: Optional[List[Tensor]] = None
```
**Now:**
```
class TrainingData(NamedTuple):
    Xs: List[Tensor]
    Ys: List[Tensor]
    Yvars: List[Tensor]
```

Differential Revision: D22833487


sdaulton:
Hmm, but a `SingleTaskGP` does not require Yvars. Initializing `TrainingData` with an empty list for `Yvars` seems less elegant than just letting it default to None.

PR #499, opened 2020-07-30
Add MANIFEST.in to include license file in source distribution

synapticarbors:
It would be helpful to distribute the license file with the source distribution.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,0 +1,1 @@
+include LICENSE


Balandat:
Good idea, but can we instead of adding a `MANIFEST.in` file just specify the `license_file` option in `setup.cfg` for this (https://packaging.python.org/guides/using-manifest-in/#how-files-are-included-in-an-sdist)?

synapticarbors:
I've always just used a `MANIFEST.in` since it always works. The `setup.cfg` should work in principle, but I never liked that you could specify `license_file` vs`license_files` and it was dependent on the version of `setuptools`. I would say `setup.cfg` is fine if you can guarantee that your build/sdist environment has an adequate version of `setuptools`.

Balandat:
Fair enough, happy to merge this is as is.

PR #505, opened 2020-08-04
Fix bug in draw_sobol_samples that did not use the proper effective dimension

### LINKED ISSUE #504: [Bug] Samples generated using draw_sobol_samples do not span the sample space
# 🐛 Bug

When generating samples using `draw_sobol_samples`, the resulting samples do not span the solution space for each q-batch. The example below shows that when drawing samples from 1 dimensional sample space, some of the q-batches only have samples from [0, 0.5]. This appears to be a result of `draw_sobol_samples` reshaping the `(n * q) x d` samples generated using the `SobolEngine` into `n x q x d`.

## To reproduce

```python
import torch
from botorch.utils import draw_sobol_samples

samples = draw_sobol_samples(
    bounds=torch.tensor([[0.], [1.]]), n=4000, q=100
)
max, _ = torch.max(samples, dim=0)
print(max)

```

Max above will include some entries as `0.4999` and `0.5000`, meaning that all samples corresponding to those q-batches come from [0, 0.5].

## Expected Behavior

All entries of max should be very close to 1, meaning that the samples span the sample space [0, 1] for each q-batch.

## System information

BoTorch 0.3.0
GPyTorch 1.1.1
PyTorch 1.5.1

Balandat:
Fixes #504

### DIFF
--- a/botorch/optim/initializers.py
+++ b/botorch/optim/initializers.py
@@ -77,10 +77,10 @@ def gen_batch_initial_conditions(
 
     q = 1 if q is None else q
     # the dimension the samples are drawn from
-    dim = bounds.shape[-1] * q
-    if dim > SobolEngine.MAXDIM and settings.debug.on():
+    effective_dim = bounds.shape[-1] * q
+    if effective_dim > SobolEngine.MAXDIM and settings.debug.on():
         warnings.warn(
-            f"Sample dimension q*d={dim} exceeding Sobol max dimension "
+            f"Sample dimension q*d={effective_dim} exceeding Sobol max dimension "
             f"({SobolEngine.MAXDIM}). Using iid samples instead.",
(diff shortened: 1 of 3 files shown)


saitcakmak:
@Balandat, I just noticed that this may cause a minor bug down the line. Suggested change:
- `botorch.optim.initializers.py` line 91 `if dim <= SobolEngine.MAXDIM:` should change to `if q * dim <= SobolEngine.MAXDIM:`

Balandat:
So if you look here you'll see that dim is actually `q * d`: https://github.com/pytorch/botorch/blob/8ca4ac2ec54af6ed2c4de77e8c368063e62b870e/botorch/optim/initializers.py#L80

Let me rename this to eff_dim to be less confusing

Issue #507, opened 2020-08-06
[Feature Request] Utilities for batch qMC (Sobol) sampling


saitcakmak:
# 🚀 Feature Request

The current qMC sampler `draw_sobol_samples` only allows for generating samples of shape `n x q x d`. It would be nice to have a method that generates low discrepancy samples of shape `n x *batch_shape x q x d` for an arbitrary batch shape. The desired output is equivalent to running a for loop over batches to generate `n x q x d` Sobol samples for each batch. 

## Motivation

The use case for this is generating the `raw_samples` / `initial_conditions` for optimization of batched GP models / acquisition functions. One can use `torch.rand` to generate `n x batch_shape x q x d` samples, however, having low discrepancy samples is more desirable for obvious reasons.

## Pitch

**Describe the solution you'd like**

A method `draw_batch_sobol_samples(bounds, n, q, batch_shape, seed)` that returns a `n x *batch_shape x q x d` tensor of samples where each batch is an independent `n x q x d` tensor of Sobol samples. The method should avoid initializing multiple `SobolEngine`, as this leads to huge overhead; as well as reshaping a big batch of Sobol samples, as this leads to #504.

**Describe alternatives you've considered**

- A seemingly obvious solution is to generate `(n * batch_size) x q x d` samples using `draw_sobol_samples`, and reshape it. This leads to the buggy samples described in #504. 
- Another obvious, though, inefficient solution is to generate `n x q x d` samples for each batch using `draw_sobol_samples` in a for loop. This leads to huge overhead due to repeated initialization of `SobolEngine`.
- Another alternative is to initialize the `SobolEngine` once, use this within a loop to generate `n x q x d` samples for each batch. To avoid issues similar to #504, after each call to `SobolEngine.draw(n)`, call `SobolEngine.fast_forward(torch.randint(some_constant), (1,))`. 
    This avoids overhead due to repeated initialization of `SobolEngine` and possibly #504. Using a for loop may get expensive as `batch_size` increases, and it may possibly lead to some unforeseen issues similar to #504. 
    It may be possible to achieve a better performance by directly modifying the `SobolEngine` source code.
    Repeated calls to `SobolEngine.draw` can also be avoided by first generating the full sequence of `randint` that are passed to `fast_forward`; calling `draw(n * batch_size + sum(randint))` to generate all samples at once; then deleting the samples corresponding to `fast_forward` before reshaping the samples to  `n x *batch_shape x q x d`.

[...]

PR #510, opened 2020-08-07
Add support for batch qMC sampling

saitcakmak:
## Motivation

See #507. Batch qMC samples are useful for generating raw_samples while optimizing batch GP models / acquisition functions.

Yes

## Test Plan

Added unit tests to verify that the samples have proper shape (passed).
Used the script in https://github.com/pytorch/botorch/issues/507#issuecomment-670188129 to verify that each batch of samples has low discrepancy.

### DIFF
--- a/botorch/utils/sampling.py
+++ b/botorch/utils/sampling.py
@@ -12,7 +12,7 @@ from __future__ import annotations
 
 import warnings
 from contextlib import contextmanager
-from typing import Generator, Optional
+from typing import Generator, Optional, Iterable
 
 import torch
 from botorch.exceptions.warnings import SamplingWarning
@@ -144,7 +144,11 @@ def construct_base_samples_from_posterior(
 
 
 def draw_sobol_samples(
-    bounds: Tensor, n: int, q: int, seed: Optional[int] = None
+    bounds: Tensor,
+    n: int,
+    q: int,
+    batch_shape: Optional[Iterable[int], torch.Size] = None,
+    seed: Optional[int] = None,
 ) -> Tensor:
     r"""Draw qMC samples from the box defined by bounds.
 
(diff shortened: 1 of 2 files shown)


botorch/utils/sampling.py
```diff
@@ -154,22 +158,30 @@ def draw_sobol_samples(
...
         q: The size of each q-batch.
+        batch_shape: The batch shape of the samples. If given, returns samples
+            of shape `n x *batch_shape x q x d`, where each batch is an
+            `n x q x d`-dim tensor of qMC samples.
         seed: The seed used for initializing Owen scrambling. If None (default),
             use a random seed.
 
     Returns:
-        A `n x q x d`-dim tensor of qMC samples from the box defined by bounds.
+        A `n [x *batch_shape] x q x d`-dim tensor of qMC samples from the box
```
Balandat, line 168: we usually just use the following notation: `n x batch_shape x q x d`, understanding that `batch_shape` is a (possibly 0-element) tuple.

botorch/utils/sampling.py
```diff
@@ -154,22 +158,30 @@ def draw_sobol_samples(
...
+    batch_size = int(torch.prod(torch.tensor(batch_shape)))
     d = bounds.shape[-1]
     lower = bounds[0]
     rng = bounds[1] - bounds[0]
     sobol_engine = SobolEngine(q * d, scramble=True, seed=seed)
-    samples_raw = sobol_engine.draw(n, dtype=lower.dtype).view(n, q, d)
-    samples_raw = samples_raw.to(device=lower.device)
+    samples_raw = sobol_engine.draw(batch_size * n, dtype=lower.dtype)
+    samples_raw = samples_raw.view(*batch_shape, n, q, d).to(device=lower.device)
+    if batch_shape != torch.Size([]):
```
Balandat, line 183: if you explicitly check `batch_shape is None` above, then batch_shape here may be a tuple or a list, in which case the check fails and this breaks. Since empty tuples/lists evaluate to False you could do `batch_shape = batch_shape or torch.Size()`

test/utils/test_sampling.py
```diff
@@ -160,13 +160,22 @@ def test_construct_base_samples_from_posterior(self):  # noqa: C901
...
+            (torch.float, torch.double),
         ):
             tkwargs = {"device": self.device, "dtype": dtype}
             bounds = torch.stack([torch.rand(d), 1 + torch.rand(d)]).to(**tkwargs)
-            samples = draw_sobol_samples(bounds=bounds, n=n, q=q, seed=seed)
-            self.assertEqual(samples.shape, torch.Size([n, q, d]))
+            samples = draw_sobol_samples(
+                bounds=bounds, n=n, q=q, batch_shape=batch_shape, seed=seed
+            )
+            batch_shape = torch.Size([]) if batch_shape is None else batch_shape
```
Balandat, line 177: as above let's do `batch_shape = batch_shape or torch.Size()`

saitcakmak:
Thanks for the feedback! I made the changes

Issue #345, opened 2019-12-24
Inefficiency in generating initial conditions while using optimize_acqf with fixed_features


saitcakmak:
## Issue description

While using `optimize_acqf` with `fixed_features`, the initial condition generation does not account for the `fixed_features`. The `fixed_features` are later applied during optimization with `gen_candidates_scipy`. I believe this could be improved by using `fixed_features` during initial condition generation to improve the quality of the initial conditions. 

## Code example

Following code can be used in debug mode along with a breakpoint at the `forward` pass of `qKnowledgeGradient` to observe the behavior. The first call of `forward` will not have `X[:, :, 0] = 0.5` fixed (this is the initial condition generation), later calls will have it fixed.

``` python
import torch
from botorch.acquisition import qKnowledgeGradient
from botorch.fit import fit_gpytorch_model
from botorch.models import SingleTaskGP
from botorch.models.transforms import Standardize
from botorch.optim import optimize_acqf
from gpytorch.mlls import ExactMarginalLogLikelihood
from torch import Tensor

train_X = torch.rand((10, 2))
train_Y = torch.sum(train_X, dim=-1, keepdim=True) + torch.randn(10, 1) * 0.1

gp = SingleTaskGP(train_X, train_Y, outcome_transform=Standardize(m=1))
mll = ExactMarginalLogLikelihood(gp.likelihood, gp)
fit_gpytorch_model(mll)

bounds = Tensor([[0], [1]]).repeat(1, 2)
acqf = qKnowledgeGradient(gp)
fixed_features = {0: 0.5}

solution, value = optimize_acqf(acqf, bounds=bounds, q=1, num_restarts=10, raw_samples=100, fixed_features=fixed_features)

print("solution = %s" % solution)
```

## System Info
Please provide information about your setup, including
- BoTorch Version: up to date with master
- GPyTorch Version: up to date with master
- PyTorch Version: 1.3.1
- Computer OS: macOS Catalina 10.15.2
- Python: 3.7.5 on Anaconda

Balandat:
Thanks for flagging this issue. You're correct, the `gen_batch_initial_conditions` helper that is used internally in `optimize_acqf` doesn't account for the fixed features. The proper thing to do here would be to pass the `fixed_features` dict down to that helper, generate a low-discrepancy set of sample points in `d - n_fix`-dimensional space, where `d` is the dim of the feature space and `n_fix` is the number of fixed features, and then insert the fixed columns at the right place. 

[...]

PR #515, opened 2020-08-13
Implementing evaluate qKG as requested in #350

### LINKED ISSUE #350: evaluate qKG [Feature Request]
Current implementation of qKG only supports evaluating X_actual along with X_fantasies. If I want to plot the qKG surface for a one or two-dimensional function, I would have to optimize w.r.t. X_fantasies for each X_actual I suppose? It would be nice to implement another function for qKG that does that

saitcakmak:
## Motivation

Resolves #350. The one-shot implementation of KG is not suitable for evaluating the KG value of the candidates. This implements an `evaluate_kg` method that returns the acquisition value of the candidates by solving the inner optimization problem.

Yes.

## Test Plan

- Added unit tests to verify that the generated samples are of appropriate shape, and do not violate bounds; and changes to the existing methods do not affect their functionality.
- Tested the code in debug mode to ensure everything works as intended.
- Verified the output by i) comparing `evaluate_kg` value of the optimizer of `optimize_acqf(qKG)` to a set of randomly drawn candidates - the optimizer should have the largest value; ii) comparing the value reported by `solution, value = optimize_acqf(qKG)` to `evaluate_kg(solution)` - `evaluate_kg` should return a larger value as it does global optimization. Base script: [test_eval_kg.txt](https://github.com/pytorch/botorch/files/5066548/test_eval_kg.txt).

## Changes

- Added `qKnowledgeGradiend.evaluate_kg()` method. This generates the fantasies, the value function, calls `gen_value_function_initial_conditions` to generate initial conditions, and uses `gen_candidates_scipy` to optimize them as a big batch. Selects the batch-wise maximizer, and returns the average over fantasies. 
- Added `gen_value_function_initial_conditions` in `botorch.optim.initializers`. Much like `gen_one_shot_kg_initial_conditions`, this first solves the current problem. The solutions to current problem are then used to generate `1 - frac_random` fraction of the raw samples, and the rest are generated using `draw_sobol_samples`. All raw samples are then evaluated, and the initial conditions are generated by passing the raw samples and their evaluations to `initialize_q_batch`.
- Modified `initialize_q_batch` in `botorch.optim.initializers` to allow for input `X` of shape `b x batch_shape x q x d` and `Y` of shape `b x batch_shape`, with a corresponding output shape of `n x batch_shape x q x d`. The batches are processed independently to ensure the quality of initial conditions corresponding to each batch.

## Things to address:

- The function / method names & call signatures could be replaced with something better.
- Default parameter values as noted with `TODO:`
- Should `gen_value_function_initial_conditions` be added to `botorch.optim.initializers.__init__.py`?
- I could not get `MockModel` to work for testing `evaluate_kg`. Where I failed is in optimizing the current problem in `gen_value_function_initial_conditions`, where the `PosteriorMean` of the `MockModel` would raise a `forward` not found error. I tried to `patch` it but couldn't get it to work. With the current tests using `SingleTaskGP`, the tests (`pytest -ra`) take 75 seconds on my desktop, of which 15 seconds is for testing `evaluate_kg` only. This should be improved.

cc @Balandat, @danielrjiang


saitcakmak:
One more thing: I left out support for constraints / fixed_features, since `make_scipy_linear_constraints` does not support batches. I figured it would make more sense to integrate that after #345 is resolved.

botorch/optim/initializers.py
```diff
@@ -226,6 +229,130 @@ def gen_one_shot_kg_initial_conditions(
...
+        raise ValueError(
+            f"frac_random must take on values in (0,1). Value: {frac_random}"
+        )
+
+    # compute maximizer of the current value function
+    value_function = _get_value_function(
+        model=current_model,
+        objective=acq_function.objective,
+        sampler=acq_function.sampler
+        if isinstance(acq_function, qSimpleRegret)
```
Balandat, line 306: let's make this work with other acquisition functions as well. Using qEI here actually works very well (see https://arxiv.org/abs/2006.15779).

```suggestion
        if isinstance(acq_function, MCAcquisitionFunction)
```

botorch/optim/initializers.py
```diff
@@ -226,6 +229,130 @@ def gen_one_shot_kg_initial_conditions(
     return ics
 
 
+def gen_value_function_initial_conditions(
+    acq_function: Union[PosteriorMean, qSimpleRegret],
```
Balandat, line 233: We can be more general here (see comment below)

```suggestion
    acq_function: AcquisitionFunction,
```

botorch/optim/initializers.py
```diff
@@ -226,6 +229,130 @@ def gen_one_shot_kg_initial_conditions(
...
+    from botorch.optim.optimize import optimize_acqf
+
+    fantasy_cands, fantasy_vals = optimize_acqf(
+        acq_function=value_function,
+        bounds=bounds,
+        q=1,
+        # TODO: set these to num_restarts and raw_samples?
+        num_restarts=options.get("num_inner_restarts", 20),
+        raw_samples=options.get("raw_inner_samples", 1024),
+        return_best_only=False,
```
Balandat, line 318: probably want to pass through `batch_limit` here

botorch/optim/initializers.py
```diff
@@ -226,6 +229,130 @@ def gen_one_shot_kg_initial_conditions(
...
+    ).permute(-1, *range(len(batch_shape)))
+    resampled = fantasy_cands[idx]
+    # add qMC samples
+    randomized = draw_sobol_samples(
+        bounds=bounds, n=raw_samples - n_value, q=1, batch_shape=batch_shape, seed=seed
+    )
+    # full set of raw samples
+    X_rnd = torch.cat([resampled, randomized], dim=0)
+
+    # evaluate the raw samples
```
Balandat, line 337: no need to use `batch_limit` here - this is to control the dimensionality fo the optimization problem in `optimize_acqf`. Here we don't optimize anything just evaluate, so the solution won't degrade if we do everything at once.

botorch/optim/initializers.py
```diff
@@ -264,27 +395,37 @@ def initialize_q_batch(X: Tensor, Y: Tensor, n: int, eta: float = 1.0) -> Tensor
...
         return X
 
-    Ystd = Y.std()
-    if Ystd == 0:
+    Ystd = Y.std(dim=0)
+    if torch.any(Ystd == 0):
         warnings.warn(
-            "All acqusition values for raw samples points are the same. "
-            "Choosing initial conditions at random.",
+            "All acquisition values for raw samples points are the same at "
```
Balandat, line 401: grammar nit: for at least one batch

test/acquisition/test_knowledge_gradient.py
```diff
@@ -213,6 +214,53 @@ def test_evaluate_q_knowledge_gradient(self):
...
+        for (
+            d,
+            q,
+            n,
+            n_f,
+            objective,
+            X_pending,
+            current_value,
+            dtype,
+        ) in itertools.product(
```
Balandat, line 233: this runs 256 different combinations which is where a lot of the long test times come from. Are some of these test combinations orthogonal and don't add value so we could prune some here?

test/acquisition/test_knowledge_gradient.py
```diff
@@ -213,6 +214,53 @@ def test_evaluate_q_knowledge_gradient(self):
...
+            (torch.float, torch.double),
+        ):
+            # basic test
+            bounds = torch.tensor([[0], [1]], device=self.device, dtype=dtype).repeat(
+                1, d
+            )
+            X_pending = X_pending.expand(-1, d) if X_pending is not None else None
+            train_X = torch.rand(n, d, device=self.device, dtype=dtype)
+            train_Y = torch.rand(n, 1, device=self.device, dtype=dtype)
+            model = SingleTaskGP(train_X, train_Y)
```
Balandat, line 250: we'll want to use a `MockModel` if we do a lot of different tests here (see comment above). We should probably  do a separate, basic tests that actually uses a real model, just to test this end-to-end, but doing this for all combos will be too slow.

test/optim/test_initializers.py
```diff
@@ -241,3 +248,48 @@ def test_gen_one_shot_kg_initial_conditions(self):
...
                 self.assertTrue(torch.all(ics[..., -n_value:, :] == 1))
+
+
+class TestGenValueFunctionInitialConditions(BotorchTestCase):
+    def test_gen_value_function_initial_conditions(self):
+        num_fantasies = 8
+        num_solutions = 3
+        num_restarts = 4
+        raw_samples = 16
+        n_train = 4
```
Balandat, line 259: let's try to get these numbers down to the minimum, every ms counts for the tests
```suggestion
        num_fantasies = 2
        num_solutions = 3
        num_restarts = 4
        raw_samples = 5
        n_train = 6
```

test/optim/test_initializers.py
```diff
@@ -241,3 +248,48 @@ def test_gen_one_shot_kg_initial_conditions(self):
...
+        dim = 2
+        for dtype in (torch.float, torch.double):
+            train_X = torch.rand(n_train, dim, device=self.device, dtype=dtype)
+            train_Y = torch.rand(n_train, 1, device=self.device, dtype=dtype)
+            model = SingleTaskGP(train_X, train_Y)
```
Balandat, line 264: will also want to use `MockModel` if we run through a lot of option combos

botorch/acquisition/knowledge_gradient.py
```diff
@@ -183,6 +183,73 @@ def forward(self, X: Tensor) -> Tensor:
...
+            options: Options for optimization of the inner problem. This includes
+                `num_restarts`, `raw_samples` and other optimization options.
+
+        Returns:
+            A Tensor of shape `b`. For t-batch b, the q-KG value of the design
+                `X_actual[b]` is averaged across the fantasy models.
+                NOTE: If `current_value` is not provided, then this is not the
+                true KG value of `X_actual[b]`.
+        """
+        while X_actual.dim() < 3:
```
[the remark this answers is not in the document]
saitcakmak: I modified `t_batch_mode_transform` to work with `kwargs` and used that instead

botorch/optim/initializers.py
saitcakmak: passed the full `options` dict to be future compatible

test/acquisition/test_knowledge_gradient.py
saitcakmak: I reduced this to only 2 combinations that likely covers everything I tested for. I may add some more based on codecov report. We could also add additional combinations using mock models as needed

test/optim/test_initializers.py
saitcakmak: It is currently testing for only 2 combinations. We could use `MockModel` to test for the generation logic. Do you think additional tests are needed?

saitcakmak:
I updated the code based on the review, and reduced the number of test combinations. The updated tests take ~200ms total.

test/utils/test_transforms.py
```diff
@@ -79,6 +79,11 @@ def q_method(self, X: Tensor) -> None:
     def q1_method(self, X: Tensor) -> None:
         return X
 
+    @t_batch_mode_transform()
+    def kw_method(self, X: Tensor, dummy_arg: Any = None):
+        assert dummy_arg is not None
```
Balandat, line 84: Consistency nit:
```suggestion
        self.assertIsNone(dummy_arg)
```

saitcakmak:
Updated! Thanks for the review!

saitcakmak:
My bad, committed from GitHub without testing. It is fixed now

saitcakmak:
I restructured `test_gen_value_function_initial_conditions` to increase the coverage without added overhead. Codecov should pass now

botorch/optim/initializers.py
```diff
@@ -226,6 +228,131 @@ def gen_one_shot_kg_initial_conditions(
...
+    if raw_samples > n_value:
+        randomized = draw_sobol_samples(
+            bounds=bounds,
+            n=raw_samples - n_value,
+            q=1,
+            batch_shape=batch_shape,
+            seed=seed,
+        )
+    else:
+        randomized = torch.empty(0, *batch_shape, 1, bounds.shape[-1])
```
Balandat, line 342: looks like this isn't covered by any test - mind adding one? In fact, looks like this will break with non-default dtype..

saitcakmak: Actually, that line is unreachable. Due to `if not 0 < frac_random < 1: raise ValueError` and `n_value = int((1 - frac_random) * raw_samples)`, `raw_samples` will be at least 1 greater than `n_value`. 

I removed that if/else and fixed the dtype issue. This made one of the test cases redundant, so I removed that too

Balandat:
Many thanks for the contribution, @saitcakmak !

saitcakmak:
Happy to help! Thanks for the reviews, @Balandat !

PR #555, opened 2020-09-16
Handle case when nb metada has no kernelspec attribute in tutorial parsing

Balandat:
This allows notebook parsing to succeed even if kernelspec is not defined in the source metadata (this is the case e.g. for `multi_fidelity_bo.ipynb`, not sure why).


Balandat:
Abandoning this in favor of fixing the tutorial notebook.

PR #565, opened 2020-09-21
Fitting a model tutorial: update URL

markus289:
The URL to the original notebook changed. Update accordingly.


Balandat:
Thanks!

Issue #573, opened 2020-10-07
Shaping of qNegativeIntegratedPosteriorVariance and optimize_acqf_function [Bug]


wjmaddox:
# 🐛 Bug

There's an indexing error that occurs at the end of an `optimize_acqf` call when optimizing `qNegativeIntegratedPosteriorVariance`. The error results because `qNegativeIntegratedPosteriorVariance` reduces over an incorrect dimension --- an input of size `batch x n x d` produces an output of size `batch x n_mc_points`. 

Edit: An error occurs in initialization if you don't use an objective.

## To reproduce

** Code snippet to reproduce **
```python
import torch

from botorch.optim import optimize_acqf
from botorch.models import SingleTaskGP
from botorch.acquisition.active_learning import qNegIntegratedPosteriorVariance
from botorch.acquisition.objective import ScalarizedObjective
from botorch.sampling.samplers import SobolQMCNormalSampler

train_X = torch.rand(10, 2)
Y = 1 - (train_X - 0.5).norm(dim=-1, keepdim=True)  # explicit output dimension
Y += 0.1 * torch.rand_like(Y)
train_Y = (Y - Y.mean()) / Y.std()

model = SingleTaskGP(train_X, train_Y)

qmc_sampler = SobolQMCNormalSampler(num_samples=32)
bounds = torch.stack([torch.zeros(2), torch.ones(2)])
qnipv = qNegIntegratedPosteriorVariance(
    model=model, 
    mc_points=train_X, 
    sampler=qmc_sampler,
    objective=ScalarizedObjective(weights=torch.ones(1))
)
```

** Stack trace/error message **
```
---------------------------------------------------------------------------
IndexError                                Traceback (most recent call last)
<ipython-input-10-75fdb3cbe1eb> in <module>
      5     num_restarts=1,
      6     raw_samples=10,  # used for intialization heuristic
----> 7     options={"batch_limit": 5, "maxiter": 200},
      8 )

~/Documents/GitHub/botorch/botorch/optim/optimize.py in optimize_acqf(acq_function, bounds, q, num_restarts, raw_samples, options, inequality_constraints, equality_constraints, fixed_features, post_processing_func, batch_initial_conditions, return_best_only, sequential, **kwargs)
    190     if return_best_only:
    191         best = torch.argmax(batch_acq_values.view(-1), dim=0)
--> 192         batch_candidates = batch_candidates[best]
    193         batch_acq_values = batch_acq_values[best]
    194 

[...]

PR #577, opened 2020-10-16
Implements output shape checks for t_batch_mode_transform

saitcakmak:
## Motivation

In most cases, the acquisition functions are intended to return an output with `shape` matching the `batch_shape` of `X`, or the `model._input_batch_shape` in the case of acquisition functions using batched models. In certain cases (#573), when the acqf attributes are mis-specified, we get an output of wrong shape, which either leads errors down the line, or could possibly lead to buggy output. This implements an `assert_output_shape` flag in `t_batch_mode_transform` (defaults to `True`), which checks the output shape to catch the error in place.

Other changes:
`qMultiFidelityKnowledgeGradient` was missing from `__init__.py`, added it.
Set `assert_output_shape=False` for `ExpectedImprovement`, since `NoisyExpectedImprovement` uses it to get output with an additional dimension.
Updated some mock tests to match the correct input / output shapes.

(Write your motivation here.)

(Write your answer here.)
Yes.

## Test Plan

Added unit tests to ensure everything works as intended. This also lead to discovering some shape mismatches in the mock tests in `test.acquisition`, which got fixed in the process.

cc @Balandat


saitcakmak:
That was an interesting failure. CircleCI fails black formatting that is enforced by my local black installation. I updated it to match what CircleCI recommends.

botorch/utils/transforms.py
```diff
@@ -166,7 +171,29 @@ def decorated(cls: Any, X: Tensor, **kwargs: Any) -> Any:
...
+                and not (
+                    output.shape == torch.Size() and X.shape[:-2] == torch.Size([1])
+                )  # output of size 1 is typically squeezed
+            ):
+                # the check above may fail when using a batched model.
+                # ensure that this is not the case
+                from botorch.acquisition import AcquisitionFunction
+
+                if (
+                    not isinstance(cls, AcquisitionFunction)
```
Balandat, line 187: Do we ever use this decorator outside of acquisition functions?

botorch/utils/transforms.py
```diff
@@ -166,7 +171,29 @@ def decorated(cls: Any, X: Tensor, **kwargs: Any) -> Any:
...
+                    or output.shape != cls.model._input_batch_shape
```
Balandat, line 188: this may raise an `AttributeError` for models that don't have an `_input_batch_shape` property.

Balandat (commented):
Thanks for putting this up and for adding solid tests. The fact that we can have batched models makes the shape checking somewhat complicated. Let me think a little bit about what the best way of doing this in a general setting is. 

Maybe @sdaulton has some thoughts on this as well.

botorch/utils/transforms.py
saitcakmak: Looks like it is currently only used for acquisition functions. I'll remove that check

saitcakmak:
Thanks for reviewing this. I updated the handling of batched models in line with the comments. 

My experience with batched models is pretty much limited to fantasy models, so I don't know if there are cases this is missing. I can update it / add tests later based on the feedback.

saitcakmak:
@Balandat I'll bring this up to date with #588. Are there other changes this needs?

botorch/utils/transforms.py
```diff
@@ -166,7 +171,25 @@ def decorated(cls: Any, X: Tensor, **kwargs: Any) -> Any:
                     f" got X with shape {X.shape}."
                 )
             X = X if X.dim() > 2 else X.unsqueeze(0)
-            return method(cls, X, **kwargs)
+            output = method(cls, X, *args, **kwargs)
+            if (
```
Balandat, line 175: This whole if clause is kind of complex, mind putting this into a helper function that returns a bool? That will clean things up. 

Also, can you migrate this from using the `_input_batch_shape` to the universal `batch_shape` interface introduced in #588?

Balandat (changes requested):

saitcakmak:
I made the changes and added & fixed some test cases.

botorch/utils/transforms.py
```diff
@@ -124,19 +124,52 @@ def normalize_indices(indices: Optional[List[int]], d: int) -> Optional[List[int
     return normalized_indices
 
 
+def _verify_output_shape(cls: Any, X: Tensor, output: Tensor) -> bool:
```
Balandat, line 127: If the first argument is an acquisition function object, let's maybe use `acqf: AcquisitionFunction` for this?

botorch/utils/transforms.py
```diff
@@ -166,7 +199,17 @@ def decorated(cls: Any, X: Tensor, **kwargs: Any) -> Any:
...
                 )
             X = X if X.dim() > 2 else X.unsqueeze(0)
-            return method(cls, X, **kwargs)
+            output = method(cls, X, *args, **kwargs)
+            if assert_output_shape and not _verify_output_shape(
+                cls=cls, X=X, output=output
+            ):
+                raise AssertionError(
+                    "Expected the output shape to match either the batch shape of "
+                    "X, or the `model._input_batch_shape` in the case of "
```
Balandat, line 208:
```suggestion
                    "X, or the `model.batch_shape` in the case of "
```

botorch/utils/transforms.py
```diff
@@ -124,19 +124,52 @@ def normalize_indices(indices: Optional[List[int]], d: int) -> Optional[List[int
...
+    """
+    try:
+        return (
+            output.shape == X.shape[:-2]
+            or (output.shape == torch.Size() and X.shape[:-2] == torch.Size([1]))
+            or output.shape == cls.model.batch_shape
+        )
+    except (AttributeError, NotImplementedError):
+        # cls does not have model or cls.model does not define `batch_shape`
+        return False
```
Balandat, line 152: So I don't know if returning `False` here is the right thing to do. Basically if we get to this point, we have made some assumptions about the acquisition function (that it operates on a model that implements a `batch_shape` property) that are not true. I think maybe being somewhat lenient here and returning `True` but raising a `RuntimeWarning` pointing this out instead may be the better approach (as in less annoying to work around if doing custom stuff)?

test/acquisition/test_knowledge_gradient.py
```diff
@@ -349,7 +349,9 @@ def test_evaluate_q_multi_fidelity_knowledge_gradient(self):
                     patch_f.assert_called_once()
                     cargs, ckwargs = patch_f.call_args
                     self.assertEqual(ckwargs["X"].shape, torch.Size([b, 1, 1]))
-            val_exp = mock_util(X, mean.squeeze(-1) - current_value).mean(dim=0)
+            val_exp = mock_util(X, mean.squeeze(-1).squeeze(-1) - current_value).mean(
```
Balandat, line 352:
```suggestion
            val_exp = mock_util(X, mean.view(mean.shape[:-2]) - current_value).mean(
```

test/acquisition/test_monte_carlo.py
```diff
@@ -761,7 +761,7 @@ def test_q_upper_confidence_bound_batch(self):
             acqf.set_X_pending(None)
             self.assertIsNone(acqf.X_pending)
             acqf.set_X_pending(X)
-            self.assertEqual(acqf.X_pending, X)
+            self.assertTrue((acqf.X_pending == X).all())
```
Balandat, line 764:
```suggestion
            self.assertTrue(torch.equal(acqf.X_pending, X))
```

Balandat (changes requested):
Looks mostly good to me, thanks for polishing this.

My main comment is on how to handle the case when the acquisition function doesn't have a model or if the model doesn't implement `batch_shape`.

The rest are mostly nits.

test/utils/test_transforms.py
```diff
@@ -17,7 +17,10 @@
     standardize,
     t_batch_mode_transform,
     unnormalize,
+    _verify_output_shape,
 )
+from botorch.models.model import Model
+from botorch.utils.testing import MockModel, MockPosterior
```
Balandat, line 23: oh and can you rearrange these inputs alphabetically? linter will complain about this.

saitcakmak: What convention do we follow for ordering inputs? Using the notation `from <package> import <module>`, this is currently sorted for the <package> and over <module> for each package. Does linter expect it to be sorted over <module> instead? Trying to get a better understanding for future reference

botorch/utils/transforms.py
saitcakmak: I am thinking in what scenario you'd get to here. Either you'd use one of the canned acquisition functions (that uses `t_batch_mode_transform`) with a custom model that doesn't define `batch_shape`, or you wrote a custom class, used `t_batch_model_transform` in it without passing `assert_output_shape = False`. In the second scenario, it is easy to fix by passing `assert_output_shape = False`, but the first one would require editing the source or redefining the acquisition function, which is annoying. Yeah, ignoring a `RuntimeWarning` sounds less annoying.

botorch/utils/transforms.py
```diff
@@ -154,7 +187,7 @@ def t_batch_mode_transform(
 
     def decorator(method: Callable[[Any, Tensor], Any]) -> Callable[[Any, Tensor], Any]:
         @wraps(method)
-        def decorated(cls: Any, X: Tensor, **kwargs: Any) -> Any:
+        def decorated(cls: Any, X: Tensor, *args: Any, **kwargs: Any) -> Any:
```
saitcakmak, line 190: @Balandat, Should this also change to `acqf: AcquisitionFunction` since that is the only use case? I think it makes sense to keep the signature consistent with `_verify_output_shape`.

test/utils/test_transforms.py
Balandat: So we have been using iSort so far - I'm about to switch over to usort in #603, but there is one edge case that has to do with FB-infra that needs to be fixed upstream in usort before I can do that. 

Basically to fix this you just need to merge the `botorch.utils.testing` imports with the existing one above and move the models import up to be in alphabetic order.

botorch/utils/transforms.py
saitcakmak: This actually leads to a circular import

PR #592, opened 2020-11-06
Hit and run sampler for uniform sampling from a polytope (#591)

jelena-markovic:
Summary:
Pull Request resolved: https://github.com/pytorch/botorch/pull/591

Hit and run sampler for uniform sampling from a constrained set described with a set of inequalities and possibly a set of equalities.

Reviewed By: Balandat

Differential Revision: D23145190

### DIFF
--- a/botorch/utils/sampling.py
+++ b/botorch/utils/sampling.py
@@ -12,8 +12,10 @@ from __future__ import annotations
 
 import warnings
 from contextlib import contextmanager
-from typing import Generator, Iterable, Optional
+from typing import Generator, Iterable, Optional, Tuple
 
+import numpy as np
+import scipy
 import torch
 from botorch.exceptions.warnings import SamplingWarning
 from botorch.posteriors.posterior import Posterior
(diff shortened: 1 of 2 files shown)


Balandat:
@jduerholt

jduerholt:
Thanks for including this. It is very helpful for us and looks much better than my "quick-and-dirty solution". I will test it in the next week.

jduerholt:
Hi @Balandat and @jelena-markovic,

thanks again for implementing the hit and run sampler for sampling uniformly from a polytope. It really helps for our use cases!

Testing it, I just found one minor bug. At the moment it is not possible to invoke the draw method twice on a just initialized sampler. See the MWE attached.

Instead of defining the new x0 as `self.x0 = samples[samples.size(-1)]` in line 484 of botorch/utils/sampling.py defining it as
`self.x0=samples[-1].reshape(-1,1)` should fix it.

It will also fix the problem that at the moment not the last element of the chain is taken as new x0.

Best,

Johannes

[mwe.py.txt](https://github.com/pytorch/botorch/files/5560214/mwe.py.txt)

Balandat:
Great, thanks for spotting this bug and providing the fix, will go in asap.

PR #594, opened 2020-11-08
Modifies qMFKG.evaluate() to work with project, expand and cost aware utility

### LINKED ISSUE #587: [Bug] Issues with qKG.evaluate()
# 🐛 Bug

i) This line raises an `AttributeError` with models that do not have an `_input_batch_shape` attribute. (#578 as an example)
https://github.com/pytorch/botorch/blob/7e2a4040e60c6351ff127d6ff65506955c2beccb/botorch/optim/initializers.py#L328

ii) Since it is a subclass of `qKnowledgeGradient`, `qMultiFidelityKnowledgeGradient` has `evaluate` method implemented. However, the current implementation ignores the `project` operator in the inner optimization, and may produce buggy output. 

## Expected Behavior

i) The batch shape could be inferred from `model.train_inputs` batch shape. My observation is `_input_batch_shape = model.train_inputs[0].shape[:-2]`, but I am not sure if this is always true. If this is true, a simple try/except should do it.

ii) It should either raise a `NotImplementedError`, or be modified to accommodate the `project` operator.

saitcakmak:
## Motivation

Modifies `qMFKG.evaluate()` to work with `project`, `expand` and `cost_aware_utility`. Partially fixes #587. 

- Introduces a `ProjectedValueFunction` that wraps the `value_function` and applies the `project` operator on the `forward` call.
- Changes `evaluate()` signature to use `X` instead of `X_actual`. Current implementation raises an exception with the decorators when called with `evaluate(X_actual=...)`.

Note: The treatment of `cost_aware_utility` assumes that it is monotone non-decreasing in `deltas`. Otherwise, optimizing the inner problem and passing through `cost_aware_utility` may not produce the correct output.

Yes

## Test Plan

Added mock unit tests. Verified the expected behavior in additional offline tests.

### DIFF
--- a/botorch/acquisition/knowledge_gradient.py
+++ b/botorch/acquisition/knowledge_gradient.py
@@ -189,12 +189,12 @@ class qKnowledgeGradient(MCAcquisitionFunction, OneShotAcquisitionFunction):
 
     @concatenate_pending_points
     @t_batch_mode_transform()
-    def evaluate(self, X_actual: Tensor, bounds: Tensor, **kwargs: Any) -> Tensor:
+    def evaluate(self, X: Tensor, bounds: Tensor, **kwargs: Any) -> Tensor:
         r"""Evaluate qKnowledgeGradient on the candidate set `X_actual` by
         solving the inner optimization problem.
 
         Args:
-            X_actual: A `b x q x d` Tensor with `b` t-batches of `q` design points
+            X: A `b x q x d` Tensor with `b` t-batches of `q` design points
                 each. Unlike `forward()`, this does not include solutions of the
                 inner optimization problem.
             bounds: A `2 x d` tensor of lower and upper bounds for each column of
@@ -206,18 +206,26 @@ class qKnowledgeGradient(MCAcquisitionFunction, OneShotAcquisitionFunction):
 
         Returns:
             A Tensor of shape `b`. For t-batch b, the q-KG value of the design
-                `X_actual[b]` is averaged across the fantasy models.
+                `X[b]` is averaged across the fantasy models.
                 NOTE: If `current_value` is not provided, then this is not the
-                true KG value of `X_actual[b]`.
+                true KG value of `X[b]`.
         """
+        if self.X_pending is not None:
+            X = torch.cat([X, match_batch_shape(self.X_pending, X)], dim=-2)
+        if hasattr(self, "expand"):
+            X = self.expand(X)
+
         # construct the fantasy model of shape `num_fantasies x b`
         fantasy_model = self.model.fantasize(
-            X=X_actual, sampler=self.sampler, observation_noise=True
+            X=X, sampler=self.sampler, observation_noise=True
         )
 
         # get the value function
         value_function = _get_value_function(
-            model=fantasy_model, objective=self.objective, sampler=self.inner_sampler
+            model=fantasy_model,
+            objective=self.objective,
+            sampler=self.inner_sampler,
+            project=getattr(self, "project", None),
         )
 
         # optimize the inner problem
@@ -245,6 +253,10 @@ class qKnowledgeGradient(MCAcquisitionFunction, OneShotAcquisitionFunction):
         if self.current_value is not None:
             values = values - self.current_value
 
+        if getattr(self, "cost_aware_utility", None) is not None:
+            values = self.cost_aware_utility(
+                X=X, deltas=values, sampler=self.cost_sampler
+            )
         # return average over the fantasy samples
         return values.mean(dim=0)
 
@@ -408,13 +420,16 @@ class qMultiFidelityKnowledgeGradient(qKnowledgeGradient):
 
         # get the value function
         value_function = _get_value_function(
-            model=fantasy_model, objective=self.objective, sampler=self.inner_sampler
+            model=fantasy_model,
+            objective=self.objective,
+            sampler=self.inner_sampler,
+            project=self.project,
         )
 
         # make sure to propagate gradients to the fantasy model train inputs
         # project the fantasy points
         with settings.propagate_grads(True):
-            values = value_function(X=self.project(X_fantasies))  # num_fantasies x b
+            values = value_function(X=X_fantasies)  # num_fantasies x b
 
         if self.current_value is not None:
             values = values - self.current_value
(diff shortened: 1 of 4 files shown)


botorch/acquisition/knowledge_gradient.py
```diff
@@ -245,6 +253,10 @@ def evaluate(self, X_actual: Tensor, bounds: Tensor, **kwargs: Any) -> Tensor:
         if self.current_value is not None:
             values = values - self.current_value
 
+        if getattr(self, "cost_aware_utility", None) is not None:
```
Balandat, line 256:
```suggestion
        if hasattr(self, "cost_aware_utility"):
```

botorch/acquisition/knowledge_gradient.py
```diff
@@ -428,16 +443,44 @@ def forward(self, X: Tensor) -> Tensor:
         return values.mean(dim=0)
 
 
+class ProjectedValueFunction(AcquisitionFunction):
+    r"""
+    Defines a wrapper around the value function that incorporates the project
+    operator.
```
Balandat, line 449:
```suggestion
    Defines a wrapper around  an `AcquisitionFunction` that incorporates the project
    operator. Typically used to handle value functions in look-ahead methods.
```

botorch/acquisition/knowledge_gradient.py
```diff
@@ -428,16 +443,44 @@ def forward(self, X: Tensor) -> Tensor:
...
+    """
+
+    def __init__(
+        self,
+        base_value_function: AcquisitionFunction,
+        project: Callable[[Tensor], Tensor],
+    ):
```
Balandat, line 456: Implied, but our convention is to add hints throughout.

```suggestion
    ) -> None:
```

botorch/acquisition/knowledge_gradient.py
```diff
@@ -428,16 +443,44 @@ def forward(self, X: Tensor) -> Tensor:
...
+        base_value_function = qSimpleRegret(
+            model=model, sampler=sampler, objective=objective
+        )
+    else:
+        base_value_function = PosteriorMean(model=model, objective=objective)
+    if project is None:
+        return base_value_function
     else:
-        return PosteriorMean(model=model, objective=objective)
+        return ProjectedValueFunction(base_value_function, project)
```
Balandat, line 483: Our convention is to call with explicit kwargs everywhere (where it makes sense...)

```suggestion
        return ProjectedValueFunction(
            base_value_function=base_value_function,
            project=project,
        )
```

botorch/acquisition/knowledge_gradient.py
```diff
@@ -428,16 +443,44 @@ def forward(self, X: Tensor) -> Tensor:
         return values.mean(dim=0)
 
 
+class ProjectedValueFunction(AcquisitionFunction):
```
Balandat, line 446: Let's make sure names are as descriptive as possible. I added clarifying note in the docstring that this is typically used for value functions.

```suggestion
class ProjectedAcquisitionFunctionFunction(AcquisitionFunction):
```

test/acquisition/test_knowledge_gradient.py
```diff
@@ -399,6 +400,62 @@ def test_evaluate_q_multi_fidelity_knowledge_gradient(self):
...
+                                torch.ones(1, 1, 1, device=self.device, dtype=dtype),
+                                torch.ones(1, device=self.device, dtype=dtype),
+                            ),
+                        ):
+                            val = qMFKG.evaluate(
+                                X=torch.zeros(1, 1, 1, device=self.device, dtype=dtype),
+                                bounds=torch.tensor([[0.0], [1.0]]),
+                                num_restarts=1,
+                                raw_samples=1,
+                            )
```
Balandat, line 446: Can use `ExitStack` to get rid of the nested contexts

```suggestion
            with ExitStack() as es:
                patch_f = es.enter_context(
                    mock.patch.object(MockModel, "fantasize", return_value=mfm),
                )
                mock_num_outputs = es.enter_context(
                    mock.patch(NO, new_callable=mock.PropertyMock),
                )
                es.enter_context(
                    mock.patch(
                       "botorch.optim.optimize.optimize_acqf",
                       return_value=(
                           torch.ones(1, 1, 1, device=self.device, dtype=dtype),
                           torch.ones(1, device=self.device, dtype=dtype),
                       ),
                   ),
                )
                es.enter_context(
                    mock.patch(
                        "botorch.generation.gen.gen_candidates_scipy",
                        return_value=(
                            torch.ones(1, 1, 1, device=self.device, dtype=dtype),
                            torch.ones(1, device=self.device, dtype=dtype),
                        ),
                    ),
                )

                mock_num_outputs.return_value = 1
                qMFKG = qMultiFidelityKnowledgeGradient(
                    model=mm,
                    num_fantasies=n_f,
                    X_pending=torch.rand(1, 1, 1, device=self.device, dtype=dtype),
                    current_value=torch.zeros(1, device=self.device, dtype=dtype),
                    cost_aware_utility=cau,
                    project=lambda X: torch.zeros_like(X),
                    expand=lambda X: torch.ones_like(X),
                )
                val = qMFKG.evaluate(
                     X=torch.zeros(1, 1, 1, device=self.device, dtype=dtype),
                     bounds=torch.tensor([[0.0], [1.0]]),
                     num_restarts=1,
                     raw_samples=1,
                 )
```

test/acquisition/test_knowledge_gradient.py
```diff
@@ -399,6 +400,62 @@ def test_evaluate_q_multi_fidelity_knowledge_gradient(self):
...
+                    patch_f.asset_called_once()
+                    cargs, ckwargs = patch_f.call_args
+                    self.assertTrue(
+                        (
+                            ckwargs["X"]
+                            == torch.ones(2, 1, 1, device=self.device, dtype=dtype)
```
Balandat, line 452: use `torch.equal` here

Balandat (commented):
Sorry, just got to this. Thanks for putting this up. A few nits mostly, o/w this looks good. 

@danielrjiang, @RaulAstudillo06 any comments / thoughts?

test/acquisition/test_knowledge_gradient.py
saitcakmak: Cool! I've never used `ExitStack` before.

saitcakmak:
Thanks for the review! I implemented the changes. Using `torch.equal` for the test revealed a small bug (concatenating pending points twice), which got fixed in the process. Turns out `torch_ones(2, 1, 1) == torch.ones(1, 3, 1)` returns a `2 x 3 x 1` tensor of `True` instead of raising an error.

test/acquisition/test_knowledge_gradient.py
```diff
@@ -3,7 +3,7 @@
 #
 # This source code is licensed under the MIT license found in the
 # LICENSE file in the root directory of this source tree.
-
+from contextlib import ExitStack
```
Balandat, line 6: Mind adding an empty line here? Our linter is complaining about this. 

```suggestion

from contextlib import ExitStack
```

Balandat (changes requested):

Balandat:
Thanks for contributing and sorry about the delay merging this in.

saitcakmak:
Thanks for the reviews. Happy to be contribute :)

PR #596, opened 2020-11-14
Fix HolderTable synthetic function bounds

milesial:
## Motivation

The 2D synthetic test function HolderTable had bounds that didn't match with its dimensionality, like every other test function does.

Yes

## Test Plan

It's just an internal attribute change, so no test here.

(CircleCI failed tests unrelated)


Balandat:
Thanks for fixing this. Test failure is unrelated, caused by upstream changes in gpytorch.

Balandat (approved):

PR #598, opened 2020-11-17
Add tutorial for TuRBO-1

dme65:
This adds a tutorial for TuRBO-1.

## Motivation

To allow people to use, improve upon, and experiment with TuRBO in BoTorch.

Yes.

## Test Plan

This only adds a notebook, so no additional tests should be needed.

## Related PRs

N/A


Balandat:
Thanks for putting this up, overall this lgtm.

A few comments:
- there is a lot of string indexing in the state - might be cleaner to use a `dataclass` container for it
- how about using https://github.com/pytorch/botorch/blob/dd17a4c19890e87460dbb50e14456c2f525aeeeb/botorch/generation/sampling.py#L59 for TS (after applying the perturbation mask)?
- let's put a note down to add handling of `SMOKE_TEST` env var to be introduced in #600

Test failure is unrelated - should disappear afterrebase.

tutorials/turbo_1.ipynb
```diff
@@ -0,0 +1,592 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 3,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "def create_initial_state(dim, batch_size):\n",
+    "    return {\n",
```
eytan, line 94: suggest using a `dataclass` or `namedtuple` here.

tutorials/turbo_1.ipynb
```diff
@@ -0,0 +1,592 @@
...
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Optimization loop\n",
+    "This simple loop runs one instance of TuRBO-1 with Thompson sampling until convergence.\n",
+    "\n",
+    "TuRBO-1 is a local optimizer that can be used for a fixed evaluation budget in a multi-start fashion.  Once TuRBO converges, `state[\"restart_triggered\"]` will be set to true and the run should be aborted.  If you want to run more evaluations with TuRBO, you simply generate a new set of initial points and then keep generating batches until convergence or when the evaluation budget has been exceeded.  It's important to note that evaluations from previous instances are discarded when TuRBO restarts.\n",
+    "\n",
+    "NOTE: We use a `FixedNoiseGP` in this case as the problem is noise-free."
```
eytan, line 246: I am not sure if we want to encourage users to input noisy observations when no observation noise is present.  Inferring the noise can help with numerical stability and help protect against model misspecification (see e.g., https://arxiv.org/pdf/1007.4580.pdf)

eytan (commented):
Looks pretty good to me, added two brief comments.

tutorials/turbo_1.ipynb
dme65: The thought here is that the `FixedNoiseGP` with a little bit of noise acts as regularization, but it makes sense to learn. I can switch over to the `SingleTaskGP` and add a constraint + prior to keep the learned noise from being large since we know the problem is noise-free.

dme65:
Thanks for fixing the failing tests on master, @Balandat! I've made some changes, what do you think?

Balandat (approved):
lgtm, thanks! Feel free to import and land.

PR #601, opened 2020-11-17
Sort imports with usort

Balandat:
We'll ensure consistent import sorting with µsort: https://github.com/facebookexperimental/usort, this batch formats all sources.


Balandat:
actually with the proper usort settings there aren't any changes. Closing this.

PR #603, opened 2020-11-18
Add ufmt check to CI and docs

Balandat:
Going forward, we will use `ufmt` to sort imports. `ufmt` includes both `black` and `usort` and runs the formatting in a single step (which will ensure there is consistency with FB-internal codeformatting tools).

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -3,3 +3,6 @@
 [build-system]
 requires = ["setuptools>=34.4", "wheel", "setuptools_scm"]
 build-backend = "setuptools.build_meta"
+
+[tool.usort]
+first_party_detection = false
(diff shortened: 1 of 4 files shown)


pyproject.toml
```diff
@@ -3,3 +3,6 @@
 [build-system]
 requires = ["setuptools>=34.4", "wheel", "setuptools_scm"]
 build-backend = "setuptools.build_meta"
+
+[tool.usort]
+first_party_detection = false
```
Balandat, line 8: This is the setting also in fb-internal repos, making this change for consistency.

Balandat:
CI failure is caused by https://github.com/facebookexperimental/usort/issues/49, we can hold off until that is fixed.

Balandat:
(this has been fixed in usort, just waiting on the next release)

Balandat:
Updated to new GitHub actions workflows. Still waiting for a µsort release to have it agree with black on the blank likes for comments between imports.

PR #604, opened 2020-11-18
modify _expand_bounds() shape checks to work with >2-dim bounds.

saitcakmak:
## Motivation

`botorch.optim.utils._expand_bounds()` raises an error when used with `bounds` with more than 2 dimensions. This PR modifies the shape checks in `_expand_bounds()` to allow more general shapes, as long as `bounds` has the same t-batch shape as `X`.

This change extends the functionality of `gen_candidates_scipy` to allow optimization with a custom bound for each entry of `initial_conditions`.

Yes.

## Test Plan

Added unit tests.

### DIFF
--- a/botorch/optim/utils.py
+++ b/botorch/optim/utils.py
@@ -153,7 +153,11 @@ def _expand_bounds(
             ebounds = bounds.view(1, -1)
         else:
             ebounds = bounds
-        if ebounds.shape[1] != X.shape[-1]:
+        if (
+            ebounds.shape[-1] != X.shape[-1]
+            or ebounds.shape[-2] not in [1, X.shape[-2]]
+            or (ebounds.dim() > 2 and ebounds.shape[:-2] != X.shape[:-2])
+        ):
             raise RuntimeError(
                 "Bounds must either be a single value or the same dimension as X"
             )
--- a/botorch/optim/utils.py
+++ b/botorch/optim/utils.py
@@ -147,19 +147,11 @@ def _expand_bounds(
     if bounds is not None:
         if not torch.is_tensor(bounds):
             bounds = torch.tensor(bounds)
-        if len(bounds.shape) == 0:
-            ebounds = bounds.expand(1, X.shape[-1])
-        elif len(bounds.shape) == 1:
-            ebounds = bounds.view(1, -1)
-        else:
-            ebounds = bounds
-        if (
-            ebounds.shape[-1] != X.shape[-1]
-            or ebounds.shape[-2] not in [1, X.shape[-2]]
-            or (ebounds.dim() > 2 and ebounds.shape[:-2] != X.shape[:-2])
-        ):
+        try:
+            ebounds = bounds.expand_as(X)
+        except RuntimeError:
             raise RuntimeError(
-                "Bounds must either be a single value or the same dimension as X"
+                "Bounds must be broadcastable to X!"
             )
         return ebounds.to(dtype=X.dtype, device=X.device)
     else:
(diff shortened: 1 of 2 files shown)


test/optim/test_utils.py
```diff
@@ -153,7 +169,7 @@ def test_get_extra_mll_args(self):
 
 class TestExpandBounds(BotorchTestCase):
     def test_expand_bounds(self):
-        X = torch.zeros(2, 3)
+        X = torch.zeros(4, 2, 3)
```
Balandat, line 172: we should probably test with and without explicit t-batch dimension

botorch/optim/utils.py
```diff
@@ -153,7 +153,11 @@ def _expand_bounds(
             ebounds = bounds.view(1, -1)
         else:
             ebounds = bounds
-        if ebounds.shape[1] != X.shape[-1]:
+        if (
+            ebounds.shape[-1] != X.shape[-1]
+            or ebounds.shape[-2] not in [1, X.shape[-2]]
+            or (ebounds.dim() > 2 and ebounds.shape[:-2] != X.shape[:-2])
```
Balandat, line 159: So this requires bounds to be defined with explicit t-batch dim. An alternative would be to actually fully support broadcasting - would that make sense her? I wrote some utilities for this a while ago that may be helpful here: https://github.com/cornellius-gp/gpytorch/blob/master/gpytorch/utils/broadcasting.py#L6

Balandat (commented):
Cool, thanks for the PR. 

What's the specific motivation for optimizing different elements w.r.t. different bounds? Is this to deal e.g. with different fidelities in look-ahead approaches?

test/optim/test_utils.py
saitcakmak: Just to clarify, do we want to test without explicit t-batch when `bounds` is <=2 dim, or even when `bounds` has a t-batch? If it's the second case, I am assuming this should raise an error if `bounds` has a t-batch and `X` doesn't.

botorch/optim/utils.py
saitcakmak: I looked into the behavior of `Tensor.max()` (or `min()`), and it is safe as long as `bounds` is broadcastable to `X`. It matches the expected behavior as well, so that should work here. I will look into a simpler condition for this.

saitcakmak:
Thanks for the quick review!

> What's the specific motivation for optimizing different elements w.r.t. different bounds?

I was implementing batch BnB, using `gen_candidates_scipy` to optimize relaxations. Naturally, each branch is optimized over a different solution space. This happened to be the only thing that raised an issue.

botorch/optim/utils.py
Balandat: You may be able to just use the broadcasting utils directly? Or I guess that doesn't cover the case when `bounds` has explicit t-batch dims but `X` doesn't?

Balandat:
> I was implementing batch BnB, using gen_candidates_scipy to optimize relaxations.

Cool, excited to see how this works.

saitcakmak:
I modified it to use `bounds.expand_as(X)`. This quite general and has the correct behavior. According to PyTorch documentation, `expand` does not create new memory, so I don't see any downside to this.

This does not handle the case when `bounds` has explicit t-batch but `X` doesn't. I verified that `_expand_bounds` is only used within `columnwise_clamp`, which is only used in `gen_candidates_*`. I don't see any use case where `bounds` would have t-batch and `X` wouldn't, so I think it is safe to ignore it. This will raise an error in that case.

Balandat (approved):
great, this lgtm

Balandat:
Thanks for contributing.

PR #643, opened 2020-12-23
Apply transform to constrain RBFKernel lengthscale in LCEMGP

neerajprad:
## Motivation

This adds a transform to `UniformPrior` in `LCEMGP` class to convert real values to constrained values in `[0, 2]`. Otherwise, with distribution validation enabled (which will be the default once https://github.com/pytorch/pytorch/pull/48743 is merged), this throws an error in `Uniform.log_prob` method. We were facing this issue in an internal FB test that uses botorch.

While this fixes the specific issue, my thinking is that this transform should ideally be applied automatically within `gpytorch.priors`. That said, my understanding of the codebase is very limited, so would love to hear others' thoughts on this.

Yes

## Test Plan

Turned `Distribution.set_default_validate_args(True)` and succesffully ran `test_contextual_multioutput.py: ContextualMultiOutputTest.testLCEMGP` with a fixed seed and increased `maxiter` (this test fails without the fix).

cc @Balandat

### DIFF
--- a/botorch/models/contextual_multioutput.py
+++ b/botorch/models/contextual_multioutput.py
@@ -86,8 +86,9 @@ class LCEMGP(MultiTaskGP):
         unif = Uniform(0.0, 2.0)
         self.task_covar_module = RBFKernel(
             ard_num_dims=n_embs,
-            lengthscale_prior=UniformPrior(unif.low, unif.high,
-                                           transform=biject_to(unif.support)),
+            lengthscale_prior=UniformPrior(
+                unif.low, unif.high, transform=biject_to(unif.support)
+            ),
         )
         self.to(train_X)
 
--- a/botorch/models/contextual_multioutput.py
+++ b/botorch/models/contextual_multioutput.py
@@ -13,9 +13,7 @@ from gpytorch.distributions.multivariate_normal import MultivariateNormal
 from gpytorch.kernels.rbf_kernel import RBFKernel
 from gpytorch.lazy import InterpolatedLazyTensor, LazyTensor
 from gpytorch.likelihoods.gaussian_likelihood import FixedNoiseGaussianLikelihood
-from gpytorch.priors.torch_priors import UniformPrior
 from torch import Tensor
-from torch.distributions import Uniform, biject_to
 from torch.nn import ModuleList
 
 
--- a/botorch/models/contextual_multioutput.py
+++ b/botorch/models/contextual_multioutput.py
@@ -8,6 +8,7 @@ from typing import List, Optional
 
 import torch
 from botorch.models.multitask import MultiTaskGP
+from gpytorch.constraints import Interval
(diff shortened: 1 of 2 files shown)


botorch/models/contextual_multioutput.py
```diff
@@ -80,8 +81,14 @@ def __init__(
...
         )
+        # TODO: Ideally, the transform should be applied automatically within
+        # gpytorch.priors
+        unif = Uniform(0.0, 2.0)
         self.task_covar_module = RBFKernel(
-            ard_num_dims=n_embs, lengthscale_prior=UniformPrior(0.0, 2.0)
+            ard_num_dims=n_embs,
+            lengthscale_prior=UniformPrior(
+                unif.low, unif.high, transform=biject_to(unif.support)
+            ),
```
Balandat, line 91: It seems problematic to have a prior here that makes the MLL non-differentiable (non-continuous even) as there is no restriction to the support of the optimizer. I would expect this to result in a bunch of problems (cc @qingfeng10, @bletham).

What we should instead do here is impose an `Interval` constraint that achieves this. Once we do this, then there will be no prior necessary (no prior = uniform). 

So this should be 
```
self.task_covar_module = RBFKernel(
    ard_num_dims=n_embs,
    lengthscale_constraint=Interval(0.0, 2.0),
)
```

where `Interval` is `gpytorch.constraints.Interval`. This would then use the default sigmoid transform. 

The alternative is to specify `transform=None` on the constraint and handle the bounds natively by the scipy optimizer. With L-BFGS-B this works generally better (o/w the optimizer runs off to +/- inf and doesn't really know what to do for lack of gradients). In `fit_gpytorch_torch` we handle this by explicit clamping (or "projecting onto the feasible set" if you want to be all fancy). We extract the bounds for this automatically here: https://github.com/pytorch/botorch/blob/dd17a4c19890e87460dbb50e14456c2f525aeeeb/botorch/optim/numpy_converter.py#L69-L74

```
self.task_covar_module = RBFKernel(
    ard_num_dims=n_embs,
    lengthscale_constraint=Interval(0.0, 2.0, transform=None, inital_value=1.0),
)
```
(the `initial_value` here instructs the optimizer to use 1.0 as the initial condition for MAP estimation).

Balandat (changes requested):

botorch/models/contextual_multioutput.py
neerajprad: > It seems problematic to have a prior here that makes the MLL non-differentiable (non-continuous even) as there is no restriction to the support of the optimizer.

Thanks for your detailed explanation, @Balandat. My understanding was that the `transform` in `UniformPrior` should be differentiable and will transform from the unconstrained real space (support of the optimizer) to the `Interval(0, 2)`, but you are right that we should directly use the `Interval` constraint here. I'll make the change based on your second suggestion here.

PR #646, opened 2020-12-24
Handle outcome_transform Standardize in HigherOrderGP

saitcakmak:
## Motivation

Using `outcome_transform=Standardize(...)` with `HigherOrderGP` leads to errors in `forward` pass due to shape mismatch. This raises a warning when `Standardize` is used and uses the corresponding `FlattenedStandardize` instead.

Yes.

## Test Plan

Added unit tests.


botorch/models/higher_order_gp.py
```diff
@@ -384,6 +385,19 @@ def __init__(
...
+                outcome_transform, FlattenedStandardize
+            ):
+                warnings.warn(
+                    "HigherOrderGP does not support the outcome_transform "
+                    "`Standardize`! Using `FlattenedStandardize` with `output_shape="
+                    f"{train_Y.shape[-2:]} instead.",
+                    RuntimeWarning,
+                )
+                outcome_transform = FlattenedStandardize(
+                    output_shape=train_Y.shape[-2:],
```
saitcakmak, line 398: Just to verify, `FlattenedStandardize` `output_shape` is supposed to be `q x d` dimensions? That's what I understood from existing tests. Adding docstrings to`FlattenedStandardize` may be helpful here (could add in this PR).

Balandat: So it's supposed to be whatever the output dimension of the training data is (which can be multi-dimensional). That is `train_Y` will be `batch_shape x n x output_shape`, where `output_shape` is of dim 1+ and could have an arbitrary number of dimensions (in practice this will be soon hit scalability limits of course) and `batch_shape` is at most one-dimensional (support for multi-level batched models is not implemented yet). To indicate whether a batch dim is present, use `first_dim_is_batch=True`.

This API is probably not ideal - for consistency I would suggest replacing `first_dim_is_batch` by a `num_output_dims` arg that then indicates `len(output_shape)` and thus makes things future-proof for additional batch dimensions should we ever need them. For now we can error out if the passed batch shapes are multi-dimensional. @wjmaddox  any thoughts here? 

Definitely agree on the need for better docstrings though/

@wjmaddox

saitcakmak: I updated it to infer the `output_shape` from `train_Y`. Also made the shapes of `train_X` and `train_Y` explicit in the docstring.

> for consistency I would suggest replacing first_dim_is_batch by a num_output_dims arg that then indicates len(output_shape)

`first_dim_is_batch` only comes into play in the constructor, so this should be easy to implement. I could make the change and raise an error (`NotImplementedError` ?) if the implied `batch_shape` is multi-dimensional. Does this sound appropriate?

Balandat: Yes that sounds like the right thing to do. Thanks.

wjmaddox: I think that a `num_output_dims` arg is probably the best bet because it ought to be pretty straightforward to allow for multiple batched output dims if we know how many output dimensions there are.

saitcakmak: It looks like `num_output_dims` is not necessary at all. Since `train_X` is `batch_shape x n x d` and `train_Y` is `batch_shape x n x output_shape`, we can just infer `num_output_dims = train_Y.dim() - train_X.dim() + 1`. 

Is this correct or is it possible to have `train_X: n x d` with `train_Y: batch_shape x n x output_shape`?

wjmaddox: I believe that situation (batched Y, non-batched X) occurs when fantasizing, so you do need `num_output_dims`.

saitcakmak: I didn't think of fantasizing but it looks like `get_fantasy_model` expands the inputs (X) with the batch shape as well. So, this doesn't happen there either. I guess the question is does HOGP support batched Y with non-batched X. The answer to that also seems to be no, since `_validate_tensor_args` (of `GPyTorchModel`) throws an error when I try this.

saitcakmak: It looks like inferring `num_output_dims = train_Y.dim() - train_X.dim() + 1` is safe. So, I removed `first_dim_is_batch` and replaced it with an inferred `num_output_dims`.

botorch/models/higher_order_gp.py
```diff
@@ -361,39 +362,60 @@ def __init__(
...
 
         if input_transform is not None:
             input_transform.to(train_X)
 
+        # infer the dimension of `output_shape`.
+        num_output_dims = train_Y.dim() - train_X.dim() + 1
+        if num_output_dims not in [train_Y.dim() - 1, train_Y.dim() - 2]:
+            raise NotImplementedError(
+                "HigherOrderGP currently only supports 1-dim `batch_shape`."
+            )
```
Balandat, line 389: why not just check that `len(train_X.shape[:-2])` is at most of length 1? seems less convoluted

botorch/models/higher_order_gp.py
```diff
@@ -361,39 +362,60 @@ def __init__(
...
+                    RuntimeWarning,
+                )
+                outcome_transform = FlattenedStandardize(
+                    output_shape=train_Y.shape[-num_output_dims:],
+                    batch_shape=train_Y.shape[: -1 - num_output_dims],
+                )
             train_Y, _ = outcome_transform(train_Y)
 
-        if first_dim_is_batch:
+        if num_output_dims == train_Y.dim() - 2:
```
Balandat, line 408: same here, lets infer `batch_shape = X.shape[:-2]` above and then use that.

botorch/models/higher_order_gp.py
```diff
@@ -361,39 +362,60 @@ def __init__(
...
+            # if using a batched model (with 1-dim batch)
```
Balandat, line 409: you can make the code below generic without requiring the if/else if you use negative indexing, e.g.
```
self._aug_batch_shape = train_Y.shape[:-num_output_dims - 1]
```

test/models/test_higher_order_gp.py
```diff
@@ -71,9 +90,19 @@ def test_posterior(self):
...
         train_y = randn(10, 4, 5, device=self.device)
+
+        # test handling of Standardize
+        with self.assertWarns(RuntimeWarning):
+            model = HigherOrderGP(
+                train_X=train_x, train_Y=train_y, outcome_transform=Standardize(m=5)
+            )
+            self.assertIsInstance(model.outcome_transform, FlattenedStandardize)
+            self.assertEqual(model.outcome_transform.output_shape, train_y.shape[1:])
+            self.assertEqual(model.outcome_transform.batch_shape, Size())
```
Balandat, line 101:
```suggestion
      self.assertIsInstance(model.outcome_transform, FlattenedStandardize)
      self.assertEqual(model.outcome_transform.output_shape, train_y.shape[1:])
      self.assertEqual(model.outcome_transform.batch_shape, Size())
```

Balandat (commented):
I think inferring it from the inputs this way makes sense. Some minor suggestions for streamlining this even more.

botorch/models/higher_order_gp.py
```diff
@@ -361,41 +362,55 @@ def __init__(
...
+                warnings.warn(
+                    "HigherOrderGP does not support the outcome_transform "
+                    "`Standardize`! Using `FlattenedStandardize` with `output_shape="
+                    f"{train_Y.shape[- num_output_dims:]} and batch_shape="
+                    f"{train_Y.shape[: -1 - num_output_dims]} instead.",
+                    RuntimeWarning,
+                )
+                outcome_transform = FlattenedStandardize(
+                    output_shape=train_Y.shape[-num_output_dims:],
+                    batch_shape=train_Y.shape[: -1 - num_output_dims],
```
Balandat, line 405: could just reuse the previously defined `batch_shape` here

Balandat (approved):

saitcakmak:
Brought this up to date with master. Changes appear to be compatible with #648

PR #651, opened 2020-12-27
Run CI tests on py 3.9

Balandat:
Also:
- use default conda version
- don't repeatedly upload sphinx coverage from multiple jobs
- explicitly test against py 3.7


Balandat:
Hmm interestingly this segfaults on py3.9 upon successfully completing unit tests. I can't reproduce this locally on my Mac though.

Issue #649, opened 2020-12-26
Segmentation fault on larger dataset using MaxPosteriorSampling


dong-tony:
## Issue description

This is part of the project I'm working on which I mentioned in Facebook/Ax#459. 

tldr i'm working on a chemistry problem where the search space is high dimensional (30+) and composed of discrete points (molecules). I'm still learning about this process so I'm experimenting with many approaches. One of the things I want to try is simply doing thompson sampling using `MaxPosteriorSampling` on my entire dataset (~49000 points), calculate the performance of the returned molecules (5 is the number i'm trying right now), and feed it back to the optimization loop manually. 

Everything runs fine when I was testing my code out on smaller subsets of the data, however when I tried running the full scale test I get a segmentation fault. I did a few more tests and seems like everything can still run up until ~42000 points, and after that it just throws the error again. Is this just an intrinsic limitation of the c code under the hood, or is there something I can do to solve this? Could it be that I'm running slightly older versions (see below)? I'm limited because i'm using a IBM cpu based computing cluster, which has the amount of memory I need to support this calculation, and seems like PyTorch 1.5 is the max they support currently.

Thanks and happy holidays!

## Code example

```
batch_size = 5

X = X_ini  # initial training set
Y = Y_ini
X_smiles = X_ini_smiles

train_Y = (Y - Y.mean()) / Y.std()

model = SingleTaskGP(X, train_Y)
mll = ExactMarginalLogLikelihood(model.likelihood, model)
fit_gpytorch_model(mll)

X_remain = torch.tensor(enu_features) # enu_features is a numpy array of shape (49000, 35)
thompson_sampling = MaxPosteriorSampling(model=model)
X_next = thompson_sampling(X_remain, num_samples=batch_size)
```

## System Info
- BoTorch Version 0.3.0
- GPyTorch Version 1.1.1
- PyTorch Version 1.5.0
- Linux RedHat 7.6

saitcakmak:
I can't replicate the segmentation fault, but trying the same thing leads me to out of memory error. Segmentation fault maybe your CPUs way of complaining that it doesn't have enough memory. My rough estimate is that you'd need ~400GB memory to jointly sample from `49000` points. Do you need to jointly sample from that many points? If so, do you need to use "exact" sampling, or would approximate methods be fine?

[...]

PR #654, opened 2020-12-29
Add large-scale Thompson sampling tutorial

dme65:
**Note**: This PR is work in progress so no need to review it yet, but I'm putting it up as it's relevant to #649.

This PR adds a tutorial for using large-scale Thompson sampling using Cholesky, random Fourier features (RFFs), and contour integral quadrature (CIQ). We can potentially add additional methods such as optimizing a sample path using a quasi-Newton optimizer, but that would require `RBFKernelGrad` to support incomplete gradient information.

We want to wait until #652 is landed before merging this so we can delete some of the custom TS code that I added to avoid using Cholesky in the posterior calls when using RFFs and CIQ. I'll also make some changes to be consistent with the `SingleTaskGP` priors, but for now I just added some simple bounds to the hyperparameters.


eytan:
Just out of curiosity, how difficult would it be to modify `RBFKernelGrad` to support incomplete gradient information?  Apologies for the naive question here, but why would we need to modify `RBFKernelGrad` to support doing gradient descent over a RFF model?

dme65:
> Just out of curiosity, how difficult would it be to modify `RBFKernelGrad` to support incomplete gradient information? Apologies for the naive question here, but why would we need to modify `RBFKernelGrad` to support doing gradient descent over a RFF model?

For the first question, pretty easy since it's just a matter of selecting a subset of the rows in the blocks with the first and second order derivative information. Would be nice to get that merged into GPyTorch at some point. That functionality will be useful if we want to optimize the sample path of an exact GP with a gradient-based optimizer, but it will also lead to a lot of numerical issues. It sounds like there are better alternatives here like the decoupled sampling you suggested offline.

Balandat:
The tutorial run fails for lack of pykeops. You will want to edit the `[tutorials]` option in `setup.py` (https://github.com/pytorch/botorch/blob/master/setup.py#L34), and then also edit the CI installation scripts for pip (https://github.com/pytorch/botorch/blob/master/scripts/install_via_pip.sh) and conda (https://github.com/pytorch/botorch/blob/master/scripts/install_via_conda.sh).

tutorials/thompson_sampling.ipynb
```diff
@@ -0,0 +1,396 @@
...
+    "    return hart6(unnormalize(x, hart6.bounds))"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 4,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "def get_initial_points(dim, n_pts, seeed=None):\n",
```
saitcakmak, line 90:
```suggestion
    "def get_initial_points(dim, n_pts, seed=None):\n",
```
just a minor typo

dme65:
> Can we do best value achieved rather than regret for consistency with the other tutorials?

We can do that! The main reason for using regret is to be able to put things on a log-scale, but we can skip that part.

> I wonder if it's worth demonstrating space complexities in addition to the time complexities (at least, if that isn't too hard to do). Seems that running out of memory is a recurring issue for users of TS.

Yeah, I will add some comments regarding memory usage. Anything with Cholesky is quadratic in the number of sample points while KeOps allows you to drop this to linear when used with CIQ.  RFFs will be linear without KeOps.

Balandat:
This is great. A few additional comments: 

1. Maybe you can switch out items 2. and 3. in the introduction? 2 talks about CIQ before it is introduced in 3.

2. Do you mind adding the timings / memory footprint to this similar as in https://github.com/saitcakmak/gp-sampling/blob/master/notebooks/ciq_thompson_sampling.ipynb ? 

3. In the TS code, you can clean up repeated code by using an ExitStack:
```
    # Thompson sample
    with ExitStack() as es:
        if sampler == "cholesky":
            es.enter_context(gpts.max_cholesky_size(float("inf")))
        elif sampler == "ciq":
            es.enter_context(gpts.fast_computations(covar_root_decomposition=True))
            es.enter_context(gpts.max_cholesky_size(0))
            es.enter_context(gpts.ciq_samples(True))
            es.enter_context(gpts.minres_tolerance(5e-3))
            es.enter_context(gpts.num_contour_quadrature(15))
        elif sampler == "rff":
            es.enter_context(gpts.fast_computations(covar_root_decomposition=True))
    
        thompson_sampling = MaxPosteriorSampling(model=model, replacement=False)
        X_next = thompson_sampling(X_cand, num_samples=batch_size)
```

dme65:
> This is great. A few additional comments:
> 
> 1. Maybe you can switch out items 2. and 3. in the introduction? 2 talks about CIQ before it is introduced in 3.
> 2. Do you mind adding the timings / memory footprint to this similar as in https://github.com/saitcakmak/gp-sampling/blob/master/notebooks/ciq_thompson_sampling.ipynb ?
> 3. In the TS code, you can clean up repeated code by using an ExitStack:
> 
> ```
>     # Thompson sample
>     with ExitStack() as es:
>         if sampler == "cholesky":
>             es.enter_context(gpts.max_cholesky_size(float("inf")))
>         elif sampler == "ciq":
>             es.enter_context(gpts.fast_computations(covar_root_decomposition=True))
>             es.enter_context(gpts.max_cholesky_size(0))
>             es.enter_context(gpts.ciq_samples(True))
>             es.enter_context(gpts.minres_tolerance(5e-3))
>             es.enter_context(gpts.num_contour_quadrature(15))
>         elif sampler == "rff":
>             es.enter_context(gpts.fast_computations(covar_root_decomposition=True))
>     
>         thompson_sampling = MaxPosteriorSampling(model=model, replacement=False)
>         X_next = thompson_sampling(X_cand, num_samples=batch_size)
> ```

Good comments! Let me know if the latest version looks good to you.

saitcakmak:
> Do you mind adding the timings / memory footprint to this similar as in https://github.com/saitcakmak/gp-sampling/blob/master/notebooks/ciq_thompson_sampling.ipynb ?

FYI, I'll be sharing an updated version of this notebook with you guys soon. No need to look into it yet.

Issue #667, opened 2021-01-20
Questions regarding contextual GPs / LCEMGP


saitcakmak:
I want to use the contextual GPs, in particular the `LCEMGP`, for a project. The existing documentation, i.e., the docstrings, do not present any examples, which makes it challenging to understand how to use these models with all their features. A tutorial notebook would be very helpful for me and others that may want to use these models in the future.

In particular, I'd like to understand how to use `context_cat_feature`, `context_emb_feature` and `embs_dim_list`, and the assumptions behind these. From the LCEMGP docstring, it seems like it is assumed that the contexts take values `0, ..., n_contexts-1`, though it is never made explicit. Below are some thoughts & questions I had while going through the code.

Reading the source code, it becomes somewhat clear that the contexts are assumed to be integers (since it converts them to `dtype=torch.long` when setting `self.all_tasks`). Though, this may lead to a bug when (carelessly) using fractional context values as it later assigns `context_cat_feature = all_tasks.unsqueeze(-1)` where `all_tasks` still has the same dtype as the input. A simple input validation could alert the user to this mistake. This is also an issue with `MultiTaskGP`.

Another thing I noticed is that `task_feature` is `int` (presumably due to it being a subclass of `MultiTaskGP`), which implies a single task / context variable, whereas the treatment of tasks within the constructor is written allowing for multiple task indices, e.g., the treatment of `self.emb_dims`. I guess this could purely be intended to support `context_cat_features` when `k >= 1`. In this case, if I have 2D contextual variables, let's say with 3 and 5 categories per dimension respectively, would I feed this into `LCEMGP` by first flattening the 2D contexts into a 1D variable with 15 categories, then use the `context_cat_features` to convert it back to 2D? Is it implemented this way to allow for subclassing `MultiTaskGP`, or am I missing something important here?

saitcakmak:
A far-fetched addition: Is there anything preventing me from using `LCEMGP` (or some variant of it) with purely categorical inputs? I know `LCEMGP` will run into issues when `x_basic` is a 0-element tensor. If I were to implement a variant that doesn't have this issue, is there any reason that I shouldn't use such a model?

Balandat:
I'll let @qingfeng10 answer your specific questions regarding the inputs and assumptions, but in the meantime it may be helpful to look at how the model is hooked up in Ax: https://github.com/facebook/Ax/blob/master/ax/models/torch/cbo_lcea.py

[...]

PR #669, opened 2021-01-20
Implements `fantasize` method for MultiTaskGPyTorchModel

saitcakmak:
## Motivation

As reported in #667, the `Model.fantasize` is not compatible with multi-task models. This over-writes the `fantasize` method in `MultiTaskGPyTorchModel`. 

The main differences in this implementation: i) the task indices are separated from `X` before evaluating the `posterior`, ii) the output of `posterior` is limited to only the task indices that are present in `X`, ii) in case of multiple output Indices in `X`, the output of `MTMVN` is post-processed to select the output corresponding to each task.

Note: I kept the `observation_noise: bool = True` as the default (to be consistent with `Model.fantasize(...)`), but `MultiTaskGPyTorchModel.posterior(...)` doesn't support `observation_noise` atm. So, the default behavior is to raise a `NotImplementedError` within the `posterior` call.

Yes.

## Test Plan

Added unit tests. 

Note: I added the tests in `test_multitask.py` rather than `test_gpytorch.py`, since i) `test_multitask.py` has utilities already implemented for generating MT models for testing, ii) `test_gpytorch.py` did not have any existing tests for `MultiTaskGPyTorchModel` (despite it being implemented in `gpytorch.py`). If this is not preferred, I can move them.

cc @qingfeng10

### DIFF
--- a/botorch/models/gpytorch.py
+++ b/botorch/models/gpytorch.py
@@ -20,6 +20,7 @@ from copy import deepcopy
 from typing import Any, Iterator, List, Optional, Tuple, Union
 
 import torch
+from botorch import settings
 from botorch.exceptions.errors import BotorchTensorDimensionError
 from botorch.exceptions.warnings import BotorchTensorDimensionWarning
 from botorch.models.model import Model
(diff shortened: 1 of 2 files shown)


saitcakmak:
I don't know how I missed this all along, but this PR treats `MultiTaskGP` as a single-output model. Might be because I was searching for a single-output contextual model with embeddings in the first place. I am closing this since the treatment is wrong.

PR #679, opened 2021-02-09
Add osy test problem

wjmaddox:
This PR adds the OSY multi-objective test problem (two objectives, six constraints, six dimensions).

## Motivation

It's a higher dimensional output multi-objective test problem that is useful for testing out MTGPs for constrained multi-objective bayes opt.

yes.

## Test Plan

unit tests. osy is added.

## Related PRs

N/A. Pulled out of https://github.com/wjmaddox/botorch.

### DIFF
--- a/botorch/test_functions/multi_objective.py
+++ b/botorch/test_functions/multi_objective.py
@@ -30,7 +30,7 @@ References
     A. Osyczka, S. Kundu. 1995. A new method to solve generalized multicriteria 
     optimization problems using the simple genetic algorithm. In Structural
     Optimization 10. 94–99.
-    
+
 .. [Tanabe2020]
     Ryoji Tanabe, Hisao Ishibuchi, An easy-to-use real-world multi-objective
     optimization problem suite, Applied Soft Computing,Volume 89, 2020.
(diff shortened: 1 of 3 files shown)


botorch/test_functions/multi_objective.py
```diff
@@ -601,3 +606,54 @@ def evaluate_slack_true(self, X: Tensor) -> Tensor:
...
+        (0.0, 6.0),
+        (1.0, 5.0),
+        (0.0, 10.0),
+    ]
+    _ref_point = [-75.0, 75.0]
+
+    def __init__(
+        self,
+        dim: int = 6,
+        num_objectives: int = 2,
```
Balandat, line 635: looks like this is always in dim 6 with 2 objectives? Let's remove these args then (these are for test functions where one can select the dimensionality).

Balandat: In fact, you can just omit the constructor altogether then since the superclass constructor will do all you need

Balandat (changes requested):

botorch/test_functions/multi_objective.py
wjmaddox: I originally built the constructor to have negate=True as the default option, but that was breaking the unit tests.

Balandat (approved):

sdaulton:
lgtm. Let's add this to botorch/test_functions/__init__.py

PR #680, opened 2021-02-09
Memory Fixes to Higher Order GP Model

wjmaddox:
Detaching the latent dimensions of the HOGP should reduce the memory overhead a bit as we don't need to keep the computation graphs around.

Also fixes a small bug where the full joint covariance was not picking up the computation graph; resolved by cloning the lazy tensor here.

## Motivation

Reduce memory overhead.

yes.

## Test Plan

unit tests.

## Related PRs

n/a.

### DIFF
--- a/botorch/models/higher_order_gp.py
+++ b/botorch/models/higher_order_gp.py
@@ -413,7 +413,9 @@ class HigherOrderGP(BatchedMultiOutputGPyTorchModel, ExactGP):
             self._is_custom_likelihood = True
 
         super().__init__(
-            train_X, train_Y.view(*self._aug_batch_shape, -1), likelihood=likelihood,
+            train_X, 
+            train_Y.view(*self._aug_batch_shape, -1), 
+            likelihood=likelihood,
         )
 
         if covar_modules is not None:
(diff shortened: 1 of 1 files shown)


botorch/models/higher_order_gp.py
```diff
@@ -650,7 +654,7 @@ def posterior(
                 pred_variance = self.make_posterior_variances(joint_covar)
 
                 full_covar = KroneckerProductLazyTensor(
-                    full_covar, *joint_covar.lazy_tensors[1:]
+                    full_covar, *[x.detach() for x in joint_covar.lazy_tensors[1:]]
```
Balandat, line 657: I guess this will preclude computing gradients of quantities computed on the posterior w.r.t. the latents? I think that's fine since we usually don't do that, but we should probably add a comment here just to be safe.

botorch/models/higher_order_gp.py
```diff
@@ -659,7 +663,7 @@ def posterior(
                 covar = cm(param)
                 if covar.batch_shape != batch_shape:
                     covar = BatchRepeatLazyTensor(covar, batch_shape)
-                joint_covar_list.append(covar)
+                joint_covar_list.append(covar.detach())
```
Balandat, line 666: why not move this up and do `covar = cm(param).detach()`? That would avoid calling this on a `LazyTensor`

botorch/models/higher_order_gp.py
wjmaddox: I think it's unlikely we'd need to do that for now, but the updated comment makes it clear.

Balandat (approved):

PR #684, opened 2021-02-16
Github Actions deploy workflow

Balandat:
Adds automated testing and deployment of python and conda packages as well as the stable website build.
This workflow is triggered upon creating a new tag.


scripts/build_and_verify_py_packages.sh
```diff
@@ -18,7 +18,7 @@ path="${cur_dir}/dist/botorch-${version}"
 # install wheel and verify import works and version is correct
 pip uninstall -y botorch
 pip install "${path}-py3-none-any.whl"
-wheel_version=$(python -c "import botorch; print(botorch.__version__)")
+wheel_version=$(python -c "import botorch; print(botorch.__version__)" | tail -n 1)
```
Balandat, line 21: these tails are necessary b/c pykeops thinks it's a good idea to print upon import

PR #713, opened 2021-02-20
Update large-scale TS notebook

dme65:
- Re-ran notebook after #701 which caused CIQ to not be used
- Added Lanczos as a baseline
- Added a smoke test setting for the notebook


Balandat (approved):
thanks for updating.

Issue #626, opened 2020-12-11
Error when using FixedFeatureAcquisitionFunction and project with qMultifidelityKnowledgeGradient


r-ashwin:
Related tp #625 but opening a new issue since that one was closed.

versions `torch/gpytorch/botorch 1.7.0-cpu/1.3.0/0.3.3`

Reproducible example

```
import math
import torch

import botorch
botorch.__version__
from botorch.fit import fit_gpytorch_model
from botorch.models import SingleTaskGP
from botorch.utils import standardize
from gpytorch.mlls import ExactMarginalLogLikelihood

d = 3
bounds  = torch.stack([torch.zeros(d), torch.ones(d)])

train_X = bounds[0] + (bounds[1] - bounds[0]) * torch.rand(15, d)
train_Y = torch.sin(2 * math.pi * train_X[:, [0]]) 
# train_Y = standardize(train_Y + 0.0 * torch.randn_like(train_Y))

model   = SingleTaskGP(train_X, train_Y)
mll     = ExactMarginalLogLikelihood(model.likelihood, model)
fit_gpytorch_model(mll);

def project(x):
    return torch.cat([x[..., :-1], torch.full_like(x[..., -1:], .5)], dim=-1)

from botorch.acquisition import qKnowledgeGradient, qMultiFidelityKnowledgeGradient
torch.manual_seed(123)
num_fantasies = 64
qKG           = qMultiFidelityKnowledgeGradient(model, num_fantasies=num_fantasies, project=project)

from botorch.optim import optimize_acqf
from botorch.utils.sampling import manual_seed

with manual_seed(1234):
    candidates, acq_value = optimize_acqf(
        acq_function=qKG, 
        bounds=bounds,
        q=1,
        num_restarts=5,
        raw_samples=25,
        fixed_features={d-1:1.}
    )
```
```
from botorch.acquisition import FixedFeatureAcquisitionFunction
columns = [2]
values  = [2.]
qKG_FF = FixedFeatureAcquisitionFunction(qKG, 3, columns, values)
```

```
from botorch.optim import optimize_acqf
from botorch.utils.sampling import manual_seed

ff_bounds = bounds[...,:-1].reshape(2,-1)
with manual_seed(1234):
    candidates, acq_value = optimize_acqf(
        acq_function=qKG_FF, 
        bounds=ff_bounds,
        q=1,
        num_restarts=5,
        raw_samples=25,
    )
```

```
---------------------------------------------------------------------------
[...]

PR #718, opened 2021-02-23
Introduce base_sample_shape property to Posterior objects

Balandat:
This allows base sample shapes that are different from the posterior's `event_shape`. This is relevant e.g if for GPyTorch MVN posteriors the covariance is singular, or if a low-rank approximation of the root decomposition is used. 
See the corresponding GPyTorch PR here: https://github.com/cornellius-gp/gpytorch/pull/1502

Also does some cleanup and moves the `HigherOrderGPPosterior` into the `posteriors` module.

### DIFF
--- a/botorch/acquisition/utils.py
+++ b/botorch/acquisition/utils.py
@@ -269,12 +269,12 @@ def prune_inferior_points(
     with torch.no_grad():
         posterior = model.posterior(X=X)
     if sampler is None:
-        if posterior.event_shape.numel() > SobolEngine.MAXDIM:
+        if posterior.base_sample_shape.numel() > SobolEngine.MAXDIM:
             if settings.debug.on():
                 warnings.warn(
-                    f"Sample dimension q*m={posterior.event_shape.numel()} exceeding "
-                    f"Sobol max dimension ({SobolEngine.MAXDIM}). Using iid samples "
-                    "instead.",
+                    f"Sample dimension q*m={posterior.base_sample_shape.numel()} "
+                    f"exceeding Sobol max dimension ({SobolEngine.MAXDIM}). "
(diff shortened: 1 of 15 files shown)


Balandat:
One thing that is missing is properly handling the situation if the size of the base samples changes in the context of SAA. Specifically, we may start optimizing for some initial `X` at which the covariance is singular, and the root decomposition is low rank so that `base_sample_shape < event_shape`. As we take a gradient step the covariance may gain a rank, and subsequently `base_sample_shape` may grow. 

It seems that the proper way to handle this is to have the `base_samples` only ever grow in size but never shrink. That's straightforward with IID samples but not so much with QMC samples.

saitcakmak:
> It seems that the proper way to handle this is to have the base_samples only ever grow in size but never shrink. That's straightforward with IID samples but not so much with QMC samples.

Would passing in full `event_shape` Sobol samples and subsetting it dimension-wise to get to the `base_sample_shape` be an appropriate solution here? As discussed in https://github.com/pytorch/botorch/issues/626#issuecomment-744055636 taking a subset of dimensions of Sobol samples preserves the digital net property. So, it should be fine if this was your original concern.

Balandat:
> Would passing in full event_shape Sobol samples and subsetting it dimension-wise to get to the base_sample_shape be an appropriate solution here? As discussed in #626 (comment) taking a subset of dimensions of Sobol samples preserves the digital net property. So, it should be fine if this was your original concern.

So it would but that would require dealing with both `event_shape` and `base_sample_shape`, and ideally we could keep this interface as simple as possible. Also `base_sample_shape` may be larger than `event_shape`, e.g. when using Matheron's rule (which we use when sampling from HOGPs)

PR #728, opened 2021-03-03
Fix HigherOrderGP dtype bug

saitcakmak:
## Motivation

HigherOrderGP would raise `RuntimeError: expected scalar type Float but found Double` when used with double. This fixes the bug by setting the model `to(train_X)` (instead of `train_X.device`).

Yes.

## Test Plan

Updated unit tests. Updated tests would fail without the change.


Balandat:
Thanks a lot for putting this up and for expanding the test coverage!

Balandat (approved):

PR #735, opened 2021-03-07
Use torch.linalg module from PyTorch 1.8

Balandat:
Also bumps version requirement to PyTorch>=1.8


botorch/utils/sampling.py
```diff
@@ -379,7 +379,7 @@ def __init__(
 
         if equality_constraints is not None:
             self.C, self.d = equality_constraints
-            U, S, V = torch.svd(self.C, some=False)
+            U, S, V = torch.linalg.svd(self.C)
```
IvanYashchuk, line 382: `torch.linalg.svd` is different from `torch.svd` to be NumPy compatible. The third returned tensor is actual now is the conjugate transpose of `V`.
```suggestion
            U, S, Vh = torch.linalg.svd(self.C)
            V = Vh.conj().transpose(-2, -1)
```

botorch/sampling/qmc.py
```diff
@@ -136,7 +136,7 @@ def __init__(
         )
         # compute Cholesky decomp; if it fails, do the eigendecomposition
         try:
-            self._corr_matrix = torch.cholesky(cov).transpose(-1, -2)
+            self._corr_matrix = torch.linalg.cholesky(cov).transpose(-1, -2)
         except RuntimeError:
             eigval, eigvec = torch.symeig(cov, eigenvectors=True)
```
IvanYashchuk, line 141: symeig can be replaced with linalg.eigh
```suggestion
            eigval, eigvec = torch.linalg.eigh(cov)
```

IvanYashchuk (commented):
Hey @Balandat it's great to see the transition to the new linalg module! I left a suggestion on using eigh and svd.

Balandat:
Hmm @IvanYashchuk looks like `symeig` and `linalg.eigh` use different algos under the hood? Or at least `linalg.eigh`  is not clamping the evals? This is messing with our tests...

```
A = torch.tensor(
    [[1, 0, 1], [0, 1, 1], [1, 1, 2]],
    dtype=torch.float,
)
evals_sym, evecs_sym = torch.symeig(A, eigenvectors=True)
evals_eigh, evecs_eigh = torch.linalg.eigh(A)
```

`evals_sym = tensor([0., 1., 3.])`
`evals_eigh = tensor([-2.2915e-07,  1.0000e+00,  3.0000e+00])`

IvanYashchuk:
Yes, they call different LAPACK routines: `torch.linalg.eigh` uses the one based on divide and conquer algorithm (`syevd`) same as NumPy and CUDA path while `torch.symeig` uses slower and the most robust QR based (`syev`). We should probably add `driver=` option to choose the underlying algorithm as SciPy for example supports.

Here the discrepancy is caused by upper/uplo argument the defaults are opposite:
```python
In [13]: torch.symeig(A, eigenvectors=True, upper=False)
Out[13]: 
torch.return_types.symeig(
eigenvalues=tensor([-2.2915e-07,  1.0000e+00,  3.0000e+00]),
eigenvectors=tensor([[ 5.7735e-01,  7.0711e-01, -4.0825e-01],
        [ 5.7735e-01, -7.0711e-01, -4.0825e-01],
        [-5.7735e-01,  6.7987e-08, -8.1650e-01]]))
In [14]: torch.linalg.eigh(A, "U")
Out[14]: 
torch.return_types.linalg_eigh(
eigenvalues=tensor([0., 1., 3.]),
eigenvectors=tensor([[ 0.5774,  0.7071,  0.4082],
        [ 0.5774, -0.7071,  0.4082],
        [-0.5774,  0.0000,  0.8165]]))
```

Balandat:
> We should probably add driver= option to choose the underlying algorithm as SciPy for example supports.

Sounds like a good idea.

Interesting to see this discrepancy coming from `upper`...

PR #737, opened 2021-03-08
Added GIBBON Notebook

henrymoss:
Hi all,

I have written a short notebook that demonstrates the use of GIBBON within BoTorch.  I have focused primarily on nice visualisations and a direct comparison with your batch MES.

Let me know if you want me to add/change anything.

Henry


Balandat (commented):
This looks great, thanks for putting this up. 

1. Would you mind selecting the parameters that strongly affect the computation time (like the size of the candidate set) based on the `SMOKE_TEST` env var handling as we have in the other notebooks (e.g. https://github.com/pytorch/botorch/blob/master/tutorials/constrained_multi_objective_bo.ipynb)? This will allow us to run this tutorial as part of every CI run to make sure it doesn't break b/c of any code changes. The value should be tiny (e.g. 4 points) so that the CI run can just check everything runs properly. 
2. Could you please add the tutorial to our list of tutorials (similar as for [the MES one](https://github.com/pytorch/botorch/blob/master/website/tutorials.json#L69-L72)) so that it will automatically be built as part of the website? This should have been caught by CI, but the tutorial parsing was missing (adding this in #738)
3. I also left a few formatting nits inline.

henrymoss:
Made the requested changes, including formatting, SMOKE and adding to the list of tutorials.

Balandat:
Many thanks for making the changes!

PR #741, opened 2021-03-11
DelaunayPolytopeSampler

Balandat:
Introduces a new `DelaunayPolytopeSampler` that is based on tesselating polytope with simplexes obtained via Delaunay triangulation. This allows extremely efficient uniform sampling after spending an initial upfront cost for performing the vertex enumeration. This is done using scipy (and QHull under the hood).

To re-use as much existing code as possible, this makes `PolytopeSampler` an abstract base class, and implements the previous `PolytopeSampler` as `HitAndRunPolytopeSampler`.

Basic example nb/illustration:
[delaunay_sampler.ipynb.txt](https://github.com/pytorch/botorch/files/6121119/delaunay_sampler.ipynb.txt)

### DIFF
--- a/botorch/utils/sampling.py
+++ b/botorch/utils/sampling.py
@@ -6,11 +6,18 @@
 
 r"""
 Utilities for MC and qMC sampling.
+
+References
+
+.. [Trikalinos2014polytope]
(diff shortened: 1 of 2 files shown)


Balandat:
@jduerholt you might like this one

PR #757, opened 2021-03-31
update test functions to use negative indexing

saitcakmak:
## Motivation

With the existing implementation, some test functions do not support evaluations with `X` having multi-dim `batch_shape`, e.g., `X.shape = [2, 3, d]`. This PR updates the implementation of such test functions to use negative indexing, thus supporting multi-dim `batch_shape`.

Yes

## Test Plan

Extended test coverage by adding `batch_shape = [2, 3]` in `BaseTestProblemBaseTestCase.test_forward()`. Made sure that the updated implementation agrees with the function definitions in https://www.sfu.ca/~ssurjano/optimization.html.


Balandat (approved):
great, thanks for this PR!

Issue #765, opened 2021-04-09
[Bug] gen_candidates_* fails with qKnowledgeGradient


wjmaddox:
# 🐛 Bug

There's a second backwards error that occurs when trying to use `gen_candidates_torch` with `qKnowledgeGradient`. I'm pretty confident that something isn't getting destroyed in the fantasy model after we evaluate it, but don't know what it is. `gen_candidates_torch` does work with other MC acquisition functions like `qExpectedImprovement`. 

## To reproduce

** Code snippet to reproduce **
```python
import torch

from botorch.models import SingleTaskGP
from botorch.generation.gen import gen_candidates_torch
from botorch.test_functions import Beale
from botorch.acquisition import qKnowledgeGradient, qExpectedImprovement
from botorch.models.transforms import Standardize
from botorch.optim.fit import fit_gpytorch_torch

from gpytorch.mlls import ExactMarginalLogLikelihood

torch.random.manual_seed(20)

neg_hartmann6 = Beale(negate=True)

bounds = neg_hartmann6.bounds

dtype = torch.double # if dtype=="double" else torch.float
device = torch.device("cuda:0" if torch.cuda.is_available() else "cpu")

def generate_initial_data(n, fn, NOISE_SE, device, dtype):
    # generate training data
    train_x = torch.rand(n, fn.dim, device=device, dtype=dtype)
    exact_obj = fn(train_x).unsqueeze(-1)  # add output dimension
    train_obj = exact_obj + NOISE_SE * torch.randn_like(exact_obj)

    train_obj = (train_obj - train_obj.mean() ) / train_obj.std()
    best_observed_value = train_obj.max().item()
    return train_x, train_obj, best_observed_value

# call helper functions to generate initial training data and initialize model
train_x_ei, train_obj_ei, best_observed_value_ei = generate_initial_data(
        n=50, fn=neg_hartmann6, NOISE_SE=0.2, device=device, dtype=dtype
)

model = SingleTaskGP(train_x_ei, train_obj_ei,)

bounds  = bounds.to(device)

xvals = torch.rand(5, 67, 2, device = device, dtype=dtype)
xvals =  bounds[0] + (bounds[1] - bounds[0]) * xvals

kg_exact = qKnowledgeGradient(exact, current_value=best_observed_value_ei, num_fantasies=64)
# the ei equivalent works
# kg_exact = qExpectedImprovement(exact, best_f=best_observed_value_ei)

gen_candidates_torch(xvals.detach(), kg_exact)
[...]

PR #766, opened 2021-04-11
Backprop wrt candidates in gen_candidates_torch

wjmaddox:
## Motivation

Bugfix for #765. The current error was occurring when trying to optimize qKnowledgeGradient with gen_candidates_torch. The fix is to backprop only wrt candidates like what occurs in `gen_candidates_scipy`. 

Yes.

## Test Plan

Not sure how to setup a unit test here.

## Related PRs

N/A.


saitcakmak:
> Not sure how to setup a unit test here.

In #765 the bug was in optimizing qKG. The current test suite only optimizes qEI (thus missing the issue). So, adding qKG as an additional test case may be an option

botorch/generation/gen.py
```diff
@@ -219,11 +219,12 @@ def gen_candidates_torch(
...
             print("Iter: {} - Value: {:.3f}".format(i, -(loss.item())))
         loss_trajectory.append(loss.item())
-        param_trajectory["candidates"].append(candidates.clone())
+        param_trajectory["candidates"].append(candidates.clone().detach())
 
         def closure():
             bayes_optimizer.zero_grad()
-            loss.backward()
+            output_grad = torch.autograd.grad(loss, candidates)[0]
+            candidates.grad = output_grad
```
Balandat, line 227: hmm there seems to be some inconsistency here between the unconstrained candidates, the clamped candidates, and those were `fixed_features` is applied. Specifically, `bayes_optimizer` uses `clamped_candidates` as its parameter (l. 207), but here we're taking the gradient w.r.t. `candidates`. There also seems to be some inconsistency below where we compute `clamped_candidates` but return `candidates` (and evaluate the acquisition function on those). 

We probably want to combine the clamping and the "fixing" of features into a single wrapper so we can always work with a single `projected_candidates` tensor to avoid this kind of confusion.

Balandat:
Is this issue addressed by #839? If not maybe @vishwakftw can grab the relevant changes and incorporate them?

wjmaddox:
I don't know, probably the inconsistency you pointed out above is. Will checkout #839 now and see if that resolves the issue.

wjmaddox:
Re-running the command in #765 on top of #839 still fails so this should probably stay open. Maybe @vishwakftw can resolve the second gen conversation here since I don't have a great idea of what should be fixed / transformed?

botorch/generation/gen.py
```diff
@@ -288,19 +286,18 @@ def gen_candidates_torch(
...
+            output_grad = torch.autograd.grad(loss, clamped_candidates)[0]
+            clamped_candidates.grad = output_grad
             return loss
 
         bayes_optimizer.step(closure)
         with torch.no_grad():
             clamped_candidates = columnwise_clamp(
                 X=clamped_candidates, lower=lower_bounds, upper=upper_bounds
-            )
+            ).requires_grad_(True)
```
Balandat, line 300: Hmm I've never thought about what happens when you set `requires_grad` within a `no_grad` context. Will this work properly?

test/generation/test_gen.py
```diff
@@ -72,15 +72,28 @@ def test_gen_candidates(self, gen_candidates=gen_candidates_scipy, options=None)
...
+
+                candidates, _ = gen_candidates(
+                    initial_conditions=ics,
+                    acquisition_function=acqf,
+                    lower_bounds=0,
+                    upper_bounds=1,
+                    options=options or {},
+                )
+                if isinstance(acqf, qKnowledgeGradient):
+                    candidates = candidates[0]
```
Balandat, line 94:
```suggestion
                if isinstance(acqf, qKnowledgeGradient):
                    candidates = acqf.extract_candidates(candidates)
```

botorch/generation/gen.py
```diff
@@ -288,19 +286,18 @@ def gen_candidates_torch(
         loss = -acquisition_function(clamped_candidates).sum()
         if verbose:
             print("Iter: {} - Value: {:.3f}".format(i, -(loss.item())))
-        loss_trajectory.append(loss.item())
-        param_trajectory["candidates"].append(clamped_candidates.clone())
```
Balandat, line 292: I guess this code has been dead for a while then?

test/generation/test_gen.py
```diff
@@ -96,18 +109,30 @@ def test_gen_candidates_with_none_fixed_features(
...
+                candidates, _ = gen_candidates(
+                    initial_conditions=ics,
+                    acquisition_function=acqf,
+                    lower_bounds=0,
+                    upper_bounds=1,
+                    fixed_features={1: None},
+                    options=options or {},
+                )
+                if isinstance(acqf, qKnowledgeGradient):
+                    candidates = candidates[0]
```
Balandat, line 132:
```suggestion
                    candidates = acqf.extract_candidates(candidates)
```

test/generation/test_gen.py
```diff
@@ -121,18 +146,32 @@ def test_gen_candidates_with_fixed_features(
...
+                    initial_conditions=ics,
+                    acquisition_function=acqf,
+                    lower_bounds=0,
+                    upper_bounds=1,
+                    fixed_features={1: 0.25},
+                    options=options,
+                )
+
+                if isinstance(acqf, qKnowledgeGradient):
+                    candidates = candidates[0]
```
Balandat, line 170:
```suggestion
                    candidates = acqf.extract_candidates(candidates)
```

Balandat (approved):

botorch/generation/gen.py
wjmaddox: Yes.

botorch/generation/gen.py
wjmaddox: Seems to be since it's not actually a differentiable operation. Wasn't passing the unit tests before.

PR #783, opened 2021-05-06
Implement construct_inputs for acquisition functions

Balandat:
Summary: Implement the `construct_inputs` method from D28021531 for most acquisition funcitons

Differential Revision: D28170202

### DIFF
--- a/botorch/acquisition/acquisition.py
+++ b/botorch/acquisition/acquisition.py
@@ -12,10 +12,12 @@ from __future__ import annotations
 
 import warnings
 from abc import ABC, abstractmethod
-from typing import Optional
+from typing import Any, Dict, Optional
 
+from botorch.acquisition.objective import AcquisitionObjective
 from botorch.exceptions import BotorchWarning
 from botorch.models.model import Model
+from botorch.utils.containers import TrainingData
 from torch import Tensor
 from torch.nn import Module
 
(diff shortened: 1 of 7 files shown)


botorch/acquisition/acquisition.py
```diff
@@ -64,6 +66,37 @@ def forward(self, X: Tensor) -> Tensor:
...
+        **kwargs: Any
+    ) -> Dict[str, Any]:
+        r"""Construct kwargs for the `AcquisitionFunction` constructor.
+
+        Args:
+            model: The model to be used in the acquisition function.
+            training_data: A TrainingData object contraining the model's
+                training data. Used e.g. to extract inputs such as `best_f`
+                for expected improvement acquisition functions.
+            objective: The objective to in the acquisition function.
```
saitcakmak, line 84: Nice feature! Just two typos - "contraining" & "the objective to in". These repeat in all `construct_inputs` docstrings.

saitcakmak:
I was also wondering if the `training_data` is needed as an additional argument. We could easily extract this from the model, and untransform it to get the original data where needed. I was just curious if there's a good reason for not doing it this way that I am missing

botorch/acquisition/analytic.py
```diff
@@ -120,6 +106,53 @@ def __init__(
...
+        Returns:
+            A dict mapping kwarg names of the constructor to values.
+
+        Example:
+            >>> kwargs = ExpectedImprovement.construct_inputs(
+                    model, training_data,
+                )
+            >>> EI = ExpectedImprovement(**kwargs)
+        """
+        best_f = training_data.Y.max(-2).values.squeeze(-1)
```
saitcakmak, line 134: If using a multi-output model with a `ScalarizedObjective`, I think we need to apply the same scalarization here before taking the max.

Balandat:
Superseded by #788

PR #801, opened 2021-05-20
Make `raw_samples` optional.

vishwakftw:
If `batch_initial_conditions` is passed, there is no need to set `raw_samples` as this argument is only used for the underlying initialization strategy when `batch_initial_condition` is `None`.

Test Plan:

Expect current tests to pass without issue.

### DIFF
--- a/botorch/optim/optimize.py
+++ b/botorch/optim/optimize.py
@@ -33,7 +33,7 @@ def optimize_acqf(
     bounds: Tensor,
     q: int,
     num_restarts: int,
-    raw_samples: int,
+    raw_samples: Optional[int] = None,
     options: Optional[Dict[str, Union[bool, float, int, str]]] = None,
     inequality_constraints: Optional[List[Tuple[Tensor, Tensor, float]]] = None,
     equality_constraints: Optional[List[Tuple[Tensor, Tensor, float]]] = None,
@@ -52,7 +52,8 @@ def optimize_acqf(
         q: The number of candidates.
         num_restarts: The number of starting points for multistart acquisition
             function optimization.
-        raw_samples: The number of samples for initialization.
+        raw_samples: The number of samples for initialization. This is required
+            if `batch_initial_conditions` is not specified.
         options: Options for candidate generation.
         inequality constraints: A list of tuples (indices, coefficients, rhs),
             with each tuple encoding an inequality constraint of the form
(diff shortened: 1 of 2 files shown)


botorch/optim/optimize.py
```diff
@@ -33,7 +33,7 @@ def optimize_acqf(
     bounds: Tensor,
     q: int,
     num_restarts: int,
-    raw_samples: int,
+    raw_samples: Optional[int] = None,
```
saitcakmak, line 36: Shouldn't there be a default value or an error message when `raw_samples` is `None` and `batch_initial_conditions` is not specified? Otherwise, this will lead to a `TypeError` in `gen_batch_initial_conditions`, which is not super user friendly.

vishwakftw: I think an error message is more appropriate here.

botorch/optim/optimize.py
```diff
@@ -152,6 +153,11 @@ def optimize_acqf(
         return X, acq_value
 
     if batch_initial_conditions is None:
+        if raw_samples is None:
+            raise ValueError(
+                "Expected raw_samples to be set when batch_initial_conditions is None"
```
Balandat, line 158:
```suggestion
                "Must specify `raw_samples` when `batch_initial_conditions` is `None`"
```

botorch/optim/optimize.py
```diff
@@ -342,7 +349,8 @@ def optimize_acqf_list(
         bounds: A `2 x d` tensor of lower and upper bounds for each column of `X`.
         num_restarts:  Number of starting points for multistart acquisition
             function optimization.
-        raw_samples: Number of samples for initialization.
+        raw_samples: Number of samples for initialization. This is required
+            if `batch_initial_conditions` isn't set.
```
Balandat, line 353:
```suggestion
            if `batch_initial_conditions` is not specified.
```

botorch/optim/optimize.py
```diff
@@ -419,7 +427,8 @@ def optimize_acqf_mixed(
         q: The number of candidates.
         num_restarts:  Number of starting points for multistart acquisition
             function optimization.
-        raw_samples: Number of samples for initialization.
+        raw_samples: Number of samples for initialization. This is required
+            if `batch_initial_conditions` isn't set.
```
Balandat, line 431:
```suggestion
            if `batch_initial_conditions` is not specified.
```

botorch/optim/optimize.py
```diff
@@ -152,6 +153,11 @@ def optimize_acqf(
         return X, acq_value
 
     if batch_initial_conditions is None:
+        if raw_samples is None:
+            raise ValueError(
```
Balandat, line 157: this isn't covered by the tests, can you please add a `with self.assertRaises` check for this?

Balandat (commented):
some nits, test coverage

Balandat (approved):

PR #681, opened 2021-02-09
Add support for transforms in MTGPs

wjmaddox:
This PR adds support for input and outcome transforms for MTGPs as currently implemented in BoTorch.

## Motivation

Helps improve bayesian optimization loops using MTGPs including contextual ones by allowing these types of transforms.

### Have you read the [Contributing Guidelines on pull requests](https://github.com/pytorch/botorch/blob/master/CONTRIBUTING.md#pull-requests)?

yes.

## Test Plan

Current unit tests; need to dive into these.

## Related PRs

See #679 as well.

PR #819, opened 2021-06-09
Modify input transforms to support one-to-many transforms

saitcakmak:
Summary:
This modifies the input transform to support one-to-many transforms that are needed for robust BO.

Changes:
- `transform_inputs` is applied in `model.forward` if the model is in `train` mode, otherwise it is applied in the `posterior` call. Moving it to `posterior` is necessary to make one-to-many transforms work, applying in `model.forward` is necessary to make sure certain transforms get applied at training.
- If the transform has `transform_on_train=True`, at the end of model fit using `fit_gpytorch_model`, `model.train_inputs` is updated with the transformed version. This makes sure that the `train_inputs` are transformed at both training and eval if and only if `transform_on_train=True`.
- With this change, `transform_on_preprocess` attribute becomes redundant and is removed. 
- Implements a `fantasize` flag and a `transform_on_fantasize` attribute to allow for turning off certain one-to-many transforms during `model.fantasize` call. Example use case: In the risk measure setting, we have a model trained on `(x, w)` inputs, for which we want to draw posterior samples of the risk measure (computed over w's) for a given x. This is achieved by using an input transform that appends w's to the given x, evaluating `posterior` on the resulting `(x, w)` pairs, and applying the risk measure (as an objective) to the resulting samples. So, in this case we would have the input transform taking in `x` and giving out a bunch of `(x, w)`s. However, when we want to fantasize on this model, we would call `fantasize` with an `(x, w)`, which if transformed would turn to `(x, w, w)` - not good. `transform_on_fantasize=False` along with the `fantasize` flag makes sure this doesn't happen.

Note: This does not touch `PairwiseGP` model. `PairwiseGP` implements its own `posterior` , uses input transforms manually on train and test inputs, and should not be affected by these changes.

Differential Revision: D28944189

### DIFF
--- a/botorch/fit.py
+++ b/botorch/fit.py
@@ -128,6 +128,7 @@ def fit_gpytorch_model(
         for w in ws:
             has_optwarning |= issubclass(w.category, OptimizationWarning)
             warnings.warn(w.message, w.category)
+        # TODO: this counts hitting `maxiter` as an optimization failure!
         if not has_optwarning:
             _set_transformed_inputs(mll=mll)
             mll.eval()
@@ -136,6 +137,7 @@ def fit_gpytorch_model(
         logging.log(logging.DEBUG, f"Fitting failed on try {retry}.")
 
     warnings.warn("Fitting failed on all retries.", OptimizationWarning)
+    _set_transformed_inputs(mll=mll)
     return mll.eval()
 
 
--- a/botorch/models/gp_regression.py
+++ b/botorch/models/gp_regression.py
@@ -145,7 +145,8 @@ class SingleTaskGP(BatchedMultiOutputGPyTorchModel, ExactGP):
         self.to(train_X)
 
     def forward(self, x: Tensor) -> MultivariateNormal:
-        x = self.transform_inputs(x)
+        if self.training:
+            x = self.transform_inputs(x)
         mean_x = self.mean_module(x)
         covar_x = self.covar_module(x)
         return MultivariateNormal(mean_x, covar_x)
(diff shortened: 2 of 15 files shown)


saitcakmak:
Open question: This currently applies the input transforms at two places, in `model.forward(...)` at the training time, and in `posterior(...)` at `eval`. Applying them at `posterior` call is necessary to avoid issues in `ExactGP.__call__` when the transform changes the input shape. However, it is not a good design to have them be applied in two separate places, so we need a better design.

Proposal: Implement a `Model.__call__` method, apply input transforms in there then call `super().__call__(...)`. 
Issues: This works with `ExactGP` models (since inheritance favors `Model` over `ExactGP`) but not with `ApproximateGP`. The inheritance there is typically like `ExampleVariationalGP(ApproximateGP, GPyTorchModel)` with `ApproximateGP.__call__` being favored over `Model.__call__`.

cc @wjmaddox @Balandat @sdaulton

Balandat:
@wjmaddox you've thought about similar things before, wondering if you have any thoughts on this.

saitcakmak:
> Second, I don't remember running into any issues when implementing a PCA transform that did implement many-to-one transforms for outcomes (not inputs), other than the sample shape issue in #681.

Yeah, outcome transforms don't have this issue. I have also tested some many-to-one outcome transforms successfully (implemented as MC objectives in #821 instead). It is changing the shape of `x` in `forward` (i.e. taking `n x q x d` and returning an MVN with `even_shape = n x q' x m`) that leads to issues.

PR #820, opened 2021-06-09
Implements AppendFeatures

saitcakmak:
Summary:
Implements AppendFeatures for built-in risk measure support in BoTorch.

AppendFeatures:
- Appends a given set of features to the input for joint posterior evaluation.
- For a `batch x q x d`-dim input `X` and a `n_f x d_f`-dim `feature_set`, this results in a `batch x (q * n_f) x (d + d_f)`-dim tensor of appended inputs.

Differential Revision: D28981162

### DIFF
--- a/botorch/models/transforms/input.py
+++ b/botorch/models/transforms/input.py
@@ -92,7 +92,7 @@ class InputTransform(ABC):
             A `batch_shape x n x d`-dim tensor of un-transformed inputs.
         """
         raise NotImplementedError(
-            f"{self.__class__.__name__} does not implement the `untransform` method"
+            f"{self.__class__.__name__} does not implement the `untransform` method."
         )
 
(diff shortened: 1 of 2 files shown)


saitcakmak:
We need a better name for this transform. Any suggestions @RaulAstudillo06 @danielrjiang ?

eytan:
In terms of alternative names for AddWTransform, I wonder if AddContextDistributionTransform Or AppendContextDistrivutionTransform or AppendContexts (adding Transform to the class name feels kind of redundant since it’s already in the transform module (cc @Balandat ). 

In general some more concrete examples or motivation in the doc strings and or comments could be helpful.

saitcakmak:
Thanks @eytan, great suggestions! I like `Add / Append` `ContextVariables`. Thoughts @sdaulton ?

eytan:
also cc @bletham

PR #823, opened 2021-06-10
Notebook: BO of risk measures

saitcakmak:
Summary: A tutorial notebook implementing rhoKG from [Cakmak2020risk]_ for risk averse Bayesian optimization with environmental variables.

Differential Revision: D29018010

### DIFF
--- a/tutorials/risk_averse_bo_with_environmental_variables.ipynb
+++ b/tutorials/risk_averse_bo_with_environmental_variables.ipynb
@@ -1,0 +1,551 @@
+{
+  "metadata": {
+    "bento_stylesheets": {
+      "bento/extensions/flow/main.css": true,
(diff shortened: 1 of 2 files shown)


eytan:
Hi Sait,

thanks for the nice tutorial.  A few high level comments:
1. It would be helpful to provide just a bit more context about risk averse BayesOpt is, some of the notation, e.g. that W represents some distribution of contexts for which f(x|w) may vary that we wish to minimize risk over. 
2. More generally, some additional motivation for risk averse Bo and when you’d want to use it could be helpful. This intro might include defining or describing at a high level what CVaR is and how it is computed. 
3. The above reader should help ease readers into a high level description of how the algorithm works and sketch out some of the logic of how we go about augmenting the inputs with the context distribution, and how that gets picked up by the risk averse BO utility functions.

Issue #543, opened 2020-09-11
Mixed Deterministic and GPyTorch Model


serizba:
Hi!

I would like to know if it is possible to have a multi-output model that uses a different type of model for each output. For example, in a problem with two outputs, it would be interesting to model the first output with Gaussian Processes and the second output with a Deterministic Model.

## Motivation

Main motivation is to easily model problems where one output is a black box and other outputs are well known functions. For example, in a HyperParameter optimization problem for a Neural Network, we may be interested in optimize the network in terms of accuracy and size of the network (preferring small ones). For this specific problem it would be interesting to use Gaussian Processes for the accuracy surrogate model and a Deterministic model for the size of the network (which can be computed easily).

I read the documentation and I did not found any way of achieving such feature. I even tries to implement it by myself but I am not able to mix a Deterministic and a Gaussian posterior correctly.

Thanks for your great work!

sdaulton:
Hi @serizba! A `ModelListGP` (https://github.com/pytorch/botorch/blob/master/botorch/models/model_list_gp_regression.py#L22) should do the trick. For example, you can pass a GPyTorchModel and a DeterministicModel as:
```
deterministic_model = GenericDeterministicModel(f=lambda x: torch.sin(x))
gp = SingleTaskGP(train_x, train_y)
model = ModelListGP(deterministic_model, gp)
```

sdaulton:
Yeah we should definitely add support for this functionality. cc @Balandat

Balandat:
Yeah we can look into making this work for convenience. For things like this we'll usually just use a separate model since if things are deterministic we don't really need to estimate / model covariances between those and the GP-modeled outcomes. But I agree it would be nice to have this adhere to the standard multi-output model API.

sdaulton:
Yeah we need a ModelList in botorch that is not tethered to gpytorch

serizba:
Thanks for your quick reply, yes, definitely it would be an interesting addition.

@Balandat said that for things like this it is preferable to use a separate model, but, for instance, my use case is to integrate the model in Ax, which is expecting a single Model, and therefore I would need ModelList.

For now I have created a simple workaround: creating a model that returns a Multivariate Gaussian Posterior, but I change the fitted mean with my deterministic value and I set the variance to zero.

Thanks!

Balandat:
great, glad you were able to unblock yourself here. 

> my use case is to integrate the model in Ax, which is expecting a single Model,

Nice, we definitely want to enable this.

[...]

PR #829, opened 2021-06-17
Introduce ModelList and PosteriorList

Balandat:
Summary:
This introudces a `ModelList` container for using heterogeneous models. This is very similar to `ModelListGP`, but more generic in the sense that it works with general objects that implement the `Model` API, not just GPs.

The particular use case I have in mind for this is to have a multi-output model where one or more outputs are deterministic functions of the input. Using `ModelList`, we can just throw some `DeterministicModel` objects into a `ModelList`, together with GP models. This is useful e.g. for doing Multi-Objective optimization when some objectives are deterministic functions of the input, which is the case in some Chemistry problems or in the context of optimizing sparsity-promoting policies. In that case, we should be able to simply stick the `ModelList` into existing MOO acquisition functions and everything should (famous last words) work.

Differential Revision: D29187914


Balandat:
This addresses #543

PR #838, opened 2021-06-22
Fix flaky test_warp_transform test

Balandat:
The data points generated could ended up near zero, which causes numerical issues. This also adjusts the eps in the Warp transform based on the dtype.


Balandat:
Coverage regression was caused by #681

PR #858, opened 2021-07-06
Fix qMFKG evaluate bug

saitcakmak:
Summary: In `qMFKG.evalute` the use of `if hasattr(self, "cost_aware_utility")` leads to a bug when `cost_aware_utility=None` (ends up calling `None` as a function). This replaces that with `if getattr(self, "cost_aware_utility", None) is not None`, which handles both not having a `cost_aware_utility` attribute (qKG) and `cost_aware_utility=None` (qMFKG).

Differential Revision: D29565578


Balandat (approved):
Thanks!

Issue #878, opened 2021-07-23
[Bug] Large-Scale HOGP Posterior Sampling Fail due to shape mismatch


wjmaddox:
# 🐛 Bug

Posterior sampling fails with the HOGP on crossing the Cholesky / Lanczos barrier due to a base sample shape mismatch on decomposing the joint train / test covariance.

## To reproduce

** Code snippet to reproduce **
```python
import torch 
from botorch.models import HigherOrderGP
from botorch.sampling import IIDNormalSampler

model = HigherOrderGP(torch.randn(100, 3), torch.randn(100, 3, 12, 15, 25))
test_x = torch.randn(701, 3) # torch.randn(700, 3) is okay
posterior = model.posterior(test_x)
sampler = IIDNormalSampler(num_samples=32)
sampler(posterior)
```

** Stack trace/error message **
```
botorch/posteriors/higher_order.py in rsample(self, sample_shape, base_samples)
    169         covariance_matrix = self.joint_covariance_matrix
    170         covar_root = covariance_matrix.root_decomposition().root
--> 171         samples = covar_root.matmul(base_samples)
    172 
    173         # now pluck out Y_x and X_x

RuntimeError: Incompatible dimensions for matmul: torch.Size([10813500, 1350000]) and torch.Size([10813500, 32])
```

## Expected Behavior

Base samples should adapt properly to the root decomposition size.

## System information

**Please complete the following information:**
- botorch master
- pytorch 1.10+

## Additional context

Unit test should also vary the cholesky / lanczos threshold to catch this. 

I'll put up a PR shortly, but flagging now.

wjmaddox:
Will be closed by #883 which is an expansion of #879.

PR #879, opened 2021-07-23
Allow base samples to have smaller shape than joint covar root in HOGP Posterior

wjmaddox:
Summary: resolves botorch #878

Differential Revision: D29886795

### DIFF
--- a/botorch/models/higher_order_gp.py
+++ b/botorch/models/higher_order_gp.py
@@ -44,7 +44,7 @@ from gpytorch.likelihoods import (
 )
 from gpytorch.models import ExactGP
 from gpytorch.priors.torch_priors import GammaPrior, MultivariateNormalPrior
-from gpytorch.settings import fast_pred_var, skip_posterior_variances
+from gpytorch.settings import fast_pred_var, max_cholesky_size, skip_posterior_variances
 from torch import Tensor
 from torch.nn import ModuleList, Parameter, ParameterList
 
@@ -447,7 +447,15 @@ class HigherOrderGP(BatchedMultiOutputGPyTorchModel, ExactGP):
                 )
             else:
                 train_inputs = self.train_inputs[0]
+            # TODO: use LazyTensor.cat_rows utility instead of forming the joint which
+            # we would have to decompose
             full_covar = self.covar_modules[0](torch.cat((train_inputs, X), dim=-2))
+
+            # in order to have continuous gradients wrt X we need to compute a dense
+            # cholesky wrt X, rather than relying on Lanczos so we pre-compute it here
+            # rather than in the rsample call where it is more difficult to control
+            with max_cholesky_size(10000):
+                _ = full_covar.root_decomposition(method="cholesky")
 
             if no_pred_variance:
                 pred_variance = mvn.variance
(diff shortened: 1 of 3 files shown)


Balandat:
So one potential issue here is that `R(x) @ base_samples[..., :R(x).shape[-1], ']` may not be continuous / diff'able at `x` when `R(x)` is computed with Lanczos (correct me if I'm wrong). If not it's still fine for sampling, but we won't be able to differentiate through posterior samples.

wjmaddox:
It looks to be okay to me from this snippet:

```python
import gpytorch
import torch
from gpytorch.kernels import MaternKernel

x = torch.randn(3000, 3, requires_grad = True)
kernel = MaternKernel(ard_num_dims=3)
root_lt = kernel(x).root_decomposition().root
root_lt.shape # this is 3000 x 100 (so lanczos)

root_lt.matmul(torch.randn(100, 5)).norm().backward()
x.grad.norm() # nonzero
```

Balandat:
Oh so this is the gradient at a single `x`. Sure, you can compute that. What I'm saying is that b/c of Lanczos, the root decomp (and therefore the matmul after) is not continuous in `x`. So if you try to optimize some function of `x` (say an acquisition function) you're screwed, unless there is some kind of regularization. The issue is that for `RR^T = M` there are many approximate low rank roots, and nothing forces them to be continuous in the elements of `M` without any regularization. 

```
import gpytorch
import torch
from gpytorch.kernels import MaternKernel
​
kernel = MaternKernel(ard_num_dims=3)
​rhs = torch.randn(100, 5)

x1 = torch.randn(3000, 3, requires_grad = True)
x2 = (1e-6 + x1.detach().clone()).requires_grad_(True)

y1 = kernel(x1).root_decomposition().root.matmul(rhs)
y1.norm().backward()
y1, x1.grad

y2 = kernel(x2).root_decomposition().root.matmul(rhs)
y2.norm().backward()
y2, x2.grad  # way different from y1. x1.grad

torch.norm(y1-y2)
tensor(106.1002, grad_fn=<CopyBackwards>)
```

Cholesky: 
```
x1_chol = torch.randn(100, 3, requires_grad = True)
x2_chol = (1e-6 + x1_chol.detach().clone()).requires_grad_(True)
y1_chol = kernel(x1_chol).root_decomposition().root.matmul(rhs)
y1_chol.norm().backward()
y1_chol, x1_chol.grad

y2_chol = kernel(x2_chol).root_decomposition().root.matmul(rhs)
y2_chol.norm().backward()
y2_chol, x2_chol.grad  # same-ish as y1. x1.grad

torch.norm(y1_chol - y2_chol)
tensor(2.0724e-05, grad_fn=<CopyBackwards>)
```

wjmaddox:
Ahh good point, for now, I can just throw in a larger cholesky flag (say 5k) at the expense of memory. 

However, from your example, the correct thing to do is to cache the training covariance (and its root decomposition) and then use `LazyTensor.cat_rows` to construct the training covariance, which should stay continuous wrt `x` because the root decompositions stay pretty much exact. This is what's done in the MTGP posterior code already [here](https://github.com/pytorch/botorch/blob/32c066e5362d35d8d59b72f54fd65a6d2c978420/botorch/models/multitask.py#L574), but the HOGP lazy covariance is just a [full covariance](https://github.com/pytorch/botorch/blob/32c066e5362d35d8d59b72f54fd65a6d2c978420/botorch/models/higher_order_gp.py#L450) because it was implemented before `cat_rows` went in. 

I can set that up in another PR, which should have improved memory overhead anyways.

wjmaddox:
Closed by #883 which subsumes this PR.

PR #881, opened 2021-07-26
Add `maximize` flag to PosteriorMean 

lukasfro:
## Motivation

The `PosteriorMean` acquisition function does not provide the convenience of a `maximize` flag as opposed to the other analytical acquisition function such as `UpperConfidenceBound`. The current state of the API therefore requires to use the following argument instead `objective=ScalarizedObjective(weights=torch.tensor([-1.0])`.

Yes

## Test Plan

The proposed change is trivial. I simply extended the unit test for the `PosteriorMean` acquisition function.

## Related PRs

### DIFF
--- a/botorch/acquisition/analytic.py
+++ b/botorch/acquisition/analytic.py
@@ -163,6 +163,23 @@ class PosteriorMean(AnalyticAcquisitionFunction):
         >>> pm = PM(test_X)
     """
 
+    def __init__(
+        self,
+        model: Model,
+        objective: Optional[ScalarizedObjective] = None,
+        maximize: bool = True,
+    ) -> None:
+        r"""Single-outcome Posterior Mean.
+
+        Args:
+            model: A fitted single-outcome GP model (must be in batch mode if
+                candidate sets X will be)
+            objective: A ScalarizedObjective (optional).
+            maximize: If True, consider the problem a maximization problem.
+        """
+        super().__init__(model=model, objective=objective)
+        self.maximize = maximize
+
     @t_batch_mode_transform(expected_q=1)
     def forward(self, X: Tensor) -> Tensor:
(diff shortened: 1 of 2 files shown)


Balandat (approved):
This looks good to me, thanks for the PR.

lukasfro:
Hi @Balandat, is there anything I can do regarding the failing internal tests?

Balandat:
@lukasfro seems to have been a transient issue, this should be good from your end.

botorch/acquisition/analytic.py
```diff
@@ -163,6 +163,23 @@ class PosteriorMean(AnalyticAcquisitionFunction):
...
+        objective: Optional[ScalarizedObjective] = None,
+        maximize: bool = True,
+    ) -> None:
+        r"""Single-outcome Posterior Mean.
+
+        Args:
+            model: A fitted single-outcome GP model (must be in batch mode if
+                candidate sets X will be)
+            objective: A ScalarizedObjective (optional).
+            maximize: If True, consider the problem a maximization problem.
```
saitcakmak, line 178: My concern about this is that it is not clear what happens internally to consider it a maximization / minimization problem. It may be nice to make it explicit that this negates the posterior mean if False.

It may help avoid the scenario where one optimizes `PosteriorMean(..., maximize=False)` and uses the reported value as the minimum value of the posterior mean (when it is actually -1 * minimum.).

lukasfro: @saitcakmak Thanks for the suggestion. If I understand you correctly, you are proposing to just change the docstring?

saitcakmak: Yeah, I think adding a note warning about this within the docstring should be sufficient.

PR #882, opened 2021-07-26
Ignore categorical dimensions when validating training inputs in MixedSingleTaskGP

vishwakftw:
Summary: In MixedSingleTaskGP, the Categorical Kernel operates on the categorical dimensions (`cat_dims`). The scaling of the data doesn't matter here since the distance is computed via an equality check.

Differential Revision: D29920517

### DIFF
--- a/botorch/models/gp_regression.py
+++ b/botorch/models/gp_regression.py
@@ -97,7 +97,10 @@ class SingleTaskGP(BatchedMultiOutputGPyTorchModel, ExactGP):
         if outcome_transform is not None:
             train_Y, _ = outcome_transform(train_Y)
         self._validate_tensor_args(X=transformed_X, Y=train_Y)
-        validate_input_scaling(train_X=transformed_X, train_Y=train_Y)
+        ignore_X_dims = getattr(self, "_ignore_X_dims_scaling_check", None)
+        validate_input_scaling(
+            train_X=transformed_X, train_Y=train_Y, ignore_X_dims=ignore_X_dims
+        )
         self._set_dimensions(train_X=train_X, train_Y=train_Y)
         train_X, train_Y, _ = self._transform_tensor_args(X=train_X, Y=train_Y)
         if likelihood is None:
--- a/botorch/models/utils.py
+++ b/botorch/models/utils.py
@@ -133,7 +133,11 @@ def check_no_nans(Z: Tensor) -> None:
 
 
 def check_min_max_scaling(
-    X: Tensor, strict: bool = False, atol: float = 1e-2, raise_on_fail: bool = False
+    X: Tensor,
+    strict: bool = False,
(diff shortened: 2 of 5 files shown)


botorch/models/utils.py
```diff
@@ -133,20 +133,31 @@ def check_no_nans(Z: Tensor) -> None:
 
 
 def check_min_max_scaling(
-    X: Tensor, strict: bool = False, atol: float = 1e-2, raise_on_fail: bool = False
+    X: Tensor,
+    ignore_dims: Optional[List[int]] = None,
```
Balandat, line 137: Mind adding this as the last arg so that in case folks call this with positional args it doesn't break stuff?

botorch/models/gp_regression.py
```diff
@@ -97,7 +97,10 @@ def __init__(
         if outcome_transform is not None:
             train_Y, _ = outcome_transform(train_Y)
         self._validate_tensor_args(X=transformed_X, Y=train_Y)
-        validate_input_scaling(train_X=transformed_X, train_Y=train_Y)
+        ignore_X_dims = getattr(self, "cat_dims", None)
```
Balandat, line 100: Let's make this more generic so it can be used by other models as well? We could do sth like `_ignore_X_dims_scaling`?

botorch/models/utils.py
```diff
@@ -133,20 +133,31 @@ def check_no_nans(Z: Tensor) -> None:
...
     """
+    ignore_dims = ignore_dims or []
+    check_dims = list(set(range(X.shape[-1])) - set(ignore_dims))
+    if len(check_dims) == 0:
+        return None
+
     with torch.no_grad():
-        Xmin, Xmax = torch.min(X, dim=-1)[0], torch.max(X, dim=-1)[0]
+        X_check = X[..., check_dims]
+        Xmin, Xmax = torch.min(X_check, dim=-1)[0], torch.max(X_check, dim=-1)[0]
```
Balandat, line 160:
```suggestion
        Xmin = torch.min(X_check, dim=-1).values
        Xmax = torch.max(X_check, dim=-1).values
```

Balandat (commented):
This mostly looks good, some nits. The main comment is making attribute containing the dims to be ignored more generic.

botorch/models/gp_regression.py
vishwakftw: Sounds good!

botorch/models/utils.py
vishwakftw: Oh, my bad.

botorch/models/utils.py
vishwakftw: Yep, that makes sense.

Balandat (approved):

PR #890, opened 2021-07-30
Implements PosteriorTransform

saitcakmak:
Summary: Implements PosteriorTransform for analytic transforms on the posteriors. This replaces the existing analytic objectives, e.g., `ScalarizedObjective`, and adds support for use of other analytic posterior transforms within the MC acquisition functions, such as an expectation over the `q` batch.

Differential Revision: D30003828


saitcakmak:
This is the design candidate for #860. Thoughts @Balandat, @sdaulton ?

Balandat:
Overall this looks great. We should be careful not break backwards-compatibility in a big way though. Here is my suggestion:

1) keep `AcquisitionObjective` but clearly mark it as DEPRECATED in the docstring

2) Add this:
```
class ScalarizedObjective(ScalarizedPosteriorTransform, AcquisitionObjective):
    """DEPRECATED - Use ScalarizedPosteriorTransform instead"""
    def __init__(self, weights: Tensor, offset: float = 0.0) -> None:
        warnings.warn(
            "ScalarizedObjective is deprecated and will be removed in the next version. "
            "Use ScalarizedPosteriorTransform instead."
        )
        super().__init__(weights=weights, offset=offset)    
```
so it remains importable/instantiable. 

3) Add a `**kwargs` to the init of all acquisition functions that expected an `AcquisitionObjective`, and use sth like the following to reduce deprecation boilerplate code:
```
def _deprecate_acqf_objective(
    cls: Type[AcquisitionFunction],
    posterior_transform: Optional[PosteriorTransform],
    objective: Optional[AquisitionObjective],
) -> Optional[PosteriorTransform]:
    if objective is None:
        return posterior_transform
    warnings.warn(
        "The objective argument to {cls.__name__} is deprecated and will be removed "
        "in the next version. Use posterior_transform instead.
    )
    return PosteriorTransform(weights=objective.weights, offset=objective.offset)
```
and call it at the beginning of each constructor via
```
posterior_transform = _deprecate_acqf_objective(
    cls=self.__class__,
    posterior_transform=posterior_transform,
    objective=kwargs.get("objective"),
)
```

botorch/acquisition/input_constructors.py
```diff
@@ -130,19 +135,32 @@ def construct_inputs_analytic_base(
...
+            "The `objective` argument to AnalyticAcquisitionFunctions is deprecated "
+            "and will be removed in the next version. Use `posterior_transform` "
+            "instead.",
+            DeprecationWarning,
+        )
+    return {
+        "model": model,
+        "posterior_transform": posterior_transform,
+        **{k: v for k, v in kwargs.items() if k == "objective"},
+    }
```
Balandat, line 156: Would you want to instantiate `posterior_transform` here from the objective, as you do in the deprecation helper? Doing this and raising the warning here would make more sense than deferring that to the constructor.

botorch/acquisition/input_constructors.py
```diff
@@ -634,20 +690,25 @@ def construct_inputs_qNEHVI(
 def get_best_f_analytic(
     training_data: TrainingData,
     objective: Optional[AcquisitionObjective] = None,
```
Balandat, line 692: could also raise a warning here and automatically generate the proper `PosteriorTransform` (same below)

botorch/acquisition/knowledge_gradient.py
```diff
@@ -87,11 +87,11 @@ def __init__(
             sampler: The sampler used to sample fantasy observations. Optional
                 if `num_fantasies` is specified.
             objective: The objective under which the samples are evaluated. If
-                `None` or a ScalarizedObjective, then the analytic posterior mean
-                is used, otherwise the objective is MC-evaluated (using
-                inner_sampler).
+                `None`, then the analytic posterior mean is used. Otherwise, the
+                objective is MC-evaluated (using inner_sampler).
+            posterior_transform: An optional PosteriorTransform.
```
Balandat, line 92: can you explain how this is used with the objective (i.e. transform, then sample, then application of objective)?

saitcakmak:
Ooops... Didn't know renaming the branch would close the PR :)

botorch/acquisition/monte_carlo.py
```diff
@@ -65,17 +71,14 @@ def __init__(
...
             sampler = SobolQMCNormalSampler(num_samples=512, collapse_batch_dims=True)
         self.add_module("sampler", sampler)
-        if objective is None:
-            if model.num_outputs != 1:
-                raise UnsupportedError(
-                    "Must specify an objective when using a multi-output model."
-                )
-            objective = IdentityMCObjective()
-        elif not isinstance(objective, MCAcquisitionObjective):
+        if objective is None and posterior_transform is None and model.num_outputs != 1:
```
sdaulton, line 74: same issue as mentioned above

botorch/acquisition/monte_carlo.py
```diff
@@ -212,7 +225,11 @@ def __init__(
...
         super().__init__(
-            model=model, sampler=sampler, objective=objective, X_pending=X_pending
+            model=model,
+            sampler=sampler,
+            objective=objective,
+            posterior_transform=posterior_transform,
+            X_pending=X_pending,
         )
         if prune_baseline:
             X_baseline = prune_inferior_points(
```
sdaulton, line 235: Need to pass `posterior_transform` to `prune_inferior_points`

botorch/models/higher_order_gp.py
```diff
@@ -502,8 +504,10 @@ def posterior(
             )
             if hasattr(self, "outcome_transform"):
                 posterior = self.outcome_transform.untransform_posterior(posterior)
-
-            return posterior
+            if posterior_transform is not None:
+                return posterior_transform(posterior)
```
sdaulton, line 508: Hmm do we need to be raising an exception in `botorch.posteriors.gpytorch.scalarize_posterior` for HOGP?

botorch/models/gpytorch.py
```diff
@@ -150,7 +156,10 @@ def posterior(
         posterior = GPyTorchPosterior(mvn=mvn)
         if hasattr(self, "outcome_transform"):
             posterior = self.outcome_transform.untransform_posterior(posterior)
-        return posterior
+        if posterior_transform is not None:
+            return posterior_transform(posterior)
+        else:
```
sdaulton, line 161: Let's remove `else` (many below) for conciseness.

sdaulton (changes requested):

botorch/acquisition/acquisition.py
```diff
@@ -32,6 +33,34 @@ def __init__(self, model: Model) -> None:
         super().__init__()
         self.add_module("model", model)
 
+    def _deprecate_acqf_objective(
```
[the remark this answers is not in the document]
Balandat: Not really necessary I guess. But we want to deprecate this after the next release anyway, and so I think we can just delete that code then.

botorch/acquisition/monte_carlo.py
Balandat: good catch

botorch/models/higher_order_gp.py
Balandat: Hmm yeah let me do this for now to be safe. May want to think through this some more down the road. cc @wjmaddox

botorch/acquisition/knowledge_gradient.py
```diff
@@ -122,12 +127,30 @@ def __init__(
...
+                    objective=objective,
+                )
+                objective = None
+            else:
+                raise RuntimeError(
+                    "Got both a non-MC objective (DEPRECATED) and a posterior "
+                    "transform. Use only a posterior transform instead."
+                )
+
+        if objective is None and posterior_transform is None and model.num_outputs != 1:
```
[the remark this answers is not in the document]
Balandat: Hmm good catch. Right now we don't have any posterior transforms that don't scalarize, but we do want to allow that behavior (e.g. for scaling outputs individually etc). Not a fan of the dummy eval idea though, but I guess I could add a `scalarizes` class property to PosteriorTransform that we read in this case.

botorch/models/higher_order_gp.py
wjmaddox: What's the current error with the HOGP and these types of transforms?

Balandat: I think I'll just raise a `NotImplementedError` here if `posterior_transform` is not `None` for now.

Balandat: > What's the current error with the HOGP and these types of transforms?

Not sure if it technically raises an error, but basically `scalarize_posterior` assumes a standard `GPyTorchPosterior` and will always return a `GPyTorchPosterior` and in some cases may try to instantiate the covariance. I don't think we have a use case for this at this point so maybe it's just safer to error out and revisit this when needed.

wjmaddox: Yeah, that probably should be applied to `KroneckerMultitaskGP` then as well for the same reason (forming covariances is expensive).

Balandat: Addressed in #1076

Issue #893, opened 2021-07-30
Fitting the model directly with `fit_gpytorch_scipy/torch` leads to buggy behavior with input transforms.


saitcakmak:
## Issue description

Since the 0.5 release, one has to call `_set_transformed_inputs` after model training to update `model.train_inputs` with their transformed versions (think `model.train_inputs = model.transform_inputs(model.train_inputs)`). This is done at the end of `fit_gpytorch_model`, so everything works great if one uses `fit_gpytorch_model` to train their model. 

But if one instead uses `fit_gpytorch_scipy/torch` directly to fit the model (e.g., #874 - saw other examples of this as well), `_set_transformed_inputs` does not get called automatically, the user is probably not aware of this need either, so we end up with a buggy behavior. Without `_set_transformed_inputs`, we end up calculating the posterior using non-transformed training inputs with transformed test inputs, along with hyper-parameters that were trained using transformed training inputs.

## Code example

```
import torch
from botorch import fit_gpytorch_model
from botorch.models import SingleTaskGP
from botorch.models.transforms.input import Normalize
from botorch.optim.fit import fit_gpytorch_torch
from gpytorch import ExactMarginalLogLikelihood


scale = 5
t_X = torch.linspace(0, scale, 10).view(-1, 1)
t_Y = torch.randn(10, 1)
model = SingleTaskGP(t_X, t_Y, input_transform=Normalize(d=1))
mll = ExactMarginalLogLikelihood(model.likelihood, model)
fit_gpytorch_model(mll)
print("Train inputs with `fit_gpytorch_model`: ", model.train_inputs)

model = SingleTaskGP(t_X, t_Y, input_transform=Normalize(d=1))
mll = ExactMarginalLogLikelihood(model.likelihood, model)
fit_gpytorch_torch(mll)
print("Train inputs with `fit_gpytorch_torch`: ", model.train_inputs)
```

Output (we want to see them normalized here):
```
Train inputs with `fit_gpytorch_model`:  (tensor([[0.0000],
        [0.1111],
        [0.2222],
        [0.3333],
        [0.4444],
        [0.5556],
        [0.6667],
        [0.7778],
        [0.8889],
        [1.0000]]),)

Train inputs with `fit_gpytorch_torch`:  (tensor([[0.0000],
        [0.5556],
        [1.1111],
        [1.6667],
        [2.2222],
        [2.7778],
        [3.3333],
        [3.8889],
        [4.4444],
        [5.0000]]),)
```

[...]

PR #894, opened 2021-07-31
Move `train_inputs` transforms to `model.train/eval` calls.

saitcakmak:
Summary:
This moves `_set_transformed_inputs` call from `fit_gpytorch_model` to `model.eval()`. In addition, this adds a `_revert_to_original_inputs` method, which is called in `model.train()`.
Addresses #893.

Differential Revision: D30030256

### DIFF
--- a/botorch/fit.py
+++ b/botorch/fit.py
@@ -15,13 +15,10 @@ import warnings
 from copy import deepcopy
 from typing import Any, Callable
 
-from botorch.exceptions.errors import BotorchError, UnsupportedError
+from botorch.exceptions.errors import UnsupportedError
 from botorch.exceptions.warnings import BotorchWarning, OptimizationWarning
 from botorch.models.converter import batched_to_model_list, model_list_to_batched
-from botorch.models.gp_regression import HeteroskedasticSingleTaskGP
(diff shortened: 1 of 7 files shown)


botorch/models/deterministic.py
```diff
@@ -76,6 +76,18 @@ def posterior(
...
+        r"""This is a no-op since DeterministicModel does not have a `train_inputs`
+        attribute and does not go through training.
+        """
+        pass  # pragma: no cover
+
+    def _revert_to_original_inputs(self) -> None:
+        r"""This is a no-op since DeterministicModel does not have a `train_inputs`
+        attribute and does not go through training.
+        """
+        pass  # pragma: no cover
```
Balandat, line 89: It would be good if one wouldn't have to overwrite this in subclasses that don't have a `train_inputs` attribute. Can we just modify the function in `Model` to be a nullup if there is no `train_inputs` attribute?

Balandat (commented):
Nice.

botorch/models/deterministic.py
saitcakmak: Yeah, makes sense. I did that and added a warning so that this doesn't turn into another gotcha down the line.

PR #896, opened 2021-08-02
Modify qNEHVI to support one-to-many input transforms

saitcakmak:
Summary: Moves the `objective` call from `_compute_qehvi` to `forward` in order to support one-to-many input transforms. Also adds a `constraint_objective` that is applied to `samples` before evaluating the constraints.

Differential Revision: D29336709

### DIFF
--- a/botorch/acquisition/multi_objective/monte_carlo.py
+++ b/botorch/acquisition/multi_objective/monte_carlo.py
@@ -45,6 +45,7 @@ from botorch.models.deterministic import DeterministicModel
 from botorch.models.model import Model
 from botorch.models.model_list_gp_regression import ModelListGP
 from botorch.models.multitask import KroneckerMultiTaskGP, MultiTaskGP
+from botorch.models.transforms.input import InputPerturbation
 from botorch.posteriors import DeterministicPosterior
 from botorch.posteriors.gpytorch import GPyTorchPosterior
 from botorch.posteriors.posterior import Posterior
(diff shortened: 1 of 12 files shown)


test/acquisition/multi_objective/test_monte_carlo.py
```diff
@@ -556,6 +570,11 @@ def setUp(self):
             device=self.device,
         )
 
+    @mock.patch.object(
+        IdentityMCMultiOutputObjective,
+        "_verify_output_shape",
+        mock.PropertyMock(return_value=False),
+    )
```
saitcakmak, line 577: I went with this blanket patch to avoid having to repeat it in many places. NEHVI in general requires 2 different posterior shapes (only for baseline and baseline + current X), which does not work when using `MockPosterior`. However, it seems that this blanket patch ended up hiding some bugs that were only caught thanks to the tutorials. Implementing a new mock posterior that adapts to the input shape is probably a better way to go here.

botorch/acquisition/multi_objective/monte_carlo.py
```diff
@@ -708,9 +730,6 @@ def _set_sampler(
                 self.sampler.base_samples[..., :end_idx, :] = current_base_samples.view(
                     view_shape
                 ).expand(expanded_shape)
-                # update cached subset indices
-                # Note: this also stores self.q = q
-                self._cache_q_subset_indices(q=q)
```
sdaulton, line 713: Why is this removed?

saitcakmak: It's redundant since `_compute_qnehvi` has this call. Both `self.q` and `q_subset_indices` are used only after that (or in the `q!=self.q` check).

sdaulton (approved):

PR #902, opened 2021-08-11
Fixes input transform bug in fantasize call

### LINKED ISSUE #899: [Bug] qNegIntegratedPosteriorVariance with input transforms
# 🐛 Bug
Hi first off thanks a lot for this fantastic tool. Not sure if this is a bug or just misusage on my side…
I've observed an odd behaviour when I tried to use qNegIntegratedPosteriorVariance with a FixedNoiseGP that uses an `input_transform`. It seems that `optimize_acqf` returns only suggestions within [0,1]^d, so in the "post-transform" parameter space. When using no `input_transform` (i.e. non-optimal priors?) the suggestions are given over the desired bounds. Also when using a different acquisition function everything looks okay. Do I need to do an explicit back-transformation of the suggested parameters?

The plots below show the two scenarios (blue dots are the samples, orange are the mc_points for NIPV). With `input_transform` all but the first 2 random samples are in the same spot.

Thanks!

nipv_trans
nipv_notrans

## To reproduce

```python
test_func = Branin()
ndim = 2
xmin, xmax = (-5., 10.)

l_bounds = [xmin]*ndim
u_bounds = [xmax]*ndim
sampler = qmc.Halton(d=ndim, scramble=False)
sample = sampler.random(n=1024)
sample = qmc.scale(sample, l_bounds, u_bounds)
mc_points = torch.from_numpy(sample)

X_data = torch.from_numpy(np.random.rand(2,2))*(xmax-xmin)+xmin
y_data = test_func(X_data).reshape(-1,1)
bounds = torch.tensor([[xmin]*ndim,[xmax]*ndim])

maes = []
for i in tqdm(range(20)):
    y_noise = torch.ones_like(y_data)*1e-6
    model = FixedNoiseGP(X_data, y_data,y_noise,
                         #input_transform=None,
                         input_transform=Normalize(ndim, bounds=bounds),
                         outcome_transform=Standardize(1),
                         )
    fit_gpytorch_model(mll)
    acq_func = qNegIntegratedPosteriorVariance(model=model,mc_points=mc_points)
    #acq_func = qNoisyExpectedImprovement(model, X_baseline=X_data)
    
    candidates, _ = optimize_acqf(
        acq_function=acq_func,
        bounds=bounds,
        q=1,
        num_restarts=10,
        raw_samples=512,  # used for intialization heuristic
    )
    new_x = candidates.detach()
    exact_obj = test_func(new_x).unsqueeze(-1)
    X_data = torch.cat([X_data, new_x])
    y_data = torch.cat([y_data,exact_obj])
```

## System information

-   botorch 0.5.0
-  gpytorch 1.5.0
-  pytorch 1.9.0+cu102
-  Linux

saitcakmak:
Summary:
Fixes #899.
`fantasize` call was adding the fantasy `X` to `train_inputs` without applying the input transforms. This adds a `transform_inputs` call to fix that. In addition, the `FixedNoiseGP` fantasize call was not using the `fantasize` flag (to disable one-to-many input transforms), which is also fixed here.

Differential Revision: D30254090

### DIFF
--- a/botorch/models/gp_regression.py
+++ b/botorch/models/gp_regression.py
@@ -17,6 +17,7 @@ from botorch import settings
 from botorch.models.gpytorch import BatchedMultiOutputGPyTorchModel
 from botorch.models.transforms.input import InputTransform
 from botorch.models.transforms.outcome import Log, OutcomeTransform
+from botorch.models.utils import fantasize as fantasize_flag
 from botorch.models.utils import validate_input_scaling
 from botorch.sampling.samplers import MCSampler
 from botorch.utils.containers import TrainingData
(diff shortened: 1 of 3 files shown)


Balandat (approved):
thanks for the fix!

PR #903, opened 2021-08-11
Implements ExpectationPosteriorTransform

saitcakmak:
Summary:
Implements ExpectationPosteriorTransform, which transforms the batch x (q * n_w) x m posterior to a batch x q x m posterior of the expectation over the n_w points.
Unlike the RiskMeasureMCObjective, this avoids the posterior sampling over q * n_w points, which leads to significant speed-ups for large q * n_w.

Differential Revision: D29277116

### DIFF
--- a/botorch/acquisition/multi_objective/multi_output_risk_measures.py
+++ b/botorch/acquisition/multi_objective/multi_output_risk_measures.py
@@ -95,7 +95,13 @@ class MultiOutputRiskMeasureMCObjective(
 
 
 class MultiOutputExpectation(MultiOutputRiskMeasureMCObjective):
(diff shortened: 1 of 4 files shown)


Balandat:
@saitcakmak same here, is there a plan to merge this in?

saitcakmak:
@Balandat, I was thinking of merging this in after `PosteriorTransform` since that actually makes this useful. Currently it adds things to NEHVI which will be redundant once that one is merged in. Otherwise, its use is very limited as an `AcquisitionObjective`.

Balandat:
Great. I’ve rebased and polished the PosteriorTransform PR internally and plan to merge it in after our upcoming maintenance release.

PR #917, opened 2021-08-21
fix outcome transform bug in batched_to_model_list

### LINKED ISSUE #916: [Bug] batch_to_model_list errors when using outcome transforms.
# 🐛 Bug

The `batch_to_model_list` helper throws a `RuntimeError` when used with a model that has an outcome transform.

## To reproduce

```python
import torch
from botorch.models.converter import batched_to_model_list

train_X = torch.rand(10, 3)
train_Y = torch.randn(10, 2)

octf = Standardize(m=2)

model = SingleTaskGP(
    train_X, train_Y, outcome_transform=octf,
).eval()

batched_to_model_list(batch_model=model)
```

```
  File "/Users/saitcakmak/PycharmProjects/botorch_test/debug_batch_to_modellist.py", line 15, in <module>
    batched_to_model_list(batch_model=model)
  File "/Users/saitcakmak/.conda/envs/botorch_test/lib/python3.8/site-packages/botorch/models/converter.py", line 252, in batched_to_model_list
    model.load_state_dict(sd)
  File "/Users/saitcakmak/.conda/envs/botorch_test/lib/python3.8/site-packages/torch/nn/modules/module.py", line 1406, in load_state_dict
    raise RuntimeError('Error(s) in loading state_dict for {}:\n\t{}'.format(
RuntimeError: Error(s) in loading state_dict for SingleTaskGP:
	Unexpected key(s) in state_dict: "outcome_transform.stdvs", "outcome_transform._stdvs_sq", "outcome_transform.means". 
```

## Expected Behavior

It should convert the batched model into a model list.

## System information

BoTorch / GPyTorch: Current main/master branches.

## Proposed Fix

I think it can be resolved by passing in the subsetted outcome transform in here https://github.com/pytorch/botorch/blob/e4dcf603fdaf83f0e5f8b9b392f943c89dfff7eb/botorch/models/converter.py#L251. I think we also need to remove the outcome transform keys from `sd` to avoid reverting the work done by `subset_output`. So, it would be something like:
```python
if outcome_transform is not None:
    octf = outcome_transform.subset_output(idcs=[i])
    sd = {k: v for k, v in sd.items() if "outcome_transform" not in k}
else:
    octf = None
model = batch_model.__class__(input_transform=input_transform, outcome_transform=octf, **kwargs)
```

Happy to put this up as a PR if the solution seems reasonable.

saitcakmak:
## Motivation

Closes #916. Assigns the subsetted outcome transform to each model in model list. 

Also adds an unsupported error for model list to batch conversion when used with outcome transforms (outcome transforms are currently ignored by the converter).

Yes.

## Test Plan

Unit tests.

### DIFF
--- a/botorch/models/converter.py
+++ b/botorch/models/converter.py
@@ -265,7 +265,7 @@ def batched_to_model_list(batch_model: BatchedMultiOutputGPyTorchModel) -> Model
             kwargs["outcome_transform"] = octf
             # Update the outcome transform state dict entries.
             sd = {
-                **{k: v for k, v in sd.items() if "outcome_transform" not in k},
+                **sd,
                 **{"outcome_transform." + k: v for k, v in octf.state_dict().items()},
             }
         else:
--- a/botorch/models/converter.py
+++ b/botorch/models/converter.py
@@ -21,6 +21,7 @@ from botorch.models.gp_regression_mixed import MixedSingleTaskGP
 from botorch.models.gpytorch import BatchedMultiOutputGPyTorchModel
 from botorch.models.model_list_gp_regression import ModelListGP
 from botorch.models.transforms.input import InputTransform
+from botorch.models.transforms.outcome import OutcomeTransform
 from torch import Tensor
 from torch.nn import Module
 
@@ -49,7 +50,7 @@ def _get_module(module: Module, name: str) -> Module:
 
 def _check_compatibility(models: ModelListGP) -> None:
     """Check if a ModelListGP can be converted."""
-    # check that all submodules are of the same type
+    # Check that all submodules are of the same type.
     for modn, mod in models[0].named_modules():
         mcls = mod.__class__
         if not all(isinstance(_get_module(m, modn), mcls) for m in models[1:]):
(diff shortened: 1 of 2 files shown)


botorch/models/converter.py
```diff
@@ -248,6 +258,18 @@ def batched_to_model_list(batch_model: BatchedMultiOutputGPyTorchModel) -> Model
...
             kwargs.update(batch_model._init_args)
+        # NOTE: Adding outcome transform to kwargs to avoid the multiple
+        # values for same kwarg issue with SingleTaskMultiFidelityGP.
+        if outcome_transform is not None:
+            octf = outcome_transform.subset_output(idcs=[i])
+            kwargs["outcome_transform"] = octf
+            # Update the outcome transform state dict entries.
+            sd = {
+                **{k: v for k, v in sd.items() if "outcome_transform" not in k},
+                **{"outcome_transform." + k: v for k, v in octf.state_dict().items()},
```
Balandat, line 269: I don't think it's necessary to exclude `"outcome_transform"` in the first line here - should be overwritten by the content of the state dicts?

Balandat (approved):
this looks great, many thanks for the PR.

botorch/models/converter.py
saitcakmak: Good point!

PR #919, opened 2021-08-23
Adding proximal acquisition function wrapper

roussel-ryan:
## Motivation

The goal of this acquisition function is to bias a GP optimization towards smooth optimization through the input domain. The proximal AF multiplies the base acquisition function by a squared exponential with a user defined lengthscale, centered at the most recently observed training point (assumed to be model.train_inputs[-1]). If the associated lengthscale is short the algorithm makes small jumps in input space, if it is long it is not strongly biased. This method differs from simply restricting the max travel size in input space by allowing large travel distances if the predicted value is large enough. See https://journals.aps.org/prab/abstract/10.1103/PhysRevAccelBeams.24.062801 for discussion and analysis.

This becomes relevant when using Bayesian optimization techniques on optimizing physical systems, where there is a cost associated with changing input parameters. 

Yes, I've tried my best to satisfy all requirements although there are possibly errors (first time contributing to a major project like this).

## Test Plan

Test script test_proximal.py has been added to test/acquisition. I can also provide numerical proof that this works with a simple script but I was unsure where to include. Result is show below
figure_1

Please comment if I need to change anything, thanks!

## Related PRs
None

### DIFF
--- a/botorch/acquisition/proximal.py
+++ b/botorch/acquisition/proximal.py
@@ -13,10 +13,9 @@ from __future__ import annotations
 
 import torch
 from botorch.acquisition.analytic import AnalyticAcquisitionFunction
+from botorch.utils import t_batch_mode_transform
 from torch import Tensor
 from torch.nn import Module
-
-from botorch.utils import t_batch_mode_transform
 
 
 class ProximalAcquisitionFunction(AnalyticAcquisitionFunction):
@@ -48,10 +47,13 @@ class ProximalAcquisitionFunction(AnalyticAcquisitionFunction):
         Module.__init__(self)
         self.acq_func = acq_function
 
-        self.register_buffer('proximal_weights', proximal_weights)
+        self.register_buffer("proximal_weights", proximal_weights)
 
         # check to make sure that weights match the training data shape
-        assert self.proximal_weights.shape[0] == self.acq_func.model.train_inputs[0][-1].shape[-1]
+        assert (
+            self.proximal_weights.shape[0]
+            == self.acq_func.model.train_inputs[0][-1].shape[-1]
+        )
 
     @t_batch_mode_transform(expected_q=1, assert_output_shape=False)
     def forward(self, X: Tensor):
@@ -66,7 +68,9 @@ class ProximalAcquisitionFunction(AnalyticAcquisitionFunction):
         """
         d = self.proximal_weights.shape[0]
         last_X = self.acq_func.model.train_inputs[0][-1].reshape(1, 1, -1)
-        _unbroadcasted_scale_tril = torch.diag(torch.sqrt(self.proximal_weights)).reshape(1, 1, d, d)
+        _unbroadcasted_scale_tril = torch.diag(
+            torch.sqrt(self.proximal_weights)
+        ).reshape(1, 1, d, d)
 
         diff = X - last_X
         M = _batch_mahalanobis(_unbroadcasted_scale_tril, diff)
@@ -99,16 +103,20 @@ def _batch_mahalanobis(bL, bx):
     bx_new_shape += (n,)
     bx = bx.reshape(bx_new_shape)
     # Permute bx to make it have shape (..., 1, j, i, 1, n)
-    permute_dims = (list(range(outer_batch_dims)) +
-                    list(range(outer_batch_dims, new_batch_dims, 2)) +
-                    list(range(outer_batch_dims + 1, new_batch_dims, 2)) +
-                    [new_batch_dims])
+    permute_dims = (
+        list(range(outer_batch_dims))
+        + list(range(outer_batch_dims, new_batch_dims, 2))
+        + list(range(outer_batch_dims + 1, new_batch_dims, 2))
+        + [new_batch_dims]
+    )
     bx = bx.permute(permute_dims)
 
     flat_L = bL.reshape(-1, n, n)  # shape = b x n x n
     flat_x = bx.reshape(-1, flat_L.size(0), n)  # shape = c x b x n
     flat_x_swap = flat_x.permute(1, 2, 0)  # shape = b x n x c
-    M_swap = torch.triangular_solve(flat_x_swap, flat_L, upper=False)[0].pow(2).sum(-2)  # shape = b x c
+    M_swap = (
+        torch.triangular_solve(flat_x_swap, flat_L, upper=False)[0].pow(2).sum(-2)
+    )  # shape = b x c
     M = M_swap.t()  # shape = c x b
 
     # Now we revert the above reshape and permute operators.
@@ -118,5 +126,3 @@ def _batch_mahalanobis(bL, bx):
         permute_inv_dims += [outer_batch_dims + i, old_batch_dims + i]
     reshaped_M = permuted_M.permute(permute_inv_dims)  # shape = (..., 1, i, j, 1)
     return reshaped_M.reshape(bx_batch_shape)
-
-
--- a/botorch/acquisition/proximal.py
+++ b/botorch/acquisition/proximal.py
@@ -20,7 +20,7 @@ from torch.nn import Module
 
 class ProximalAcquisitionFunction(AnalyticAcquisitionFunction):
     """A wrapper around AquisitionFunctions to add proximal weighting of the acquisition function. Acquisition
-    function is weighted via a squared exponential centered at the last training point with varying lengthscales
+    function is weighted via a squared exponential centered at the last training point, with varying lengthscales
     corresponding to 'proximal_weights'. Can only be used with single batch analytical acquisition functions.
 
     Example:
@@ -36,11 +36,11 @@ class ProximalAcquisitionFunction(AnalyticAcquisitionFunction):
         acq_function: AnalyticAcquisitionFunction,
         proximal_weights: Tensor,
     ) -> None:
-        r"""Derived Acquisition Function by fixing a subset of input features.
+        r"""Derived Acquisition Function by weighted by proximity to recently observed point.
 
         Args:
             acq_function: The base acquisition function, operating on input
-                tensors `X_full` of feature dimension `d`.
+                tensors of feature dimension `d`.
             proximal_weights: Tensor used to bias locality along each axis, should be in the shape of ('d',).
 
         """
@@ -70,12 +70,12 @@ class ProximalAcquisitionFunction(AnalyticAcquisitionFunction):
         last_X = self.acq_func.model.train_inputs[0][-1].reshape(1, 1, -1)
         _unbroadcasted_scale_tril = torch.diag(
             torch.sqrt(self.proximal_weights)
-        ).reshape(1, 1, d, d)
+        ).reshape(1, d, d)
 
         diff = X - last_X
         M = _batch_mahalanobis(_unbroadcasted_scale_tril, diff)
         proximal_acq_weight = torch.exp(-0.5 * M)
-        return self.acq_func(X) * proximal_acq_weight
+        return self.acq_func(X) * proximal_acq_weight.flatten()
 
 
 def _batch_mahalanobis(bL, bx):
--- a/botorch/acquisition/proximal.py
+++ b/botorch/acquisition/proximal.py
@@ -1,0 +1,122 @@
+#!/usr/bin/env python3
+# Copyright (c) Facebook, Inc. and its affiliates.
+#
+# This source code is licensed under the MIT license found in the
+# LICENSE file in the root directory of this source tree.
+
+r"""
+A wrapper around AquisitionFunctions to fix certain features for optimization.
+This is useful e.g. for performing contextual optimization.
+"""
+
+from __future__ import annotations
+
+import torch
+from botorch.acquisition.analytic import AnalyticAcquisitionFunction
+from torch import Tensor
+from torch.nn import Module
+
+from botorch.utils import t_batch_mode_transform
+
+
+class ProximalAcquisitionFunction(AnalyticAcquisitionFunction):
+    """A wrapper around AquisitionFunctions to add proximal weighting of the acquisition function. Acquisition
+    function is weighted via a squared exponential centered at the last training point with varying lengthscales
+    corresponding to 'proximal_weights'. Can only be used with single batch analytical acquisition functions.
+
+    Example:
+        >>> model = SingleTaskGP(train_X, train_Y)  # d = 2
+        >>> EI = ExpectedImprovement(model, best_f=0.0)
+        >>> proximal_weights = torch.ones(d)
+        >>> EI_proximal = ProximalAcquisitionFunction(EI, proximal_weights)
+        >>> eip = EI_proximal(test_X)  # d' = 3
+    """
+
+    def __init__(
+        self,
+        acq_function: AnalyticAcquisitionFunction,
+        proximal_weights: Tensor,
+    ) -> None:
+        r"""Derived Acquisition Function by fixing a subset of input features.
+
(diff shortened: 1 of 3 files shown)


botorch/acquisition/proximal.py
```diff
@@ -0,0 +1,128 @@
...
+        self,
+        acq_function: AnalyticAcquisitionFunction,
+        proximal_weights: Tensor,
+    ) -> None:
+        r"""Derived Acquisition Function by weighted by proximity to recently observed point.
+
+        Args:
+            acq_function: The base acquisition function, operating on input
+                tensors of feature dimension `d`.
+            proximal_weights: Tensor used to bias locality along each axis, should be in the shape of ('d',).
```
saitcakmak, line 44:
```suggestion
            proximal_weights: A `d`-dim tensor used to bias locality along each axis.
```
Nit

botorch/acquisition/proximal.py
```diff
@@ -0,0 +1,128 @@
...
+        Module.__init__(self)
+        self.acq_func = acq_function
+
+        self.register_buffer("proximal_weights", proximal_weights)
+
+        # check to make sure that weights match the training data shape
+        assert (
+            self.proximal_weights.shape[0]
+            == self.acq_func.model.train_inputs[0][-1].shape[-1]
+        )
```
saitcakmak, line 56:
```suggestion
        if self.proximal_weights.shape[0] != self.acq_func.model.train_inputs[0][-1].shape[-1]:
            raise ValueError("`proximal_weights` must be a one dimensional tensor with shape..... 
```
Let's use an explicit error message instead.

botorch/acquisition/proximal.py
```diff
@@ -0,0 +1,128 @@
...
+from botorch.acquisition.analytic import AnalyticAcquisitionFunction
+from botorch.utils import t_batch_mode_transform
+from torch import Tensor
+from torch.nn import Module
+
+
+class ProximalAcquisitionFunction(AnalyticAcquisitionFunction):
+    """A wrapper around AquisitionFunctions to add proximal weighting of the acquisition function. Acquisition
+    function is weighted via a squared exponential centered at the last training point, with varying lengthscales
+    corresponding to 'proximal_weights'. Can only be used with single batch analytical acquisition functions.
```
saitcakmak, line 24: Since this can only be used with single batch analytic acquisition functions, should we add checks for these in the constructor?

Just curious, why is this limited to analytic acquisition functions? Does anything break if I were to use this with an MC acquisition function (along with SAA to fix base samples - botorch default)?

botorch/acquisition/proximal.py
```diff
@@ -0,0 +1,128 @@
...
+
+
+def _batch_mahalanobis(bL, bx):
+    r"""
+    Computes the squared Mahalanobis distance :math:`\mathbf{x}^\top\mathbf{M}^{-1}\mathbf{x}`
+    for a factored :math:`\mathbf{M} = \mathbf{L}\mathbf{L}^\top`.
+
+    Accepts batches for both bL and bx. They are not necessarily assumed to have the same batch
+    shape, but `bL` one should be able to broadcasted to `bx` one.
+    """
```
saitcakmak, line 88: Add Args and Returns?

botorch/acquisition/proximal.py
```diff
@@ -0,0 +1,128 @@
...
+    """A wrapper around AquisitionFunctions to add proximal weighting of the acquisition function. Acquisition
+    function is weighted via a squared exponential centered at the last training point, with varying lengthscales
+    corresponding to 'proximal_weights'. Can only be used with single batch analytical acquisition functions.
+
+    Example:
+        >>> model = SingleTaskGP(train_X, train_Y)  # d = 2
+        >>> EI = ExpectedImprovement(model, best_f=0.0)
+        >>> proximal_weights = torch.ones(d)
+        >>> EI_proximal = ProximalAcquisitionFunction(EI, proximal_weights)
+        >>> eip = EI_proximal(test_X)  # d' = 3
```
saitcakmak, line 31: Why is `d' = 3` here? I might be missing something but looks like you use the same `d` below.

test/acquisition/test_proximal.py
```diff
@@ -0,0 +1,35 @@
...
+        train_Y = train_X.norm(dim=-1, keepdim=True)
+        model = SingleTaskGP(train_X, train_Y).to(device=self.device).eval()
+        EI = ExpectedImprovement(model, best_f=0.0)
+
+        # test single point
+        proximal_weights = torch.ones(3)
+        test_X = torch.rand(1, 3, device=self.device)
+        EI_prox = ProximalAcquisitionFunction(EI, proximal_weights=proximal_weights)
+
+        ei_prox = EI_prox(test_X)
```
saitcakmak, line 26: Let's check that `ei_prox` has the correct shape and the correct value (i.e., by comparing to a manually weighted EI). Same below if possible

saitcakmak (commented):
Sounds like an interesting application! Left a few comments.

roussel-ryan:
Thanks for the comments, I indeed based it on the FixedFidelity AF but I must have forgotten to rewrite all of the comments. I'll make changes to address your comments

botorch/acquisition/proximal.py
```diff
@@ -0,0 +1,128 @@
...
+        self.register_buffer("proximal_weights", proximal_weights)
+
+        # check to make sure that weights match the training data shape
+        assert (
+            self.proximal_weights.shape[0]
+            == self.acq_func.model.train_inputs[0][-1].shape[-1]
+        )
+
+    @t_batch_mode_transform(expected_q=1, assert_output_shape=False)
+    def forward(self, X: Tensor):
```
Balandat, line 59:
```suggestion
    def forward(self, X: Tensor) -> Tensor:
```

botorch/acquisition/proximal.py
```diff
@@ -0,0 +1,128 @@
...
+
+        Args:
+            X: Input tensor of feature dimension `d` .
+
+        Returns:
+            Base acquisition function evaluated on tensor `X` multiplied by a squared exponenital centered at the last
+            training point and is weighted according to the weighting of each dimension.
+        """
+        d = self.proximal_weights.shape[0]
+        last_X = self.acq_func.model.train_inputs[0][-1].reshape(1, 1, -1)
```
Balandat, line 70: Strictly, this won't work for all models as the basic `Model` API does not enforce a model to have `train_inputs`. Can you add a check to the constructor that raises an `UnsupportedError` with an informative error message if `acqf.model` does not have a `train_inputs` attribute? You should probably also raise an error if the "batch shape", i.e. `train_inputs[0][-1].shape[:-3]` is not empty; as this will not work with batched models. If you want to make it work with batched models this `reshape` should be modified here.

botorch/acquisition/proximal.py
```diff
@@ -0,0 +1,128 @@
...
+            torch.sqrt(self.proximal_weights)
+        ).reshape(1, d, d)
+
+        diff = X - last_X
+        M = _batch_mahalanobis(_unbroadcasted_scale_tril, diff)
+        proximal_acq_weight = torch.exp(-0.5 * M)
+        return self.acq_func(X) * proximal_acq_weight.flatten()
+
+
+def _batch_mahalanobis(bL, bx):
```
Balandat, line 81: This looks suspiciously like `torch.distributions.multivariate_normal._batch_mahalanobis` - maybe just import that?

Balandat: Actually, since `_unbroadcasted_scale_tril` is always diagonal by construction, this is kind of boiling the ocean - the whole thing above just be equivalent to `torch.sum(diff ** 2 / self.proximal_weights)` and you can get rid of this function altogether.

Balandat (changes requested):
First of all, thanks a lot for the contribution. I agree that this is a pretty interesting application that makes a lot of sense. 

Comments inline. My main one is that you can substantially simplify the weighting logic here.

botorch/acquisition/proximal.py
roussel-ryan: Turns out it is not limited to single batch analytic acquisition functions, just single batch models (see additions to test_proximal

botorch/acquisition/proximal.py
roussel-ryan: I ended up using the torch.distributions version of _batch_mahalanobis, it was easier to implement the calculation in batch model than what you stated. Initially I was being over-conservative with my imports

roussel-ryan:
Thanks again for all of the comments, I have updated the pull request with changes to address the requested changes

botorch/acquisition/proximal.py
```diff
@@ -0,0 +1,109 @@
+#!/usr/bin/env python3
+# Copyright (c) Facebook, Inc. and its affiliates.
+#
+# This source code is licensed under the MIT license found in the
+# LICENSE file in the root directory of this source tree.
+
+r"""
+A wrapper around AquisitionFunctions to add proximal weighting of the
```
saitcakmak, line 8:
```suggestion
A wrapper around AcquisitionFunctions to add proximal weighting of the
```

botorch/acquisition/proximal.py
```diff
@@ -0,0 +1,109 @@
...
+acquisition function. Acquisition function is weighted via a squared exponential
+centered at the last training point, with varying lengthscales corresponding to
+'proximal_weights'. Can only be used with acqusition functions based on single
```
saitcakmak, line 11:
```suggestion
'proximal_weights'. Can only be used with acquisition functions based on single
```

botorch/acquisition/proximal.py
```diff
@@ -0,0 +1,109 @@
...
+batch models.
+
+Small values of proximal_weights corresponds to strong biasing towards recently
```
saitcakmak, line 14:
```suggestion
Small values of `proximal_weights` corresponds to strong biasing towards recently
```
just for consistency

botorch/acquisition/proximal.py
```diff
@@ -0,0 +1,109 @@
...
+class ProximalAcquisitionFunction(AcquisitionFunction):
+    """A wrapper around AquisitionFunctions to add proximal weighting of the
+    acquisition function. Acquisition function is weighted via a squared exponential
+    centered at the last training point, with varying lengthscales corresponding to
+    'proximal_weights'. Can only be used with acqusition functions based on single
+    batch models.
+
+    Small values of proximal_weights corresponds to strong biasing towards recently
+    observed points, which smoothes optimization with a small potential decrese in
+    convergence rate.
```
saitcakmak, line 39: Same as above. You could also shorten the above version since it is given in detail here.

botorch/acquisition/proximal.py
```diff
@@ -0,0 +1,109 @@
...
+        acq_function: AcquisitionFunction,
+        proximal_weights: Tensor,
+    ) -> None:
+        r"""Derived Acquisition Function weighted by proximity to recently
+        observed point.
+
+        Args:
+            acq_function: The base acquisition function, operating on input tensors
+                of feature dimension `d`.
+
```
saitcakmak, line 60:
```suggestion
```

botorch/acquisition/proximal.py
```diff
@@ -0,0 +1,109 @@
...
+        """
+        Module.__init__(self)
+
+        self.acq_func = acq_function
+
+        self.register_buffer("proximal_weights", proximal_weights)
+
+        # check model for train_inputs and single batch
+        if not hasattr(self.acq_func.model, "train_inputs"):
+            raise UnsupportedError("Acqusition function model must have train_inputs.")
```
saitcakmak, line 73:
```suggestion
            raise UnsupportedError("Acquisition function model must have `train_inputs`.")
```

botorch/acquisition/proximal.py
```diff
@@ -0,0 +1,109 @@
...
+
+        if self.acq_func.model.batch_shape != torch.Size([]):
+            raise UnsupportedError(
+                "Proximal acqusition function requires batch size " "of one."
```
saitcakmak, line 77:
```suggestion
                "Proximal acquisition function requires a single batch model."
```

test/acquisition/test_proximal.py
```diff
@@ -0,0 +1,72 @@
...
+        EI_prox = ProximalAcquisitionFunction(EI, proximal_weights=proximal_weights)
+
+        ei = EI(test_X)
+        mv_normal = MultivariateNormal(train_X[-1], torch.diag(proximal_weights))
+        test_prox_weight = torch.exp(mv_normal.log_prob(test_X)) / torch.exp(
+            mv_normal.log_prob(train_X[-1])
+        )
+
+        ei_prox = EI_prox(test_X)
+        self.assertAlmostEqual(float(ei_prox), float(ei * test_prox_weight), places=5)
```
saitcakmak, line 35: You can use `self.assertTrue(torch.equal(...))` / `torch.allclose(...)`. These check that the tensor type, device and the values match (are close) across the two inputs. It would help eliminate the float casts and the for loops below.

test/acquisition/test_proximal.py
```diff
@@ -0,0 +1,72 @@
...
+from botorch.acquisition.monte_carlo import qExpectedImprovement
+from botorch.acquisition.proximal import ProximalAcquisitionFunction
+from botorch.models import SingleTaskGP
+from botorch.utils.testing import BotorchTestCase
+from torch.distributions.multivariate_normal import MultivariateNormal
+
+
+class TestProximalAcquisitionFunction(BotorchTestCase):
+    def test_proximal(self):
+        train_X = torch.rand(5, 3, device=self.device)
```
saitcakmak, line 18: Let's wrap these tests in a `for dtype in (torch.float, torch.double):` just for good measure. I don't see anything that would break here but it helps avoid minor bugs.

saitcakmak (commented):
Thanks for the changes. Left a some additional minor comments, otherwise this lgtm.

botorch/acquisition/proximal.py
```diff
@@ -0,0 +1,109 @@
...
+        Returns:
+            Base acquisition function evaluated on tensor `X` multiplied by proximal
+            weighting.
+        """
+        d = self.proximal_weights.shape[0]
+        last_X = self.acq_func.model.train_inputs[0][-1].reshape(1, 1, -1)
+        diff = X - last_X
+        L = torch.diag(torch.sqrt(self.proximal_weights)).reshape(1, d, d)
+
+        M = _batch_mahalanobis(L, diff)
```
Balandat, line 106: I still don't think using `_batch_mahalanobis` is necessary here. Since the weight matrix is diagonal by construction, there is no need to do fancy triangular solves or anything since this simplifies to the following. 
```suggestion
        M = (diff ** 2 / self.proximal_weights).sum()
```

Balandat (changes requested):
Thanks for making the updates. Besides Sait's comments, I have the following
1. I still don't think it's necessary to use `_batch_mahalanobis` here (see inline comment) 
2. You'll need to add the new module to the sphinx sources, close to here: https://github.com/pytorch/botorch/blob/main/sphinx/source/acquisition.rst#penalized-acquisition-function-wrapper. Should be straightforward to mimic the "Penalized Acquisition Function Wrapper" element for the proximal one.
3. Please make sure test coverage is up to 100% (you're missing 3 lines currently): https://github.com/pytorch/botorch/pull/919/checks?check_run_id=3418645463

roussel-ryan:
I've added changes in response to the comments. Just a quick question, is there interest in creating a similar wrapper style class to incorporate analytical / MC constraints? I see a TODO in Expected hypervolume improvement

PR #924, opened 2021-08-30
Bugfix for specifying proximal weights in a multi-dim tensor

roussel-ryan:
Fixes bug with proximal acquisition function where a user is allowed to specify a multi-dim proximal_weights tensor. Change prevents this and adds a test to verify

### DIFF
--- a/botorch/acquisition/proximal.py
+++ b/botorch/acquisition/proximal.py
@@ -74,7 +74,8 @@ class ProximalAcquisitionFunction(AcquisitionFunction):
 
         # check to make sure that weights match the training data shape
         if (
-            self.proximal_weights.shape[0]
+            len(self.proximal_weights.shape) != 1
+            or self.proximal_weights.shape[0]
             != self.acq_func.model.train_inputs[0][-1].shape[-1]
         ):
             raise ValueError(
--- a/botorch/acquisition/proximal.py
+++ b/botorch/acquisition/proximal.py
@@ -57,13 +57,14 @@ class ProximalAcquisitionFunction(AcquisitionFunction):
 
         self.acq_func = acq_function
 
-        if acq_function.X_pending is not None:
-            raise ProximalWarning(
-                "Proximal biasing behavior will be based on pending "
-                "observation points, may result in unintuitive "
-                "behavior."
-            )
-        self.X_pending = acq_function.X_pending
+        if hasattr(acq_function, "X_pending"):
+            if acq_function.X_pending is not None:
+                raise ProximalWarning(
+                    "Proximal biasing behavior will be based on pending "
+                    "observation points, may result in unintuitive "
+                    "behavior."
+                )
+            self.X_pending = acq_function.X_pending
 
         self.register_buffer("proximal_weights", proximal_weights)
 
--- a/botorch/acquisition/proximal.py
+++ b/botorch/acquisition/proximal.py
@@ -14,7 +14,6 @@ from __future__ import annotations
 import torch
 from botorch.acquisition import AcquisitionFunction
 from botorch.exceptions.errors import UnsupportedError
-from botorch.exceptions.warnings import ProximalWarning
 from botorch.utils import t_batch_mode_transform
 from torch import Tensor
 from torch.nn import Module
@@ -59,10 +58,8 @@ class ProximalAcquisitionFunction(AcquisitionFunction):
 
         if hasattr(acq_function, "X_pending"):
             if acq_function.X_pending is not None:
-                raise ProximalWarning(
-                    "Proximal biasing behavior will be based on pending "
-                    "observation points, may result in unintuitive "
-                    "behavior."
+                raise UnsupportedError(
+                    "Proximal acquisition function requires `X_pending` to be None."
                 )
             self.X_pending = acq_function.X_pending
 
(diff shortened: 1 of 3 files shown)


roussel-ryan:
There is another bug that I'm hunting down right now, I'll amend the pull request when I've found it

botorch/acquisition/proximal.py
```diff
@@ -56,6 +57,15 @@ def __init__(
...
         self.acq_func = acq_function
 
+        if hasattr(acq_function, "X_pending"):
+            if acq_function.X_pending is not None:
+                raise ProximalWarning(
+                    "Proximal biasing behavior will be based on pending "
+                    "observation points, may result in unintuitive "
+                    "behavior."
+                )
+            self.X_pending = acq_function.X_pending
```
saitcakmak, line 67: If there are pending evaluations, wouldn't you want to do the weighting w.r.t. the last pending point? Since the motivation is to avoid changing the parameters too much between evaluations, this seems more meaningful than using the last model training point. 

The pending points are typically used when we have asynchronous parallel evaluations, or if we're using sequential greedy optimization, e.g., in qNEHVI, which also happens only if we're using batch evaluations. In the paper, you motivate the proximal idea in a sequential optimization setting where changing the parameters too much is undesirable. So, I am a bit confused on whether batch / parallel evaluations and `X_pending` make sense with the proximal weighting and if so, what is the correct interpretation and what should be used as the last point.

The idea of changing parameters from the last value and parallel evaluations doesn't sound right to me (unless we have a batch of experiment where each one is changed only a little bit from their respective last values), so the only setting I can think of for `X_pending` is where we decide on the next point to evaluate before we observed the result of the current experiment (which is at `X_pending`). 

Curious to hear if you've actually used `X_pending` with proximal weighting and how you interpreted it. If you haven't and this is only to support the `X_pending` in BoTorch, I think it'd make sense figure out a natural interpretation for it.

roussel-ryan: From what I'm guessing the pending points will be appended to the end of model.train_inputs correct? In that case the proximal center is the last pending point.

To be honest I went back and forth thinking about whether to throw an error or a warning in this case. To be honest I'm not sure if there is a use case where this is advantageous, so I left the door open slightly with the warning. Its possible that this kind of local behavior can improve local convergence in high dimensions (sort of similar to TURBO?) however I'm not sure. But, I can change it to throw an UnsupportedError if that is better.

saitcakmak: > From what I'm guessing the pending points will be appended to the end of model.train_inputs correct? In that case the proximal center is the last pending point.

No, they're not appended to the model since they're yet to be evaluated. We typically append them to `X` in the acquisition function `forward` pass using the `@concatenate_pending_points`. If you wanted to use the pending point as the center here, you'd have to manually check that it is not `None` and set `last_X` accordingly. 

I am not sure what's the best way to handle them here either. So, I'd probably just play it safe and throw an error. @Balandat thoughts?

Balandat: Consider the scenario where you can run experiments but then data post-processing takes forever even though the test bed is already available again. In that setting you'd be wasting time by sitting around waiting for the data coming in. This would be a use case for the async parallel evaluation where the meaning of the "last point" is the last thing you've tested on the testbed. 

Alternatively there may be a setting where you need to pre-commit to what configurations to evaluate (for whatever weird reason, think of some group of folks needing to manually review and green-light the settings who don't want to do that on a continual basis). This problem is kind of interesting in that you need to generate a batch of configurations, and what you'd want to weight things by is not the distance of one point to the last point, but the length of the shortest path between the last point and all suggested points.

But these settings might be a bit arcane, and I'm ok to just error out with an informative warning if we don't care (enough) about them at this point. We can always follow up.

botorch/exceptions/warnings.py
```diff
@@ -49,3 +49,10 @@ class BotorchTensorDimensionWarning(BotorchWarning):
     r"""Warning raised when a tensor possibly violates a botorch convention."""
 
     pass
+
+
+class ProximalWarning(BotorchWarning):
```
Balandat, line 54: I think this is a bit too bespoke a warning - can you just use a `BoTorchWarning` here?

test/acquisition/test_proximal.py
```diff
@@ -114,6 +121,17 @@ def test_proximal(self):
...
                 )
 
+            with self.assertRaises(ValueError):
+                ProximalAcquisitionFunction(
+                    ExpectedImprovement(model, 0.0), torch.rand(3, 3)
+                )
+
+            # test for x_pending points
+            pending_acq = DummyAcquisitionFunction(model)
+            pending_acq.set_X_pending(torch.rand(3, 3))
```
Balandat, line 131:
```suggestion
            pending_acq.set_X_pending(torch.rand(3, 3, device=self.device, dtype=dtype))
```

test/acquisition/test_proximal.py
```diff
@@ -114,6 +121,17 @@ def test_proximal(self):
                     ExpectedImprovement(model, 0.0), proximal_weights[:1]
                 )
 
+            with self.assertRaises(ValueError):
+                ProximalAcquisitionFunction(
+                    ExpectedImprovement(model, 0.0), torch.rand(3, 3)
```
Balandat, line 126:
```suggestion
                    ExpectedImprovement(model, 0.0), torch.rand(3, 3, device=self.device, dtype=dtype)
```

botorch/acquisition/proximal.py
roussel-ryan: Right now I don't have time to look into this too deeply so I'll replace the warning with an error and we can revisit later if necessary.

Issue #938, opened 2021-09-09
Difficulty with using qKG method along with optimize_acqf_discrete and optimize_acqf_mixed functions


sambitmishra98:
I want to use optimize_acqf_discrete and optimize_acqf_mixed but am finding it difficult to use with qKG method. I was able to use qKG with optimize_acqf thanks to the documentation examples. Can you guys help me by adding a simple example for these functions in the documentation?

eytan:
cc @danielrjiang, @dme65 , @saitcakmak

eytan:
@sambitmishra98  you tell us a little bit more about what you tried along with a few simple reproducible examples?  Some inline code or collabs could be helpful.

sambitmishra98:
Thanks a lot for your help on this!
I executed it in Google Colab. Below is the link: [My Code](https://colab.research.google.com/drive/10mV4bhqmcDzuOFy46Xmpm0PyhSjv6uZs?usp=sharing)

For my own practice, I basically used the following link and tried to replace `optimize_acqf` with `optimize_acqf_discrete`, hoping for the best: [Code from examples](https://botorch.org/tutorials/one_shot_kg)

For my research, I wish to implement BO to a given set of integer variables and float variables. I just want to perform BO once and obtain one candidate as output. My end goal is using `optimize_acqf_mixed` effectively, so any example using this function would help.

Thanks again!

saitcakmak:
Hi @sambitmishra98. There's a fundamental issue with the qKG that makes it prohibitively expensive to optimize over mixed or discrete input spaces. 

In qKG, for `q` candidates and `n_f` fantasies, the solution tensor is `batch x (q + n_f) x d`-dim.  Suppose that you only have the first column of the input be discrete, and you have `k` values it can take. Since you're actually optimizing over `q + n_f` solutions jointly, you end up with `k^(q + n_f)` discrete combinations to evaluate (and optimize the continuous columns over). You can see how this blows up very fast.

Going back to your original question with `optimize_acqf_mixed/discrete`, they do not support qKG optimization properly. In `optimize_acqf_mixed`, you specify `fixed_features_list` which is a list of values a given column of the `d` dim tensor can take. Since qKG input is `(q + n_f) x d`, the same value will be applied to all q-batch dimensions, forcing both your candidates and the fantasy solutions to consider the same discrete value. It also uses sequential optimization if `q > 1`, which leads to other issues I will not discuss. 
[...]

PR #939, opened 2021-09-09
Add error message for one shot acqf in optimize_acqf_discrete

saitcakmak:
## Motivation

As noted in #938, use of one-shot acquisition functions in optimize_acqf_mixed/discrete leads to errors down the line that are hard to interpret. This PR adds an explicit error message noting that this is not supported.

Yes

## Test Plan

Units.

### DIFF
--- a/botorch/optim/optimize.py
+++ b/botorch/optim/optimize.py
@@ -18,6 +18,7 @@ from botorch.acquisition.acquisition import (
     OneShotAcquisitionFunction,
 )
 from botorch.acquisition.knowledge_gradient import qKnowledgeGradient
+from botorch.exceptions import UnsupportedError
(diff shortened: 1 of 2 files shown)


saitcakmak:
I guess `optimize_acqf_mixed` has a use in the multi-fidelity setting as in the [tutorial](https://botorch.org/tutorials/discrete_multi_fidelity_bo). Maybe we could add an exception for MFKG there?

Balandat:
@danielrjiang what's your take on this? If there is a general legitimate use case for `optimize_acqf_mixed` with one-shot acquisition functions I'd rather special case the error than an allowlist so we don't accidentally disallow legitimate uses.

PR #965, opened 2021-11-03
speed up MVaR by pruning y_grid

sdaulton:
Summary: see title. This reduces the size of y_grid from `n_w x n_w` to `(1-alpha)n_w x (1-alpha)n_w`  for each data point

Differential Revision: D32140376

### DIFF
--- a/botorch/acquisition/multi_objective/multi_output_risk_measures.py
+++ b/botorch/acquisition/multi_objective/multi_output_risk_measures.py
@@ -273,10 +273,19 @@ class MVaR(MultiOutputRiskMeasureMCObjective):
         if m != 2:  # pragma: no cover
             raise ValueError("`get_mvar_set_cpu` only supports `m=2` outcomes!")
         # Generate sets of all unique values in each output dimension.
-        unique_outcomes = [Y[:, i].unique(sorted=True).tolist()[::-1] for i in range(m)]
+        # Note that points in MVaR are bounded from above by the
+        # independent VaR of each objective. Hence, we only need to
+        # consider the unique outcomes that are less than or equal to
+        # the VaR of the independent objectives
+        var_alpha_idx = ceil(self.alpha * self.n_w) - 1
+        Y_sorted = Y.topk(Y.shape[0] - var_alpha_idx, dim=0, largest=False).values
+        unique_outcomes_list = [
+            Y_sorted[:, i].unique().tolist()[::-1] for i in range(m)
+        ]
         # Convert this into a list of m dictionaries mapping values to indices.
         unique_outcomes = [
-            dict(zip(outcomes, range(len(outcomes)))) for outcomes in unique_outcomes
+            dict(zip(outcomes, range(len(outcomes))))
+            for outcomes in unique_outcomes_list
         ]
         # Initialize a tensor counting the number of points in Y that a given grid point
         # is dominated by. This will essentially be a non-normalized CDF.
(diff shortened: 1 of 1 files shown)


sdaulton:
cc @saitcakmak

botorch/acquisition/multi_objective/multi_output_risk_measures.py
```diff
@@ -273,10 +273,19 @@ def get_mvar_set_cpu(self, Y: Tensor) -> Tensor:
         if m != 2:  # pragma: no cover
             raise ValueError("`get_mvar_set_cpu` only supports `m=2` outcomes!")
         # Generate sets of all unique values in each output dimension.
-        unique_outcomes = [Y[:, i].unique(sorted=True).tolist()[::-1] for i in range(m)]
+        # Note that points in MVaR are bounded from above by the
+        # independent VaR of each objective. Hence, we only need to
+        # consider the unique outcomes that are less than or equal to
+        # the VaR of the independent objectives
+        var_alpha_idx = ceil(self.alpha * self.n_w) - 1
+        Y_sorted = Y.sort(dim=0).values[: Y.shape[0] - var_alpha_idx]
```
Balandat, line 281: Should we use `topk(largest =False)` here rather than sorting and indexing?

botorch/acquisition/multi_objective/multi_output_risk_measures.py
```diff
@@ -343,21 +362,30 @@ def get_mvar_set_gpu(self, Y: Tensor) -> Tensor:
         if Y.dim() == 2:
             Y = Y.unsqueeze(0)
         batch, m = Y.shape[0], Y.shape[-1]
+        # Note that points in MVaR are bounded from above by the
+        # independent VaR of each objective. Hence, we only need to
+        # consider the unique outcomes that are less than or equal to
+        # the VaR of the independent objectives
+        var_alpha_idx = ceil(self.alpha * self.n_w) - 1
+        n_points = Y.shape[-2] - var_alpha_idx
+        Y_sorted = Y.sort(dim=-2).values[..., :n_points, :]
```
Balandat, line 371: again, use `topk` here?

botorch/acquisition/multi_objective/multi_output_risk_measures.py
```diff
@@ -273,10 +273,19 @@ def get_mvar_set_cpu(self, Y: Tensor) -> Tensor:
         if m != 2:  # pragma: no cover
             raise ValueError("`get_mvar_set_cpu` only supports `m=2` outcomes!")
         # Generate sets of all unique values in each output dimension.
-        unique_outcomes = [Y[:, i].unique(sorted=True).tolist()[::-1] for i in range(m)]
+        # Note that points in MVaR are bounded from above by the
+        # independent VaR of each objective. Hence, we only need to
+        # consider the unique outcomes that are less than or equal to
+        # the VaR of the independent objectives
+        var_alpha_idx = ceil(self.alpha * self.n_w) - 1
+        Y_sorted = Y.topk(Y.shape[0] - var_alpha_idx, dim=0, largest=False).values
```
saitcakmak, line 281: Should we pass in `sorted=True`? In my testing, the return values are sorted regardless. But PyTorch does not explicitly document what should happen when `sorted=False`. Might be safer since the algorithm depends on the values being sorted.

saitcakmak (approved):
Lgtm!

Since this may significantly change the runtime of the algorithms, it may be worth re-thinking about when to dispatch to `get_mvar_set_cpu/gpu`.

botorch/acquisition/multi_objective/multi_output_risk_measures.py
Balandat: The default value of the `sorted` arg is `True`, I believe? https://pytorch.org/docs/stable/generated/torch.topk.html

saitcakmak: That explains it. I must've completely missed that somehow.

PR #977, opened 2021-11-12
update nehvi citation

sdaulton:
Summary: see title

Differential Revision: D32400736


Balandat (approved):

Balandat:
My guess is that the tests are failing on gpytorch master due to https://github.com/cornellius-gp/gpytorch/pull/1813. Will have to dig into that a bit to understand what exactly is going wrong here.

PR #991, opened 2021-11-23
Short-cut t_batch_mode_transform decorator on non-tensor inputs

Balandat:
This essentially makes `t_batch_mode_tarnsform` a nullop in case the argument `X` passed to the acquisition function is not a `torch.Tensor` object. This allows using acquisition functions that use models with non non-standard input types such as strings, which is the case in some applications.

Currently this just touches the decorator; in the future we should consider changing the types and signatures of the acquisition functions and models throughout to natively support this more generally.

cc @wjmaddox

### DIFF
--- a/botorch/utils/transforms.py
+++ b/botorch/utils/transforms.py
@@ -12,10 +12,12 @@ from __future__ import annotations
 
 import warnings
 from functools import wraps
-from typing import Any, Callable, List, Optional
+from typing import Any, Callable, List, Optional, TypeVar
 
 import torch
 from torch import Tensor
+
+ACQF = TypeVar("AcquisitionFunction")
 
 
 def squeeze_last_dim(Y: Tensor) -> Tensor:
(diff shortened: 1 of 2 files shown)


Balandat:
@wjmaddox is this something you need in the next release? Or is it ok for this to wait until the next one?

wjmaddox:
This would be really helpful to have in the next release, but if it's not tested, we can wait.

Balandat:
I think it should be ok to add this as a "hidden" feature, let me do some more testing and try to get that in this week.

Balandat:
@saitcakmak do you see any issues with doing this? Ultimately I'd like to support other input types better more generally, but this is kind of a stop gap measure that makes it easy to do so as long as the model is doing proper things with the input.

botorch/utils/transforms.py
```diff
@@ -202,9 +204,16 @@ def t_batch_mode_transform(
...
 
-    def decorator(method: Callable[[Any, Tensor], Any]) -> Callable[[Any, Tensor], Any]:
+    def decorator(
+        method: Callable[[ACQF, Any], Any],
+    ) -> Callable[[ACQF, Any], Any]:
         @wraps(method)
-        def decorated(acqf: Any, X: Tensor, *args: Any, **kwargs: Any) -> Any:
+        def decorated(acqf: ACQF, X: Any, *args: Any, **kwargs: Any) -> Any:
+
+            # Allow using acquisition funcitons for other inputs (e.g. lists of strings)
```
saitcakmak, line 213:
```suggestion
            # Allow using acquisition functions with other inputs (e.g. lists of strings).
```

saitcakmak (approved):
This looks fine to me. It skips the check only if `X` is not a tensor. If the `acqf / method`  does not expect that, you would get an error anyways.

PR #1007, opened 2021-12-01
Handle NotPSDError and hitting maxiter in fit_gpytorch_model

saitcakmak:
## Motivation

The current behavior of `fit_gpytorch_model` is to error out when it receives a `NotPSDError` and to count hitting `maxiter` during optimization as an optimization failure.

Labeling `maxiter` as an optimization failure leads to retrying the model fit, and likely reaching `maxiter` in each try and labeling all tries as failures. This ends up using more effort than the user intended to (by passing in `maxiter` in the first place), while offering no real benefit. The proposed solution here is to simply count hitting `maxiter` as a successful model fit and to return the fit obtained with the given budget.

The `NotPSDError` during model fit is one of the most common reasons that leads to me having to restart an experiment. What I observed is that when the first try raises `NotPSDError`, a second try starting from initial parameters sampled from the priors will most of the time produce a successful model fit. So, this proposes to catch the `NotPSDError` and retry the model fit with new initial parameter values. 

Yes

## Test Plan

Unittests.

## Related PRs

cc @sdaulton

### DIFF
--- a/botorch/fit.py
+++ b/botorch/fit.py
@@ -23,6 +23,7 @@ from botorch.optim.fit import fit_gpytorch_scipy
 from botorch.optim.utils import sample_all_priors
 from gpytorch.mlls.marginal_log_likelihood import MarginalLogLikelihood
 from gpytorch.mlls.sum_marginal_log_likelihood import SumMarginalLogLikelihood
+from gpytorch.utils.errors import NotPSDError
 
 
 FAILED_CONVERSION_MSG = (
@@ -80,7 +81,6 @@ def fit_gpytorch_model(
                 tf = mll.model.outcome_transform
                 mll.model.outcome_transform = None
             model_list = batched_to_model_list(mll.model)
-            model_ = model_list_to_batched(model_list)
             mll_ = SumMarginalLogLikelihood(model_list.likelihood, model_list)
             fit_gpytorch_model(
                 mll=mll_,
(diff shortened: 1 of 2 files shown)


botorch/fit.py
```diff
@@ -121,12 +121,22 @@ def fit_gpytorch_model(
...
+            except NotPSDError:
+                retry += 1
+                logging.log(
+                    logging.DEBUG,
+                    f"Fitting failed on try {retry} due to a NotPSDError.",
+                )
+                continue
         has_optwarning = False
         for w in ws:
+            # Do not count reaching `maxiter` as an optimization failure/
```
Balandat, line 135:
```suggestion
            # Do not count reaching `maxiter` as an optimization failure
```

test/test_fit.py
```diff
@@ -268,8 +237,9 @@ def test_fit_gpytorch_model_singular(self):
             )
             mll = ExactMarginalLogLikelihood(gp.likelihood, gp)
             mll.to(device=self.device, dtype=dtype)
-            with self.assertRaises(NotPSDError):
+            with self.assertLogs(level="DEBUG") as logs:
                 fit_gpytorch_model(mll, options=options, max_retries=2)
+            self.assertTrue(any(["NotPSDError" in log for log in logs.output]))
```
Balandat, line 242:
```suggestion
            self.assertTrue(any("NotPSDError" in log for log in logs.output))
```

test/test_fit.py
```diff
@@ -278,6 +248,18 @@ def test_fit_gpytorch_model_singular(self):
...
+                mll = self._getModel()
+                with self.assertLogs(level="DEBUG") as logs:
+                    fit_gpytorch_model(mll, max_retries=2)
+                self.assertEqual(
+                    logs.output,
+                    [
+                        "DEBUG:root:Fitting failed on try 1 due to a NotPSDError.",
+                        "DEBUG:root:Fitting failed on try 2 due to a NotPSDError.",
+                    ],
+                )
```
Balandat, line 262: The equality assertion here could be quite brittle if something benign changed (e.g. the logging prefix, or if other stuff is logged on the gpytorch end). Maybe we can just assert that "Fitting failed on try 1 due to a NotPSDError." and "Fitting failed on try 2 due to a NotPSDError." are both contained in the logs (rather than doing an exact log match)?

botorch/fit.py
```diff
@@ -121,12 +121,22 @@ def fit_gpytorch_model(
...
+            if "ITERATIONS REACHED LIMIT" in str(w.message):
+                continue
```
Balandat, line 137: Should we log this on the DEBUG level? It might be good for users to know when their optimization doesn't converge but ends with hitting the iteration limit.

Balandat (commented):
Thanks, this makes a lot of sense to me! A few comments inline, but this looks good overall.

Balandat (approved):

Balandat:
Looks like we lost a tiny bit of test coverage since we don't actually fail fitting anymore other than with the NonPSDError. Would you mind adding a test for that (e.g.. fail with some other exception)?

saitcakmak:
I didn't realize we needed `settings.debug(True)` for warnings to work. I spent some time wondering why my warning wasn't being caught... 
Also, I can't find a way to use `mock` to produce warnings. I was looking for something like `side_effect` but it doesn't seem to support warnings. If you know of a way of wrapping a function with a warning, I'd love to learn!

PR #1034, opened 2021-12-12
make risk measure more memory efficient

sdaulton:
Summary:
Using torch.quantile is way more efficient because it does not create large tensors for values and indices (which are the same size as the input).

torch.topk also yields memory improvements

Differential Revision: D33037848

### DIFF
--- a/botorch/acquisition/risk_measures.py
+++ b/botorch/acquisition/risk_measures.py
@@ -24,6 +24,7 @@ from abc import ABC, abstractmethod
 from math import ceil
 from typing import Optional
 
+import torch
 from botorch.acquisition.objective import MCAcquisitionObjective
 from torch import Tensor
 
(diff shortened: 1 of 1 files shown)


saitcakmak (approved):
Lgtm. The CVaR is pretty safe and the VaR was tested offline to make sure it has the same behavior as the original implementation.

botorch/acquisition/risk_measures.py
```diff
@@ -170,8 +192,20 @@ def forward(self, samples: Tensor, X: Optional[Tensor] = None) -> Tensor:
...
+        # this is equivalent to sorting along dim=-1 in descending order
+        # and taking the values at index self.alpha_idx. E.g.
+        # >>> sorted_res = prepared_samples.sort(dim=-1, descending=True)
+        # >>> sorted_res.values[..., self.alpha_idx]
+        # Using quantile is far more memory efficient since `torch.sort`
+        # produces values and indices tensors with shape
+        # `sample_shape x batch_shape x (q * n_w) x m`
+        return torch.quantile(
+            input=prepared_samples,
+            q=self._q,
```
Balandat, line 204: Why not just use `alpha` as the quantile here? Seem like that's what you'd want?

saitcakmak: It produces different results due to the way `quantile` is implemented. 

If you have `n_w=20, alpha=0.5`, you want the result to be `sorted_samples[..., 9]`. Using `q=alpha` with `quantile` will return `sorted_samples[..., 10]`. Technically, both are totally fine since the result is a consistent estimator of VaR as n_w -> \infty. (I'd actually go for `interpolation = "linear"` if we didn't care about the result being different). 

The reason we wanted the result to be consistent with the original implementation is that i) it is consistent with the way CVaR is implemented, ii) it is consistent with the definition of the VaR that is used in the papers we reference.

PR #1035, opened 2021-12-14
Store tensor attributes of input transforms as buffers

saitcakmak:
## Motivation

If we store the tensor attributes as `self.attr = tensor_attr`, the `Module.to(...)` calls don't affect these attributes. To actually change the device / dtype of this attribute, we manually need to call `module.attr.to(...)`, which is less than ideal particularly if the `module` is a sub-module of a `Model`. Registering the attribute as a buffer solves this issue.

Yes

## Test Plan

Unit tests.

## Related PRs


Balandat (approved):
Awesome, thanks for the PR!

PR #1026, opened 2021-12-09
Modify qNEHVI to support DeterministicModels

saitcakmak:
## Motivation

Fixes #966 by disabling `_cache_root` and `_set_sampler` when using a deterministic model with qNEHVI.

### Have you read the [Contributing Guidelines on pull requests](https://github.com/pytorch/botorch/blob/main/CONTRIBUTING.md#pull-requests)?

Yes

## Test Plan

Unit tests.

## Related PRs

(If this PR adds or changes functionality, please take some time to update the docs at https://github.com/pytorch/botorch, and link to your PR here.)
#896 - I will update that once I figure out a good way to support set valued objectives as discussed offline.
cc @sdaulton

PR #1037, opened 2021-12-15
Modify NEHVI to support MTGPs

### LINKED ISSUE #1036: [question] Can qNEHVI be used with KroneckerMultiTaskGP?
Following [this tutorial](https://botorch.org/tutorials/multi_objective_bo), I was trying to swap the `ModelListGP` model for a `KroneckerMultiTaskGP`, with this modification (leaving everything else the same):

```
def initialize_model(train_x, train_obj):
    model = KroneckerMultiTaskGP(train_x, train_obj, outcome_transform=Standardize(m=train_obj.shape[-1]))
    return mll, model
```
With this model, the BO loop fails at `optimize_qnehvi_and_get_observation()` with error:

```
AttributeError: 'TransformedPosterior' object has no attribute 'mvn'
```
Does that mean that `KroneckerMultiTaskGP` cannot be used with qNEHVI as it doesn't have the `MultitaskMultivariateNormal` attribute or am I misinterpreting this? Are any other changes necessary?

Thanks a lot!

saitcakmak:
## Motivation

As seen in #1036, qNEHVI currently does not support KroneckerMTGP, or any other GP that uses Matheron's rule sampling. This modifies the `_set_sampler` to add support for these.

With this, `HigherOrderGP` is also supported as long as it is coupled with an `objective` that reduces the `batch x q x output_shape` samples to `batch x q x m` (where `m` agrees with the `ref_point`, of course).

In addition, I sneaked in a `GenericMCMultiOutputObjective`.

Resolves #1036.

Yes.

## Test Plan

Unittests.

## Related PRs

#1026 

cc @wjmaddox

### DIFF
--- a/botorch/acquisition/multi_objective/monte_carlo.py
+++ b/botorch/acquisition/multi_objective/monte_carlo.py
@@ -40,9 +40,11 @@ from botorch.acquisition.multi_objective.utils import (
 )
 from botorch.exceptions.errors import UnsupportedError
 from botorch.exceptions.warnings import BotorchWarning
+from botorch.models.deterministic import DeterministicModel
 from botorch.models.model import Model
 from botorch.models.model_list_gp_regression import ModelListGP
 from botorch.models.multitask import MultiTaskGP
+from botorch.posteriors import DeterministicPosterior
 from botorch.posteriors.gpytorch import GPyTorchPosterior
 from botorch.posteriors.posterior import Posterior
 from botorch.sampling.samplers import MCSampler, SobolQMCNormalSampler
@@ -449,6 +451,8 @@ class qNoisyExpectedHypervolumeImprovement(qExpectedHypervolumeImprovement):
             self.p_class = FastNondominatedPartitioning
         self.register_buffer("_X_baseline", X_baseline)
         self.register_buffer("_X_baseline_and_pending", X_baseline)
+        if isinstance(model, DeterministicModel):
+            cache_root = False
         self._cache_root = cache_root
         self.register_buffer(
             "cache_pending",
@@ -687,6 +691,7 @@ class qNoisyExpectedHypervolumeImprovement(qExpectedHypervolumeImprovement):
             if (
                 self.X_baseline.shape[0] > 0
                 and self.base_sampler.base_samples is not None
+                and not isinstance(posterior, DeterministicPosterior)
             ):
                 current_base_samples = self.base_sampler.base_samples.detach().clone()
                 view_shape = (
--- a/botorch/acquisition/multi_objective/monte_carlo.py
+++ b/botorch/acquisition/multi_objective/monte_carlo.py
@@ -735,7 +735,15 @@ class qNoisyExpectedHypervolumeImprovement(qExpectedHypervolumeImprovement):
                     expanded_shape
                 )
                 if self._uses_matheron:
-                    self.sampler.base_samples[..., :end_idx] = expanded_samples
+                    n_train_samples = current_base_samples.shape[-1] // 2
+                    # The train base samples.
+                    self.sampler.base_samples[..., :n_train_samples] = expanded_samples[
+                        ..., :n_train_samples
+                    ]
+                    # The train noise base samples.
+                    self.sampler.base_samples[
+                        ..., -n_train_samples:
+                    ] = expanded_samples[..., -n_train_samples:]
                 else:
                     self.sampler.base_samples[..., :end_idx, :] = expanded_samples
                 # update cached subset indices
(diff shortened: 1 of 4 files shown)


botorch/acquisition/multi_objective/monte_carlo.py
```diff
@@ -747,7 +780,13 @@ def _get_f_X_samples(self, posterior: GPyTorchPosterior, q: int) -> Tensor:
 
         # TODO: improve efficiency for multi-task models
         samples = self.sampler(posterior)
-        return samples[..., -q:, :]
+        if isinstance(self.model, HigherOrderGP):
+            # Select the correct q-batch dimension for HOGP.
+            q_dim = -self.model._num_dimensions
+            q_idcs = torch.arange(-q, 0) + samples.shape[q_dim]
```
sdaulton, line 786: We need to set the device in arange to run this on cuda

botorch/acquisition/multi_objective/monte_carlo.py
```diff
@@ -704,10 +733,14 @@ def _set_sampler(
...
-                end_idx = current_base_samples.shape[-2]
-                self.sampler.base_samples[..., :end_idx, :] = current_base_samples.view(
-                    view_shape
-                ).expand(expanded_shape)
+                end_idx = current_base_samples.shape[-1 if self._uses_matheron else -2]
+                expanded_samples = current_base_samples.view(view_shape).expand(
+                    expanded_shape
+                )
+                if self._uses_matheron:
+                    self.sampler.base_samples[..., :end_idx] = expanded_samples
```
sdaulton, line 741: What is the shape of base samples here w/ and w/o matheron's rule?

sdaulton (commented):
Thanks for putting this up!

botorch/acquisition/multi_objective/monte_carlo.py
saitcakmak: Interesting, `gpu_tensor[cpu_idx_tensor]` works fine but `gpu_tensor.index_select(idx, cpu_idx_tensor)` doesn't.

botorch/acquisition/multi_objective/monte_carlo.py
saitcakmak: Without Matheron, it is `batch x q_full x m`. With Matheron, it is `batch x (product(output_shape) * (q_full + 2 * n_train))` where `output_shape` is the `output_shape` of HOGP or `m` for KroneckerMTGP

samuelstanton:
+1 on getting this merged, would very much like to use NEHVI with MTGPs.

botorch/acquisition/multi_objective/monte_carlo.py
sdaulton: It might be good to check that the cached posterior samples for X_baseline are the same before/after updating the sampler here, in light of the bug that I recently fixed https://github.com/pytorch/botorch/pull/1057. I haven't reviewed how the base samples are being reshaped/permuted with Matheron

saitcakmak: Good call. I found a bug and fixed it. The base samples for Matheron are structured as `train_base_samples, test_base_samples, train_noise_base_samples`. So, setting `self.sampler.base_samples[..., :end_idx] = expanded_samples` was not producing the correct samples and instead we need to split the original base samples into two and add it to the beginning and the end.

test/acquisition/multi_objective/test_monte_carlo.py
```diff
@@ -3,7 +3,6 @@
 #
 # This source code is licensed under the MIT license found in the
 # LICENSE file in the root directory of this source tree.
-
```
Balandat, line 6: Can you bring this blank line back, please? Our intern linter is complaining about this.

botorch/acquisition/multi_objective/monte_carlo.py
```diff
@@ -415,6 +426,17 @@ def __init__(
...
+            warnings.warn(
+                message=(
+                    "sampler.batch_range is not (0, -1). qNEHVI requires that the "
+                    "sampler.batch_range is (0, -1) with GPs that use Matheron's rule "
+                    "for sampling, in order to properly collapse batch dimensions. "
+                    "Overwriting the batch_range with (0, -1)!"
+                ),
+                category=BotorchWarning,
+            )
+            self.sampler.batch_range = (0, -1)
```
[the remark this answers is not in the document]
saitcakmak: So, this essentially makes SAA work correctly by making `collapse_batch_dims` function properly. If you don't do this,  when `X` is `b x q x d` the sampler base sample shape ends up being `num_samples x b x m'` where `m'` is the proper # of base samples for Matheron. This means, if I take a `q x d`-dim `X` and evaluate `X.repeat(5, 1, 1)`, I will get `5` different acquisition values.

The reason I went for overwriting is that we typically don't set the `batch_range` of the sampler and just use the default `(0, -2)` because that is what you want for every other model. So, I figured we can just correct the default here. Maybe an error is more fool-proof. Let the user fix it themselves

PR #1043, opened 2021-12-20
Fix split chunks in sample_cached_cholesky

saitcakmak:
## Motivation

In `sample_cached_cholesky` if `bottom_rows.shape[-1] - q < bottom_rows.shape[-1] / 2`, the split call produces more than two tensors, which errors out during assignment. This happens if `q` is larger than the number of baseline points. This fixes the error by ~~replacing the `torch.split` call with explicit indexing.~~ using list of chunk sizes in `split`.

Yes.

## Test Plan

Update units.

## Related PRs

cc @sdaulton


sdaulton (approved):
good catch. thanks!

PR #1047, opened 2021-12-23
Make Normalize input transform input column specific

jduerholt:
## Motivation

This PR adds the possibility to apply a `Normalize` input transform only to specific input columns specified via a list of indices (in analogy to the `Log10` transform). This makes it possible to combine it with other transforms or intentionally ignore inputs in the transform.

Yes

## Test Plan

Unit tests.

## Related PRs

### DIFF
--- a/botorch/models/transforms/input.py
+++ b/botorch/models/transforms/input.py
@@ -343,9 +343,9 @@ class Normalize(ReversibleInputTransform, Module):
                 zero errors.
         """
         super().__init__()
-        if (indices is None) or (len(indices) == 0):
-            self.register_buffer("indices", torch.arange(d))
-        else:
+        if (indices is not None) and (len(indices) == 0):
+            raise ValueError("`indices` list is empty!")
+        if (indices is not None) and (len(indices) > 0):
             indices = torch.tensor(indices, dtype=torch.long)
             if len(indices) > d:
                 raise ValueError("Can provide at most `d` indices!")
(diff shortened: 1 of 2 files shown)


botorch/models/transforms/input.py
```diff
@@ -340,6 +343,21 @@ def __init__(
                 zero errors.
         """
         super().__init__()
+        if (indices is None) or (len(indices) == 0):
+            self.register_buffer(
+                "indices", torch.tensor(list(range(d)), dtype=torch.long)
+            )
```
Balandat, line 349:
```suggestion
            self.register_buffer("indices", torch.arange(d))
```

Balandat, line 349: I wonder if a slightly better design would be to just not register an `indices` buffer and then then we just gate the indexing in the `_transform` and `_untransform` methods by sth like `if hasattr(self, "indices")):`. That would have the advantage of not doing a nullop clone + indexing operation in the all-indices case, which I anticipate to be the most common one.

jduerholt: Good, idea. I will change it accordingly.

botorch/models/transforms/input.py
```diff
@@ -340,6 +343,21 @@ def __init__(
                 zero errors.
         """
         super().__init__()
+        if (indices is None) or (len(indices) == 0):
```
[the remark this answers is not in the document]
jduerholt: I will error it out.

jduerholt:
Hi @Balandat, 

I implemented your suggested changes. `indices` are now not anymore registered as a buffer. Instead it is gated by a `hasattr` call. This change make the `equals` method a bit more complicated, but I think this is ok. 

Just have a look yourself ;-)

Best,

Johannes

P.S.: I wish you a happy new year.

Balandat:
Happy new year to you as well, and thanks for updating the PR! Will merge this soon.

PR #1050, opened 2021-12-29
Discrete KG and Continuous KG with Unittests

JuanUngredda:
This PR takes into account the revisions done last time, the continuous KG implementation and the unittests. I decided to simply send a new PR request hehe.


JuanUngredda:
Sorry! This wasn't intended for the FB group

PR #1075, opened 2022-02-09
Update tutorials to use "sample_around_best": True

saitcakmak:
## Motivation

Using `"sample_around_best": True` should lead to improved optimization performance.

Yest

## Test Plan

TODO: Re-run the notebooks, so the outputs reflect the changes.

## Related PRs

cc @sdaulton


Balandat:
> are there any cases where we wouldn't want to sample_around_best?

Probably not? I guess that requires some additional benchmarking

> if it's so commonly useful for AF optimization why hide it in a options blob?

I would like to avoid adding a long rats tail of boolean options to our interfaces, that will soon become unwieldy. If we want to enable that by default we can do that without exposing it as a separate option.

dme65:
The main issue I can imagine with this is that the initialization becomes too greedy and we fail to uncover a global optimum of the acquisition function that is far from the best point(s). This on its own may not actually be a bad thing since it is similar to what TuRBO/MORBO do and we may just end up avoiding points that we probably didn't want to pick in the first place. I think there will be a heavy dependency of the dimensionality here where this will probably always help in the high-dimensional setting, but it may make things worse for simpler low-dimensional problems. I echo what @Balandat said that some benchmarking would be interesting.

sdaulton:
Thanks for putting this up @saitcakmak!

I think there are very few situations where using `sample_around_best` would perform worse. It is worth noting that `sample_around_best` does not replace the space filling (sobol) initialization heuristic. Rather, the initial conditions consist of `raw_samples` points from the global sobol heuristic and `raw_samples` points from the `sample_around_best` heuristic. Hence, if the sobol points identify a design far from the current best with higher acquisition value than points around the current best, then that design will be preferred in the initialization heuristic. 

Using `sample_around_best` makes the heuristic slightly more greedy, so the one case where this could be suboptimal is if one of the sobol points with lower acquisition value is not selected as a starting point for gradient optimization, but is proximal to a better optima of the acquisition surface. However, there is still additional tempering since we do Boltzmann sampling on the acquisition values of the raw samples which should help mitigate this. Using more random restarts would also help mitigate this. 

The main benefit of `sample_around_best` is mitigate the issue of all raw samples having zero acquisition value. This can happen for improvement-based acquisition functions when the optima are in tiny regions of the design space. For example in ZDT1, the Pareto frontier is in a small sliver of the design space and the acquisition surface quickly becomes zero basically everywhere, expect for right around the previously evaluated pareto optimal designs. Using `sample_around_best` makes acquisition optimization significantly more robust to these scenarios (see attached plot).

zdt1_nehvi_sample_around_best

In the high-dimensional case, the `sample_around_best` heuristic only perturbs a subset of the dimensions of the best point(s) (and again this complements and does not replace the global sobol heuristic). So I would expect there to be a significant performance improvement.

The main challenge to turning this on by default is that it would require unifying an interface for extracting previously evaluated designs(or best designs) from botorch models and acquisition functions. The current error handling is fairly robust, but may not cover every single edge case for any conceivable (e.g. non-gpytorch) model. @Balandat and I have chatted about adding an `X_baseline`/`train_inputs` property to all botorch models and acquisition functions, but that change would touch a lot of code. If we want to turn this on by default, we should explore that refactor.

sdaulton:
@saitcakmak What is the status of this PR? It would great to get his merged in

saitcakmak:
@sdaulton We're waiting on the benchmarking suite to decide whether to set `sample_around_best` `True` by default.

esantorella:
> We're waiting on the benchmarking suite to decide whether to set sample_around_best True by default.

@saitcakmak  is this PR still relevant?

saitcakmak:
The PR itself is not that relevant (quite a bit outdated). Making these changes to the default configs is still worth investigating though. I never got around to properly benchmark this since we had a bunch of reproducibility bugs in the benchmarking suite back then.

One thing to keep in mind: The current implementation of `sample_around_best` doubles the number of `raw_samples`. So, turning it on and off doesn't really give us an apples to apples comparison. If we're going to benchmark this, we should make sure they use the same number of `raw_samples`.

PR #1088, opened 2022-02-17
Support heteroscedastic perturbations in InputPerturbations

saitcakmak:
Summary: Allows passing in a callable that generates the perturbations for any given `X`, thus supporting heteroscedastic perturbations.

Differential Revision: D34320700

### DIFF
--- a/botorch/models/transforms/input.py
+++ b/botorch/models/transforms/input.py
@@ -18,7 +18,7 @@ from __future__ import annotations
 
 from abc import ABC, abstractmethod
 from collections import OrderedDict
-from typing import List, Optional, Union
+from typing import Callable, List, Optional, Union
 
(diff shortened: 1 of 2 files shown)


botorch/models/transforms/input.py
```diff
@@ -1098,7 +1098,9 @@ def __init__(
 
         Args:
             perturbation_set: An `n_p x d`-dim tensor denoting the perturbations
-                to be added to the inputs.
+                to be added to the inputs. Alternatively, this can be a callable that
```
Balandat, line 1101: Can you update the type of the arg then to be `Union[Tensor, Callable[[Tensor], Tensor]`?

Balandat (approved):

PR #1105, opened 2022-03-04
Use BOTORCH_MODULAR in tutorials with Ax

saitcakmak:
Summary:
Update the Ax-related tutorials in BoTorch to use the BOTORCH_MODULAR interface.

For starters, this updates custom BoTorch model tutorial. I will update others based on feedback on the style.

Differential Revision: D34646421

### DIFF
--- a/tutorials/custom_acquisition.ipynb
+++ b/tutorials/custom_acquisition.ipynb
@@ -1,390 +1,2269 @@
 {
- "cells": [
-  {
-   "cell_type": "markdown",
-   "metadata": {
-    "collapsed": true
-   },
-   "source": [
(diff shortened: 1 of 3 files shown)


Balandat:
Thanks for updating this! I wonder if instead of the Ax Developer API we should instead (or in addition) use the Service API in this tutorial? This would probably make it more concise and easier for people to follow. @lena-kashtelyan, any thoughts on this?

Balandat:
Thanks for updating this! I wonder if instead of the Ax Developer API we should instead (or in addition) use the Service API in this tutorial? This would probably make it more concise and easier for people to follow. @lena-kashtelyan, any thoughts on this?

PR #1109, opened 2022-03-05
Fix smoke_test bug

saitcakmak:
Fixes the bug with handling of `SMOKE_TEST` environmental variable while running tutorials. The issue was that `subprocess.run` does not accept boolean environmental variables, so both `true` and `false` were being interpreted as non-empty strings by Python.

## Test Plan

Tested by adding the following in the tutorials.
```
if SMOKE_TEST:
    raise RuntimeError("Running w/ smoke test")
else:
    raise NotImplementedError("Running w/o smoke test")
```

### DIFF
--- a/scripts/run_tutorials.py
+++ b/scripts/run_tutorials.py
@@ -33,7 +33,7 @@ def parse_ipynb(file: Path) -> str:
     return script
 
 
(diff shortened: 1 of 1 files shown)


Balandat:
Thanks for fixing this.

PR #1124, opened 2022-03-23
Fix CAS tutorial

dme65:
Fix a small bug in the CAS tutorial where unsatisfactory points were incorrectly filtered out when computing ECI.


gustavomalkomes:
LGTM! Thanks for putting this together! 

you could also remove the function 
``` def _identify_samples_which_satisfy_constraints(self, X): ``` 
which it's not being used anymore.

PR #1142, opened 2022-03-29
Validate the shape of the bounds argument tensor in optimize_acqf

talesa:
## Motivation

I spent a silly amount of time debugging only to find out that my `bounds` tensor was transposed. Hope that with these 4 lines noone else will have to do that again!

Yes

## Test Plan

Tested manually, no further tests needed.


qingfeng10:
@talesa, apologize for reviewing this late! Thanks for adding the error check! The change lgtm!

PR #1151, opened 2022-03-31
Use re-usable workflows for Tutorials workflows

saitcakmak:
This is the first attempt at cleaning up GHA workflows. We currently have lots of duplicate code. By using re-usable workflows, we can significantly lower the maintenance burden. 

Test Plan:
* Nightly: https://github.com/pytorch/botorch/runs/5765190193?check_suite_focus=true
* Stable w/ stable Ax: https://github.com/pytorch/botorch/runs/5765197499?check_suite_focus=true
* Stable w/ latest Ax: https://github.com/pytorch/botorch/runs/5765197454?check_suite_focus=true
* Smoke test: https://github.com/pytorch/botorch/runs/5765209107?check_suite_focus=true

### DIFF
--- a/.github/workflows/reusable_tutorials.yml
+++ b/.github/workflows/reusable_tutorials.yml
@@ -1,0 +1,64 @@
+name: Reusable Tutorials Workflow
+
+on:
+  workflow_call:
+    inputs:
+      smoke_test:
+        required: false
+        type: boolean
+        default: true
+      use_stable_pytorch_gpytorch:
+        required: false
+        type: boolean
+        default: false
+      use_stable_ax:
+        required: false
+        type: boolean
+        default: false
+
+jobs:
+  tutorials:
+    name: Run tutorials
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - name: Set up Python
+      uses: actions/setup-python@v2
+      with:
+        python-version: 3.7
+    - name: Fetch all history for all tags and branches
+      # We need to do this so setuptools_scm knows how to set the BoTorch version.
+      run: git fetch --prune --unshallow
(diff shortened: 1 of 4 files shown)


.github/workflows/reusable_tutorials.yml
```diff
@@ -0,0 +1,64 @@
...
+      uses: actions/setup-python@v2
+      with:
+        python-version: 3.7
+    - name: Fetch all history for all tags and branches
+      # We need to do this so setuptools_scm knows how to set the BoTorch version.
+      run: git fetch --prune --unshallow
+    - if: ${{ !inputs.use_stable_pytorch_botorch }}
+      name: Install latest PyTorch & GPyTorch
+      run: |
+        pip install git+https://github.com/cornellius-gp/gpytorch.git
```
Balandat, line 35: We should probably install the pytorch nightly here? Otherwise gpytorch will resolve the requirement upon install and then install the latest compatible release.

.github/workflows/reusable_tutorials.yml
```diff
@@ -0,0 +1,64 @@
...
+        # This is so Ax's setup doesn't install a pinned BoTorch version.
+        ALLOW_BOTORCH_LATEST: true
+      run: |
+        pip install git+https://github.com/facebook/Ax.git
+    - if: ${{ always() }}
+      name: Install BoTorch with tutorials dependencies
+      env:
+        ALLOW_BOTORCH_LATEST: true
+      run: |
+        pip install .[tutorials]
```
Balandat, line 56:
```suggestion
    - name: Install BoTorch with tutorials dependencies
      env:
        ALLOW_BOTORCH_LATEST: true
      run: |
        pip install .[tutorials]
```

Balandat (approved):
This is awesome, way cleaner! Just a couple of things inline.

PR #1165, opened 2022-04-11
Fix typo in `reusable_tutorials` workflow

saitcakmak:
botorch -> gpytorch

Caught in https://github.com/pytorch/botorch/runs/5978010093?check_suite_focus=true (installs latest rather than stable)

Test plan: https://github.com/pytorch/botorch/runs/5978402785?check_suite_focus=true should install min required.


Balandat (approved):
thanks for the fix

PR #1173, opened 2022-04-13
Add expectation risk measures

saitcakmak:
Summary:
Adds MC expectation risk measures, both single-output and multi-output.
We initially skipped these thinking that one should use the analytic expectation. The work on robust MOBO made us realize that you can't use analytic expectation if you have constraints. So, adding these now to provide a full selection of risk measures out of the box in Ax.

Differential Revision: D35631768

### DIFF
--- a/botorch/acquisition/multi_objective/multi_output_risk_measures.py
+++ b/botorch/acquisition/multi_objective/multi_output_risk_measures.py
@@ -94,6 +94,26 @@ class MultiOutputRiskMeasureMCObjective(
         pass  # pragma: no cover
 
 
+class MultiOutputExpectation(MultiOutputRiskMeasureMCObjective):
+    r"""A multi-output MC expectation risk measure."""
+
+    def forward(self, samples: Tensor, X: Optional[Tensor] = None) -> Tensor:
+        r"""Calculate the expectation of the given samples. Expectation is
+        calculated over each `n_w` samples in the q-batch dimension.
+
+        Args:
+            samples: A `sample_shape x batch_shape x (q * n_w) x m`-dim tensor of
+                posterior samples. The q-batches should be ordered so that each
+                `n_w` block of samples correspond to the same input.
+            X: A `batch_shape x q x d`-dim tensor of inputs. Ignored.
+
+        Returns:
+            A `sample_shape x batch_shape x q x m`-dim tensor of expectation samples.
+        """
+        prepared_samples = self._prepare_samples(samples)
+        return prepared_samples.mean(dim=-2)
+
+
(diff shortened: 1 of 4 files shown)


Balandat (approved):
LGTM, thanks. 

> We initially skipped these thinking that one should use the analytic expectation. The work on robust MOBO made us realize that you can't use analytic expectation if you have constraints. 

Should we add a comment to the docstrings that if there are no constraints people should be using the analytic expectation instead?

botorch/acquisition/multi_objective/multi_output_risk_measures.py
```diff
@@ -94,6 +94,25 @@ def forward(self, samples: Tensor, X: Optional[Tensor] = None) -> Tensor:
...
+    def forward(self, samples: Tensor, X: Optional[Tensor] = None) -> Tensor:
+        r"""Calculate the expectation of the given samples. Expectation is
+        calculated over each `n_w` samples in the q-batch dimension.
+
+        Args:
+            samples: A `sample_shape x batch_shape x (q * n_w) x m`-dim tensor of
+                posterior samples. The q-batches should be ordered so that each
+                `n_w` block of samples correspond to the same input.
+            X: A `batch_shape x q x d`-dim tensor of inputs. Ignored.
+        Returns:
```
Balandat, line 109: nit
```suggestion
            X: A `batch_shape x q x d`-dim tensor of inputs. Ignored.
            
        Returns:
```

botorch/acquisition/multi_objective/multi_output_risk_measures.py
```diff
@@ -94,6 +94,25 @@ def forward(self, samples: Tensor, X: Optional[Tensor] = None) -> Tensor:
...
+            A `sample_shape x batch_shape x q x m`-dim tensor of expectation samples.
+        """
+        prepared_samples = self._prepare_samples(samples)
```
Balandat, line 112: since `X` is ignored, should we emit a warning if the passed in `X` is not `None`?

saitcakmak: `X` is ignored in most MC objectives, including all risk measures. Since we call the objective with `X=X` in the acquisition functions, I feel like this may end up flooding the output with warnings

saitcakmak (commented):
> Should we add a comment to the docstrings that if there are no constraints people should be using the analytic expectation instead?

Yeah, that's a good idea. That will be more efficient particularly with large `q`. I still haven't landed the analytic one, so this might be a bit premature. I'll add a TODO to that diff

PR #1185, opened 2022-04-19
Implements Multi-Fidelity GIBBON (Lower Bound MES) acquisition.

npielawski:
## Motivation

Since `qLowerBoundMaxValueEntropy` provides a cheap approximation to `qMaxValueEntropy`, this PR implements the multi-fidelity version of `qMultiFidelityMaxValueEntropy` to be able to use the approximation in a multi-fidelity setting as well.

Yes.

## Test Plan

Copied the unit test for `qMultiFidelityMaxValueEntropy` without changes (besides the class name). The `qMultiFidelityLowerBoundMaxValueEntropy` class is the same as `qMultiFidelityMaxValueEntropy` with a different compute_information_gain function, so doesn't require a different set of tests.

### DIFF
--- a/botorch/acquisition/max_value_entropy_search.py
+++ b/botorch/acquisition/max_value_entropy_search.py
@@ -795,6 +795,58 @@ class qMultiFidelityMaxValueEntropy(qMaxValueEntropy):
         return ig.mean(dim=0)  # average over the fantasies
 
 
+class qMultiFidelityLowerBoundMaxValueEntropy(
+    qMultiFidelityMaxValueEntropy, qLowerBoundMaxValueEntropy
+):
+    r"""Multi-fidelity acquisition function for General-purpose Information-Based
+    Bayesian optimization (GIBBON).
+
+    The acquisition function for multi-fidelity max-value entropy search
+    with support for trace observations. See [Takeno2020mfmves]_
+    for a detailed discussion of the basic ideas on multi-fidelity MES
+    (note that this implementation is somewhat different). This acquisition function
+    is similar to `qMultiFidelityMaxValueEntropy` but computes the information gain
+    from the lower bound described in [Moss2021gibbon].
+
+    The model must be single-outcome, unless using a PosteriorTransform.
+    The batch case `q > 1` is supported through cyclic optimization and fantasies.
+
+    Example:
+        >>> model = SingleTaskGP(train_X, train_Y)
+        >>> candidate_set = torch.rand(1000, bounds.size(1))
+        >>> candidate_set = bounds[0] + (bounds[1] - bounds[0]) * candidate_set
+        >>> MF_qGIBBON = qMultiFidelityLowerBoundMaxValueEntropy(model, candidate_set)
+        >>> mf_gibbon = MF_qGIBBON(test_X)
+    """
+
+    def _compute_information_gain(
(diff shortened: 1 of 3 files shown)


botorch/acquisition/max_value_entropy_search.py
```diff
@@ -795,6 +795,58 @@ def forward(self, X: Tensor) -> Tensor:
         return ig.mean(dim=0)  # average over the fantasies
 
 
+class qMultiFidelityLowerBoundMaxValueEntropy(
+    qMultiFidelityMaxValueEntropy, qLowerBoundMaxValueEntropy
```
Balandat, line 799: So since `qLowerBoundMaxValueEntropy` doesn't define any methods other than `_compute_information_gain` it seems like you're not actually inheriting anything by way of how this is implemented below (calling `qLowerBoundMaxValueEntropy._compute_information_gain`). So we could avoid subclassing from `qLowerBoundMaxValueEntropy` here to simplify things (unless we want to retain the class hierarchy for other reasons such as instance checks or some multiple dispatch pattern). But I think we for now don't need that and so I'd suggest not actually subclassing here. What do you think?

test/acquisition/test_max_value_entropy_search.py
```diff
@@ -286,6 +287,51 @@ def test_q_multi_fidelity_max_value_entropy(self):
             X = torch.rand(1, 2, device=self.device, dtype=dtype)
             self.assertEqual(qMF_MVE(X).shape, torch.Size([1]))
 
+    def test_q_multi_fidelity_lower_bound_max_value_entropy(self):
```
Balandat, line 290: If this is exactly the same test as `test_q_multi_fidelity_max_value_entropy`, how about changing that function signature to 
```
def test_q_multi_fidelity_max_value_entropy(self, aqcf_class=qMultiFidelityMaxValueEntropy):
   ...
```
and using `aqcf_class` in the test. Then you can just do 
```
def test_q_multi_fidelity_lower_bound_max_value_entropy(self):
    self.test_q_multi_fidelity_max_value_entropy(acqf_class=qMultiFidelityLowerBoundMaxValueEntropy)
```
and avoid a bunch of code duplication.

Balandat (commented):
Thanks a lot for contributing this, overall this looks great to me. Just a couple of minor suggestions inline.

botorch/acquisition/max_value_entropy_search.py
npielawski: Thanks for your comments!
Yes that makes sense, I will make the change. I do need to [cast](https://docs.python.org/3/library/typing.html#typing.cast) the `self` from `_compute_information_gain` however, otherwise pyright complains about the type of `self` not being an instance of `qLowerBoundMaxValueEntropy`.

Is that fine? I saw that `cast` was never used in botorch previously.
```python
return qLowerBoundMaxValueEntropy._compute_information_gain(
    cast(qLowerBoundMaxValueEntropy, self),
    X=X,
    mean_M=mean_M,
    variance_M=variance_M,
    covar_mM=covar_mM,
)
```

Balandat: Hmm so we don't use `pyright` or other type checkers by default in BoTorch at this point (whether we should is a separate discussion), so it should be fine to do this without the `cast` for now. Other than that subclassing from both as in your original commit is fine as well.

Balandat (approved):

PR #1189, opened 2022-04-21
Add Bilog outcome transform

roussel-ryan:
## Motivation

This paper https://arxiv.org/abs/2002.08526 suggests that using a bilog transformation can help improve constrained BO optimization by magnifying the region around zero. However, the bilog transformation is not (to my knowledge) analytically invertible. I have replicated most of the bilog behavior using a tanh function, which is invertible. This comes with a caveat though, as the derivatives of the transformed function go to zero if the raw values are far outside the range [-1,1]. For now I think this is an acceptable trade-off.

Yes

## Test Plan

Tests have been included.

## Related PRs

### DIFF
--- a/botorch/models/transforms/outcome.py
+++ b/botorch/models/transforms/outcome.py
@@ -644,7 +644,7 @@ class Power(OutcomeTransform):
 class Bilog(OutcomeTransform):
     r"""Bilog-transform outcomes.
 
-    The Bilog transform [eriksson2021scalable] is useful for modeling constraining
+    The Bilog transform [eriksson2021scalable]_ is useful for modeling constraining
     functions. It magnifies values near zero and flattens extreme values outside the
     domain [-1,1].
     """
--- a/botorch/models/transforms/outcome.py
+++ b/botorch/models/transforms/outcome.py
@@ -631,3 +631,127 @@ class Power(OutcomeTransform):
             posterior=posterior,
             sample_transform=lambda x: x.pow(1.0 / self.power),
         )
+
+
+class Tanh(OutcomeTransform):
+    r"""Tanh-transform outcomes.
+
+    Useful for modeling constraining functions. Magnifies values near zero and
+    flattens extreme values outside the domain [-1,1].
+    """
+
+    def __init__(self, outputs: Optional[List[int]] = None) -> None:
+        r"""Tanh-transform outcomes.
+
+        Args:
+            outputs: Which of the outputs to tanh-transform. If omitted, all
+                outputs will be transformed.
+        """
+        super().__init__()
+        self._outputs = outputs
+
+    def subset_output(self, idcs: List[int]) -> OutcomeTransform:
+        r"""Subset the transform along the output dimension.
+
+        Args:
+            idcs: The output indices to subset the transform to.
+
+        Returns:
+            The current outcome transform, subset to the specified output indices.
+        """
+        new_outputs = None
+        if self._outputs is not None:
+            if min(self._outputs + idcs) < 0:
+                raise NotImplementedError(
+                    f"Negative indexing not supported for {self.__class__.__name__} "
+                    "when subsetting outputs and only transforming some outputs."
+                )
+            new_outputs = [i for i in self._outputs if i in idcs]
+        new_tf = self.__class__(outputs=new_outputs)
+        if not self.training:
+            new_tf.eval()
+        return new_tf
+
+    def forward(
+        self, Y: Tensor, Yvar: Optional[Tensor] = None
+    ) -> Tuple[Tensor, Optional[Tensor]]:
+        r"""Tanh-transform outcomes.
+
+        Args:
+            Y: A `batch_shape x n x m`-dim tensor of training targets.
+            Yvar: A `batch_shape x n x m`-dim tensor of observation noises
+                associated with the training targets (if applicable).
+
+        Returns:
+            A two-tuple with the transformed outcomes:
+
+            - The transformed outcome observations.
+            - The transformed observation noise (if applicable).
+        """
+        Y_tf = Y.tanh()
+        outputs = normalize_indices(self._outputs, d=Y.size(-1))
+        if outputs is not None:
+            Y_tf = torch.stack(
+                [
+                    Y_tf[..., i] if i in outputs else Y[..., i]
+                    for i in range(Y.size(-1))
+                ],
+                dim=-1,
+            )
+        if Yvar is not None:
+            raise NotImplementedError(
+                "Tanh does not yet support transforming observation noise"
+            )
+        return Y_tf, Yvar
+
+    def untransform(
+        self, Y: Tensor, Yvar: Optional[Tensor] = None
+    ) -> Tuple[Tensor, Optional[Tensor]]:
+        r"""Un-transform tanh-transformed outcomes
+
+        Args:
+            Y: A `batch_shape x n x m`-dim tensor of power-transfomred targets.
+            Yvar: A `batch_shape x n x m`-dim tensor of power-transformed
+                observation noises associated with the training targets
+                (if applicable).
+
+        Returns:
+            A two-tuple with the un-transformed outcomes:
+
+            - The un-power transformed outcome observations.
+            - The un-power transformed observation noise (if applicable).
+        """
+        Y_utf = Y.atanh()
+        outputs = normalize_indices(self._outputs, d=Y.size(-1))
+        if outputs is not None:
+            Y_utf = torch.stack(
+                [
+                    Y_utf[..., i] if i in outputs else Y[..., i]
+                    for i in range(Y.size(-1))
+                ],
+                dim=-1,
+            )
+        if Yvar is not None:
+            # TODO: Delta method, possibly issue warning
+            raise NotImplementedError(
+                "Tanh does not yet support transforming observation noise"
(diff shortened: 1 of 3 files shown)


Balandat:
Thanks for the PR. 

@dme65 <- this guy would know sth about this paper :)

roussel-ryan:
I imagine so, any comments would be welcome

dme65:
Hi @roussel-ryan, thanks for adding this. Isn't the inverse of `f(x) = sign(x) * log(1 + |x|)` (which is what we used in the paper) given by `f^{-1}(y) = sign(y) * (exp(|y|) - 1)`? It may be nice to use bilog since it consistently worked well for the problems we considered in the paper, but it may also be that tanh works works just as well since the main idea is the same.

roussel-ryan:
So I replaced the tanh function with the bilog transform using torch functions. However, there is a numerical/floating point error with testing the forward + backward transformation. Any suggestions?

dme65:
Thank you for the updates, I'll review this PR later today!

botorch/models/transforms/outcome.py
```diff
@@ -631,3 +631,128 @@ def untransform_posterior(self, posterior: Posterior) -> Posterior:
...
+            Yvar: A `batch_shape x n x m`-dim tensor of observation noises
+                associated with the training targets (if applicable).
+
+        Returns:
+            A two-tuple with the transformed outcomes:
+
+            - The transformed outcome observations.
+            - The transformed observation noise (if applicable).
+        """
+        Y_tf = torch.sign(Y) * torch.log(torch.abs(Y) + torch.tensor(1.0).to(Y))
```
dme65, line 691:
```suggestion
        Y_tf = Y.sign() * (Y.abs() + 1).log()
```

botorch/models/transforms/outcome.py
```diff
@@ -631,3 +631,128 @@ def untransform_posterior(self, posterior: Posterior) -> Posterior:
...
+                observation noises associated with the training targets
+                (if applicable).
+
+        Returns:
+            A two-tuple with the un-transformed outcomes:
+
+            - The un-power transformed outcome observations.
+            - The un-power transformed observation noise (if applicable).
+        """
+        Y_utf = torch.sign(Y) * (torch.exp(torch.abs(Y)) - torch.tensor(1.0).to(Y))
```
dme65, line 724:
```suggestion
        Y_utf = Y.sign() * (Y.abs().exp() - 1)
```

botorch/models/transforms/outcome.py
```diff
@@ -631,3 +631,128 @@ def untransform_posterior(self, posterior: Posterior) -> Posterior:
...
+            )
+        if Yvar is not None:
+            # TODO: Delta method, possibly issue warning
+            raise NotImplementedError(
+                "Bilog does not yet support transforming observation noise"
+            )
+        return Y_utf, Yvar
+
+    def untransform_posterior(self, posterior: Posterior) -> Posterior:
+        r"""Un-transform the tanh-transformed posterior.
```
dme65, line 742:
```suggestion
        r"""Un-transform the bilog-transformed posterior.
```

botorch/models/transforms/outcome.py
```diff
@@ -631,3 +631,128 @@ def untransform_posterior(self, posterior: Posterior) -> Posterior:
...
+
+        Args:
+            posterior: A posterior in the tanh-transformed space.
```
dme65, line 745:
```suggestion
            posterior: A posterior in the bilog-transformed space.
```

test/models/transforms/test_outcome.py
```diff
@@ -582,3 +583,120 @@ def test_power(self, seed=0):
...
+    def test_bilog(self, seed=0):
+        torch.random.manual_seed(seed)
+
+        ms = (1, 2)
+        batch_shapes = (torch.Size(), torch.Size([2]))
+        dtypes = (torch.float, torch.double)
+
+        # test transform and untransform
+        for m, batch_shape, dtype in itertools.product(ms, batch_shapes, dtypes):
+
```
dme65, line 596: We often define `tkwargs = {"device": self.device, "dtype": dtype}` and then you can just use `**tkwargs` each time you create a new tensor instead of explicitly passing in the device and dtype each time.

botorch/models/transforms/outcome.py
```diff
@@ -631,3 +631,128 @@ def untransform_posterior(self, posterior: Posterior) -> Posterior:
...
+
+        Returns:
+            The un-transformed posterior.
+        """
+        if self._outputs is not None:
+            raise NotImplementedError(
+                "Bilog does not yet support output selection for untransform_posterior"
+            )
+        return TransformedPosterior(
```
dme65, line 754: @Balandat, @saitcakmak  do we need a `mean_transform` and `variance_transform` here as well?

botorch/models/transforms/outcome.py
```diff
@@ -631,3 +631,128 @@ def untransform_posterior(self, posterior: Posterior) -> Posterior:
...
+            posterior=posterior,
+            sample_transform=lambda x: torch.sign(x)
+            * (torch.exp(torch.abs(x)) - torch.tensor(1.0).to(x)),
```
dme65, line 757:
```suggestion
            sample_transform=lambda x: x.sign() * (x.abs().exp() - 1),
```

botorch/models/transforms/outcome.py
```diff
@@ -631,3 +631,128 @@ def untransform_posterior(self, posterior: Posterior) -> Posterior:
             posterior=posterior,
             sample_transform=lambda x: x.pow(1.0 / self.power),
         )
+
+
+class Bilog(OutcomeTransform):
+    r"""Bilog-transform outcomes.
+
+    Useful for modeling constraining functions. Magnifies values near zero and
+    flattens extreme values outside the domain [-1,1].
```
dme65, line 640: Can you also add a pointer to the SCBO paper here so people know where this transform comes from?

botorch/models/transforms/outcome.py
```diff
@@ -631,3 +631,128 @@ def untransform_posterior(self, posterior: Posterior) -> Posterior:
...
+            )
+        return Y_tf, Yvar
+
+    def untransform(
+        self, Y: Tensor, Yvar: Optional[Tensor] = None
+    ) -> Tuple[Tensor, Optional[Tensor]]:
+        r"""Un-transform bilog-transformed outcomes
+
+        Args:
+            Y: A `batch_shape x n x m`-dim tensor of power-transfomred targets.
```
dme65, line 713:
```suggestion
            Y: A `batch_shape x n x m`-dim tensor of bilog-transformed targets.
```

botorch/models/transforms/outcome.py
```diff
@@ -631,3 +631,128 @@ def untransform_posterior(self, posterior: Posterior) -> Posterior:
...
+            Yvar: A `batch_shape x n x m`-dim tensor of power-transformed
```
dme65, line 714:
```suggestion
            Yvar: A `batch_shape x n x m`-dim tensor of bilog-transformed
```

test/models/transforms/test_outcome.py
```diff
@@ -582,3 +583,120 @@ def test_power(self, seed=0):
...
+            self.assertTrue(tf.training)
+            self.assertIsNone(tf._outputs)
+
+            # no observation noise
+            Y = torch.rand(*batch_shape, 3, m, device=self.device, dtype=dtype)
+            Y_tf, Yvar_tf = tf(Y, None)
+            self.assertTrue(tf.training)
+            self.assertTrue(
+                torch.allclose(
+                    Y_tf, torch.sign(Y) * torch.log(torch.tensor(1.0) + torch.abs(Y))
```
dme65, line 608:
```suggestion
                    Y_tf, Y.sign() * (Y.abs() + 1).log()
```

test/models/transforms/test_outcome.py
```diff
@@ -582,3 +583,120 @@ def test_power(self, seed=0):
...
+            tf = Bilog(outputs=outputs)
+            self.assertTrue(tf.training)
+            # cannot normalize indices b/c we don't know dimension yet
+            self.assertEqual(tf._outputs, [-1])
+
+            # no observation noise
+            Y = torch.rand(*batch_shape, 3, m, device=self.device, dtype=dtype)
+            Y_tf, Yvar_tf = tf(Y, None)
+            self.assertTrue(tf.training)
+            self.assertTrue(torch.allclose(Y_tf[..., 1], Y[..., 1].tanh()))
```
dme65, line 670: Shouldn't be a `tanh` here

dme65 (changes requested):
I left some comments, but this generally looks good! One test still uses `tanh` so not sure if that is why you were having issues with the tests.

Balandat:
> the tests still fail on line 611 of test_outcome.py due to some numerical issue I can't seem to solve.

`torch.allclose` by default has relatively tight tolerances (`rtol=1e-05`, `atol=1e-08`). Did you try relaxing those a bit? In general this doesn't seem too dicey since you're computing `log` on a number between 1 and 2, and `exp` on a number between 0 and 1, and so this should be reasonably accurate, but maybe you need to back off a bit...

roussel-ryan:
That's fine w/ me, I feel like I would need permission to do that from you guys

PR #1191, opened 2022-04-23
Make `get_infeasible_cost` return a cost value for each outcome.

saitcakmak:
Summary: Current implementation returns a single `M` value. This modifies it to return an `m`-dim tensor of infeasible cost values, each corresponding to one of the `m` outcomes.

Differential Revision: D35847505

### DIFF
--- a/botorch/acquisition/utils.py
+++ b/botorch/acquisition/utils.py
@@ -209,22 +209,22 @@ def get_infeasible_cost(
     model: Model,
     objective: Optional[Callable[[Tensor, Optional[Tensor]], Tensor]] = None,
     posterior_transform: Optional[PosteriorTransform] = None,
-) -> float:
+) -> Tensor:
     r"""Get infeasible cost for a model and objective.
 
-    Computes an infeasible cost `M` such that `-M < min_x f(x)` almost always,
-        so that feasible points are preferred.
+    For each outcome, computes an infeasible cost `M` such that
+    `-M < min_x f(x)` almost always, so that feasible points are preferred.
 
     Args:
         X: A `n x d` Tensor of `n` design points to use in evaluating the
             minimum. These points should cover the design space well. The more
             points the better the estimate, at the expense of added computation.
-        model: A fitted botorch model.
+        model: A fitted botorch model with `m` outcomes.
         objective: The objective with which to evaluate the model output.
         posterior_transform: A PosteriorTransform (optional).
 
     Returns:
-        The infeasible cost `M` value.
+        An `m`-dim tensor of infeasible cost values.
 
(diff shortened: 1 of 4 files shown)


botorch/acquisition/utils.py
```diff
@@ -237,9 +237,13 @@ def objective(Y: Tensor, X: Optional[Tensor] = None):
             return Y.squeeze(-1)
 
     posterior = model.posterior(X, posterior_transform=posterior_transform)
-    lb = objective(posterior.mean - 6 * posterior.variance.clamp_min(0).sqrt()).min()
-    M = -(lb.clamp_max(0.0))
-    return M.item()
+    lb = objective(posterior.mean - 6 * posterior.variance.clamp_min(0).sqrt())
+    if lb.ndim < posterior.mean.ndim:
+        lb = lb.unsqueeze(-1)
+    # Take outcome-wise min. Looping in to handle batched models.
```
Balandat, line 243: So If I read this correctly, this computes the minimum across all batch dimensions? Is that what we want here? The previous implementation also did this so it's probably fine to get this in as is, but we may want to think about this some more if the batch dims correspond to different data and we want to extract them individually (this may be relevant for fully bayesian models - @dme65).

test/acquisition/test_objective.py
```diff
@@ -320,7 +320,7 @@ def test_constrained_mc_objective(self):
                 obj=constrained_obj,
                 constraints=[feasible_con, infeasible_con],
                 samples=samples,
-                infeasible_cost=0.0,
+                infeasible_cost=torch.tensor([0.0], device=self.device, dtype=dtype),
```
Balandat, line 323: you're accepting `Union[Tensor, float]` - should we test support both input types?

Balandat (approved):

botorch/acquisition/utils.py
saitcakmak: Yeah, that is an interesting thought. I guess it would depend on how different we expect each model in the batch to be.  Since they're all trained on the same set of inputs, I'd presume they'd be relatively similar, so one cost value should work for all. This would also affect the fantasized models, so maybe if we have two fantasy models, one on a large fantasy observation and the other on a small one, the cost value for one may be too conservative for the other? One could also argue that the 6 sigma we use here is too conservative to begin with. Anyways, I'm happy to revisit this later if we think it is worth exploring in more detail.

test/acquisition/test_objective.py
saitcakmak: There are something like 6 tests in there. I only changed two of them to use tensors, so both float and tensor are being tested.

PR #1197, opened 2022-04-28
Modify risk measures to accept `List[float]` for `weights`.

saitcakmak:
Summary: As in title. `List[float]` simplifies the storage on Ax end.

Differential Revision: D36010654

### DIFF
--- a/botorch/acquisition/risk_measures.py
+++ b/botorch/acquisition/risk_measures.py
@@ -22,7 +22,7 @@ see also [Hong2014review]_.
 
 from abc import ABC, abstractmethod
 from math import ceil
-from typing import Optional
+from typing import List, Optional, Union
 
 import torch
 from botorch.acquisition.objective import MCAcquisitionObjective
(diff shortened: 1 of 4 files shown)


botorch/acquisition/risk_measures.py
```diff
@@ -47,18 +47,20 @@ class RiskMeasureMCObjective(MCAcquisitionObjective, ABC):
...
-            weights: An optional `m`-dim tensor of weights for scalarizing
+            weights: An optional `m`-dim tensor or list of weights for scalarizing
                 multi-output samples before calculating the risk measure.
         """
         super().__init__()
         self.n_w = n_w
-        self.weights = weights
+        self.register_buffer(
+            "weights", torch.as_tensor(weights) if weights is not None else None
+        )
```
Balandat, line 63: I had to check what this does - if `None` this doesn't actually register a buffer at all on the module. I feel this is kind of unclear, so let's do this instead.
```suggestion
        if weights is not None:
            self.register_buffer("weights", torch.as_tensor(weights))
```

Balandat (approved):
overall this lgtm - just one nit

botorch/acquisition/risk_measures.py
saitcakmak: So, it actually registers it. It just won't display `None` buffers. If you do `self.register_buffer("weights", None)` and call `self.weights`, it will not raise an `AttributeError`. `self.weights` is `None`, and you can find `weights` under `self._buffers`. It just doesn't get included in `state_dict` or in `repr(self)`.

Balandat: hmm, interesting. Ok that sounds reasonable then

PR #1207, opened 2022-05-05
Update conda install instructions & listed dependencies

saitcakmak:
## Motivation

As noted in #1206, current conda installation instructions do not work for latest versions since the listed channels cannot resolve the `pyro-ppl` dependency. This updates the instructions to include the `conda-forge` channel as a fix.

Note: Technically `conda-forge` has `botorch` and `gpytorch` as well, so we could get away without including `-c pytorch` and `-c gpytorch`. However, `conda-forge` versions of these packages may not always be up-to-date, so keeping these channels makes sure we can always install the latest versions.

## Test Plan

`./scripts/build_docs.sh`

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,6 +59,8 @@ Optimization simply use Ax.
 - PyTorch >= 1.9
 - gpytorch >= 1.6
 - scipy
+- multiple-dispatch
+- pyro-ppl == 1.8.0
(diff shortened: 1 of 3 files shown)


Balandat (approved):
lgtm, thanks!

PR #1212, opened 2022-05-10
Proximal biasing support for multi-output SingleTaskGP models

roussel-ryan:
## Motivation
This PR adds support for multi-output SingleTaskGP models in proximal biasing.

Yes

## Test Plan
A test was added to confirm that no errors are raised


test/acquisition/test_proximal.py
```diff
@@ -179,6 +179,24 @@ def test_proximal(self):
                     ExpectedImprovement(bad_single_task, 0.0), proximal_weights
                 )
 
+            # test a multi-output SingleTaskGP model
+            train_X = torch.rand(5, 3, device=self.device, dtype=dtype)
+            train_Y = torch.rand(5, 2, device=self.device, dtype=dtype)
+
+            multi_output_model = (
+                SingleTaskGP(train_X, train_Y).to(device=self.device).eval()
```
Balandat, line 187: I don't think the `eval` should be necessary here - when called in `acq`, the `posterior` call of the model should automatically make sure that the model is in eval mode.

botorch/acquisition/proximal.py
```diff
@@ -95,10 +95,15 @@ def forward(self, X: Tensor) -> Tensor:
 
         train_inputs = model.train_inputs[0]
 
+        # if the model is ModelListGP then get the first model
         if isinstance(model, ModelListGP):
             train_inputs = train_inputs[0]
             model = model.models[0]
 
+        # if the model has more than one output get the first copy of training inputs
+        if model.num_outputs > 1:
```
Balandat, line 104: It doesn't need to be the case that a model with `num_outputs > 1` has copies of the training inputs. Maybe a more precise way of identifying the case you're after here would be to do `if isinstance(model, BatchedMultiOutputGPyTorchModel)`.

Balandat (commented):
Thanks for these edits, I only have minor comments. 

Not sure why the CI runs are getting rate-limited. Maybe it's because they are running through your repo rather through the botorch repo? I'll try re-running only failed ones to see if that fixes this.

Balandat:
Tutorial failure is unrelated and has been fixed, rebasing on the latest changes will make the test pass.

botorch/acquisition/proximal.py
roussel-ryan: I added the check for BatchedMultiOutputGPyTorchModel, but I still need the number of outputs to be greater than 1

PR #1213, opened 2022-05-11
Fix custom botorch model in Ax tutorial to support new interface

Balandat:
This also updates the synthetic function to return NaN sem values -- before it returned zero, which was then ignored by the custom model.


saitcakmak (approved):
Lgtm!

PR #1216, opened 2022-05-11
Bump coverage to 100%

saitcakmak:
99.99% just doesn't feel right.


Balandat:
Hmm shouldn't we instead patch the `test_gen_candidates_torch` test to also use a trivial test `callback`? That way we'd actually test the code rather than mask the gap in coverage...

saitcakmak:
Yeah, I was being lazy since it is such a trivial piece of code. Updated.

Balandat (approved):
gtmtl

PR #1231, opened 2022-05-22
Validate  constraints in optimize_acqf

Balandat:
In some cases we may allow not using box constraints (e.g. when optimizing Alebo). Also, in general it would be good to the optimziation raise an error if the constraint set is empty.

Will require some unit tests.

### DIFF
--- a/botorch/optim/optimize.py
+++ b/botorch/optim/optimize.py
@@ -12,6 +12,7 @@ from __future__ import annotations
 
 from typing import Any, Callable, Dict, List, Optional, Tuple, Union
 
+import numpy as np
 import torch
 from botorch.acquisition.acquisition import (
     AcquisitionFunction,
@@ -26,6 +27,7 @@ from botorch.optim.initializers import (
     gen_one_shot_kg_initial_conditions,
 )
 from botorch.optim.stopping import ExpMAStoppingCriterion
+from scipy.optimize import linprog
 from torch import Tensor
 
 INIT_OPTION_KEYS = {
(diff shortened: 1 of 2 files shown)


Balandat:
Not sure why the MacOS tests are failing, but that's unrelated to this PR.

Balandat:
One question here is whether we're ok with solving an LP on each call of `optimize_acqf` as this isn't super lightweight. But I guess it's rather minimal compare to full fledged acquisition function optimization. We could also add a `validate_constraints` option that defaults to `True` but that we set to `False` in situations like running sequential greedy optimization where we know constraints don't change between calls.

Balandat:
Coverage gap is unrelated and fixed in #1234

botorch/optim/optimize.py
```diff
@@ -707,6 +744,105 @@ def _gen_batch_initial_conditions_local_search(
...
+    #   max_x |x|_1
+    #     s.t. bounds(x)
+    #          inequality_constraints(x)
+    #          equality_constraints(x)
+    #
+    # To do this we can introduce auxiliary variables s and solve the
+    # following standard formulation:
+    #
+    #   min_(x, s) - sum_i(s_i)
+    #     s.t. -x <= s <= x
```
bletham, line 779: I don't think this is correct - we're trying to solve
```
max \sum_i |x_i|
```
and the correct transformation would be 
```
max \sum_i s_i
s_i <= |x_i|
```
The constraint `|x_i| >= s_i` is equal to `(x_i >= s_i) OR (x_i <= -s_i)` and is not convex and so won't be representable as an LP.

The constraint given here, `-x _i<= s_i <= x_i` is equivalent to `|s_i| <= x`, which is convex but is not correct for the problem we're trying to solve. In particular it will only provide the correct answer for positive values of `x_i`; if the inequality and equality constraints are unbounded towards -Inf but has volume in the positive orthant, this LP will not detect the unboundedness.

esantorella:
Is there still a need for this? Conceptually the `_validate_constraints` function still makes sense.

Balandat:
Yeah I think we still want to do this, just correctly :) I think we should be able to use a slightly different reformulation of the problem to avoid the non-convexity issue.

PR #1257, opened 2022-06-10
Scalable Constrained Bayesian Optimization

nataliemaus:
## Motivation

I created this PR to add support for Scalable Constrained Bayesian Optimization (SCBO) as described in [1] to BoTorch. Since BoTorch already supports TuRBO, relatively few modifications were needed to add support for SCBO. I have found the SCBO method to be very useful in my research, and I hope that this addition of SCBO to BoTorch will make it easier for others to use SCBO in the future. 

[1] David Eriksson and Matthias Poloczek. Scalable constrained Bayesian optimization. In International Conference on Artificial Intelligence and Statistics, pages 730–738. PMLR, 2021.
(https://doi.org/10.48550/arxiv.2002.08526)

To implement SCBO, this PR includes two additions to BoTorch 
1. A new sampling class called ConstrainedMaxPosteriorSampling that we use to implement the constrained Thompson sampling described in [1]. 
2. A tutorial Jupiter notebook that walks the user through a simple example, constrained optimization problem with SCBO. 

Yes

## Test Plan

I tested the added code in botorch.generation.sampling.ConstrainedMaxPosteriorSampling by running the tutorial notebook (tutorials/scalable_constrained_bo.ipynb) which uses ConstrainedMaxPosteriorSampling.

## Related PRs

See the SCBO tutorial included in this PR (tutorials/scalable_constrained_bo.ipynb) for a description of the added implementation. Additionally, I am more than happy to add more formal documentation, specifically for the ConstrainedMaxPosteriorSampling class, I just wasn't sure where this documentation should go in within botorch/docs? 

@jacobrgardner

### DIFF
--- a/botorch/models/converter.py
+++ b/botorch/models/converter.py
@@ -152,6 +152,8 @@ def model_list_to_batched(model_list: ModelListGP) -> BatchedMultiOutputGPyTorch
 
     # construct the batched GP model
     input_transform = getattr(models[0], "input_transform", None)
+    if input_transform is not None:
+        input_transform.train()
     batch_gp = models[0].__class__(input_transform=input_transform, **kwargs)
     adjusted_batch_keys, non_adjusted_batch_keys = _get_adjusted_batch_keys(
         batch_state_dict=batch_gp.state_dict(), input_transform=input_transform
@@ -220,6 +222,8 @@ def batched_to_model_list(batch_model: BatchedMultiOutputGPyTorchModel) -> Model
             "Conversion of MixedSingleTaskGP is currently not supported."
         )
     input_transform = getattr(batch_model, "input_transform", None)
+    if input_transform is not None:
+        input_transform.train()
     outcome_transform = getattr(batch_model, "outcome_transform", None)
     batch_sd = batch_model.state_dict()
 
@@ -324,6 +328,8 @@ def batched_multi_output_to_single_output(
             "Conversion of models with custom likelihoods is currently unsupported."
         )
     input_transform = getattr(batch_mo_model, "input_transform", None)
+    if input_transform is not None:
+        input_transform.train()
     batch_sd = batch_mo_model.state_dict()
 
     # TODO: add support for outcome transforms.
--- a/test/models/test_converter.py
+++ b/test/models/test_converter.py
@@ -19,7 +19,7 @@ from botorch.models.converter import (
     batched_to_model_list,
     model_list_to_batched,
 )
-from botorch.models.transforms.input import Normalize
+from botorch.models.transforms.input import AppendFeatures, Normalize
 from botorch.models.transforms.outcome import Standardize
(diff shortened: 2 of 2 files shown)


Balandat:
Thanks for the PR! 

@dme65 seems like the right guy for reviewing this :)

dme65:
Many thanks for the PR @nataliemaus! This looks great to me from a quick glance. I'll review it more thoroughly in the next couple of days.

dme65:
Thanks again for this PR and for the many comments and explanations in the tutorial!

I left some comments  on `ConstrainedMaxPosteriorSampling`. Can you also add some tests, similar to what is in https://github.com/pytorch/botorch/blob/main/test/generation/test_sampling.py?

You need to add the tutorial under https://github.com/pytorch/botorch/blob/main/website/tutorials.json to fix the doc error and do some import sorting to fix the lint error.

Some comments on the tutorial:
- I think there is a bug in the GP models for the constraints. The training data is standardized, e.g., `train_C2 = (C2_turbo - C2_turbo.mean()) / C2_turbo.std()` before fitting the model which will change the meaning of the constraints `c(x) <= 0`.  It would be better to pass in `outcome_transform=Standardize(m=1)` to `SingleTaskGP ` which will standardize the data for you before fitting the model and then unstandardize the posterior later on. I think fixing this will improve the performance as SCBO should perform similarly to TuRBO which such simple constraints.
- Given that the models are fitted exactly the same way, you can create a simple helper method `get_fitted_model(X, Y)` that you can call for the objective and for each of the constraints. This will get rid of a big block of duplicated code.
- You can replace `_turbo` by `_scbo` in the variable names. You can also drop this completely as I only did this in the TuRBO tutorial since I was comparing the performance with EI and wanted to separate the points collected by each method.
- Note that SCBO doesn't actually rescale the trust region according to the lengthscales.
- The logic to only update the state if `Valid_Y_next.numel() > 0` doesn't look right. You would want to pass in both the objective and constraint values to `update_state ` and keep track of the objective and constraint values for the current best point. That way you can compare feasible points via their objective function values and infeasible points via their constraint violations (a feasible point is always better than an infeasible point) and track successes and failures even in cases where the constraints aren't satisfied.

Nits:
- Bayesina -> Bayesian
- [−5,10]20 -> [−5, 10]^20

nataliemaus:
@dme65 Thank you for the feedback! I think I have now made all of the necessary changes and addressed all of your comments. Let me know what you think of the updated tutorial.

dme65:
Thanks for making all these changes @nataliemaus! Looks like the lint + tests are failing, can you fix those? Can you also change the notebook to use `torch.double` by default (we do that for the other tutorials)? This one should be good to go after that.

dme65:
@nataliemaus Looks like L330 isn't covered by the updated tests!

dme65 (approved):
Thanks!

PR #1292, opened 2022-07-08
Remove deprecated usage of setuptools_scm in setup.py

Balandat:
This removes the deprecated use of `use_setuptools_scm` in `setup.py` so that we can use newer setuptools versions. 

Previously we parsed the `USE_SCM_LOCAL_VERSION` env var as part of `setup.py` in order to programmatically set the `local_scheme` setting, but this kind of dynamism is not possible in the pyproject.toml configuration. Instead, we now simply use the `SETUPTOOLS_SCM_PRETEND_VERSION` env var in order to provide a specific version name without the github hash (which we strip from the full name in a bash script).

What this means is that for untagged commits, `setuptools_scm` will always create a version string that includes the github hash and date (and overriding this will require to manually specifying the `SETUPTOOLS_SCM_PRETEND_VERSION` env var).

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -53,7 +53,7 @@ jobs:
         pip install --pre torch -f https://download.pytorch.org/whl/nightly/cpu/torch_nightly.html
         pip install git+https://github.com/cornellius-gp/gpytorch.git
         pip install .[test]
-        pip install --upgrade setuptools wheel
+        pip install --upgrade setuptools setuptools_scm wheel
     - name: Build packages (wheel and source distribution)
       run: |
         # set a pseudo version that strips the commit hash to enable upload to pypi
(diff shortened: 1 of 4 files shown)


setup.py
```diff
@@ -80,16 +80,6 @@
     long_description_content_type="text/markdown",
     python_requires=">=3.7",
     setup_requires=["setuptools-scm"],
```
saitcakmak, line 82: I don't think you need this anymore.
```suggestion
```

.github/workflows/nightly.yml
```diff
@@ -55,14 +55,14 @@ jobs:
         pip install .[test]
         pip install --upgrade setuptools wheel
```
saitcakmak, line 56:
```suggestion
        pip install --upgrade setuptools wheel setuptools_scm
```
`no_local_version=$(python -m setuptools_scm | cut -d "+" -f 1)` will not work without it

saitcakmak (approved):
Thanks! Let's run nightly workflow to make sure it all works

.github/workflows/nightly.yml
Balandat: good catch!

PR #1297, opened 2022-07-12
Skip bound validation in optimize_acqf if inequality constraints are specified

Balandat:
In some cases we may allow not using box constraints (e.g. when optimizing Alebo). This removes a check that currently breaks Alebo optimization. A proper solution should actually validate the constraint set, this was started in https://github.com/pytorch/botorch/pull/1231 but there are some nontrivial issues with this, so this provides a quick fix in the meantime.


saitcakmak (approved):
Lgtm!

PR #1304, opened 2022-07-13
Bump newest python version in GHA workflows to 3.10

Balandat:
We should make sure that we're py3.10 compatible.


saitcakmak (approved):
Lgtm pending CI runs.

Balandat:
Failures seem to be due to some issues with pyro on py3.10.

Balandat:
Closing this in favor of #1380

PR #1312, opened 2022-07-15
Fix automated release process by installing setuptools_scm

Balandat:
#1292 made changes to `setup.py`, which require `setuptools_scm` to be installed for the version generation to work properly. While this change was made to the nightly workflow, it was not made to the deployment ones. This PR fixes that.


saitcakmak (approved):
Thanks!

PR #1318, opened 2022-07-18
Properly invoke `setuptools_scm`

Balandat:
Previously, running things like `python setup.py --version` would just return 0.0.0 if `setuptools_scm` was not installed. 

This use of setuptools_scm is deprecated, this PR changes things to run `python -m setuptools_scm` as suggested here: https://github.com/pypa/setuptools_scm#pyprojecttoml-usage

This way, a missing setuptools_scm installation will raise an error ("No module named...") rather than "inferring" a version 0.0.0.

### DIFF
--- a/scripts/build_and_verify_py_packages.sh
+++ b/scripts/build_and_verify_py_packages.sh
@@ -6,6 +6,11 @@
 
 # Get version number (created dynamically via setuptools-scm)
 version=$(python -m setuptools_scm)
+if [[ $? != "0" ]]; then
+  echo "Determininig version via setuptools_scm failed."
+  echo "Make sure that setuptools_scm is installed in your python environment."
+  exit 1
+fi
 cur_dir=$(pwd)
 
 # set up temporary working directory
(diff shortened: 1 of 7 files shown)


scripts/build_and_verify_py_packages.sh
```diff
@@ -5,7 +5,7 @@
 # LICENSE file in the root directory of this source tree.
 
 # Get version number (created dynamically via setuptools-scm)
-version=$(python setup.py --version)
+version=$(python -m setuptools_scm)
```
saitcakmak, line 8: Is it possible to error out and exit if this fails? The default behavior of the bash script is to print out the error message (i.e., `No module named setuptools_scm`) and keep executing the rest of the script. I think things will fail either way, but it would make it easier to pinpoint the root cause of the failure.

saitcakmak (approved):
Should we add `setuptools_scm` to `DEV_REQUIRES` as well?

scripts/build_and_verify_py_packages.sh
Balandat: Sure, we could just regex our way around this and see if `version` is a `No module...` string. Will do that.

Balandat:
> Should we add setuptools_scm to DEV_REQUIRES as well?

Hmm why would we? What makes this necessary? To ensure people who do development have setuptools_scm installed? It's a bit too late at the point of executing setup.py to require that since at that point the damage is already done and if setuptools_scm is not installed the version is incorrectly inferred.

PR #1313, opened 2022-07-15
Add check for __init__ docstring from flake8-docstrings to GH Actions; fix docstrings

esantorella:
## Motivation
See #1306 for context.

When a class lacks an `__init__` docstring and has a super class, Sphinx will render the super class's docstring, which makes the documentation on botorch.org confusing and often incorrect. 

[x] Add D107 check from `flake8-docstrings` to prevent this issue from recurring, and [several others](https://github.com/pytorch/botorch/pull/1313#issuecomment-1186030721) that seem useful
[x] Check that the newly added checks (and nothing else) are correctly triggered in Actions
[x] Move `args` block from class to `__init__` docstrings (see #1306 for context)
[x] Add `__init__` docstrings where they are missing
[x] Update [CONTRIBUTING.md](https://github.com/pytorch/botorch/blob/main/CONTRIBUTING.md#pull-requests) to explain usage and installation of `flake8` with `flake8-docstrings`
[x] Build website and read it over

#1306 also requires making ABCs and other classes that should not be instantiated private. That is not done in this PR.

### Have you read the [Contributing Guidelines on pull requests](https://github.com/pytorch/botorch/blob/main/CONTRIBUTING.md#pull-requests)?

Yes. Violated the guideline to fork so that I can test Actions.

## Test Plan
[x] [Actions run](https://github.com/pytorch/botorch/runs/7364978964?check_suite_focus=true#logs) identifies classes missing `__init__` docstrings
[x] After fixing the docstrings, there are no more lint issues in Actions

Issue #1306, opened 2022-07-14
[Bug] Sphinx pulls documentation from ancestor classes for some Models, making documentation incorrect


esantorella:
# 🐛 Bug
The phrase "Initializes internal Module state, shared by both nn.Module and ScriptModule." occurs nonsensically 12 times in BoTorch [Sphinx-rendered documentation]((https://botorch.org/api/models.html)). This phrase does not occur in BoTorch; it comes from the docstrings for the `__init__` for `torch.nn.Module`, which is an ancestor of the affected classes.

There are several other similar issues:
* Documentation under `SingleTaskMultiFidelityGP` lists two separate sets of parameters, return type, and examples. The second set of parameters contains the _union_ of the parameters for this class and for its super class, `SingleTaskGP`. The example is from `SingleTaskGP`. The parameters listed are especially confusing, because there is no method that has that signature.
* `FixedNoiseMultiFidelityGP`, `SACGP`, `LCEAGP`, `LCEMGP`, and `FixedNoiseLCEMGP` all have similar issues to the above.
* [`ModelListGPyTorchModel`](https://botorch.org/api/models.html#botorch.models.gpytorch.ModelListGPyTorchModel) includes a sentence (starting with "A multi-output") and example that are from the `__init__` docstring for `ModelList`, one of `ModelListGPyTorchModel`'s superclasses.

![image](https://user-images.githubusercontent.com/9137425/179120285-3c94d34f-7fae-4a35-9c61-84f46730f62a.png)

## To reproduce

Read the botorch.org and [source code on botorch.org](https://botorch.org/api/_modules/botorch/models/gpytorch.html#ModelListGPyTorchModel).

## Expected Behavior

I would have expected documentation under a class to only contain material from the class and `__init__` docstrings for that class. I think this is what the authors of the docstrings expected.

## System information

**Please complete the following information:**
- BoTorch Version: 0.6.4 (current version on botorch.org -- I'm not running any code; this is all in the browser.)
- Computer OS: MacOS Monterey 12.4
- Browser: Chrome

## Additional context

This is not a surprise, since [Sphinx does render superclass documentation when `autoclass_content='both'`](https://github.com/sphinx-doc/sphinx/issues/7782), which matches BoTorch's [Sphinx config](https://github.com/pytorch/botorch/blob/3318c06333166328c95d24edd1055d4ef4bded70/sphinx/source/conf.py).

I think what's happening is:
[...]

PR #1319, opened 2022-07-18
Don't render ABCs with Sphinx

esantorella:
# Motivation

Currently, ABCs and other classes that are not intended to be used by end users or instantiated are listed in the Sphinx-rendered documentation on botorch.org. This creates several issues:

1) Spinx renders the documentation for the `__init__` method of a superclass even if the class can't be instantiated. See #1306
2) The website becomes cluttered with classes that a user is very unlikely to be looking for.

This PR adds `:meta private:` to the docstrings of several classes that are not intended to be instantiated.

Yes.

## Test Plan

[x] Built botorch.org locally and looked over the model documentation to make sure it looks okay.
[x] Checked the Sphinx-rendered documentation for instances of the phrase "Initializes internal Module state, shared by both nn.Module and ScriptModule", which had been incorrectly inserted numerous times. The only place this still occurs is in `botorch.models.kernels.categorical.CategoricalKernel`. This one is tricky because the issue originates in gpytorch, which provides `CategoricalKernel`'s superclass.
[x] Checked the Sphinx-rendered documentation for the issues mentioned in #1306 with `SingleTaskMultiFidelityGP`, `FixedNoiseMultiFidelityGP`, SACGP`, `LCEAGP`,` LCEMGP`, and `FixedNoiseLCEMGP` are no longer occurring.

## Related PRs

Depends on #1313


botorch/acquisition/acquisition.py
```diff
@@ -4,9 +4,7 @@
 # This source code is licensed under the MIT license found in the
 # LICENSE file in the root directory of this source tree.
 
-r"""
-Abstract base module for all botorch acquisition functions.
-"""
+"""Abstract base module for all botorch acquisition functions."""
```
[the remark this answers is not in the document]
esantorella: Nope should be fine either way. I think I just changed it while debugging something.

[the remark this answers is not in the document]
Balandat: Cool let's leave them as raw strings throughout then.

esantorella:
The "files changed" looks weird since it's including a few commits that have already been merged. I think we can ignore that for now and see if the issue persists after #1313 is incorporated.

Balandat:
> The "files changed" looks weird since it's including a few commits that have already been merged. I think we can ignore that for now and see if the issue persists after https://github.com/pytorch/botorch/pull/1313 is incorporated.

I believe this is b/c the branch for this PR originates off the branch that at the time was still an open PR, and not off the `main` branch. You could either rebase the branch onto the current head of main (then you'd have to force-push though), or merge the current `main` branch into your branch. I think this should resolve the weirdness about changed files (and would be good to do so this PR contains only the changes relative the current `main`).

botorch/acquisition/acquisition.py
```diff
@@ -101,7 +101,11 @@ def forward(self, X: Tensor) -> Tensor:
 
 
 class OneShotAcquisitionFunction(AcquisitionFunction, ABC):
-    r"""Abstract base class for acquisition functions using one-shot optimization"""
+    r"""
+    Abstract base class for acquisition functions using one-shot optimization
```
saitcakmak, line 105:
```suggestion
    r"""Abstract base class for acquisition functions using one-shot optimization.
```
Is there a particular reason for having these in a new line? I think we generally use the line starting with `r"""` for the one-line description.

botorch/models/likelihoods/pairwise.py
```diff
@@ -21,7 +21,20 @@
 
 
 class PairwiseLikelihood(Likelihood, ABC):
-    """Pairwise likelihood base class for pairwise preference GP (e.g., PairwiseGP)"""
+    """
```
saitcakmak, line 24:
```suggestion
    r"""
```

botorch/models/kernels/linear_truncated_fidelity.py
```diff
@@ -97,6 +68,36 @@ def __init__(  # noqa C901
         covar_module_biased: Optional[Kernel] = None,
         **kwargs: Any,
     ) -> None:
+        """
```
saitcakmak, line 71:
```suggestion
        r"""
```
for consistency

botorch/models/pairwise_gp.py
```diff
@@ -943,6 +943,14 @@ class PairwiseLaplaceMarginalLogLikelihood(MarginalLogLikelihood):
     See (12) from [Chu2005preference]_.
     """
 
+    def __init__(self, likelihood, model: GP):
+        """
```
saitcakmak, line 947:
```suggestion
        r"""
```

saitcakmak (approved):
LGTM! A few nits in line.

botorch/acquisition/acquisition.py
esantorella: Sorry, I somehow missed these comments... this one isn't showing up in the "Files changed" pane, and it looks like I made the opposite change?

In any case, PEP 257 says "The summary line may be on the same line as the opening quotes or on the next line." pydocstyle additionally specifies "One-line docstring should fit on one line with quotes." but doesn't seem to care if a multi-line docstring starts with a newline. This is rendering fine.

botorch/models/likelihoods/pairwise.py
esantorella: PEP 257:
> For consistency, always use """triple double quotes""" around docstrings. Use r"""raw triple double quotes""" if you use any backslashes in your docstrings. 

The CI is now enforcing this because this PR enabled the D301 lint, which checks that any docstring with a backslash in it is raw. I just tried adding a backslash to this docstring and running `flake8` and it did complain.

PR #1321, opened 2022-07-19
Add `maximize` option in subclasses of `MCAcquisitionFunction`

jiayuewan:
Summary: Add  `maximize` option in subclasses of `MCAcquisitionFunction`, including `qExpectedImprovement`, `qProbabilityOfImprovement`, `qSimpleRegret`, and `qUpperConfidenceBound`.

Differential Revision: D37902500


jiayuewan:
Makes sense to me. Thanks for the comment!

Issue #1227, opened 2022-05-18
[Bug] Equality constraint violated in optimization of ACQF


jduerholt:
# 🐛 Bug

In some cases the optimizer within `optimize_acqf` spits out candidates that violate the defined equality constraints provided to `optimize_acqf`. I build an MWE for it. You can find it below. If one reduces the number of points provided to the acqf, the chances that the constraint is fulfilled increase. Perhaps you have to run the MWE several times to observe the mentioned behavior, because the constraint is sometimes fulfilled and sometimes not. I attach the example data (totally made up data) as an csv file. 

## To reproduce

```python
import torch
from torch import tensor
from botorch.models import SingleTaskGP
from botorch.fit import fit_gpytorch_model
from gpytorch.mlls import ExactMarginalLogLikelihood
from botorch.models.transforms.input import Normalize
from botorch.models.transforms.outcome import Standardize
from botorch.optim import optimize_acqf
from botorch.models import ModelListGP
from botorch.acquisition.multi_objective import qNoisyExpectedHypervolumeImprovement
import pandas as pd

torch.manual_seed(1)

tkwargs = {'dtype': torch.float64, 'device': 'cpu'}

df_data = pd.read_csv("equality_constraint.csv")
#df_data = pd.read_csv("equality_constraint.csv").loc[:10]
train_X = torch.from_numpy(df_data[["a", "b", "c", "d"]].values).to(**tkwargs)
train_Y = torch.from_numpy(df_data[["alpha", "beta"]].values).to(**tkwargs)

lower = tensor([0.1,0.3,0.1,30.])
upper = tensor([0.6,0.7,0.7,70.])

bounds = torch.stack((lower, upper)).to(**tkwargs)

models = []
for i, feat in enumerate(["alpha", "beta"]):
    gp = SingleTaskGP(train_X, train_Y[:,i].unsqueeze(-1), input_transform=Normalize(d=4), outcome_transform=Standardize(m=1))
    mll = ExactMarginalLogLikelihood(gp.likelihood, gp)
    fit_gpytorch_model(mll)
    models.append(gp)
model = ModelListGP(*models)

acqf = qNoisyExpectedHypervolumeImprovement(model=model, ref_point=df_data[["alpha", "beta"]].min().values.tolist(), X_baseline=train_X)

candidate, acq_value = optimize_acqf(
[...]

Issue #1169, opened 2022-04-12
[Bug] `ABNORMAL_TERMINATION_IN_LNSRCH` in `fit_gpytorch_scipy` is not being caught, leading to poor model fits


crasanders:
# 🐛 Bug

I unfortunately don't have a concise example to share, but we make use of `fit_gpytorch_model` in [aepsych](https://github.com/facebookresearch/aepsych). `fit_gpytorch_model` calls `fit_gpytorch_scipy`, which calls `scipy.minimize`, which sometimes fails to converge with `"ABNORMAL_TERMINATION_IN_LNSRCH"`.  For some reason this is not being caught by the `with warnings.catch_warnings(record=True) as ws:` line in `fit_gpytorch_model`, which leads to poor model fits on our end.

## System information

**Please complete the following information:**
-  Botorch: 0.6.2
-  GPyTorch: 1.6.0
-  PyTorch: 1.9.1
-  Mac OS 12.3.1

saitcakmak:
That's an interesting error, one I haven't seen before. A google search brought up this [thread](https://stackoverflow.com/questions/34663539/scipy-optimize-fmin-l-bfgs-b-returns-abnormal-termination-in-lnsrch) which suggests that this may be related to function / gradient mismatch. It may just be a numerical issue in grad computation, but it might be worth looking into whether there's a bigger issue going on with the gradients in this case.

Is the `scipy.minimize` output something like `{'grad': array([ -1.36050603e-05]), 'task': b'ABNORMAL_TERMINATION_IN_LNSRCH', 'funcalls': 51, 'nit': 4, 'warnflag': 2}` (found in google search results) when this happens? If so, it should be easy to look into task and filter it that way.

crasanders:
The `scipy.minimize` output looks like this: 
```
fun: nan
 hess_inv: <3196x3196 LbfgsInvHessProduct with dtype=float64>
      jac: array([ 0.09521612,  0.35479467, -0.02726126, ...,  0.00897098,
       -0.05011029,  0.70100166])
  message: 'ABNORMAL_TERMINATION_IN_LNSRCH'
     nfev: 6
      nit: 3
     njev: 6
   status: 2
  success: False
        x: array([ 0.65288807, -0.28074463,  0.23027564, ...,  8.96318507,
        0.92354893,  0.42418053])
```
[...]

PR #1325, opened 2022-07-25
Error handling and restarting in optimization

esantorella:
## Motivation

Several issues were discussed in #1227:
- `scipy.optimize.minimize` and functions that call it can fail without warnings or errors.
- We allow use of `sample_around_best` with equality constraints even though it likely would not produce valid solutions.

Changes in this PR:
- Issue a warning in `gen_candidates_scipy` if Scipy reports that optimization failed.
- Raise an `UnsupportedError` in `gen_batch_inital_conditions` if the user tries to use the `"sample around best"` option in combination with equality constraints
- `optimize_acqf` catches optimization warnings from `gen_candidates_scipy`. If there are user-provided `batch_initial_conditions`, it raises a warning and suggests trying different initial conditions. Otherwise, it generates new initial conditions and tries again, generating another warning if it didn't work even on the second try.

I'd be curious to hear if people think that re-trying with new initial conditions could create a performance regression. That could happen if Scipy has been overly aggressive in tagging optimization runs as "failed" when they were actually good enough.

Yes

## Test Plan

- Test in `test_gen.py` for the optimization warning from `gen_candidates_scipy`
- Test that the appropriate `UnsupportedError` is raised in `gen_batch_initial_conditions`
Several tests that the appropriate warnings are raised and appropriate re-try happens in `optimize_acqf`:
- A case where it fails due to bad user-provided ICs
- A case where it fails due to bad ICs due to a bad seed, then succeeds on retry
- A case where it fails due to bad ICs due to a bad seed and fails again on retry

### DIFF
--- a/botorch/generation/gen.py
+++ b/botorch/generation/gen.py
@@ -17,6 +17,7 @@ from typing import Any, Callable, Dict, List, NoReturn, Optional, Tuple, Type, U
 import numpy as np
 import torch
 from botorch.acquisition import AcquisitionFunction
+from botorch.exceptions.warnings import OptimizationWarning
 from botorch.generation.utils import _remove_fixed_features_from_optimization
 from botorch.optim.parameter_constraints import (
     _arrayify,
@@ -210,6 +211,19 @@ def gen_candidates_scipy(
         callback=options.get("callback", None),
         options={k: v for k, v in options.items() if k not in ["method", "callback"]},
     )
+
+    if "success" not in res.keys():
+        warnings.warn(
+            "Optimization failed within `scipy.optimize.minimize` with no "
+            "status returned to `res.`",
+            OptimizationWarning,
+        )
+    elif not res.success:
+        warnings.warn(
+            f"Optimization failed within `scipy.optimize.minimize` with status "
+            f"{res.status}.",
+            OptimizationWarning,
+        )
     candidates = fix_features(
         X=torch.from_numpy(res.x).to(initial_conditions).reshape(shapeX),
         fixed_features=fixed_features,
--- a/botorch/generation/gen.py
+++ b/botorch/generation/gen.py
@@ -212,18 +212,22 @@ def gen_candidates_scipy(
         options={k: v for k, v in options.items() if k not in ["method", "callback"]},
     )
 
-    if "success" not in res.keys():
-        warnings.warn(
-            "Optimization failed within `scipy.optimize.minimize` with no "
-            "status returned to `res.`",
-            OptimizationWarning,
-        )
+    if "success" not in res.keys() or "status" not in res.keys():
+        with warnings.catch_warnings():
+            warnings.simplefilter("always", category=OptimizationWarning)
+            warnings.warn(
+                "Optimization failed within `scipy.optimize.minimize` with no "
+                "status returned to `res.`",
+                OptimizationWarning,
+            )
     elif not res.success:
-        warnings.warn(
-            f"Optimization failed within `scipy.optimize.minimize` with status "
-            f"{res.status}.",
-            OptimizationWarning,
-        )
+        with warnings.catch_warnings():
+            warnings.simplefilter("always", category=OptimizationWarning)
+            warnings.warn(
+                f"Optimization failed within `scipy.optimize.minimize` with status "
+                f"{res.status}.",
+                OptimizationWarning,
+            )
     candidates = fix_features(
         X=torch.from_numpy(res.x).to(initial_conditions).reshape(shapeX),
         fixed_features=fixed_features,
--- a/botorch/optim/optimize.py
+++ b/botorch/optim/optimize.py
@@ -284,9 +284,7 @@ def optimize_acqf(
             f"warning(s):\n{[w.message for w in ws]}\nTrying again with a new "
             "set of initial conditions."
         )
-        with warnings.catch_warnings():
-            warnings.simplefilter("always", category=OptimizationWarning)
-            warnings.warn(first_warn_msg, OptimizationWarning)
+        warnings.warn(first_warn_msg, RuntimeWarning)
 
         if not initial_conditions_provided:
             batch_initial_conditions = _gen_initial_conditions()
@@ -297,13 +295,11 @@ def optimize_acqf(
                 (issubclass(w.category, OptimizationWarning) for w in ws)
             )
             if optimization_warning_raised:
-                with warnings.catch_warnings():
-                    warnings.simplefilter("always", category=OptimizationWarning)
-                    warnings.warn(
-                        "Optimization failed on the second try, after generating a "
-                        "new set of initial conditions.",
-                        OptimizationWarning,
-                    )
+                warnings.warn(
+                    "Optimization failed on the second try, after generating a "
+                    "new set of initial conditions.",
+                    RuntimeWarning,
+                )
 
     if post_processing_func is not None:
         batch_candidates = post_processing_func(batch_candidates)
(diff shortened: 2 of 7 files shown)


botorch/generation/gen.py
```diff
@@ -17,6 +17,7 @@
 import numpy as np
 import torch
 from botorch.acquisition import AcquisitionFunction
+from botorch.exceptions.warnings import OptimizationWarning
```
saitcakmak, line 20: There's an issue with botorch warnings getting filtered out in some cases (https://github.com/pytorch/botorch/blob/main/botorch/settings.py#L58). See #1169 for relevant issue & #1170, #1178 for fixes. Let's not use them in critical cases such as this one. `RuntimeWarning` might be a better alternative.

botorch/optim/optimize.py
```diff
@@ -239,11 +249,60 @@ def optimize_acqf(
...
-        batch_candidates_list.append(batch_candidates_curr)
-        batch_acq_values_list.append(batch_acq_values_curr)
-        logger.info(f"Generated candidate batch {i+1} of {len(batched_ics)}.")
-    batch_candidates = torch.cat(batch_candidates_list)
-    batch_acq_values = torch.cat(batch_acq_values_list)
+
+        for i, batched_ics_ in enumerate(batched_ics):
+            # optimize using random restart optimization
+            with warnings.catch_warnings(record=True) as ws:
+                warnings.simplefilter("always", category=OptimizationWarning)
```
saitcakmak, line 256: This is only effective within the context. So it'll let you catch the ones raised in `gen_candidates_scipy` here.

botorch/optim/optimize.py
```diff
@@ -239,11 +249,60 @@ def optimize_acqf(
...
+        batch_candidates = torch.cat(batch_candidates_list)
+        batch_acq_values = torch.cat(batch_acq_values_list)
+        return batch_candidates, batch_acq_values, opt_warnings
+
+    batch_candidates, batch_acq_values, ws = _optimize_batch_candidates()
+
+    optimization_warning_raised = any(
+        (issubclass(w.category, OptimizationWarning) for w in ws)
+    )
+    if optimization_warning_raised:
```
saitcakmak, line 274: But the warnings in here may not get raised.

saitcakmak (approved):
Thanks! A series of comments in line regarding the behavior of BoTorch warnings. 

Thoughts on raising a warning vs an error if everything failed? We use warnings in `fit_gpytorch_model`, so having warnings here is consistent with that. It will let you keep the optimization going, though it may end up with a pathological behavior of never evaluating feasible points. No strong preference either way. cc @Balandat

Balandat:
> Thoughts on raising a warning vs an error if everything failed? We use warnings in fit_gpytorch_model, so having warnings here is consistent with that. It will let you keep the optimization going, though it may end up with a pathological behavior of never evaluating feasible points. No strong preference either way.

Let's stick with warnings for now. There may be situations in which every once in a while the acquisition optimization fails but most of the time everything works fine. In those cases a warning will, well, just warn, rather than error and potentially break a bunch of downstream code. We can reconsider this later, but for now let's take one step at a time :)

esantorella:
> Let's stick with warnings for now. There may be situations in which every once in a while the acquisition optimization fails but most of the time everything works fine. In those cases a warning will, well, just warn, rather than error and potentially break a bunch of downstream code. We can reconsider this later, but for now let's take one step at a time :)

On a similar note, are there any concerns about restarting with new initial conditions if `minimize` fails on the first try? If `minimize` is being overly sensitive in saying it failed (perhaps if a tolerance has been set too small), then could this create a substantial performance drag?

Balandat:
> On a similar note, are there any concerns about restarting with new initial conditions if minimize fails on the first try? If minimize is being overly sensitive in saying it failed (perhaps if a tolerance has been set too small), then could this create a substantial performance drag?

I am not sure, we'd have to look at a range of cases to get a sense for how prevalent these optimization failures of the acquisition function are. Maybe we can use @mpolson64's benchmarking setup to do that?

botorch/generation/gen.py
```diff
@@ -212,18 +212,22 @@ def f(x):
...
 
-    if "success" not in res.keys():
-        warnings.warn(
-            "Optimization failed within `scipy.optimize.minimize` with no "
-            "status returned to `res.`",
-            OptimizationWarning,
-        )
+    if "success" not in res.keys() or "status" not in res.keys():
+        with warnings.catch_warnings():
+            warnings.simplefilter("always", category=OptimizationWarning)
```
saitcakmak, line 217: I'm not a fan of having these two lines repeated so many times in the code. Probably not a concern here, but this also prevents the user from suppressing these warnings if they wanted to.

botorch/optim/optimize.py
```diff
@@ -272,24 +272,23 @@ def _optimize_batch_candidates() -> Tuple[Tensor, Tensor, List[Warning]]:
...
+            " Suggested remediation: Try again with different "
+            "`batch_initial_conditions`, or don't provide `batch_initial_conditions.`"
+            if initial_conditions_provided
+            else "Optimization failed in `gen_candidates_scipy` with the following "
+            f"warning(s):\n{[w.message for w in ws]}\nTrying again with a new "
+            "set of initial conditions."
+        )
+        with warnings.catch_warnings():
+            warnings.simplefilter("always", category=OptimizationWarning)
+            warnings.warn(first_warn_msg, OptimizationWarning)
```
saitcakmak, line 289: How about changing the type of this warning and the one below to something like `RuntimeWarning` rather than using the context & filter?

botorch/generation/gen.py
esantorella: I don't think it does -- I suppressed these in a few tests where they weren't relevant.

esantorella:
> > On a similar note, are there any concerns about restarting with new initial conditions if minimize fails on the first try? If minimize is being overly sensitive in saying it failed (perhaps if a tolerance has been set too small), then could this create a substantial performance drag?
> 
> I am not sure, we'd have to look at a range of cases to get a sense for how prevalent these optimization failures of the acquisition function are. Maybe we can use @mpolson64's benchmarking setup to do that?

That sounds good to me. Any opinion on testing it through the benchmarking setup vs. just landing this and waiting to see what happens in the nightly run?

botorch/generation/gen.py
saitcakmak: Hmm, this still raises the warning for me:
```
import warnings

warnings.simplefilter("ignore")
with warnings.catch_warnings():
    warnings.simplefilter("always")
    warnings.warn("dummy warning", RuntimeWarning)
```

saitcakmak:
> That sounds good to me. Any opinion on testing it through the benchmarking setup vs. just landing this and waiting to see what happens in the nightly run?

I think this is benign enough that we could just land it, but it doesn't hurt to be cautious. It's super easy to kick off a benchmark run. I'll share the instructions offline

botorch/generation/gen.py
esantorella: Oh, yeah, good catch. I was using `warnings.catch_warnings(record=True) as ws` to prevent the warning from being written to stdout, but it still seems to be added to `ws` regardless of any `ignore` filtering.

Balandat:
> I think this is benign enough that we could just land it, but it doesn't hurt to be cautious. It's super easy to kick off a benchmark run. I'll share the instructions offline

Yeah there is no particular rush getting this in right away, so let's run the benchmarks to be sure.

PR #1327, opened 2022-07-29
Expanded docstrings for models in modules "model",  "gp_regression", "deterministic", and "cost"

esantorella:
NOTE: I am making several PRs each expanding on model documentation. I am trying to identify the best subject-matter expert for each as a reviewer; please let me know if someone else would be a better fit. In this case this is @danielrjiang for cost modeling, but a lot of the documentation in this PR here is for very commonly used models so many people would be good reviewers.

## Motivation

It can be hard for users to figure out which model to use because docstrings are not always as thorough as they could be.

There will be several PRs with docstrings for different model modules.

## Test Plan

- Asked experts for clarification on unclear model documentation
- Built the docs and confirmed that it rendered as expected with Sphinx

### DIFF
--- a/botorch/models/cost.py
+++ b/botorch/models/cost.py
@@ -7,11 +7,10 @@
 r"""
 Cost models to be used with multi-fidelity optimization.
 
-Cost models are deterministic models, useful for defining known cost functions
-when the cost of an evaluation is heterogeneous in fidelity. For a full worked
-example, see the
-`tutorial <https://botorch.org/tutorials/multi_fidelity_bo>`_
-on continuous multi-fidelity Bayesian Optimization.
+Cost are useful for defining known cost functions when the cost of an evaluation
+is heterogeneous in fidelity. For a full worked example, see the
+`tutorial <https://botorch.org/tutorials/multi_fidelity_bo>`_ on continuous
+multi-fidelity Bayesian Optimization.
 """
 
 from __future__ import annotations
@@ -24,12 +23,16 @@ from torch import Tensor
 
 
 class AffineFidelityCostModel(DeterministicModel):
-    r"""Affine cost model operating on fidelity parameters.
+    r"""Deterministic, affine cost model operating on fidelity parameters.
 
     For each (q-batch) element of a candidate set `X`, this module computes a
     cost of the form
 
         cost = fixed_cost + sum_j weights[j] * X[fidelity_dims[j]]
+
+    For a full worked example, see the
+    `tutorial <https://botorch.org/tutorials/multi_fidelity_bo>`_ on continuous
+    multi-fidelity Bayesian Optimization.
 
     Example:
         >>> from botorch.models import AffineFidelityCostModel
--- a/botorch/models/cost.py
+++ b/botorch/models/cost.py
@@ -6,6 +6,12 @@
 
 r"""
 Cost models to be used with multi-fidelity optimization.
+
+Cost models are deterministic models, useful for defining known cost functions
+when the cost of an evaluation is heterogeneous in fidelity. For a full worked
+example, see the
+`tutorial <https://botorch.org/tutorials/multi_fidelity_bo>`_
+on continuous multi-fidelity Bayesian Optimization.
 """
 
 from __future__ import annotations
@@ -24,6 +30,14 @@ class AffineFidelityCostModel(DeterministicModel):
     cost of the form
 
         cost = fixed_cost + sum_j weights[j] * X[fidelity_dims[j]]
+
+    Example:
+        >>> from botorch.models import AffineFidelityCostModel
+        >>> from botorch.acquisition.cost_aware import InverseCostWeightedUtility
+        >>> cost_model = AffineFidelityCostModel(
+        >>>    fidelity_weights={6: 1.0}, fixed_cost=5.0
+        >>> )
+        >>> cost_aware_utility = InverseCostWeightedUtility(cost_model=cost_model)
     """
 
     def __init__(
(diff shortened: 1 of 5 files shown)


botorch/models/gp_regression.py
```diff
@@ -6,6 +6,26 @@
...
+covariates are normalized to the unit cube and outcomes are standardized (zero
+mean, unit variance).
+
+These models all work in batch mode (each batch having its own hyperparameters).
+When the training observations include multiple outputs, these models use
+batching to model outputs independently.
+
+These models all support multiple outputs. However, as single-task models,
+`SingleTaskGP`, `FixedNoiseGP`, and `HeteroskedasticSingleTaskGP` should be
+used only when the output(s) are independent and all use the same training data.
```
saitcakmak, line 25:
```suggestion
used only when the outputs are independent and all use the same training data.
```

botorch/models/gp_regression.py
```diff
@@ -6,6 +6,26 @@
...
+If outputs are independent and outputs have different training data, use the
+ModelListGP. When modeling correlations between outputs, use a multi-task
```
saitcakmak, line 27:
```suggestion
`ModelListGP`. When modeling correlations between outputs, use a multi-task
```

botorch/models/gp_regression.py
```diff
@@ -6,6 +6,26 @@
...
+model like MultiTaskGP.
```
saitcakmak, line 28:
```suggestion
model like `MultiTaskGP`.
```

botorch/models/gp_regression.py
```diff
@@ -159,13 +179,26 @@ def forward(self, x: Tensor) -> MultivariateNormal:
...
 
     This model works in batch mode (each batch having its own hyperparameters).
 
+    An example of a case in which noise levels are known is online
+    experimentation, where noise can be measured using the variability of
+    different observations from the same arm, or provided by outside software.
+    Another use case is simulation optimization, where the evaluation can
+    provide variance estimates, perhaps from bootstrapping. In any case, these
+    noise levels must be provided to `FixedNoiseGP` as `train_Yvar`.
+
```
saitcakmak, line 197:
```suggestion

    `FixedNoiseGP` is also commonly used when the observations are known to be noise-free.
    Noise-free observations can be modeled using arbitrarily small noise values, such as
    `train_Yvar=torch.full_like(train_Y, 1e-6)`.

```

botorch/models/model.py
```diff
@@ -246,16 +253,24 @@ def train(self, mode: bool = True) -> Model:
 
 
 class ModelList(Model):
-    r"""Container for a list of models."""
+    r"""A multi-output Model represented by a list of independent models.
+
+    A multi-output Model represented by a list of independent models. Any
```
saitcakmak, line 258: Repeats the same sentence
```suggestion
    All
```

botorch/models/model.py
```diff
@@ -246,16 +253,24 @@ def train(self, mode: bool = True) -> Model:
...
+    BoTorch models are acceptable as inputs. The cost of this flexibility is
+    that ModelList does not support all methods that may be implemented by its
```
saitcakmak, line 260:
```suggestion
    that `ModelList` does not support all methods that may be implemented by its
```

saitcakmak (approved):
Thanks!

botorch/models/cost.py
```diff
@@ -6,6 +6,12 @@
 
 r"""
 Cost models to be used with multi-fidelity optimization.
+
+Cost models are deterministic models, useful for defining known cost functions
```
Balandat, line 10: I don't think cost models necessarily need to be deterministic, they could also be probabilistic. In fact, there are some applications in which we make cost observations and model them with a GP model. It just so happens that the models in this module are deterministic.

botorch/models/deterministic.py
```diff
@@ -5,9 +5,19 @@
...
+Deterministic models are useful for defining known cost functions for
+cost-aware acquisition utilities. Cost-aware optimization is desirable when
+evaluations have a cost that is heterogeneous, either in the inputs `X` or in a
+particular fidelity parameter that directly encodes the fidelity of the
+observation. `GenericDeterministicModel` supports arbitrary deterministic
+functions, while `AffineFidelityCostModel` is a particular cost model for
+multi-fidelity optimization. Other use cases of deterministic models include
+representing approximate GP sample paths, e.g. random Fourier features obtained
+with `get_gp_samples`, which allows them to be substituted in acquisition
+functions or in other places where a `Model` is expected.
```
Balandat, line 20:
```suggestion
Deterministic models are useful for expressing known input-output relationships
within the BoTorch Model API. This is useful e.g. for multi-objective optimization
with known objective functions (e.g. the number of parameters of a Neural Network
in the context of Neural Architecture Search is usually a known function of the 
architecture configuration), or to encode cost functions for
cost-aware acquisition utilities. Cost-aware optimization is desirable when
evaluations have a cost that is heterogeneous, either in the inputs `X` or in a
particular fidelity parameter that directly encodes the fidelity of the
observation. `GenericDeterministicModel` supports arbitrary deterministic
functions, while `AffineFidelityCostModel` is a particular cost model for
multi-fidelity optimization. Other use cases of deterministic models include
representing approximate GP sample paths, e.g. random Fourier features obtained
with `get_gp_samples`, which allows them to be substituted in acquisition
functions or in other places where a `Model` is expected.
```

botorch/models/gp_regression.py
```diff
@@ -159,13 +179,26 @@ def forward(self, x: Tensor) -> MultivariateNormal:
...
     r"""A single-task exact GP model using fixed noise levels.
 
-    A single-task exact GP that uses fixed observation noise levels. This model
-    also uses relatively strong priors on the Kernel hyperparameters, which work
+    A single-task exact GP that uses fixed observation noise levels, differing from
+    `SingleTaskGP` only in that noise levels are provided rather than inferred.
+    This model also uses relatively strong priors on the Kernel
+    hyperparameters, which work
     best when covariates are normalized to the unit cube and outcomes are
     standardized (zero mean, unit variance).
```
Balandat, line 187:
```suggestion
    This model also uses relatively strong priors on the Kernel hyperparameters,
    which work best when covariates are normalized to the unit cube and outcomes
    are standardized (zero mean, unit variance).
```

Balandat (approved):
Some nits, o/w looks great. Thanks!

botorch/models/cost.py
esantorella: Thanks! `class AffineFidelityCostModel(DeterministicModel)` is the only class in the module so there's no need to be overly general in the module docstring. I'll move some of this material out of here and integrate into the `AffineFidelityCostModel` docstring.

botorch/models/gp_regression.py
esantorella: This is just about the placement of line breaks? FWIW this doesn't affect how it renders (although the suggested change does look better in code).

I wish there was a way to automatically fix docstring line length since this was tedious, but it looks like the tools that exist for this (`docstringformatter` and `pydocstringformatter`) make too many other changes at the same time.

PR #1332, opened 2022-07-30
Use method='highs' in scipy.optimize.linprog

### LINKED ISSUE #1331: [Bug] One unit test failing on main, Python 3.9 only
# 🐛 Bug

One unit test is failing. This happened for MacOS, Ubuntu, and Windows, but on Python 3.9 only (not 3.7). This happened on [main](https://github.com/pytorch/botorch/runs/7587401356?check_suite_focus=true), as well as in PRs #1327 , #1328 , #1329 , and #1330, none of which could have plausibly broken this test.

## To reproduce

[Actions](https://github.com/pytorch/botorch/runs/7587401356?check_suite_focus=true)

```
=================================== FAILURES ===================================
___________________ TestSampleUtils.test_find_interior_point ___________________

self = <test.utils.test_sampling.TestSampleUtils testMethod=test_find_interior_point>

    def test_find_interior_point(self):
        # basic problem: 1 <= x_1 <= 2, 2 <= x_2 <= 3
        A = np.concatenate([np.eye(2), -np.eye(2)], axis=0)
        b = np.array([2.0, 3.0, -1.0, -2.0])
        x = find_interior_point(A=A, b=b)
        self.assertTrue(np.allclose(x, np.array([1.5, 2.5])))
        # problem w/ negatives variables: -2 <= x_1 <= -1, -3 <= x_2 <= -2
        b = np.array([-1.0, -2.0, 2.0, 3.0])
        x = find_interior_point(A=A, b=b)
        self.assertTrue(np.allclose(x, np.array([-1.5, -2.5])))
        # problem with bound on a single variable: x_1 <= 0
        A = np.array([[1.0, 0.0]])
        b = np.zeros(1)
        x = find_interior_point(A=A, b=b)
        self.assertLessEqual(x[0].item(), 0.0)
        # unbounded problem: x >= 3
        A = np.array([[-1.0]])
        b = np.array([-3.0])
        x = find_interior_point(A=A, b=b)
>       self.assertAlmostEqual(x.item(), 6.201544, places=4)
E       AssertionError: 5.0 != 6.201544 within 4 places (1.2015440000000002 difference)

test/utils/test_sampling.py:311: AssertionError
```

## Expected Behavior

`x.item()` should be close to  6.201544 rather than being 5.0.

## System information

GitHub Actions on current commit (147cf755db8f2da5e22788a9155372432f0c1ae6)

## Additional context

The last overnight run passed, so I think #1318 was the only commit recent enough to do this.

Balandat:
Fixes #1331.

From the scipy 1.9.0 release notes:

>The default method of scipy.optimize.linprog is now 'highs', not
'interior-point' (which is now deprecated), so callback functions and
some options are no longer supported with the default method. With the
default method, the x attribute of the returned OptimizeResult is
now None (instead of a non-optimal array) when an optimal solution
cannot be found (e.g. infeasible problem).

Since scipy 1.9.0 requires python 3.8+, we're seeing different behavior between on 3.7 vs. 3.8+. This PR explicitly passes in the new highs default method, making the behavior consistent across scipy versions.

### DIFF
--- a/test/utils/test_sampling.py
+++ b/test/utils/test_sampling.py
@@ -308,7 +308,7 @@ class TestSampleUtils(BotorchTestCase):
         A = np.array([[-1.0]])
         b = np.array([-3.0])
         x = find_interior_point(A=A, b=b)
-        self.assertAlmostEqual(x.item(), 6.201544, places=4)
+        self.assertAlmostEqual(x.item(), 5.0, places=4)
 
     def test_get_polytope_samples(self):
         tkwargs = {"device": self.device}
--- a/botorch/utils/sampling.py
+++ b/botorch/utils/sampling.py
@@ -476,7 +476,13 @@ def find_interior_point(
     A_ub[-1, -1] = -1.0
 
     result = scipy.optimize.linprog(
-        c=c, A_ub=A_ub, b_ub=b_ub, A_eq=A_eq, b_eq=b_eq, bounds=(None, None)
+        c=c,
+        A_ub=A_ub,
+        b_ub=b_ub,
+        A_eq=A_eq,
+        b_eq=b_eq,
+        bounds=(None, None),
+        method="highs",
     )
 
     if result.status == 3:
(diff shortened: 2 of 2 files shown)


test/utils/test_sampling.py
```diff
@@ -308,7 +308,7 @@ def test_find_interior_point(self):
         A = np.array([[-1.0]])
         b = np.array([-3.0])
         x = find_interior_point(A=A, b=b)
-        self.assertAlmostEqual(x.item(), 6.201544, places=4)
+        self.assertAlmostEqual(x.item(), 5.0, places=4)
```
esantorella, line 311: If any `x>=3` is okay, should we just check `x>=3` rather than a specific value?

esantorella (approved):
LGTM. Looks like nothing was truly broken and the test was only failing because it was overly picky.

test/utils/test_sampling.py
Balandat: Yeah that would work as well and would not have resulted in a test failure. It's kind of nice to know about different behaviors that may occur under different versions, ideally the behavior of our functions is the same independent of the versions of the libraries used (so long as they are compatible). So far this hasn't been much of an issue so I'd like to to the value comparison; if this gets too noisy we can chance that policy.

PR #1336, opened 2022-08-02
Support batched models in `RandomFourierFeatures`

saitcakmak:
Summary: Expands `get_gp_samples` & `RandomFourierFeatures` to work with batched models.

Differential Revision: D38002555

### DIFF
--- a/botorch/utils/gp_sampling.py
+++ b/botorch/utils/gp_sampling.py
@@ -118,7 +118,7 @@ class RandomFourierFeatures(Module):
         Args:
             kernel: The GP kernel.
             input_dim: The input dimension to the GP kernel.
-            num_rff_features: The number of fourier features.
+            num_rff_features: The number of Fourier features.
             sample_shape: The shape of a single sample. For a single-element
                 `torch.Size` object, this is simply the number of RFF draws.
         """
@@ -134,9 +134,8 @@ class RandomFourierFeatures(Module):
             outputscale = kernel.outputscale.detach().clone()
         if not isinstance(base_kernel, (MaternKernel, RBFKernel)):
             raise NotImplementedError("Only Matern and RBF kernels are supported.")
-        elif len(base_kernel.batch_shape) > 0:
-            raise NotImplementedError("Batched kernels are not supported.")
         super().__init__()
+        self.kernel_batch_shape = base_kernel.batch_shape
         self.register_buffer("outputscale", outputscale)
 
         self.register_buffer("lengthscale", base_kernel.lengthscale.detach().clone())
@@ -157,6 +156,7 @@ class RandomFourierFeatures(Module):
             * pi
             * torch.rand(
                 *self.sample_shape,
+                *self.kernel_batch_shape,
                 num_rff_features,
                 dtype=base_kernel.lengthscale.dtype,
                 device=base_kernel.lengthscale.device,
@@ -175,14 +175,16 @@ class RandomFourierFeatures(Module):
         Args:
             kernel: The GP base kernel.
             input_dim: The input dimension to the GP kernel.
-            num_rff_features: The number of fourier features.
+            num_rff_features: The number of Fourier features.
             sample_shape: The sample shape of weights.
         Returns:
-            A `(sample_shape) x input_dim x num_rff_features`-dim tensor of weights
+            A tensor of weights with shape
+            `(*sample_shape, *kernel_batch_shape, input_dim, num_rff_features)`.
         """
         sample_shape = torch.Size() if sample_shape is None else sample_shape
         weights = torch.randn(
             *sample_shape,
+            *self.kernel_batch_shape,
             input_dim,
             num_rff_features,
             dtype=base_kernel.lengthscale.dtype,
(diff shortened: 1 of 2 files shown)


botorch/utils/gp_sampling.py
```diff
@@ -178,11 +178,13 @@ def _get_weights(
             num_rff_features: The number of fourier features.
             sample_shape: The sample shape of weights.
         Returns:
-            A `(sample_shape) x input_dim x num_rff_features`-dim tensor of weights
+            A `(sample_shape) x (kernel_batch_shape) x input_dim x num_rff_features`-dim
+            tensor of weights.
```
esantorella, line 182:
```suggestion
            A tensor of weights with shape
            `(*sample_shape, *kernel_batch_shape, input_dim, num_rff_features)`.
```

Are the parentheses supposed to represent that `sample_shape` and `kernel_batch_shape` may not be present, or do they indicate that these might be tuples?

Also technically this is the shape, not the dimension (the dimension is >=2).

botorch/utils/gp_sampling.py
```diff
@@ -197,44 +199,59 @@ def _get_weights(
         return weights
 
     def forward(self, X: Tensor) -> Tensor:
-        """
-        Get fourier basis features for the provided inputs.
-        Note that if `sample_shape` has been passed, then the rightmost
-        subset of the batch shape of the input should be `sample_shape`.
+        """Get fourier basis features for the provided inputs.
```
esantorella, line 202:
```suggestion
        """Get Fourier basis features for the provided inputs.
```

botorch/utils/gp_sampling.py
```diff
@@ -197,44 +199,59 @@ def _get_weights(
...
+
+        Note that the right-most subset of the batch shape of `X`
+        should be `sample_shape x kernel_batch_shape` if using either
+        the `sample_shape` argument or a batched kernel.
```
esantorella, line 206: What if you're using the `sample_shape` argument but not a batched kernel or vice versa? Is the "right-most subset" always 2 elements or can it be 1? Some examples of what `X.shape` should look like would clarify.

botorch/utils/gp_sampling.py
```diff
@@ -197,44 +199,59 @@ def _get_weights(
         return weights
 
     def forward(self, X: Tensor) -> Tensor:
-        """
-        Get fourier basis features for the provided inputs.
-        Note that if `sample_shape` has been passed, then the rightmost
-        subset of the batch shape of the input should be `sample_shape`.
+        """Get fourier basis features for the provided inputs.
+
+        Note that the right-most subset of the batch shape of `X`
```
esantorella, line 204:
```suggestion
        Note that this will error unless the right-most subset of the batch shape of `X`
```
(making it clear the user doesn't need to check manually)

botorch/utils/gp_sampling.py
```diff
@@ -341,13 +343,15 @@ def get_deterministic_model_list(
...
 
-def get_weights_posterior(X: Tensor, y: Tensor, sigma_sq: float) -> MultivariateNormal:
+def get_weights_posterior(X: Tensor, y: Tensor, sigma_sq: Tensor) -> MultivariateNormal:
     r"""Sample bayesian linear regression weights.
 
     Args:
-        X: a `(batch_shape) x n x num_rff_features`-dim tensor of inputs
-        y: a `(batch_shape) x n`-dim tensor of outputs
-        sigma_sq: the noise variance
+        X: A `(batch_shape) x n x num_rff_features`-dim tensor of inputs.
```
esantorella, line 350:
```suggestion
        X: A tensor of inputs with shape `(*batch_shape, n, num_rff_features)`.
```
Similar to above, I know we've been writing documentation like this throughout, but I think using Python code to explain this would be clearer

test/utils/test_gp_sampling.py
```diff
@@ -403,16 +413,19 @@ def test_get_deterministic_model(self):
...
         n_samples = 5
         for dtype, m in product((torch.float, torch.double), (1, 2)):
             tkwargs["dtype"] = dtype
-            for batch_shape_w, batch_shape_x in product(
-                [torch.Size([]), torch.Size([3])], repeat=2
+            for batch_shape_w, batch_shape_x in (
+                (torch.Size([]), torch.Size([])),
+                (torch.Size([]), torch.Size([3])),
+                (torch.Size([3]), torch.Size([3])),
+                (torch.Size([3]), torch.Size([5, 3])),
```
esantorella, line 428: Using `with self.subTest(batch_shape_w=batch_shape_w, batch_shape_x=batch_shape_x):` as the line below would make it easier to figure out where failures are coming from

esantorella (approved):

botorch/utils/gp_sampling.py
Balandat: > Are the parentheses supposed to represent that sample_shape and kernel_batch_shape may not be present, or do they indicate that these might be tuples?

They may be tuples, but we use parentheses are to indicate that they may not be present (or equivalently be empty tuples).

test/utils/test_gp_sampling.py
saitcakmak: Cool! I didn't know this existed

Issue #732, opened 2021-03-05
[Feature Request] Docs listing available models and their properties


saitcakmak:
# 🚀 Feature Request

I find myself searching through the docstrings or the API reference whenever I am not sure what model to use. These typically have enough information to make a decision but reading through them is not the most user friendly thing. So, I was thinking having a page in docs that lists all model types packaged into BoTorch, along with a short discussion on underlying assumptions and use cases would be a handy reference. I'm sure I'd use it, and I think it'd also be a good reference for anyone that's new to BoTorch.

Balandat:
So we do have a high-level description of some model types here: https://botorch.org/docs/models

But this is not particularly detailed and also incomplete. We could either update that page, or add a new one that goes into some more detail and lists all the models.

saitcakmak:
> So we do have a high-level description of some model types here: https://botorch.org/docs/models

Not sure how I missed that page. I knew something like that existed somewhere but couldn't find it at a quick look. I must have been looking under tutorials.

> But this is not particularly detailed and also incomplete. We could either update that page, or add a new one that goes into some more detail and lists all the models.

Yes, I was thinking something more detailed and complete. Adding a new page with a short paragraph (or a few bullet points) for each model sounds like a good option

saitcakmak:
We've made a good bit of progress here thanks to @esantorella's efforts. Closing the issue.

PR #1337, opened 2022-08-02
Model docs

esantorella:
## Motivation

* Added material to  `docs/models.md`, which is rendered [here](https://botorch.org/docs/models), adding a short summary for each model and extra context.  See #732 
* Made sure a couple classes have docstrings 
* Fixed broken links

## Test Plan
* I collected a list of models that are either are in `botorch.models.__init__.py` or are non-abstract subclasses of `Model`, and made sure they are are all documented in `models.md`. This does exclude some models, such as contextual models.
* Built the website and read it over
* Clicked every link and made sure it worked

### DIFF
--- a/botorch/models/approximate_gp.py
+++ b/botorch/models/approximate_gp.py
@@ -67,7 +67,7 @@ NEG_INF = -(torch.tensor(float("inf")))
 
 
 class ApproximateGPyTorchModel(GPyTorchModel):
-    """
+    r"""
     Botorch wrapper class for various (variational) approximate GP models in
     GPyTorch.
 
--- a/botorch/models/approximate_gp.py
+++ b/botorch/models/approximate_gp.py
@@ -67,6 +67,14 @@ NEG_INF = -(torch.tensor(float("inf")))
 
 
 class ApproximateGPyTorchModel(GPyTorchModel):
+    """
+    Botorch wrapper class for various (variational) approximate GP models in
+    GPyTorch.
+
+    This can either include stochastic variational GPs (SVGPs) or
+    variational implementations of weight space approximate GPs.
+    """
+
     def __init__(
         self,
         model: Optional[ApproximateGP] = None,
@@ -76,14 +84,10 @@ class ApproximateGPyTorchModel(GPyTorchModel):
         **kwargs,
     ) -> None:
         r"""
-        Botorch wrapper class for various (variational) approximate GP models in
-        gpytorch. This can either include stochastic variational GPs (SVGPs) or
-        variational implementations of weight space approximate GPs.
-
         Args:
             model: Instance of gpytorch.approximate GP models. If omitted,
                 constructs a `_SingleTaskVariationalGP`.
-            likelihood: Instance of a GPyYorch likelihood. If omitted, uses a
+            likelihood: Instance of a GPyTorch likelihood. If omitted, uses a
                 either a `GaussianLikelihood` (if `num_outputs=1`) or a
                 `MultitaskGaussianLikelihood`(if `num_outputs>1`).
             num_outputs: Number of outputs expected for the GP model.
--- a/botorch/models/deterministic.py
+++ b/botorch/models/deterministic.py
@@ -216,7 +216,7 @@ class PosteriorMeanModel(DeterministicModel):
 
 
 class FixedSingleSampleModel(DeterministicModel):
-    """
+    r"""
     A deterministic model defined by a single sample `w`.
 
     Given a base model `f` and a fixed sample `w`, the model always outputs
--- a/botorch/models/deterministic.py
+++ b/botorch/models/deterministic.py
@@ -201,9 +201,10 @@ class AffineDeterministicModel(DeterministicModel):
 
 
 class PosteriorMeanModel(DeterministicModel):
-    def __init__(self, model: Model) -> None:
-        r"""A deterministic model that always return the posterior mean.
+    """A deterministic model that always returns the posterior mean."""
 
+    def __init__(self, model: Model) -> None:
+        r"""
         Args:
             model: The base model.
         """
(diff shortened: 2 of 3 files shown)


botorch/models/deterministic.py
```diff
@@ -215,15 +216,18 @@ def forward(self, X: Tensor) -> Tensor:
 
 
 class FixedSingleSampleModel(DeterministicModel):
-    def __init__(self, model: Model, w: Optional[Tensor] = None) -> None:
-        r"""A deterministic model defined by a single sample w.
+    """
```
Balandat, line 219:
```suggestion
    r"""
```

botorch/models/approximate_gp.py
```diff
@@ -67,6 +67,14 @@
 
 
 class ApproximateGPyTorchModel(GPyTorchModel):
+    """
```
Balandat, line 70:
```suggestion
    r"""
```

docs/models.md
```diff
@@ -9,49 +9,64 @@ In BoTorch, a `Model` maps a set of design points to a posterior probability
...
+With the exception of some of the analytic acquisition functions in the
+[`botorch.acquisition.analytic`](../api/acquisition.html#analytic-acquisition-function-api)
+module, BoTorch’s Monte Carlo-based acquisition functions are compatible with
+any model that conforms to the `Model` interface, whether user-implemented or provided.
+
+Under the hood,
+BoTorch models are PyTorch `Modules` that implement the that
+implement the light-weight
+[`Model`](../api/models.html#model-apis) interface. When working with GPs,
+When working with GPs, [`GPyTorchModel`](../api/models.html#module-botorch.models.gp_regression)
```
Balandat, line 24:
```suggestion
Under the hood, BoTorch models are PyTorch `Modules` that implement
the light-weight [`Model`](../api/models.html#model-apis) interface. 
When working with GPs, [`GPyTorchModel`](../api/models.html#module-botorch.models.gp_regression)
```

docs/models.md
```diff
@@ -9,49 +9,64 @@ In BoTorch, a `Model` maps a set of design points to a posterior probability
...
+different outputs independently rather than
+building a joint model are not multi-task.
 * If a model is both, we refer to it as a multi-task-multi-output (MTMO) model.
 
+### Noise: Homoskedastic, fixed, and heteroskedastic
+Noise can be treated in several different ways:
+
+* *Homoskedastic*: Noise is not provided as an input and is inferred, with a
+constant variance that does not depend on `X`. Many models, such as
+`SingleTaskGP`, take this approach.
```
Balandat, line 63:
```suggestion
constant variance that does not depend on `X`. Many models, such as
`SingleTaskGP`, take this approach. Use these models if you know that
your observations are noisy, but not how noisy.
```

docs/models.md
```diff
@@ -9,49 +9,64 @@ In BoTorch, a `Model` maps a set of design points to a posterior probability
...
+
+* *Fixed*: Noise is provided as an input and is not fit. In “fixed” models like `FixedNoiseGP`, noise cannot be predicted out-of-sample because it has not been modeled.
```
Balandat, line 65:
```suggestion
* *Fixed*: Noise is provided as an input and is not fit. In “fixed noise” models
like `FixedNoiseGP`, noise cannot be predicted out-of-sample because it has
not been modeled. Use these models if you have estimates of the noise in
your observations (e.g. observations may be averages over individual samples
in which case you would provide the mean as observation and and the standard
error of he mean as the noise estimate), or if you know your observations are
noiseless (by passing zero a zero noise level). 
```

docs/models.md
```diff
@@ -60,38 +75,76 @@ BoTorch provides several GPyTorch models to cover most standard BO use cases:
...
 All of the above models use Matérn 5/2 kernels with Automatic Relevance
 Discovery (ARD), and have reasonable priors on hyperparameters that make them
 work well in settings where the **input features are normalized to the unit
 cube** and the **observations are standardized** (zero mean, unit variance).
 
+## Other useful models
+
+* [`ModelList`](../api/models.html#botorch.models.model.ModelList): a multi-output model in which outcomes
+  are modeled independently, as in `ModelListGP`, but the component models do not all
+  need to be GPyTorch models.
```
Balandat, line 121:
```suggestion
* [`ModelList`](../api/models.html#botorch.models.model.ModelList): a multi-output model container
  in which outcomes are modeled independently by individual `Model`s (as in `ModelListGP`, but the
  component models do not all   need to be GPyTorch models).
```

docs/models.md
```diff
@@ -60,38 +75,76 @@ BoTorch provides several GPyTorch models to cover most standard BO use cases:
...
+* [`HigherOrderGP`](../api/models.html#botorch.models.higher_order_gp.HigherOrderGP): A GP model with
+  matrix-valued predictions, such as images or grids of images.
+* [`PairwiseGP`](../api/models.html#module-botorch.models.pairwise_gp): A probit-likelihood GP that
+  learns via pairwise comparison data, useful for preference learning.
+* [`ApproximateGPyTorchModel`](../api/models.html#botorch.models.approximate_gp.ApproximateGPyTorchModel): for
+  efficient computation when data is large or responses are non-Gaussian.
+* [Deterministic models](../api/models.html#module-botorch.models.deterministic), such as
+  [`AffineDeterministicModel`](../api/models.html#botorch.models.deterministic.AffineDeterministicModel),
+  [`AffineFidelityCostModel`](../api/models.html#botorch.models.cost.AffineFidelityCostModel),
+  [`FixedSingleSampleModel`](../api/models.html#botorch.models.deterministic.FixedSingleSampleModel),
```
Balandat, line 135: Not sure we need FixedSingleSampleModel here, this one is kind of special and a bit of a hack.

Balandat (approved):
Thanks! A few nits.

botorch/models/approximate_gp.py
esantorella: We don't need to manually check this now that we have D301 enabled, which will error if there's a backslash in a non-raw docstring

docs/models.md
esantorella: sounds good, also it's not in `models/__init__.py`

esantorella, line 135:
```suggestion
```

Issue #1333, opened 2022-07-31
[Bug] outcome_transform ignored for ModelListGP.fantasize


JackBuck:
# 🐛 Bug
The `outcome_transform` is not applied when using `ModelListGP.fantasize` meaning that the scale of the fantasy models can be magnified/shrunk.

## To reproduce

**Code snippet to reproduce**
```python
import torch
from botorch.models import ModelListGP, SingleTaskGP
from botorch.models.transforms import Standardize
from botorch.utils.sampling import manual_seed
from botorch.sampling import SobolQMCNormalSampler


# Training data
with manual_seed(1234):
    X = torch.rand(20, 1)
    Y = -100*X

# GP model
models_1d = [
    SingleTaskGP(X, Y[:, i].unsqueeze(-1), outcome_transform=Standardize(m=1))
    for i in range(Y.shape[-1])
]
model = ModelListGP(*models_1d)
model.eval()

# Fantasy models
target_x = torch.tensor([[0.5]])
sampler = SobolQMCNormalSampler(100, seed=9876)
fantasy_models = model.fantasize(target_x, sampler=sampler)

# Print results
print(f"Original model posterior mean: {model.posterior(target_x).mean.item()}")
print(f"Fantasy model posterior mean (average): {fantasy_models.posterior(target_x).mean.mean().item()}")

```

**Output**
We can see in the printed output that the fantasy models have a very different posterior mean.
```
Original model posterior mean: -48.430667877197266
Fantasy model posterior mean (average): -118.550048828125
```
If we remove the `outcome_transform` then the two means are instead very close, which is the expected behaviour.
```
Original model posterior mean: -44.10302734375
Fantasy model posterior mean (average): -44.10468673706055
```

## Expected Behavior
I would expect the `outcome_transform` to be applied in the correct place so that it does not introduce a bias in the fantasy models.

## System information

**Please complete the following information:**
- BoTorch version: 0.6.5
- GPyTorch version: 1.8.0
- PyTorch version: 1.12.0
- Python version: 3.10.5
- Computer OS: Ubuntu 20.04.4 LTS (Focal Fossa)

## Additional context
I believe the problem is in the `condition_on_observations` method. For a `SingleTaskGP`, this method contains the lines:
```python
        if hasattr(self, "outcome_transform"):
            # we need to apply transforms before shifting batch indices around
            Y, noise = self.outcome_transform(Y, noise)
```
[...]

PR #1338, opened 2022-08-03
outcome_transform ignored for ModelListGP.fantasize

esantorella:
This constitutes a very minimal solution to #1333 that I mean to improve. It could be tested more thoroughly and efficiently, and I (or any helpful reviewers :) ) should spend a little more time thinking about the logic.

## Motivation

#1333 

Yes

## Test Plan

[x] Test that with a large number of samples, the posterior mean of the original models is close to the average posterior mean of fantasy models.
[ ] Make this test more comprehensive -- try different dtypes and different types of `ModelListGP` component models
[x] Make this test more efficient -- ideally it shouldn't need such a large number of samples

### DIFF
--- a/test/models/test_model_list_gp_regression.py
+++ b/test/models/test_model_list_gp_regression.py
@@ -8,8 +8,6 @@ import itertools
 import warnings
 
 import torch
-from botorch.sampling import SobolQMCNormalSampler
-from botorch.utils import manual_seed
 from botorch.acquisition.objective import ScalarizedPosteriorTransform
 from botorch.exceptions.errors import BotorchTensorDimensionError
 from botorch.exceptions.warnings import OptimizationWarning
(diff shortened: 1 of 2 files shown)


test/models/test_model_list_gp_regression.py
```diff
@@ -312,3 +314,31 @@ def test_fantasize(self):
...
+                    for i in range(Y.shape[-1])
+                ]
+                model = ModelListGP(*models_1d)
+                model.eval()
+
+                # Fantasy models
+                target_x = torch.tensor([[0.5]])
+                # TODO: make this more efficient
+                sampler = SobolQMCNormalSampler(1000, seed=9876)
+                fantasy_models = model.fantasize(target_x, sampler=sampler)
```
Balandat, line 340: Actually doing an MC estimate here is nice in the sense that it tests this code e2e, but as you say it's not very efficient.

One alternative to comparing the statistics here would be to just mock things. E.g. one idea would be to just mock the `get_fantasy_model` call [here](https://github.com/pytorch/botorch/blob/main/botorch/models/model_list_gp_regression.py#L98) and then check that the passed in `targets` are actually transformed versions of the ones passed into `condition_on_observations`. 

Or you could mock even more and just check that every model in the list that has an outcome transform ha the outcome transform called on the appropriate index of the passed in samples `Y`.

test/models/test_model_list_gp_regression.py
```diff
@@ -312,3 +312,69 @@ def test_fantasize(self):
...
+            FixedNoiseGP(X, Y, yvar, outcome_transform=Standardize(m=1))
+        )
+
+        model.posterior(torch.zeros((1, 1)))
+
+        fant = model.fantasize(X, sampler=IIDNormalSampler(n_fants, seed=0), noise=yvar)
+
+        fant_mean = fant.posterior(X).mean.mean(0).detach().numpy().flatten()
+        self.assertAlmostEqual(fant_mean[0], y_at_low_x, delta=5e-4)
+        self.assertAlmostEqual(fant_mean[1], y_at_high_x, delta=1e-3)
```
Balandat, line 380: I know you're fixing the seed here but I recall some issues with that not working reliably, did you run a stress test of this to make sure this won't be flaky here?

test/models/test_model_list_gp_regression.py
```diff
@@ -312,3 +312,69 @@ def test_fantasize(self):
...
+        yvar = torch.full_like(Y, 1e-4)
+        model = ModelListGP(
+            FixedNoiseGP(X, Y, yvar, outcome_transform=Standardize(m=1))
+        )
+
+        model.posterior(torch.zeros((1, 1)))
+
+        fant = model.fantasize(X, sampler=IIDNormalSampler(n_fants, seed=0), noise=yvar)
+
+        fant_mean = fant.posterior(X).mean.mean(0).detach().numpy().flatten()
```
Balandat, line 378: No need to make into a numpy array, can just make it into a list of python scalars. 
```suggestion
        fant_mean = fant.posterior(X).mean.mean(0).flatten().tolist()
```

Balandat (approved):
Thanks, changes look good to me!

test/models/test_model_list_gp_regression.py
esantorella: yeah, Actions says this is not working reliably

test/models/test_model_list_gp_regression.py
esantorella: Yeah it would be nice to have some more e2e MC checks, but I don't love it as a unit test due to the inefficiency and high potential to become flaky from minor numerical issues.

Balandat (approved):
gtmtl, thanks!

Issue #1225, opened 2022-05-17
Problem with Polytope Sampler


jduerholt:
Hi,

I encountered a problem with the `get_polytope_samples` when the individual variables live in different dimensions. Here is a small didactic example:

```
from torch import tensor
from botorch.utils.sampling import get_polytope_samples
import matplotlib.pyplot as plt

samples = get_polytope_samples(
    n=100,
    bounds = tensor([[0.1,0.3,0.1,30.],[0.6,0.7,0.7,700.]]),
    equality_constraints=[(tensor([1, 2, 0]), tensor([1., 1., 1.]), 1.0)]
)

fig, ax = plt.subplots()
ax.hist(samples.numpy()[:,-1])
plt.show()
```
The bounds are as follows:

- 0.1 - 0.6
- 0.3 - 0.7
- 0.1 - 0.7
- 30 - 700

In addition an equality constraint is applied on the first three variables with all coefficients  equal to 1 and an RHS of 1. 

Sampling 100 samples leads to the following histogram for the fourth variables (bounds: 30-700). The distribution is not uniform at all and the values stay in a very slim interval.
![image](https://user-images.githubusercontent.com/28678090/168780883-01fbc131-a269-4589-a560-4e8a6b565bc6.png)

In the next step, I scaled all variables to into the 0 - 1 range, including the equality constraint by using the formula below: 

![image](https://user-images.githubusercontent.com/28678090/168782109-100fea64-4f9e-488f-b32e-4c175d67814f.png)
where `xi*` is the scaled variable and `s_i` is the difference between upper and lower bound for the variable. The code below generates 100 samples for this case including scaling back into the original bounds:

```
lower = tensor([0.1,0.3,0.1,30.])
upper = tensor([0.6,0.7,0.7,700.])

space = upper - lower

scaled_samples = get_polytope_samples(
    n=100,
    bounds = tensor([[0., 0., 0., 0.],[1., 1., 1., 1.]]),
    equality_constraints=[(tensor([1, 2, 0]), tensor([0.4000, 0.6000, 0.5000]), 0.5)]
)

samples = lower + scaled_samples*space

fig, ax = plt.subplots()
ax.hist(samples.numpy()[:,-1])
plt.show()
```

Now, the fourth variables is sampled in much more uniform way: 
![image](https://user-images.githubusercontent.com/28678090/168782936-f037a553-e204-4d55-9d0d-82d21a81344d.png)
[...]

PR #1341, opened 2022-08-05
Fix problem with polytope sampler

jduerholt:
## Motivation

As discussed in issue https://github.com/pytorch/botorch/issues/1225, the polytope sampler runs into problems if the variables live in different dimensions. This PR fixes the issue using the approach mentioned in the issue above.

Yes

## Test Plan

Unit tests

### DIFF
--- a/botorch/utils/sampling.py
+++ b/botorch/utils/sampling.py
@@ -846,9 +846,7 @@ def normalize_linear_constraints(
 
     new_constraints = []
     for index, coefficient, rhs in constraints:
-        # the index is cast to int64 since in some tests the index is
-        # given as a indexing incompatible dtype
-        lower, upper = bounds[:, index.to(dtype=torch.int64)]
+        lower, upper = bounds[:, index]
         s = upper - lower
         new_constraints.append(
             (index, s * coefficient, (rhs - torch.dot(coefficient, lower)).item())
--- a/botorch/utils/sampling.py
+++ b/botorch/utils/sampling.py
@@ -832,7 +832,7 @@ class DelaunayPolytopeSampler(PolytopeSampler):
 
 def normalize_linear_constraints(
     bounds: Tensor, constraints: List[Tuple[Tensor, Tensor, float]]
-):
+) -> List[Tuple[Tensor, Tensor, float]]:
     r"""Normalize linear constraints to the unit cube.
 
     Args:
--- a/botorch/utils/sampling.py
+++ b/botorch/utils/sampling.py
@@ -845,10 +845,14 @@ def normalize_linear_constraints(
     """
 
     new_constraints = []
-    for c in constraints:
-        lower, upper = bounds[:, index]
+    for index, coefficient, rhs in constraints:
+        # the index is cast to int64 since in some tests the index is
+        # given as a indexing incompatible dtype
+        lower, upper = bounds[:, index.to(dtype=torch.int64)]
         s = upper - lower
-        new_constraints.append((c[0], s * c[1], float(c[2] - torch.sum(c[1] * lower))))
+        new_constraints.append(
+            (index, s * coefficient, (rhs - torch.dot(coefficient, lower)).item())
+        )
     return new_constraints
 
 
@@ -860,7 +864,6 @@ def get_polytope_samples(
     seed: Optional[int] = None,
     thinning: int = 32,
     n_burnin: int = 10_000,
-    normalize_bounds: bool = True,
 ) -> Tensor:
     r"""Sample from polytope defined by box bounds and (in)equality constraints.
 
@@ -881,9 +884,6 @@ def get_polytope_samples(
         seed: The random seed.
         thinning: The amount of thinning.
         n_burnin: The number of burn-in samples for the Markov chain sampler.
-        normalize_bounds: transform the constraints to the [0,1] bounds
-            representing the unit cube, run the sampling and then transform
-            back to the original bounds. Defaults to True.
 
     Returns:
         A `n x d`-dim tensor of samples.
@@ -891,11 +891,11 @@ def get_polytope_samples(
     # create tensors representing linear inequality constraints
     # of the form Ax >= b.
     if inequality_constraints:
+        # normalize_linear_constraints is called to solve this issue:
+        # https://github.com/pytorch/botorch/issues/1225
         A, b = sparse_to_dense_constraints(
             d=bounds.shape[-1],
-            constraints=normalize_linear_constraints(bounds, inequality_constraints)
-            if normalize_bounds
-            else inequality_constraints,
+            constraints=normalize_linear_constraints(bounds, inequality_constraints),
         )
         # Note the inequality constraints are of the form Ax >= b,
         # but PolytopeSampler expects inequality constraints of the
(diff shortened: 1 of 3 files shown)


botorch/utils/sampling.py
```diff
@@ -838,6 +861,7 @@ def get_polytope_samples(
     seed: Optional[int] = None,
     thinning: int = 32,
     n_burnin: int = 10_000,
+    normalize_bounds: bool = True,
```
Balandat, line 864: Let's just do this by default and get rid of the option to simplify he interface? I don't really see a benefit for not doing this, worst case there is a bit of overhead but that's mostly negligible compared to the cost of drawing the samples.

test/utils/test_sampling.py
```diff
@@ -289,6 +290,28 @@ def test_sparse_to_dense_constraints(self):
...
+            bounds = torch.tensor(
+                [[0.1, 0.3, 0.1, 30.0], [0.6, 0.7, 0.7, 700.0]], **tkwargs
+            )
+            new_constraints = normalize_linear_constraints(bounds, constraints)
+            expected_coefficients = torch.tensor([0.4000, 0.6000, 0.5000], **tkwargs)
+            self.assertTrue(
+                torch.allclose(new_constraints[0][1], expected_coefficients)
+            )
+            expected_rhs = 0.5
+            self.assertTrue(np.allclose(new_constraints[0][-1], expected_rhs))
```
Balandat, line 313:
```suggestion
            self.assertAlmostEqual(new_constraints[0][-1], expected_rhs)
```

botorch/utils/sampling.py
```diff
@@ -878,17 +907,26 @@ def get_polytope_samples(
...
     else:
         dense_equality_constraints = None
     polytope_sampler = HitAndRunPolytopeSampler(
         inequality_constraints=dense_inequality_constraints,
-        bounds=bounds,
+        bounds=torch.tensor(
+            [list(torch.zeros(bounds.shape[-1])), list(torch.ones(bounds.shape[-1]))]
+        ).to(bounds)
+        if normalize_bounds
+        else bounds,
```
Balandat, line 922: would do sth like this for better readability:
```
normalized_bounds  = torch.zeros_like(bounds)
normalized_bounds[1, :] = 1.0
polytope_sampler = HitAndRunPolytopeSampler(...
```

Balandat (commented):
Thanks for the PR, overall this looks great. My main comment is that we should just do this by default, no need to give the user the option to shoot themselves in the foot (unless there are any downsides to this that I'm not seeing right now).

botorch/utils/sampling.py
```diff
@@ -830,6 +830,29 @@ def draw(self, n: int = 1, seed: Optional[int] = None) -> Tensor:
...
+            `\sum_i (X[indices[i]] * coefficients[i]) >= rhs` or
+            `\sum_i (X[indices[i]] * coefficients[i]) = rhs`.
+    """
+
+    new_constraints = []
+    for c in constraints:
+        lower = bounds[0, c[0].to(torch.int64)]
+        upper = bounds[1, c[0].to(torch.int64)]
+        s = upper - lower
+        new_constraints.append((c[0], s * c[1], float(c[2] - torch.sum(c[1] * lower))))
```
Balandat, line 852:
```suggestion
        new_constraints.append((c[0], s * c[1], float(c[2] - torch.dot(c[1], lower))))
```

botorch/utils/sampling.py
```diff
@@ -867,7 +894,9 @@ def get_polytope_samples(
     if inequality_constraints:
         A, b = sparse_to_dense_constraints(
             d=bounds.shape[-1],
-            constraints=inequality_constraints,
+            constraints=normalize_linear_constraints(bounds, inequality_constraints)
```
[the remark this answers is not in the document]
jduerholt: Done ;)

botorch/utils/sampling.py
```diff
@@ -830,6 +830,29 @@ def draw(self, n: int = 1, seed: Optional[int] = None) -> Tensor:
...
+        bounds (Tensor): A `2 x d`-dim tensor containing the box bounds.
+        constraints (List[Tuple[Tensor, Tensor, float]]): A list of
+            tuples (indices, coefficients, rhs), with each tuple encoding
+            an inequality constraint of the form
+            `\sum_i (X[indices[i]] * coefficients[i]) >= rhs` or
+            `\sum_i (X[indices[i]] * coefficients[i]) = rhs`.
+    """
+
+    new_constraints = []
+    for c in constraints:
```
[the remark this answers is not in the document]
jduerholt: good idea, I implemented it as suggest.

botorch/utils/sampling.py
jduerholt: I agree, I removed the keyword.

botorch/utils/sampling.py
jduerholt: Good idea.

botorch/utils/sampling.py
jduerholt: Done.

botorch/utils/sampling.py
```diff
@@ -830,6 +830,32 @@ def draw(self, n: int = 1, seed: Optional[int] = None) -> Tensor:
...
+            tuples (indices, coefficients, rhs), with each tuple encoding
+            an inequality constraint of the form
+            `\sum_i (X[indices[i]] * coefficients[i]) >= rhs` or
+            `\sum_i (X[indices[i]] * coefficients[i]) = rhs`.
+    """
+
+    new_constraints = []
+    for index, coefficient, rhs in constraints:
+        # the index is cast to int64 since in some tests the index is
+        # given as a indexing incompatible dtype
```
Balandat, line 850: It looks like you fixed this in the tests also? Can we get rid of this comment and the casting below then? I'd like to avoid a situation where someone is confusing the order or the `indices` and `coefficient` tensors and suddenly ends up with something that doesn't error out but produces garbage results.

jduerholt: There was another wrong type definition in one of the intializer tests. I also corrected this and removed the explicit cast.

test/optim/test_initializers.py
```diff
@@ -275,14 +275,14 @@ def test_gen_batch_initial_conditions_constraints(self):
             bounds = torch.tensor([[0, 0], [1, 1]], device=self.device, dtype=dtype)
             inequality_constraints = [
                 (
-                    torch.tensor([1], device=self.device, dtype=dtype),
+                    torch.tensor([1], device=self.device, dtype=torch.int64),
```
esantorella, line 278:
```suggestion
                    torch.tensor([1], device=self.device, dtype=dtype),
```
Let me know if I'm missing something, but tests still pass for me with the old version so I'd rather leave it how it is

test/optim/test_initializers.py
```diff
@@ -275,14 +275,14 @@ def test_gen_batch_initial_conditions_constraints(self):
...
                     torch.tensor([-4], device=self.device, dtype=dtype),
                     torch.tensor(-3, device=self.device, dtype=dtype),
                 )
             ]
             equality_constraints = [
                 (
-                    torch.tensor([0], device=self.device, dtype=dtype),
+                    torch.tensor([0], device=self.device, dtype=torch.int64),
```
esantorella, line 285:
```suggestion
                    torch.tensor([0], device=self.device, dtype=dtype),
```

test/optim/test_initializers.py
esantorella, line 278: Hmm, so non-int dtypes used to be acceptable in `gen_batch_initial_conditions` and `get_polytope_samples`, but now a non-int will cause a failure in `normalize_linear_constraints`. I'd prefer not to break backwards compatibility, but this should be fine for now since it isn't breaking any tests. I'll add this to the issue.

esantorella (approved):
Thanks so much for finding the bug and for fixing it! This looks great to me.

test/optim/test_initializers.py
Balandat: I think it's fine to break backward-compatibility here; it's kind of dicey to use floats to index tensor positions - more likely than not code that was doing that didn't actually mean to do that (that was the case with a type bug in our test suite as well).

PR #1343, opened 2022-08-09
Fix unit tests failues caused by upstream breaking changes

Balandat:
https://github.com/cornellius-gp/gpytorch/pull/2082 changed the shape of the mean constant (and did some other things). These changes fix the resulting unit test breakages.


saitcakmak (approved):
Thanks!

PR #1344, opened 2022-08-10
Add a `skip_expand` option to AppendFeatures

saitcakmak:
Summary: Adds a `skip_expand` option to `AppendFeatures` transform, which skips the `expand` operation in the `forward` pass, allowing it to be mixed with an `InputPerturbation` transform without multiplying the size of the output tensor. When used together, each perturbation and each appended feature will appear only once in the output tensor per q-batch.

Differential Revision: D38594609


Balandat (approved):
lgtm

PR #1351, opened 2022-08-16
fixing `test_fit_gpytorch_model` for Python versions before 3.8

SebastianAment:
Summary: [The previous PR](https://github.com/pytorch/botorch/pull/1350) introduced a new test that failed when run on Python 3.7. This was because it used the `kwargs` property of the [`call_args` property of mocks](https://docs.python.org/3/library/unittest.mock.html#unittest.mock.Mock.call_args), which was introduced in Python version 3.8. This changes the syntax to be compatible with earlier Python versions.

Differential Revision: D38762058


saitcakmak (approved):
Lgtm, pending lint fix.

PR #1353, opened 2022-08-17
[Codemod] Use time.monotonic instead of time.time to measure duration

Balandat:
Following best practices to avoid results being affected by system clock changes.


saitcakmak (approved):
Lgtm

Balandat:
Hmm interesting looks like this is failing on windows. Seems unlikely that this is caused by this change though? Running a test suite on main to verify.

Balandat:
Confirmed that conda windows failure is unrelated: https://github.com/pytorch/botorch/actions/runs/2879093795. Landing.

saitcakmak:
It is a bit weird that it suddenly started failing only on windows and only when using conda. Only if the error message way a little bit informative...

Balandat:
Yeah, presumably some conda package update was broken - let's just see if it goes away by itself, if it doesn't by early next week we can investigate further.

PR #1373, opened 2022-08-31
[Don't merge] Profiling utilities, test problems, and results

esantorella:
This may not need to be merged at all, but I'd love feedback on the content of the test problems -- are these relevant setups? Are parameters reasonable?

* botorch/profiling.py: Added profiling tools, defined test problems, and added notes on profiling results so far.
* run_profiling.py: Simple CLI for working with the functionality in botorch/profiling.py. This could be made much more user-friendly and easy to test with click or argparse, but that seems like overkill.
* test/test_profiling.py: Makes sure all of the test problems run.

Yes

## Test Plan

Unit tests


botorch/profiling.py
```diff
@@ -0,0 +1,247 @@
...
+@memory_profile
+def run_qnei(q: int = 100, dim: int = 10, n: int = 50, batch_dim: int = 5) -> None:
+    """
+    Run qNoisyExpectedImprovement.
+
+    Becomes memory-intensive as 'q' increases, risking OOMs. My Mac starts freezing
+    up around q = 10000. Interestingly, there doesn't seem to be any _deallocation_
+    within qNEI.__call__. Memory allocation increases a lot the first time it's called
+    and then slightly on subsequent calls. This is either due to caching or poor garbage
+    collection.
```
Balandat, line 72: This is most likely due to the caching. The main places I can think ofis the caching of the kernel matrices and decompositions on the gpytorch end.

This is not at all straightforward to track down, as this is happening via  `qNEI.__call__()/forward()` -> `botorch.models.model.Model.posterior()` -> `gpytorch.models.exact_gp.ExacGP.__call__` -> gpytorch's `PredictionStrategy` abstraction -> some linear operator manipulations -> ultimately calling something like [`LinearOperator._cholesky()`](https://github.com/cornellius-gp/linear_operator/blob/2e66c243a8ffc4359512bc270cec3f796a304688/linear_operator/operators/_linear_operator.py#L457-L458) which is cached by means of the [`@cached`](https://github.com/cornellius-gp/linear_operator/blob/2e66c243a8ffc4359512bc270cec3f796a304688/linear_operator/utils/memoize.py#L9-L14) decorator. 

Caching the Cholesky decomposition of `K(X, X) + sigma^2 I` (the train-train covariance plus noise diagonal) is something that definitely needs to happen. Some of this caching will also happen when sampling from the posterior covariance. That one is typically rather small in the `qEI` case, but due to how we compute `qNEI` it may be sizeable for `qNEI` and in general we should be fine to evict this from the cache. However, during acquisition function optimization we have to re-compute a new posterior covariance for each evaluation during the gradient-based optimization, so the impact of being smart about this here seems relatively small.

I do think there is a potential optimization opportunity though by exploiting the fact that we always compute the posterior covariance on `(X, X_test)`, and so we can likely cache large parts of the kernel matrices here (for the full covariance we are actually recomputing the kernel of `(X, X, X_test)`. That would require some pretty deep changes to gpytorch though (might well be worth doing this).

botorch/profiling.py
```diff
@@ -0,0 +1,247 @@
...
+def run_fit_fully_bayesian_model_nuts(dim: int = 100, n: int = 200) -> None:
+    """
+    Profile `fit_fully_bayesian_model_nuts` for time.
+
+    * Almost all of the time in `fit_fully_bayesian_model_nuts` is spent in
+        `mcmc.run`.
+    * Almost all of the time in `mcmc.run` is spent in `MCMC.sampler.run`.
+    * Almost all of the time in `MCMC.sampler.run` is in `_gen_samples`
+    * The time in `_gen_samples` is mainly spent in `kernel.sample`, with some
+        spent in `kernel.setup`.
```
Balandat, line 160: Yeah, the vast majority of the time *should* (would be good verify) be spent here: https://github.com/pytorch/botorch/blob/main/botorch/models/fully_bayesian.py#L159-L168

This is computing the train-train covariance under the new parameter (lengthscale, outputscale) samples and then samples from a MVN with that covariance (which means an `O(n^3)` Cholesky decomposition where `n` is the number of training examples.

botorch/profiling.py
```diff
@@ -0,0 +1,247 @@
...
+    213                                             n_train: int = 2,
+    214                                             train_batch_dim: int = 1,
+    215                                             n_points: int = 1,
+    216                                         ) -> None:
+    222    176.0 MiB      0.2 MiB           1       train_X, train_Y = _get_train_x_and_y(n_train, dim, train_batch_dim)
+    223
+    224    177.1 MiB      1.1 MiB           1       model = SingleTaskGP(train_X, train_Y)
+    225    178.9 MiB      1.8 MiB           1       posterior = model.posterior(train_X[0, :, :])
+    226                                             # memory_profile(posterior.sample)(sample_shape=torch.Size([b, n_points]))
+    227    224.9 MiB     46.0 MiB           1       posterior.sample(sample_shape=torch.Size([b, n_points]))
```
Balandat, line 236: this here is explained by the lazy evaluation of the covariance in the MVN that is contained in the `Posterior` object generated by the `posterior()` call.  `model.posterior` will generate that `Posterior` which has an `MVN` whose covariance matrix is implicitly represented as a `LinearOperator`. Upon the first `posterior.sample` call, that covariance matrix is actually instantiated and decomposed via a Cholesky decomposition. This decomposition is cached. Since all subsequent `sample` calls just do a matrix-vector multiplication with that cached Cholesky decomposition on the same `Posterior` and hence `MVN` object, this doesn't really need to allocate memory and is also very fast.

PR #1400, opened 2022-09-14
Deprecate `weights` argument of risk measures in favor of a `preprocessing_function`.

saitcakmak:
Summary:
Deprecates the `weights` argument of risk measures in favor of a `preprocessing_function`. This is superior in that it allows better modification of the samples before computing the risk measures.

This supports use cases such as filtering non-objective outcomes or applying feasibility weighting all within the risk measure itself. As a result, it helps avoid a number of if/else blocks when implementing robust optimization support in Ax.

Differential Revision: D39493308

### DIFF
--- a/botorch/acquisition/risk_measures.py
+++ b/botorch/acquisition/risk_measures.py
@@ -20,12 +20,22 @@ see also [Hong2014review]_.
     Computer Simulation, 2014.
 """
 
+import warnings
 from abc import ABC, abstractmethod
 from math import ceil
-from typing import List, Optional, Union
+from typing import Callable, List, Optional, Union
 
 import torch
-from botorch.acquisition.objective import MCAcquisitionObjective
+from botorch.acquisition.multi_objective.objective import (
+    IdentityMCMultiOutputObjective,
+    WeightedMCMultiOutputObjective,
+)
+from botorch.acquisition.objective import (
+    IdentityMCObjective,
+    LinearMCObjective,
+    MCAcquisitionObjective,
+)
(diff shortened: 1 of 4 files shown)


botorch/acquisition/risk_measures.py
```diff
@@ -49,20 +59,47 @@ class RiskMeasureMCObjective(MCAcquisitionObjective, ABC):
...
         """
         super().__init__()
         self.n_w = n_w
-        self.register_buffer(
-            "weights", torch.as_tensor(weights) if weights is not None else None
-        )
+        if weights is not None:
+            warnings.warn(
+                "`weights` argument of risk measures is deprecated and will be removed "
+                " in a future version. Use a `preprocessing_function` instead.",
```
Balandat, line 84:
```suggestion
                "in a future version. Use a `preprocessing_function` instead.",
```

test/acquisition/multi_objective/test_multi_output_risk_measures.py
```diff
@@ -498,7 +518,7 @@ def test_set_baseline_Y(self):
         mars.set_baseline_Y(model=model, X_baseline=X_baseline)
         self.assertTrue(torch.equal(mars.baseline_Y, torch.tensor([[-0.5, -0.5]])))
 
-    def test_get_Y_normalization_bounds(self):
+    def fest_get_Y_normalization_bounds(self):
```
Balandat, line 521: ftest...

Balandat (approved):
lgtm, couple of nits

PR #1431, opened 2022-09-28
Allow latest botorch version for Ax in website deployment flows

Balandat:
Prior to this change, installing Ax could have resulted in removing the installed botorch version and downgrading it to the version pinned in the Ax setup.py. This change avoids this issue.

### DIFF
--- a/.github/workflows/deploy_on_release.yml
+++ b/.github/workflows/deploy_on_release.yml
@@ -82,9 +82,9 @@ jobs:
     - name: Install dependencies
       shell: bash -l {0}
       run: |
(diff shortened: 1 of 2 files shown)


.github/workflows/deploy_on_release.yml
```diff
@@ -82,9 +82,9 @@ jobs:
     - name: Install dependencies
       shell: bash -l {0}
       run: |
-        conda install -y setuptools_scm multipledispatch conda-build conda-verify anaconda-client
+        conda install -y setuptools_scm conda-build conda-verify anaconda-client
+        conda install -y scipy sphinx pytest flake8 multipledispatch
```
Balandat, line 86: this change is just to logically group the deps together.

PR #1439, opened 2022-10-03
Botorch closures

j-wilson:
Summary:
**Changelog:**
- Enable user-defined loss closures.
- `fit_gptorch_torch` rewrite
- Add `fit_gyptorch_mll` dispatch for ApproximateGPs

Differential Revision: D39101211


botorch/fit.py
```diff
@@ -267,6 +278,56 @@ def _fit_fallback(
...
+    r"""Fallback method for fitting approximate Gaussian processes.
+
+    Args:
+        closure: Loss closure for `optimizer`.
+        optimizer: The underlying optimization algorithm to run. Default to
+            `fit_gpytorch_scipy` when `closure=None` and the model's internal
+            training set has no more than `full_batch_cutoff` observations; otherwise,
+            defaults to `fit_gpytorch_torch`.
+        data_loader: Convience keyword for passing in a DataLoader instance or dict of
+            keyword arguments passed to `get_data_loader` to obtain one. May only be
```
Balandat, line 302: It feels a bit weird that `data_loader` can either be an instance or a set of kwargs passed to a factory function.

botorch/fit.py
```diff
@@ -267,6 +278,56 @@ def _fit_fallback(
     raise ModelFittingError("All attempts to fit the model have failed.")
 
 
+@dispatcher.register(_ApproximateMarginalLogLikelihood, object, object)
+def _fit_fallback_approximate(
+    mll: _ApproximateMarginalLogLikelihood,
+    _: Type[Likelihood],
+    __: Type[ApproximateGPyTorchModel],
+    *,
+    closure: Optional[TLossClosure] = None,
```
Balandat, line 287: Should we call this `loss_closure` to be more clear/explicit?

botorch/optim/closures.py
```diff
@@ -0,0 +1,98 @@
...
+    ExactMarginalLogLikelihood,
+    MarginalLogLikelihood,
+    SumMarginalLogLikelihood,
+)
+from torch import Tensor
+from torch.utils.data import DataLoader
+
+dispatcher = Dispatcher("get_loss_closure")
+NoneType = type(None)
+TLossClosure = Callable[[], Tensor]
```
Balandat, line 25: Is this the right type if you're forwarding kwargs to the MLL?

botorch/optim/closures.py
```diff
@@ -0,0 +1,98 @@
...
+):
+    r"""Fallback closure for external training data."""
+
+    def _closure(**kwargs: Any) -> Tensor:
+        tensors = next(iter(data_loader))
+        if not isinstance(tensors, Sequence):
+            raise TypeError
+
+        tensors_iter = iter(tensors)
+        model_output = mll.model(*islice(tensors_iter, len(mll.model.train_inputs)))
```
Balandat, line 46: So this will ignore all excess inputs provided by `tensors_iter` that are at a position later than the length of the training inputs. Is this what we want here or should we error out if these are not compatible in size? If we silently drop those this could be hard to debug down the road.

botorch/optim/fit.py
```diff
@@ -179,119 +183,80 @@ def store_iteration(xk):
...
-    i = 0
-    stop = False
-    stopping_criterion = ExpMAStoppingCriterion(
-        **_filter_kwargs(ExpMAStoppingCriterion, **optim_options)
-    )
-    train_inputs, train_targets = mll.model.train_inputs, mll.model.train_targets
-    while not stop:
-        optimizer.zero_grad()
+    if isinstance(torch_optimizer, (type, partial)):
+        torch_optimizer = torch_optimizer(params=list(param_dict.values()))
```
Balandat, line 234: I am not the biggest fan of the pattern of allowing this to be either an instance or a factory function. Not sure if there is an elegant alternative though.

botorch/optim/fit.py
```diff
@@ -179,119 +183,80 @@ def store_iteration(xk):
...
+        closure: Callable used to evaluate the loss.
+        bounds: A dictionary mapping parameter names to tuples of lower and upper
+            bounds.
+        step_limit: Optional integer specifying the maximum number of GD steps.
+        stopping_criterion: A StoppingCriterion for GD.
+        torch_optimizer: A `torch.optim.Optimizer` instance or a factory that takes
+            a list of parameters and returns an `Optimizer` instance.
+        torch_scheduler: A `torch.optim.lr_scheduler._LRScheduler` instance or a factory
+            that takes a `Optimizer` instance and returns a `_LRSchedule` instance.
+        callable: Optional function taking an `OptimizationIteration` as its sole
```
Balandat, line 208:
```suggestion
        callback: Optional function taking an `OptimizationIteration` as its sole
```

botorch/optim/fit.py
```diff
@@ -179,119 +183,80 @@ def store_iteration(xk):
...
+        torch_optimizer.step()
+        for name, limits in bounds_dict.items():
+            param = param_dict[name]
+            param.data = param.clamp(*limits)
+
+        if torch_scheduler is not None:
+            torch_scheduler.step()
+
+    loss = None if loss is None else loss.detach().cpu().item()
+    return mll, {"fopt": loss, "wall_time": monotonic() - start_time}
```
Balandat, line 262: should we include `iterations` here for backwards compatibility? I guess this is sufficiently deep in the stack that we don't need to worry too much about this

botorch/optim/numpy_converter.py
```diff
@@ -42,6 +45,21 @@ class TorchAttr(NamedTuple):
...
+def _handle_numerical_errors(
+    error: RuntimeError, x: np.ndarray
+) -> Tuple[float, np.ndarray]:
+    if isinstance(error, NotPSDError):
+        raise error
+    error_message = error.args[0] if len(error.args) > 0 else ""
+    if (
+        isinstance(error, NanError)
+        or "singular" in error_message  # old pytorch message
+        or "input is not positive-definite" in error_message  # since pytorch #63864
```
Balandat, line 57: #63864 has long made it into the releases and is part of 1.11 our min version requirement. 

```suggestion
        or "input is not positive-definite" in error_message
```

botorch/optim/numpy_converter.py
```diff
@@ -240,3 +258,37 @@ def set_params_with_array(
...
     return module
+
+
+def get_loss_closure_numpy(
+    mll: MarginalLogLikelihood,
+    property_dict: Dict[str, TorchAttr],
+    torch_closure: Optional[TLossClosure] = None,
+    parameter_setter: Callable[[np.ndarray], Any] = set_params_with_array,
+) -> Callable[[], Tuple[np.ndarray, np.ndarray]]:
+    if torch_closure is None:
```
Balandat, line 269: add a docstring here?

botorch/optim/numpy_converter.py
```diff
@@ -240,3 +258,37 @@ def set_params_with_array(
...
+        mll.zero_grad()
+        try:  # catch linear algebra errors in gpytorch
+            loss = torch_closure()
+            loss.backward()
+        except RuntimeError as e:
+            return _handle_numerical_errors(error=e, x=x)
+
+        i = 0
+        param_dict = dict(mll.named_parameters())
+        grad = np.zeros(sum([tattr.shape.numel() for tattr in property_dict.values()]))
```
Balandat, line 283:
```suggestion
        grad = np.zeros(sum(tattr.shape.numel() for tattr in property_dict.values()))
```

botorch/optim/fit.py
```diff
@@ -179,119 +183,80 @@ def store_iteration(xk):
...
+        if stopping_criterion and stopping_criterion.evaluate(fvals=loss.detach()):
+            break
+
+        loss.backward()
+        torch_optimizer.step()
+        for name, limits in bounds_dict.items():
+            param = param_dict[name]
+            param.data = param.clamp(*limits)
+
+        if torch_scheduler is not None:
```
jacobrgardner, line 258: The docstring suggests `torch_scheduler` could also be a factory, but I don't see where that's checked, and here it's assumed to already have `step`

jacobrgardner:
This looks like a pretty reasonable solution to me! I actually like the ability to roll my own loss closures and optimizers here, and think it's worth the extra engineering. @nataliemaus has code that sometimes wants to train a GP end to end with a VAE included in the ELBO, and sometimes only wants to train the GP, so this should let us switch even more code into using BoTorch routines.

botorch/optim/fit.py
j-wilson: Would factory only be better or worse? Note that instances can still be passed as `lambda **_: instance`, but the typing is now more explicit.

botorch/optim/fit.py
j-wilson: Callbacks Are All You Need. Wilson et al., 2022.


Jokes aside: Is it just a backward compatibility thing or are you thinking that some conveniences would be useful here?

botorch/optim/fit.py
jacobrgardner: My $0.02: being able to pass an Instance is nicer than the factory, if only because then I don't necessarily need all the parameters I want to optimize to live somewhere on `mll`. There's even a reasonable argument to be made for accepting lists of Optimizers and stepping all of them, to support e.g. NGD or different Optimizers / learning rates for the GP vs a DNN

botorch/optim/numpy_converter.py
j-wilson: `sum` accepts iterators.

botorch/optim/fit.py
j-wilson: > being able to pass an Instance is nicer than the factory, if only because then I don't necessarily need all the parameters I want to optimize to live somewhere on mll.

So, here's what we can do. We can split up `fit_gpytorch_torch` into a pair of methods: i) a generic torch-based gradient descent method and ii) a wrapper that provides mll-based model fitting conveniences. This seems to address your first point.

> There's even a reasonable argument to be made for accepting lists of Optimizers and stepping all of them, to support e.g. NGD or different Optimizers / learning rates for the GP vs a DNN

Interesting idea. I like that this eliminates the need for `step_limit`. If you wanted to go further, you could pass in a list of `OptimizationStep` objects with slots for `closure`, `optimizer`, etc . Unclear to me whether we run into any scheduling issues here.

botorch/optim/closures.py
j-wilson: They aren't dropped actually. We expand `tensors_iter` when calling `mll` below and this consumes the remainder of the iterator.

botorch/optim/closures.py
j-wilson: Not really, no. If we end up allowing `kwargs` to be passed to loss closures (debatable since `torch.jit.script` does not support this), we'll make `TLossClosure` a `typing.Protocol`.

PR #1446, opened 2022-10-10
Fix matplotlib error in BOPE tutorial

saitcakmak:
Summary: Matplotlib 3.6.1 has a bug with NaN handling (https://github.com/matplotlib/matplotlib/issues/24127), leading to errors when running the tutorial. This implements a workaround, replacing `yerr` with NaNs with `None`.

Differential Revision: D40232488


Balandat:
Also, is there a reason to switch `fit_gpytorch_mll` back to `fit_gpytorch_model`?

saitcakmak:
Good catch. I must've used an older version of the notebook, will fix.

PR #1447, opened 2022-10-10
Buxfix for Proximal acquisition function wrapper for negative base acquisition functions

roussel-ryan:
## Motivation
This PR fixes a major issue when using the ```ProximalAcquisitionFunction``` with base acquisition functions that are not strictly positive. This PR fixes it by applying a Softplus transformation to the base acquisition function values (using optional beta = 1,0 argument) before multiplying by proximal weighting.

Yes

## Test Plan
Tests have been updated with correct (softplus transformed) values.


saitcakmak (approved):
Thanks, lgtm! The tutorial failure is unrelated (#1446), so please ignore it.

SebastianAment:
This is a higher level design question, rather than a comment on the particular fix. 

@roussel-ryan, is there a reason against simply adding a quadratic to an arbitrary acquisition function? With this fix, `ProximalAcquisitionFunction` will compute
```
log(1+exp(acquisition(x))) * exp(-Q(x)),
```
where `Q(x) = (x-x_{n-1}).t() * proximal_weights^{-1} * (x-x_{n-1}))`. Similar biasing could be achieved with the simpler expression
```
acquisition(x) + Q(x).
```
In addition, the latter would not change the relative magnitude of the acquisition function to the proximal term if the acquisition is negative. If it takes a large negative value, `softplus` might also lead to numerical instabilities. 
Would love to hear your thoughts!

Balandat:
Thanks for fixing this. One thing that we should keep in mind here is that if the acquisition function values are small (as is often the case later during the optimization for things like Expected Improvement) this could substantially impact the behavior for functions that are indeed guaranteed to be positive. So in that case it seems that the choice of a large `beta` would be necessary (and in general this highlights the importance of the choice of `beta`). Would it make sense to allow passing in `beta` as optional and fall back to the current behavior?

roussel-ryan:
@SebastianAment In response to large negative values for the base acquisition function I'm hoping that using a standardize outcome transform will prevent extreme values. It's not a perfect solution but I think for most cases the proposed solution is sufficient. As for adding the biasing, instead of multiplying it you mentioned that the relative scale of each term would remain the same, but I think it presents the biasing from working when the acquisition function is large (or vice versa). Unless I'm misunderstanding what you wrote?

roussel-ryan:
@Balandat I could potentially remove the transform when all of the acquisition function values are found to be positive. Alternatively I could set a large beta as the default or raise an error when the acqf values are negative. We need to prevent it from ever being used with negative acqf because the issue can be quite easy to overlook

SebastianAment:
> As for adding the biasing, instead of multiplying it you mentioned that the relative scale of each term would remain the same, but I think it presents the biasing from working when the acquisition function is large (or vice versa)

That makes sense. The multiplicative version lets us choose weights proportional to the scale of the acquisition functions, which we might not know a priori but could update during a run. It's also equivalent to the additive version by taking the log of the acquisition function, and that might be more easily applied to negative acquisition values. 
Thanks for the quick response, enjoyed reading your paper!

Balandat:
I am not sure how widely used this is. The main thing I'm worried about is that the default behavior for non-negative but small-in-magnitude acquisition functions may change quite a bit with this change. How about we (i) make `beta` optional and if `None` do not apply the softplus, and (ii) raise an error if any of the acquisition values are negative, pointing the user to set a beta value?

roussel-ryan:
@Balandat made the change you suggested

botorch/acquisition/proximal.py
```diff
@@ -49,6 +53,7 @@ def __init__(
         acq_function: AcquisitionFunction,
         proximal_weights: Tensor,
         transformed_weighting: bool = True,
+        beta: float = None,
```
Balandat, line 56:
```suggestion
        beta: Optional[float] = None,
```

test/acquisition/test_proximal.py
```diff
@@ -86,6 +92,33 @@ def test_proximal(self):
...
+                            proximal_test_X = input_transform(test_X)
+
+                    mv_normal = MultivariateNormal(last_X, torch.diag(proximal_weights))
+                    test_prox_weight = torch.exp(
+                        mv_normal.log_prob(proximal_test_X)
+                    ) / torch.exp(mv_normal.log_prob(last_X))
+
+                    ei_prox_beta = EI_prox_beta(test_X)
+                    self.assertTrue(torch.allclose(ei_prox_beta, ei * test_prox_weight))
+                    self.assertTrue(ei_prox_beta.shape == torch.Size([1]))
```
Balandat, line 120: nit
```suggestion
                    self.assertEqual(ei_prox_beta.shape, torch.Size([1]))
```

test/acquisition/test_proximal.py
```diff
@@ -86,6 +92,33 @@ def test_proximal(self):
...
+                    proximal_test_X = test_X.clone()
+                    if transformed_weighting:
+                        if input_transform is not None:
+                            last_X = input_transform(train_X[-1])
+                            proximal_test_X = input_transform(test_X)
+
+                    mv_normal = MultivariateNormal(last_X, torch.diag(proximal_weights))
+                    test_prox_weight = torch.exp(
+                        mv_normal.log_prob(proximal_test_X)
+                    ) / torch.exp(mv_normal.log_prob(last_X))
```
Balandat, line 116:
```suggestion
                    test_prox_weight = torch.exp(
                        mv_normal.log_prob(proximal_test_X) -
                        mv_normal.log_prob(last_X)
                    )
```

botorch/acquisition/proximal.py
```diff
@@ -27,15 +27,19 @@
 
 class ProximalAcquisitionFunction(AcquisitionFunction):
     """A wrapper around AcquisitionFunctions to add proximal weighting of the
-    acquisition function. Acquisition function is weighted via a squared exponential
-    centered at the last training point, with varying lengthscales corresponding to
-    `proximal_weights`. Can only be used with acquisition functions based on single
-    batch models.
+    acquisition function. Then the acquisition function is
+    weighted via a squared exponential centered at the last training point,
```
Balandat, line 31:
```suggestion
    acquisition function. The acquisition function is
    weighted via a squared exponential centered at the last training point,
```

botorch/acquisition/proximal.py
```diff
@@ -27,15 +27,19 @@
...
+    with varying lengthscales corresponding to `proximal_weights`. Can only be used
+    with acquisition functions based on single batch models. Acquisition functions
+    must be positive or `beta` is specified to apply a SoftPlus transform before
```
Balandat, line 34:
```suggestion
    must be positive or `beta` must be specified to apply a SoftPlus transform before
```

test/acquisition/test_proximal.py
```diff
@@ -133,6 +166,16 @@ def test_proximal(self):
...
 
+                    # test acquisition function with
+                    # negative values w/o softplus transform specified
+                    negative_acqf = NegativeAcquisitionFunction(model)
+                    bad_neg_prox = ProximalAcquisitionFunction(
+                        negative_acqf, proximal_weights=proximal_weights
+                    )
+
+                    with self.assertRaises(RuntimeError):
+                        bad_neg_prox(test_X)
```
Balandat, line 177: Can you use `assertRaisesRegex` in order to check for the proper error message here?

Balandat (commented):
Thanks! A couple of nits, o/w this lgtm.

Balandat (approved):
No problem, thanks for the contribution!

@saitcakmak I think this is ready for a re-import and merge!

PR #1452, opened 2022-10-17
Let Pyre know that `AcquisitionFunction.model` is a `Model`

esantorella:
## Motivation

Pyre is not smart enough to understand that calling `self.add_module('model', model)` makes `self.model` have the type of `model`, which is true due to some fairly complex underlying logic inherited from `torch.nn.Module`. However, PyTorch is smart enough to properly `add_module` if we just do `self.model = model`. This also works for tensors, but only if the tensor is explicitly registered as a buffer (by name, not necessarily by value) before assignment.

Yes

## Test Plan
- Unit tests should be unaffected
- Pyre error count drops from 1379 to 1309 (down 5%).
- Added explicit tests that `_modules` and `_buffers` are properly initialized

### DIFF
--- a/botorch/acquisition/acquisition.py
+++ b/botorch/acquisition/acquisition.py
@@ -36,7 +36,7 @@ class AcquisitionFunction(Module, ABC):
             model: A fitted model.
         """
         super().__init__()
-        self.add_module("model", model)
+        self.model: Model = model
 
     @classmethod
     def _deprecate_acqf_objective(
--- a/botorch/acquisition/analytic.py
+++ b/botorch/acquisition/analytic.py
@@ -462,13 +462,22 @@ class ConstrainedExpectedImprovement(AnalyticAcquisitionFunction):
                 con_upper_inds.append(k)
                 con_upper.append(constraints[k][1])
         # tensor-based indexing is much faster than list-based advanced indexing
-        self.register_buffer("con_lower_inds", torch.tensor(con_lower_inds))
-        self.register_buffer("con_upper_inds", torch.tensor(con_upper_inds))
-        self.register_buffer("con_both_inds", torch.tensor(con_both_inds))
-        # tensor indexing
-        self.register_buffer("con_both", torch.tensor(con_both, dtype=torch.float))
-        self.register_buffer("con_lower", torch.tensor(con_lower, dtype=torch.float))
-        self.register_buffer("con_upper", torch.tensor(con_upper, dtype=torch.float))
+        for k in [
+            "con_lower_inds",
+            "con_upper_inds",
+            "con_both_inds",
+            "con_both",
+            "con_lower",
+            "con_upper",
+        ]:
+            self.register_buffer(k, tensor=None)
+
+        self.con_lower_inds = torch.tensor(con_lower_inds)
+        self.con_upper_inds = torch.tensor(con_upper_inds)
+        self.con_both_inds = torch.tensor(con_both_inds)
+        self.con_both = torch.tensor(con_both)
+        self.con_lower = torch.tensor(con_lower)
+        self.con_upper = torch.tensor(con_upper)
 
     def _compute_prob_feas(self, X: Tensor, means: Tensor, sigmas: Tensor) -> Tensor:
         r"""Compute feasibility probability for each batch of X.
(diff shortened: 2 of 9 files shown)


botorch/acquisition/analytic.py
```diff
@@ -470,6 +470,14 @@ def _preprocess_constraint_bounds(
         self.register_buffer("con_lower", torch.tensor(con_lower, dtype=torch.float))
         self.register_buffer("con_upper", torch.tensor(con_upper, dtype=torch.float))
 
+        # All this is to make Pyre happy
```
Balandat, line 473: :(

botorch/acquisition/acquisition.py
```diff
@@ -37,6 +37,7 @@ def __init__(self, model: Model) -> None:
         """
         super().__init__()
         self.add_module("model", model)
+        self.model: Model = self.model
```
Balandat, line 40: What if you just do `self.model = model` here? I believe that if `model` is a `torch.nn.Module` then the module registration should happen automagically under the hood.

botorch/acquisition/analytic.py
Balandat: Ideally this wouldn't be necessary, not sure though whether that's possible

botorch/acquisition/knowledge_gradient.py
```diff
@@ -150,12 +150,15 @@ def __init__(
                     "If using a multi-output model without an objective, "
                     "posterior_transform must scalarize the output."
                 )
-        self.sampler = sampler
+        self.sampler: MCSampler = sampler
         self.objective = objective
         self.posterior_transform = posterior_transform
         self.set_X_pending(X_pending)
+        # TODO: Set self.X_pending directly in the __init__ to make this look more
```
Balandat, line 157: there is some handling of error emission etc in the `set_X_pending` method - ideally this wouldn't have to be repeated in each `__init__` method

botorch/acquisition/knowledge_gradient.py
```diff
@@ -338,7 +341,7 @@ def __init__(
         project: Callable[[Tensor], Tensor] = lambda X: X,
         expand: Callable[[Tensor], Tensor] = lambda X: X,
         valfunc_cls: Optional[Type[AcquisitionFunction]] = None,
-        valfunc_argfac: Optional[Callable[[Model, Dict[str, Any]]]] = None,
+        valfunc_argfac: Optional[Callable[[Model], Dict[str, Any]]] = None,
```
Balandat, line 344: oops

botorch/acquisition/acquisition.py
esantorella: Yep that works! Making this change and adding tests for proper initialization.

botorch/acquisition/analytic.py
esantorella: New version, hopefully less ugly

esantorella: New version. Still verbose, but more explicit about what's going on and less redundant.

botorch/acquisition/analytic.py
```diff
@@ -462,13 +462,23 @@ def _preprocess_constraint_bounds(
...
+            "con_lower_inds",
+            "con_upper_inds",
+            "con_both_inds",
+            "con_both",
+            "con_lower",
+            "con_upper",
+        ]:
+            self.register_buffer(k, tensor=None)
+
+        # All this is to make Pyre happy
```
esantorella, line 475:
```suggestion
```

botorch/acquisition/analytic.py
Balandat: > I think it should be possible to do all of this in one line per assignment with self.con_lower = torch.tensor(con_lower) and it should be registered as a buffer.

This only works for modules, not for buffers. To make something a buffer, you actually have to use `register_buffer` (not sure if there is a way to make a Buffer object or sth like that and then assign this though). The reason for this is that there are plenty of cases where you'll want to make a tensor an instance attribute but not a buffer (so that it doesn't get added to the `state_dict`). The reason we have these tensors as buffer here is so that a `module.to(dtype=, device=)` call automatically moves them over to the right device/dtype if needed.

PR #1458, opened 2022-10-19
Add entropy search acquisition functions

benmltu:
Hi,

I have provided some implementations of entropy search acquisition functions used in a recent paper (https://arxiv.org/abs/2210.02905). This PR includes the acquisition function and the necessary utilities. I have included a notebook that describes how to use these methods.

I was not sure what were the best places to add these acquisition functions, so I put them all in the multi-objective folder. Nevertheless, they should work for single-objective optimization as well.

Thanks,
Ben


sdaulton:
Thanks for putting this up. I will review shortly. Note that BoTorch requires 100% test coverage, so you will need to write unit tests before this PR can be merged in.

sdaulton:
Regarding testing, my comment was not intended to be criticizing. I am super excited about your PR for these AFs. The code looks quite clean. I just wanted to flag the need for unit tests since that will likely be a fair bit of work. There are many examples of testing other AFs, e.g. https://github.com/pytorch/botorch/blob/main/test/acquisition/multi_objective/test_monte_carlo.py. Let me know if you have any questions on testing. Thanks again for the PR!

botorch/acquisition/multi_objective/predictive_entropy_search.py
```diff
@@ -0,0 +1,1629 @@
...
+    multivariate normal distribution with mean mu and covariance Sigma, the natural
+    mean is Sigma^{-1} mu and the natural covariance is Sigma^{-1}.
+
+    Args:
+        X: A `batch_shape x R x d`-dim Tensor.
+        model: The fitted model.
+        observation_noise: If true the posterior is computed with observation noise.
+        jitter: The jitter added to the covariance matrix.
+        natural: If true we compute the natural statistics as well.
+    Return:
```
sdaulton, line 679: Probably need to follow this format for the autodocs: https://github.com/pytorch/botorch/blob/970664c2d8516dca7c583858a10794f78c84020f/botorch/optim/optimize.py#L110-L118

sdaulton (commented):
Thanks for putting this up. This is a partial initial pass (I didn't make it all the way through the PR). 

One high level comment is that we should make sure that this is dtype/device agnostic, so that we can run it on a GPU--the main blocker at the moment is initializing tensors (e.g. `torch.ones()`) without specifying dtype and device. Typically these can be fetched from some other tensor (e.g. X).

Another comment is that much of the code is shared between JES and MES (for example in `__init__`. It would be good to consolidate this into one common base class.

Also, it would be nice to highlight in the docstrings what functions are differentiable/not for easy of use.

eytan:
@benmltu , thanks so much for the PR—the code here is very high quality, as are these ES implementations.  This will be super valuable to the community.  I am also looking forward to seeing if there are ways to further accelerate JES through GPU-based computation once some of these type issues have been resolved.

hvarfner:
I figured I'd just chime in here. I'm also extremely grateful for this your job, thanks a lot. 

I've went through the code, tried to run it and to implement my own stuff. As far as plugging in different versions of JES, our differences are very minor and can easily be accounted for within the current structure (and probably not even two different classes).

Otherwise, I only have a couple of comments: 
- the `sampling_noise` argument is unused.
- pymoo 0.6.0 does not seem to support the pareto sampling of optima (0.5.0 was fine) 
- Single-objective MES has its sampling of maxima inside the acquisition function, while these implementations have the optima as arguments to be passed to the acquisition function. I personally prefer your approach, since one may want to experiment with sampling procedures quite a bit.

Once again, thanks a lot!

benmltu:
Thanks for all the comments, I've updated the code to address many of the issues. Main changes: 

- Unit tests are now included and everything should be dtype/device agnostic.
- JES and MES have been refactored to share a common base class.
- The acquisition functions for the multi-objective and single-objective PES and JES have been separated. I didn't do this for MES because there are already exist many different single-objective implementations. From inspection, the existing implementations do more of less same thing with some minor differences arising in how the max-values are set and how batches are handled.
- The code has been updated to pymoo 0.6.0.
- The descriptions of the methods has been updated---there were some artefacts from earlier versions that was noted in the comments.
- Constraint handling for PES has been removed. The EP updates relating to the inequality constraints appear to be unstable in some settings, it is hard to see how to fix this one without going through all the equations again.

Balandat (commented):
Excellent, thanks so much for the contribution and the many updates in response to the review. This is really great to see.

One high-level comment is that In general we're trying to keep the dependencies of botorch as lightweight as possible, so it would be great if we could make `pymoo` an optional dependency. One way to do that would be to do something like 

```
try: 
    import pymoo
    
    PYMOO_INSTALLED = True
except ImportError:
    PYMOO_INSTALLED = False
```

at the beginning of the `sample_pareto.py` module, move the imports from pymoo into the functions inside the module, and if `PYMOO_INSTALLED` is `False` just raise an informative exception prompting the user to install pymoo if they would like to the newly added entropy search acquisition functions.

esantorella:
@benmltu FYI, I merged the "main" branch into this one, so please do a "git pull" before doing any local development. I also imported it into an internal Meta repo to see if internal tools catch any issues.

botorch/utils/sample_pareto.py
```diff
@@ -0,0 +1,281 @@
...
+from torch import Tensor
+
+
+def pareto_solver(
+    model: GenericDeterministicModel,
+    bounds: Tensor,
+    num_objectives: int,
+    num_generations: int = 100,
+    pop_size: int = 100,
+    num_offsprings: int = 10,
```
sdaulton, line 46: Hmm are these good defaults? 10 offsprings seems quite small. Why not use the pymoo defaults?

botorch/utils/sample_pareto.py
```diff
@@ -0,0 +1,281 @@
...
+    num_rff_features: int = 512,
+    max_tries: int = 3,
+    num_greedy: int = 0,
+    X_baseline: Optional[Tensor] = None,
+    ref_point: Optional[Tensor] = None,
+) -> Tuple[Tensor, Tensor]:
+    r"""Computes the Pareto optimal set and front from samples of the GP.
+
+    (i) Samples are generated using random Fourier features.
```
sdaulton, line 133: This need not be RFFs. It could also use decoupled sampling. Let's add a TODO to generalize the GP sampling bit

botorch/utils/sample_pareto.py
```diff
@@ -0,0 +1,281 @@
...
+        - A `num_pareto_samples x num_pareto_points x d`-dim Tensor containing the
+            collection of Pareto optimal inputs.
+        - A `num_pareto_samples x num_pareto_points x M`-dim Tensor containing the
+            collection of Pareto optimal objectives.
+    """
+    tkwargs = {"dtype": bounds.dtype, "device": bounds.device}
+    M = model.num_outputs
+    d = bounds.shape[-1]
+
+    if M == 1:
```
sdaulton, line 172: It seems that this function should simply raise an exception if `M==1`.

botorch/utils/sample_pareto.py
```diff
@@ -0,0 +1,281 @@
...
+            num_tries = num_tries + 1
+
+        # If maximum number of retries exceeded throw out a runtime error.
+        if ratio > 1:
+            error_text = (
+                "Only found "
+                + str(num_pareto_generated)
+                + " Pareto efficient points instead of "
+                + str(num_pareto_points)
+                + "."
```
sdaulton, line 223: let's use f-strings

botorch/utils/sample_pareto.py
```diff
@@ -0,0 +1,281 @@
...
+            )
+            raise RuntimeError(error_text)
+
+        # Randomly truncate the Pareto set and front
+        if num_greedy == 0:
+            indices = torch.randperm(num_pareto_generated)[:num_pareto_points]
```
sdaulton, line 229: does this not lead to device errors?

botorch/utils/sample_pareto.py
```diff
@@ -0,0 +1,281 @@
...
+                # Update the pareto front.
+                pending_pareto_front_i = torch.cat(
+                    [
+                        pending_pareto_front_i,
+                        pareto_front_i[best_index : best_index + 1, :],
+                    ],
+                    dim=0,
+                )
+
+            indices = torch.tensor(indices)
```
sdaulton, line 276: same here. does this not lead to device errors?

botorch/utils/sample_pareto.py
```diff
@@ -0,0 +1,281 @@
...
+                    .prod(dim=-1)
+                    .sum(dim=-1)
+                )
+
+                # Zero out the pending points.
+                hvi[indices] = 0
+
+                # Store best index.
+                best_index = torch.argmax(hvi).tolist()
+                indices = indices + [best_index]
```
sdaulton, line 265:
```suggestion
                indices.append(best_index)
```

botorch/utils/sample_pareto.py
```diff
@@ -0,0 +1,281 @@
...
+                    )
+                    .prod(dim=-1)
+                    .sum(dim=-1)
+                )
+
+                # Zero out the pending points.
+                hvi[indices] = 0
+
+                # Store best index.
+                best_index = torch.argmax(hvi).tolist()
```
sdaulton, line 264:
```suggestion
                best_index = torch.argmax(hvi).item()
```

botorch/utils/sample_pareto.py
```diff
@@ -0,0 +1,281 @@
...
+
+                # `1 x num_boxes x M`
+                lo = hypercell_bounds[0].unsqueeze(0)
+                up = hypercell_bounds[1].unsqueeze(0)
+
+                # Compute the sample hypervolume improvement.
+                hvi = (
+                    torch.max(
+                        torch.min(pareto_front_i.unsqueeze(-2), up) - lo,
+                        torch.zeros(lo.shape).to(bounds),
```
sdaulton, line 254:
```suggestion
                        torch.zeros(lo.shape, dtype=bounds.dtype, device=bounds.device),
```

botorch/utils/sample_pareto.py
```diff
@@ -0,0 +1,281 @@
...
+                    ref_point=ref_point, Y=pending_pareto_front_i
+                )
+                hypercell_bounds = partitioning.hypercell_bounds
+
+                # `1 x num_boxes x M`
+                lo = hypercell_bounds[0].unsqueeze(0)
+                up = hypercell_bounds[1].unsqueeze(0)
+
+                # Compute the sample hypervolume improvement.
+                hvi = (
```
sdaulton, line 251: It seems like this would be handy to have as an instance method of `BoxDecomposition`s (e.g. a method to compute the incremental HVI of a new point --- or more generally `(batch_shape x) 1 x M` new points)

botorch/utils/sample_pareto.py
```diff
@@ -0,0 +1,281 @@
+#!/usr/bin/env python3
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+#
+# This source code is licensed under the MIT license found in the
+# LICENSE file in the root directory of this source tree.
+
+r"""
+Some methods for sampling the Pareto optimal points. This relies on the pymoo
```
sdaulton, line 8: It seems like this should live in botorch/utils/multi_objective/sample_pareto.py

botorch/acquisition/joint_entropy_search.py
```diff
@@ -0,0 +1,104 @@
...
+
+        Args:
+            model: A fitted single-outcome model.
+            optimal_inputs: A `num_samples x d`-dim tensor containing the sampled
+                optimal inputs of dimension `d`. We assume for simplicity that each
+                sample only contains one optimal set of inputs.
+            optimal_outputs: A `num_samples x 1`-dim Tensor containing the optimal
+                set of objectives of dimension `1`.
+            maximize: If true, we consider a maximization problem.
+            hypercell_bounds:  A `num_samples x 2 x J x 1`-dim Tensor containing the
```
sdaulton, line 58: Hmm these calling these hypercell bounds is odd, since they are not hyperrectangles, but rather lines... I guess it makes more sense in the constrained case

botorch/acquisition/joint_entropy_search.py
```diff
@@ -0,0 +1,104 @@
...
+from botorch.acquisition.multi_objective.joint_entropy_search import (
+    qLowerBoundMultiObjectiveJointEntropySearch,
+)
+from botorch.acquisition.multi_objective.utils import compute_sample_box_decomposition
+from botorch.models.model import Model
+from botorch.utils.transforms import concatenate_pending_points, t_batch_mode_transform
+from torch import Tensor
+
+
+class qLowerBoundJointEntropySearch(qLowerBoundMultiObjectiveJointEntropySearch):
```
sdaulton, line 26: It's seems kind of odd to have the single objective version inherit from the multi-objective version (just an observation, not a criticism)

botorch/acquisition/multi_objective/joint_entropy_search.py
```diff
@@ -0,0 +1,746 @@
...
+            self.hypercell_bounds = hypercell_bounds
+            self.num_pareto_samples = hypercell_bounds.shape[0]
+
+        # Condition the model on the sampled pareto optimal points.
+        # TODO: Apparently, we need to make a call to the posterior otherwise
+        #  we run into a gpytorch runtime error:
+        #  "Fantasy observations can only be added after making predictions with a
+        #  model so that all test independent caches exist."
+
+        if condition_on_samples:
```
sdaulton, line 123: This is specific for ` qLowerBoundMultiObjectiveJointEntropySearch`? If so, let's move it into `qLowerBoundMultiObjectiveJointEntropySearch`

botorch/acquisition/joint_entropy_search.py
benmltu: I agree, I wasn't too keen on this but it was a bit involved doing it the other way round, especially for PES

benmltu:
I've added a small update now which removes the pymoo optimizer and replaces it with a basic random search optimizer.

eytan:
> I've added a small update now which removes the pymoo optimizer and replaces it with a basic random search optimizer.

How does this affect the performance of the algorithm (both in terms of anytime optimization performance and runtime?)

benmltu:
> > I've added a small update now which removes the pymoo optimizer and replaces it with a basic random search optimizer.
> 
> How does this affect the performance of the algorithm (both in terms of anytime optimization performance and runtime?)

The performance of these entropy search methods are definitely sensitive to how the samples are optimized---“better” samples lead to better results. From some tests I ran this morning, I see a slight drop in performance from using the random search but a definite improvement in run time. The cost of pymoo on my machine is around an additional 2secs per iteration.
jes_lb_perf
mes_lb_perf
walltimes

botorch/acquisition/multi_objective/predictive_entropy_search.py
```diff
@@ -0,0 +1,1179 @@
...
+
+    Returns:
+        A `batch_shape x param_shape`-dim Tensor containing the updated natural
+        factor.
+    """
+    bs = damping_factor.shape
+    fs = old_factor.shape
+
+    df = damping_factor
+    for i in range(len(fs[len(bs) :])):
```
esantorella, line 993:
```suggestion
    for _ in range(len(fs[len(bs) :])):
```
Addressing a complaint from an internal linter

botorch/acquisition/multi_objective/utils.py
```diff
@@ -161,3 +174,224 @@ def prune_inferior_points_multi_objective(
...
+    model: Model,
+    bounds: Tensor,
+    num_samples: int,
+    num_points: int,
+    optimizer: Callable[
+        [GenericDeterministicModel, Tensor, int, bool, Any], Tuple[Tensor, Tensor]
+    ] = random_search_optimizer,
+    num_rff_features: int = 512,
+    maximize: bool = True,
+    optimizer_kwargs: dict = None,
```
esantorella, line 340:
```suggestion
    optimizer_kwargs: Optional[Dict[str, Any]] = None,
```
addressing a complaint from a Meta internal linter

botorch/acquisition/multi_objective/joint_entropy_search.py
```diff
@@ -0,0 +1,739 @@
...
+            * self.model.num_outputs
+            * (1 + torch.log(2 * pi * torch.ones(1, **tkwargs)))
+        )
+        # The variance initially has shape `batch_shape x (q*M) x (q*M)`
+        # prior_entropy has shape `batch_shape`.
+        initial_entropy = add_term + 0.5 * torch.logdet(
+            initial_posterior_plus_noise.mvn.covariance_matrix
+        )
+
+        posterior_statistics["initial_entropy"] = initial_entropy
```
esantorella, line 373:
```suggestion
        # Compute the prior entropy term depending on `X`.
        initial_posterior_plus_noise = self.initial_model.posterior(
            X, observation_noise=True
        )

        # Additional constant term.
        add_term = (
            0.5
            * self.model.num_outputs
            * (1 + torch.log(2 * pi * torch.ones(1, **tkwargs)))
        )
        # The variance initially has shape `batch_shape x (q*M) x (q*M)`
        # prior_entropy has shape `batch_shape`.
        initial_entropy = add_term + 0.5 * torch.logdet(
            initial_posterior_plus_noise.mvn.covariance_matrix
        )

        posterior_statistics = {"initial_entropy": initial_entropy}
```
A Meta-internal linter complains about the `dict()` call, and also let's define `posterior_statistics` closer to where it's used

botorch/acquisition/multi_objective/max_value_entropy_search.py
```diff
@@ -175,3 +182,196 @@ def forward(self, X: Tensor) -> Tensor:
...
+            - "observation_noise": A `batch_shape x num_pareto_samples x q x 1 x M`
+                -dim Tensor containing the observation noise at the input `X`.
+            - "posterior_with_noise": The posterior distribution at `X` which
+                includes the observation noise. This is used to compute the marginal
+                log-probabilities with respect to `p(y| x, D_n)` for `x` in `X`.
+        """
+        tkwargs = {"dtype": X.dtype, "device": X.device}
+        CLAMP_LB = torch.finfo(tkwargs["dtype"]).eps
+
+        posterior_statistics = dict()
```
esantorella, line 263:
```suggestion
```
see below

botorch/acquisition/multi_objective/max_value_entropy_search.py
```diff
@@ -175,3 +182,196 @@ def forward(self, X: Tensor) -> Tensor:
...
+            0.5
+            * self.model.num_outputs
+            * (1 + torch.log(2 * pi * torch.ones(1, **tkwargs)))
+        )
+        # The variance initially has shape `batch_shape x (q*M) x (q*M)`
+        # prior_entropy has shape `batch_shape x num_fantasies`
+        initial_entropy = add_term + 0.5 * torch.logdet(
+            posterior_plus_noise.mvn.covariance_matrix
+        )
+        posterior_statistics["initial_entropy"] = initial_entropy
```
esantorella, line 284:
```suggestion
        posterior_statistics = {"initial_entropy": initial_entropy}
```
A Meta-internal linter complains about the `dict()` call, and also let's define `posterior_statistics` closer to where it's used

botorch/acquisition/multi_objective/utils.py
```diff
@@ -161,3 +174,224 @@ def prune_inferior_points_multi_objective(
...
+    tkwargs = {"dtype": bounds.dtype, "device": bounds.device}
+    M = model.num_outputs
+    d = bounds.shape[-1]
+    if M == 1:
+        if num_points > 1:
+            raise UnsupportedError(
+                "For single-objective optimization `num_points` should be 1."
+            )
+    if optimizer_kwargs is None:
+        optimizer_kwargs = dict()
```
esantorella, line 380:
```suggestion
        optimizer_kwargs = {}
```
A Meta-internal linter prefers this (C408 from flake8-comprehensions, if anyone is curious https://pypi.org/project/flake8-comprehensions/)

Balandat:
> The performance of these entropy search methods are definitely sensitive to how the samples are optimized---“better” samples lead to better results. From some tests I ran this morning, I see a slight drop in performance from using the random search but a definite improvement in run time. The cost of pymoo on my machine is around an additional 2secs per iteration.

Thanks a lot for adding these benchmarks! I think it's fine to go ahead with the random search solution for now (but add a TODO in code to improve this) so we can get this merged in soon. Then we can figure out the pymoo dependency (or other approaches) in a follow-up. 

There is also this tutorial error in the CI: 

> RuntimeError: The following tutorial files are present, but are not linked on the website: information_theoretic_acquisition_functions.ipynb.

You can address this by adding the tutorial to this file: https://github.com/pytorch/botorch/blob/main/website/tutorials.json

Balandat:
Looks like the recent large scale refactors on our end caused some merge conflicts - once those are resolve this should be able to go in!

sdaulton:
I ran the tutorial and noticed the batch AF values are often negative (printed below with no modification to the tutorial). What causes this? Is this because the approximation of the batch information gain is a loose lower bound?

Screen Shot 2022-11-22 at 4 09 21 PM

benmltu:
> I ran the tutorial and noticed the batch AF values are often negative (printed below with no modification to the tutorial). What causes this? Is this because the approximation of the batch information gain is a loose lower bound?
> 
> The 4 AF values printed below for MES and JES are the joint AF value of candidates 1,..., j for j=1,..., q=4 (obtained via sequential greedy optimization)
> 
> Screen Shot 2022-11-22 at 4 09 21 PM

That's right, the upper bound H[p(A_1,...,A_q)] <= H[p(A_1)] + ... + H[p(A_q)] can be loose because it essentially ignores the correlation between the A_i's. As q increases, the upper bound becomes larger so the overall lower bound acq can become negative.

sdaulton:
Interesting. It's kind of funky 1) have the approximate joint utility of the optimal q-batch of points be worse (and monotonically decreasing it appears) as q increases (as new points are sequentially added to the batch), 2) have the optimal approximate information gain be negative.

Neither of these would be the case the prior joint entropy used the same approximation. Is that something you considered?

sdaulton (approved):
Thanks again for adding these! @esantorella, let's re-import and make sure there are no remaining issues

benmltu:
> 
> Neither of these would be the case the prior joint entropy used the same approximation. Is that something you considered?
> 
Yeah this was something that we considered early on, the acquisition function reduces down to something like: `qJES(X) = JES(x_1) + JES(x_2) + ... + JES(x_q)`. If you do batch optimization (not sequential) on this, then the solution is typically concentrated at one point i.e. X*=(x*,...,x*). To improve on this, we tried sequential optimization where we also conditioned on the pending points. That worked but it can be quite expensive. It turned out to be cheaper just to keep the initial joint entropy term in order to encourage the diversity in the batch.

esantorella:
This has been landed! Thanks so much @benmltu !

Balandat:
This is amazing! Thank you so much for your excellent contribution!

Issue #1459, opened 2022-10-20
[Feature Request] Fantasize SingleTaskVariationalGP and/or HeteroskedasticSingleTaskGP


JordanJar8:
# 🚀 Feature Request

An implementation of fantasize method for `SingleTaskVariationalGP` and/or `HeteroskedasticSingleTaskGP`

## Motivation

**Is your feature request related to a problem? Please describe.**
I'm trying to do some **active learning** with a `SingleTaskVariationalGP` and `HeteroskedasticSingleTaskGP` on a heteroskedastic regression task (~20 000 samples expected at the end) using `qNegIntegratedPosteriorVariance` as acquisition function. It requires to fantasize the surrogate model to compute fast posterior covariances but this method isn't implemented yet.

## Pitch

**Describe the solution you'd like**
An implementation of fantasize method that return fantasy model for `SingleTaskVariationalGP` and `HeteroskedasticSingleTaskGP`

**Describe alternatives you've considered**
I've tried to implement myself fantasize method for both models but didn't achieve yet. A little help in how BoTorch fantasize works would be very much appreciated :)

**Are you willing to open a pull request?** (See [CONTRIBUTING](../CONTRIBUTING.md))
Yes, why not. I've already tried to modify botorch lib on my PC.

## Additional context

None

saitcakmak:
Hi @JordanJar8. I am not really familiar with the details, but there's some support in GPyTorch for fantasizing variational GPs. This was implemented in https://github.com/cornellius-gp/gpytorch/pull/1874 and a tutorial is found in https://docs.gpytorch.ai/en/stable/examples/08_Advanced_Usage/SVGP_Model_Updating.html#SVGP-Model-Updating

If you need 20k samples, it is unlikely that `HeteroskedasticSingleTaskGP` or some other `ExactGP` would scale up that much (I got `SingleTaskGP` to work with 10-12k samples for some testing, but it gets too close to OOM), so variational GPs might be a better choice. To fantasize with `HeteroskedasticSingleTaskGP`, I believe you also need to fantasize on the noise model, which makes it a bit more challenging. @Balandat or @Ryan-Rhys might have more thoughts on this.

I also want to point out is that `FixedNoiseGP` can also handle heteroscedastic noise observations and supports fantasizing out of the box. It might have a bit more difficulty extrapolating far from where the observations are, but if you have a large observation budget, that shouldn't be too much of a problem, I think.

JordanJar8:
[...]

PR #1462, opened 2022-10-26
Pull out "fantasize" function so it isn't so widely inherited -- for discussion!

esantorella:
## Motivation

The `Model` base class has a `fantasize` method, but it isn't actually possible to `fantasize` from all models. For some models, `fantasize` fails with a `NotImplementedError` even though the docstring implies it should work. This is confusing (e.g. #1459 ). Another disadvantage is that codecov doesn't require tests for all of the many `fantasize` methods that exist, only for the base one -- which can't be called, since it is abstract.

This PR removes the "fantasize" method from the `Model` base class. Instead, there is a `fantasize` function that is called by the few classes that do actually use and/or test `fantasize`.

Pros:
- decreases the "surface area" of BoTorch and prevents misuse of methods that we didn't really intend to exist
- Allows codecov to surface where we aren't actually testing methods (I expect it to fail)

Cons:
-  `fantasize` still exists but doesn't work in `HeteroskedasticSingleTaskGP` because it inherits from `SingleTaskGP` (a further refactor can fix that)
- Could remove `fantasize` from classes where it _should_ exist (despite not being used or tested within BoTorch)
- Somewhat more verbose

Yes

## Test Plan
[x] Codecov
[x] Unit tests should pass
[x] No new Pyre errors introduced

### DIFF
--- a/botorch/models/approximate_gp.py
+++ b/botorch/models/approximate_gp.py
@@ -37,7 +37,6 @@ from botorch.models.transforms.input import InputTransform
 from botorch.models.transforms.outcome import OutcomeTransform
 from botorch.models.utils import validate_input_scaling
 from botorch.posteriors.gpytorch import GPyTorchPosterior
-from botorch.sampling import MCSampler
 from gpytorch.constraints import GreaterThan
 from gpytorch.distributions import MultivariateNormal
 from gpytorch.kernels import Kernel, MaternKernel, ScaleKernel
@@ -142,11 +141,6 @@ class ApproximateGPyTorchModel(GPyTorchModel):
         if self.training:
             X = self.transform_inputs(X)
         return self.model(X)
-
-    def fantasize(self, X, sampler=MCSampler, observation_noise=True, *args, **kwargs):
-        raise NotImplementedError(
-            "Fantasization of approximate GPs has not been implemented yet."
-        )
 
 
 class _SingleTaskVariationalGP(ApproximateGP):
--- a/botorch/models/fully_bayesian.py
+++ b/botorch/models/fully_bayesian.py
@@ -33,17 +33,16 @@ References:
 
 import math
 from abc import abstractmethod
-from typing import Any, Dict, List, Mapping, Optional, Tuple, Union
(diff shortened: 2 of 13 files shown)


saitcakmak:
I personally like the old pattern a bit better. Having to redefine a fantasize method just to call a common function feels like extra work to me. As mentioned in the internal thread, using a Mixin could make this a bit cleaner.

saitcakmak:
> fantasize still exists but doesn't work in HeteroskedasticSingleTaskGP because it inherits from SingleTaskGP (a further refactor can fix that)

This would still be an issue with the Mixin approach, unless we define a `SingleTaskGPBase` and add the mixin only to `SingleTaskGP` and `FixedNoiseGP`. I think that's too much work for how little it achieves though.

Balandat:
> Another disadvantage is that codecov doesn't require tests for all of the many fantasize methods that exist, only for the base one -- which can't be called, since it is abstract.

Yeah that's an interesting observation. We definitely want to make sure that all methods defined on models are actually tested. I'm not sure though that we need to define the function explicitly on each subclass though - a different solution to this could be to refactor the unit tests for the models into a more generic test harness that enforces testing of all methods defined at the level of the associated model class. What do you think about that approach? 

cc @lena-kashtelyan why has been thinking about doing the same for Ax.

esantorella:
> Yeah that's an interesting observation. We definitely want to make sure that all methods defined on models are actually tested. I'm not sure though that we need to define the function explicitly on each subclass though - a different solution to this could be to refactor the unit tests for the models into a more generic test harness that enforces testing of all methods defined at the level of the associated model class. What do you think about that approach?
> 
> cc @lena-kashtelyan why has been thinking about doing the same for Ax.

Sounds awesome if there's a good way to do it!

Another option, now that this PR is using a `FantasizeMixin`, would be to test that the `fantasize` method works for each non-abstract elementsof `FantasizeMixin.__subclasses__`. Unfortunately, that isn't an easy test to write because each subclass needs to be instantiated in a different way.

PR #1463, opened 2022-10-26
sampling.pathwise

j-wilson:
This PR contains code for efficiently sampling functions from (approximate) GP priors and posteriors. The functionality introduced here is largely exposed by three high-level methods:

- `gen_kernel_features`: Generates a feature map that represents (or approximates) a kernel.

- `draw_kernel_feature_paths`: Draws functions from a Bayesian-linear-model-based approximation to a GP prior. By default, uses random Fourier features (RFF) to represent stationary priors.

- `draw_matheron_paths`: Generates draws from an approximate GP posterior using Matheron's rule. By default, this method combines draws from an RFF-based approximate prior with exact Gaussian updates. For details, see [[Wilson et al., 2020]](https://arxiv.org/abs/2002.09309#) and [[Wilson et al., 2021]](https://arxiv.org/pdf/2011.04026.pdf).

Please let us know if you run into issues. As always, contributions are welcomed.

### DIFF
--- a/botorch/optim/closures/model_closures.py
+++ b/botorch/optim/closures/model_closures.py
@@ -12,8 +12,8 @@ from itertools import chain, repeat
 from typing import Any, Callable, Dict, Optional, Sequence, Tuple
 
 from botorch.optim.closures.core import ForwardBackwardClosure
-from botorch.optim.utils import TNone
 from botorch.utils.dispatcher import Dispatcher, type_bypassing_encoder
+from botorch.utils.types import NoneType
 from gpytorch.mlls import (
     ExactMarginalLogLikelihood,
     MarginalLogLikelihood,
@@ -151,9 +151,9 @@ def _get_loss_closure_fallback_external(
     return closure
 
 
-@GetLossClosure.register(MarginalLogLikelihood, object, object, TNone)
+@GetLossClosure.register(MarginalLogLikelihood, object, object, NoneType)
 def _get_loss_closure_fallback_internal(
-    mll: MarginalLogLikelihood, _: object, __: object, ___: TNone, **ignore: Any
+    mll: MarginalLogLikelihood, _: object, __: object, ___: NoneType, **ignore: Any
 ) -> Callable[[], Tensor]:
     r"""Fallback loss closure with internally managed data."""
 
@@ -165,9 +165,9 @@ def _get_loss_closure_fallback_internal(
     return closure
 
 
-@GetLossClosure.register(ExactMarginalLogLikelihood, object, object, TNone)
+@GetLossClosure.register(ExactMarginalLogLikelihood, object, object, NoneType)
 def _get_loss_closure_exact_internal(
-    mll: ExactMarginalLogLikelihood, _: object, __: object, ___: TNone, **ignore: Any
+    mll: ExactMarginalLogLikelihood, _: object, __: object, ___: NoneType, **ignore: Any
 ) -> Callable[[], Tensor]:
     r"""ExactMarginalLogLikelihood loss closure with internally managed data."""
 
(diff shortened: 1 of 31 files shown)


hvarfner:
Thanks for putting this up - it seems to work like a charm. One question regarding the sampling: It raises a `NotImplementedError` for `FixedNoiseGaussianLikelihood` - should the sampler not be used in conjunction with `FixedNoiseGP`?

It seems to work fine (as in that it runs if I remove the `isinstance` check) but I haven't checked the correctness of the samples.

j-wilson:
@hvarfner Good catch.  It should work for FixedNoiseGaussianLikelihoods/FixedNoiseGPs. I simply forgot that `_GaussianLikelihoodBase` is the base class, not `GaussianLikelihood`. I'll patch the instance check.

Other than that, there's a trivial dispatch in `prior_samples/bayes_linear.py` for SingleTaskGPs that simply passes `model.num_outputs` to a generic method intended for ExactGPs. Long story short: I've moved the `model.num_outputs` logic into the generic method and removed the SingleTaskGP-specific dispatch. This should ensure that `num_outputs` gets set correctly for multioutput FixedNoiseGPs.

j-wilson:
@hvarfner Should be fixed; just needed to specify `device` and `dtype` when allocating weights in `prior_samplers/bayes_linear.py`.

Edge case: May fail for non-float64 ApproximateGPs because GPyTorch's internal [Cholesky factor cache](https://github.com/cornellius-gp/gpytorch/blob/3fb89cfbcf2bc98c89f8a5ea8c8a1dbdf41e277a/gpytorch/variational/variational_strategy.py#L88) is not guaranteed to have the same dtype as the model.

botorch/sampling/pathwise/basis/expansions.py
```diff
@@ -0,0 +1,92 @@
...
+        self._initialized = Parameter(
+            torch.zeros([], dtype=torch.bool), requires_grad=False
+        )
+
+    def forward(self, x: Tensor) -> Tensor:
+        if not self._initialized:
+            self.initialize(input_shape=x.shape[-1:])
+
+        x = x if self.input_transform is None else self.input_transform(x)
+        z = x @ self.weight.transpose(-2, -1)
```
hvarfner, line 50: I'm not 100% certain, but it seems like there's a
`x = x.unsqueeze(1)`
needed here. In the MultiOutput / FullyBayesian case, we get:

`self.weight: num_outputs * 1 * num_dim * num_features`
`x: num_outputs *  num_queries * num_features`
'z: x @ self.weight.transpose(-2, -1) --> 
num_outputs * **num_outputs** * num_queries * num_features

With the proposed change, we get:
'z: x @ self.weight.transpose(-2, -1) --> 
num_outputs * **1** * num_queries * num_features

In the single output case, the functionality is unchanged.

j-wilson: Hey! Thanks for pointing this out. I think there's a good chance you're right that there's something funny going here. Are you using the latest version of this PR (there was a bug in a previous version where `num_outputs` was not being passed)? If so I would sooner expect 
`self.weight: batch_shape x num_outputs x num_dim x num_features`,
which in many cases would resolve to
`self.weight: num_outputs x num_outputs x num_dim x num_features`.

Basically, my code (erroneously) views `model.batch_shape` and multi-output dimensions as separate things whereas BoTorch absorbs the multi-output dimension as part of the batch shape. Time allowing, I'll look into this and hopefully get back to you with a fix. If you happen to have a simple example that reproduces the issue that would be helpful.

j-wilson: @hvarfner This issue should be resolved in the latest version of the PR. Note that `model.covar_module.batch_shape` now dictates the batch shape. Among other things, this means that we no longer maintain an explicit output dimension in the non-batch setting.

dylanmkennedy:
Great stuff! This is very useful for applications that my group and I at SLAC National Accelerator Laboratory are working on. Are there plans to release a version with these pathwise sampling methods included anytime soon? Thanks!

j-wilson:
> Great stuff! This is very useful for applications that my group and I at SLAC National Accelerator Laboratory are working on. Are there plans to release a version with these pathwise sampling methods included anytime soon? Thanks!

Hi @dylanmkennedy. Sorry for the delay. Yes, I plan on finalizing an initial release the pathwise sampling code in the near future (hopefully later this week).

PR #1478, opened 2022-11-05
Do not count hitting maxiter as optimization failure & update default maxiter

saitcakmak:
Summary:
We currently retry optimization upon hitting maxiter, which typically leads to optimizing the same thing twice. This diff updates the warning handling to avoid this behavior.

The default maxiter is also updated to 2_000, which is a steep change from scipy defaults of 15_000 for L-BFGS-B and 100 for SLSQP. This is just a temporary setting to bridge the gap between the two until we come up with better stopping conditions for the optimizers.

Differential Revision: D41053812

### DIFF
--- a/test/generation/test_gen.py
+++ b/test/generation/test_gen.py
@@ -69,7 +69,7 @@ class TestBaseCandidateGeneration(BotorchTestCase):
 class TestGenCandidates(TestBaseCandidateGeneration):
     def test_gen_candidates(self, gen_candidates=gen_candidates_scipy, options=None):
         options = options or {}
-        options = {**options, "maxiter": 5}
+        options = {**options, "maxiter": options.get("maxiter", 5)}
         for double in (True, False):
             self._setUp(double=double)
             acqfs = [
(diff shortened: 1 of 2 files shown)


test/generation/test_gen.py
```diff
@@ -69,7 +69,7 @@ def _setUp(self, double=False, expand=False):
 class TestGenCandidates(TestBaseCandidateGeneration):
     def test_gen_candidates(self, gen_candidates=gen_candidates_scipy, options=None):
         options = options or {}
-        options = {**options, "maxiter": 5}
+        options["maxiter"] = options.get("maxiter", 5)
```
Balandat, line 72: This is problematic since if options is passed as a dict, this method will now modify the `options` dict in-place. We generally should avoid these kinds of side effects to function arguments (unless we specifically want this to happen).

test/generation/test_gen.py
```diff
@@ -227,6 +213,15 @@ def test_gen_candidates_scipy_warns_opt_failure(self):
         )
         self.assertTrue(expected_warning_raised)
 
+    def test_gen_candidates_scipy_no_maxiter_warning(self):
+        # Check that no warnings are raised on hitting maxiter.
```
Balandat, line 217: We should probably log this? Could be at the info or maybe even the warning logging level.

PR #1482, opened 2022-11-08
Use latest gpytorch & linear operator in tutorials workflow

saitcakmak:
Title. Current implementation was setting the environment variable on the wrong step


Balandat:
ugh my bad, thanks for the fix!

PR #1492, opened 2022-11-10
Fix test_stable GHA workflow

Balandat:
We do some grepping magic to extract the versions from `botorch.egg-info/requires.txt` after running `python setup.py egg_info`. This was not properly working when pinning a version to an exact match. This fixes that by allowing both `>=` and `==` version specification in the grep rexexp.


saitcakmak:
Looks like the tutorials workflow also needs to be updated

Balandat:
Yup, already done, new test run: https://github.com/pytorch/botorch/actions/runs/3439212041

PR #1496, opened 2022-11-11
Expanding on docstrings in BoTorch `Model`

esantorella:
Summary: Added content to docstrings in BoTorch `Model` to help people who may either be unfamiliar with torch `Module` or with what its methods (such as "forward") mean in the context of BoTorch models.

Differential Revision: D41221005

### DIFF
--- a/botorch/models/model.py
+++ b/botorch/models/model.py
@@ -46,8 +46,21 @@ TFantasizeMixin = TypeVar("TFantasizeMixin", bound="FantasizeMixin")
 class Model(Module, ABC):
     r"""Abstract base class for BoTorch models.
 
-    Model cannot be used directly; it only defines an API for other BoTorch
+    `Model` cannot be used directly; it only defines an API for other BoTorch
     models.
+
+    `Model` subclasses `torch.nn.Module`. While a `Module` is most typically
+    encountered as a representation of a neural network layer, it can be used more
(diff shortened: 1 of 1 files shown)


botorch/models/model.py
```diff
@@ -46,9 +46,22 @@
...
+    `Model` subclasses `torch.nn.Module`. While a `Module` is most typically
+    encountered as a representation of a neural network layer, it can be used more
+    generally: see `documentation<NN>`_ on custom NN Modules.
+
+    Subclassing `Module` provides several pieces of  useful functionality: Attributes
+    of `Tensor` or `Module` type are automatically registered so they can be moved
+    and/or cast with `to`, automatically differentiated, and used with CUDA. Also,
+    `Model` subclasses are callable, provided that they implement a `forward` method. In
+    BoTorch, this typically computes the prior latent distribution at a point, e.g.
+    `multivariate_normal = SingleTaskGP(x)`.
```
Balandat, line 61: This may be a bit confusing - for GPyTorch models `__call__` will compute the prior latent distribution if the module is in `train` mode, but the posterior latent distribution if the model is in `eval` mode. 

In general (at least for vanilla exact GPs, things are a bit different when doing more advanced things like mini-batch training of SVGPs) the end user typically does not need to interface with the `__call__` method in botorch - we introduced the `posterior()` API to be unambiguous in that regard.

PR #1500, opened 2022-11-12
Remove deprecated AcquisitionObjective

Balandat:
This has been marked deprecated for a while, time to do some cleanup.

Looks like `AnalyticMultiOutputObjective` still uses this though, so we'll need to remove `AnalyticMultiOutputObjective` and  update things to use posterior transforms instead.


Balandat:
@sdaulton looks like `AnalyticMultiOutputObjective` was never actually properly deprecated - I haven't looked into the details here, any concerns from your end to update this to use posterior transforms?

Balandat:
This has been done as part of #1865

PR #1504, opened 2022-11-15
Make behavior of Cholesky caching more clear

esantorella:
## Motivation

General context: Caching is confusing and can lead to subtle issues. I have been trying to understand it better in order to reduce memory usage and improve runtime, since I've been seeing cache misses and tensors persisting longer than necessary. This PR doesn't fix that, but does make things a tiny bit more transparent.

Two things are making the "_cache_root_decomposition" method harder to understand than necessary:

1. It sets `self._baseline_L` and returns `None` rather than just returning `baseline_L`, so when someone sees a call to `_cache_root_decomposition` they will not immediately realize `self._baseline_L` has been set.
2. It is uses two different kinds of caching: it sets `self._baseline_L` and  it also invisibly uses LinearOperator's caching. 

This PR makes things more transparent by
* Adding comments
* Returning `baseline_L` rather than setting it as a side effect

Yes

## Test Plan

Unit tests

### DIFF
--- a/botorch/acquisition/cached_cholesky.py
+++ b/botorch/acquisition/cached_cholesky.py
@@ -103,7 +103,7 @@ class CachedCholeskyMCAcquisitionFunction(ABC):
     ) -> Tensor:
         r"""Cache Cholesky of the posterior covariance over f(X_baseline).
 
-        Because `LazyTensor.root_decomposition` is decorated with LinearOperator's
+        Because `LinearOperator.root_decomposition` is decorated with LinearOperator's
         @cached decorator, this function is doing a lot implicitly:
 
         1) Check if a root decomposition has already been cached to `lazy_covar`.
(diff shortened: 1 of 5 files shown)


botorch/acquisition/cached_cholesky.py
```diff
@@ -97,12 +97,24 @@ def _setup(
...
 
-    def _cache_root_decomposition(
+    def _compute_root_decomposition(
         self,
         posterior: Posterior,
-    ) -> None:
+    ) -> Tensor:
         r"""Cache Cholesky of the posterior covariance over f(X_baseline).
 
+        Because `LazyTensor.root_decomposition` is decorated with LinearOperator's
```
Balandat, line 106:
```suggestion
        Because `LinearOperator.root_decomposition` is decorated with LinearOperator's
```

Balandat (approved):
Thanks, this makes things a lot cleaner!

PR #1506, opened 2022-11-17
Update tutorials to work with Posterior & MCSampler refactor

saitcakmak:
Summary: See title. Also deletes the empty files leftover by the refactor.

Differential Revision: D41366067


esantorella:
FYI this is a different tutorials failure than we were seeing yesterday.

saitcakmak:
@esantorella I think I deleted too much of the notebook metadata. This should be easy to fix.

PR #1516, opened 2022-11-22
add support for categoricals in Round input transform and use STEs

sdaulton:
Summary: see title

Differential Revision: D41477456


sdaulton:
Note: this changes the default behavior of `Round` to have `approximate=False`.

PR #1526, opened 2022-11-28
make eta configurable

jduerholt:
## Motivation

I recently looked into the output constraint implementation in botorch and figured out that it behave like our custom `Objective` implementation for handling of constraints, namely by multiplying by sigmoids. Currently, the only difference is that we work often with different `eta` values per constraint. I think this would be a nice feature also for `botorch`. 

This PR is still work in progress, as the `apply_constraints` method is used at a lot of different occasions though-out the codebase, and my question is if one want to keep backwards compatibility. In the current PR, I kept the backwards compatibility and made `eta` of type `Union[float, torch.Tensor]`. If one does this one has to always catch if just a float is provided and transform the float in a tensor of the same length as the list of constraint callables.

Another option would be to set `eta` as optional with default `None` and then just generate a tensor with the old default of 10e-3. 

Which solution would you prefer?

Depending on your suggestion, I would finalize the PR and implement the functionality of different `eta`s per constraint though-out the whole codebase. 

Yes.

## Test Plan

Unit tests.

### DIFF
--- a/botorch/acquisition/multi_objective/monte_carlo.py
+++ b/botorch/acquisition/multi_objective/monte_carlo.py
@@ -103,6 +103,12 @@ class MultiObjectiveMCAcquisitionFunction(AcquisitionFunction, MCSamplerMixin, A
                 `sample_shape x batch-shape x q x m` to a Tensor of dimension
                 `sample_shape x batch-shape x q`, where negative values imply
                 feasibility.
+            eta: The temperature parameter for the sigmoid function used for the
+                differentiable approximation of the constraints. In case of a float the
+                same eta is used for every constraint in constraints. In case of a
+                tensor the length of the tensor must match the number of provided
+                constraints. The i-th constraint is then estimated with the i-th
+                eta value.
             X_pending:  A `m x d`-dim Tensor of `m` design points that have
                 points that have been submitted for function evaluation
                 but have not yet been evaluated.
--- a/botorch/acquisition/multi_objective/monte_carlo.py
+++ b/botorch/acquisition/multi_objective/monte_carlo.py
@@ -85,6 +85,7 @@ class MultiObjectiveMCAcquisitionFunction(AcquisitionFunction, MCSamplerMixin, A
         sampler: Optional[MCSampler] = None,
         objective: Optional[MCMultiOutputObjective] = None,
         constraints: Optional[List[Callable[[Tensor], Tensor]]] = None,
+        eta: Optional[Union[Tensor, float]] = 1e-3,
         X_pending: Optional[Tensor] = None,
     ) -> None:
         r"""Constructor for the MCAcquisitionFunction base class.
@@ -128,6 +129,10 @@ class MultiObjectiveMCAcquisitionFunction(AcquisitionFunction, MCSamplerMixin, A
             )
         self.add_module("objective", objective)
         self.constraints = constraints
+        if constraints:
+            if type(eta) != Tensor:
+                eta = torch.full((len(constraints),), eta)
+            self.register_buffer("eta", eta)
         self.X_pending = None
         if X_pending is not None:
             self.set_X_pending(X_pending)
@@ -153,7 +158,7 @@ class qExpectedHypervolumeImprovement(MultiObjectiveMCAcquisitionFunction):
         objective: Optional[MCMultiOutputObjective] = None,
         constraints: Optional[List[Callable[[Tensor], Tensor]]] = None,
         X_pending: Optional[Tensor] = None,
-        eta: float = 1e-3,
+        eta: Optional[Union[Tensor, float]] = 1e-3,
     ) -> None:
         r"""q-Expected Hypervolume Improvement supporting m>=2 outcomes.
 
@@ -189,7 +194,11 @@ class qExpectedHypervolumeImprovement(MultiObjectiveMCAcquisitionFunction):
                 been evaluated. Concatenated into `X` upon forward call. Copied and set
                 to have no gradient.
             eta: The temperature parameter for the sigmoid function used for the
-                differentiable approximation of the constraints.
+                differentiable approximation of the constraints. In case of a float the
+                same eta is used for every constraint in constraints. In case of a
+                tensor the length of the tensor must match the number of provided
+                constraints. The i-th constraint is then estimated with the i-th
+                eta value.
         """
         if len(ref_point) != partitioning.num_outcomes:
             raise ValueError(
@@ -207,9 +216,9 @@ class qExpectedHypervolumeImprovement(MultiObjectiveMCAcquisitionFunction):
             sampler=sampler,
             objective=objective,
             constraints=constraints,
+            eta=eta,
             X_pending=X_pending,
         )
-        self.eta = eta
         self.register_buffer("ref_point", ref_point)
         cell_bounds = partitioning.get_hypercell_bounds()
         self.register_buffer("cell_lower_bounds", cell_bounds[0])
@@ -357,7 +366,7 @@ class qNoisyExpectedHypervolumeImprovement(
         objective: Optional[MCMultiOutputObjective] = None,
         constraints: Optional[List[Callable[[Tensor], Tensor]]] = None,
         X_pending: Optional[Tensor] = None,
-        eta: float = 1e-3,
+        eta: Optional[Union[Tensor, float]] = 1e-3,
         prune_baseline: bool = False,
         alpha: float = 0.0,
         cache_pending: bool = True,
@@ -400,7 +409,11 @@ class qNoisyExpectedHypervolumeImprovement(
                 have points that have been submitted for function evaluation, but
                 have not yet been evaluated.
             eta: The temperature parameter for the sigmoid function used for the
-                differentiable approximation of the constraints.
+                differentiable approximation of the constraints. In case of a float the
+                same eta is used for every constraint in constraints. In case of a
+                tensor the length of the tensor must match the number of provided
+                constraints. The i-th constraint is then estimated with the i-th
+                eta value.
             prune_baseline: If True, remove points in `X_baseline` that are
                 highly unlikely to be the pareto optimal and better than the
                 reference point. This can significantly improve computation time and
@@ -431,6 +444,7 @@ class qNoisyExpectedHypervolumeImprovement(
             sampler=sampler,
             objective=objective,
             constraints=constraints,
+            eta=eta,
         )
         self._setup(model=model, cache_root=cache_root)
 
@@ -450,7 +464,6 @@ class qNoisyExpectedHypervolumeImprovement(
             )
         self.register_buffer("ref_point", ref_point)
         self.alpha = alpha
-        self.eta = eta
         self.q_in = -1
         self.q_out = -1
         self.q_subset_indices = BufferDict()
--- a/botorch/acquisition/multi_objective/multi_fidelity.py
+++ b/botorch/acquisition/multi_objective/multi_fidelity.py
@@ -46,7 +46,7 @@ class MOMF(qExpectedHypervolumeImprovement):
         sampler: Optional[MCSampler] = None,
         objective: Optional[MCMultiOutputObjective] = None,
         constraints: Optional[List[Callable[[Tensor], Tensor]]] = None,
-        eta: Optional[Union[Tensor, float]] = 10e-3,
+        eta: Optional[Union[Tensor, float]] = 1e-3,
         X_pending: Optional[Tensor] = None,
         cost_call: Callable[Tensor, Tensor] = None,
         **kwargs: Any,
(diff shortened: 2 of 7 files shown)


botorch/acquisition/objective.py
```diff
@@ -472,6 +472,8 @@ def __init__(
         """
         super().__init__(objective=objective)
         self.constraints = constraints
+        if type(eta) != Tensor:
+            eta = [eta for _ in range(len(constraints))]
         self.register_buffer("eta", torch.as_tensor(eta))
```
saitcakmak, line 477: Nit: You can use `torch.full` to construct the tensor in one step.
```suggestion
            eta = torch.full(len(constraints), eta)
        self.register_buffer("eta", eta)
```

botorch/utils/objective.py
```diff
@@ -83,9 +83,11 @@ def apply_constraints_nonnegative_soft(
     Returns:
         A `n_samples x b x q (x m')`-dim tensor of feasibility-weighted objectives.
     """
+    if type(eta) != Tensor:
+        eta = torch.as_tensor([eta for _ in range(len(constraints))])
```
saitcakmak, line 87:
```suggestion
        eta = torch.full(len(constraints), eta)
```

botorch/utils/objective.py
```diff
@@ -118,7 +120,7 @@ def apply_constraints(
     constraints: List[Callable[[Tensor], Tensor]],
     samples: Tensor,
     infeasible_cost: float,
-    eta: float = 1e-3,
+    eta: Union[List[float], float] = 1e-3,
```
saitcakmak, line 123: Why is it `List[float]` here and `Tensor` everywhere else?

botorch/utils/objective.py
```diff
@@ -83,9 +83,11 @@ def apply_constraints_nonnegative_soft(
     Returns:
         A `n_samples x b x q (x m')`-dim tensor of feasibility-weighted objectives.
     """
+    if type(eta) != Tensor:
+        eta = torch.as_tensor([eta for _ in range(len(constraints))])
     obj = obj.clamp_min(0)  # Enforce non-negativity with constraints
-    for constraint in constraints:
-        constraint_eval = soft_eval_constraint(constraint(samples), eta=eta)
+    for constraint, e in zip(constraints, eta):
+        constraint_eval = soft_eval_constraint(constraint(samples), eta=e)
```
saitcakmak, line 90: What happens here if `eta` is a single element tensor? I think this would ignore all the constraints after the first one. We should either add a check that they're the same length and error out, or extend the if statement above to catch single element tensors.

saitcakmak (changes requested):
Thanks for putting this together. I think being backwards compatible would be preferable, so let's accept both Tensor & float. I left a few comments in line. This should be good to go if you update the docstrings (to explain what the tensor version is) and add units.

jduerholt:
Thanks for the quick review, I will go through the codebase and change it accordingly. Best, Johannes

botorch/utils/objective.py
jduerholt: I forgot to update it, will change it ;)

botorch/utils/objective.py
jduerholt: I forgot to update it, I will change it.

botorch/utils/objective.py
jduerholt: I think the best is to just error out.

jduerholt:
Another thing which I just recognized: I think the doc string for `soft_eval_constraint` is wrong. It is stated: `As eta grows larger, this approximates the Heaviside step function.` The opposite is the case: as eta decreases, it approximates the Heaviside step function, or?

```python

from botorch.utils.objective import soft_eval_constraint, apply_constraints
import matplotlib.pyplot as plt
import numpy as np
import torch

x = np.linspace(-10,10,500)
xt = torch.from_numpy(x).unsqueeze(-1)

fig, ax = plt.subplots()

y = soft_eval_constraint(xt, eta=1e-3 )
ax.plot(xt, y, label="eta = 10e-3")
y = soft_eval_constraint(xt, eta=1e-2 )
ax.plot(xt, y, label="eta = 10e-2")
y = soft_eval_constraint(xt, eta=1e-1 )
ax.plot(xt, y, label="eta = 10e-1")
y = soft_eval_constraint(xt, eta=1 )
ax.plot(xt, y, label="eta = 1")

ax.legend()
plt.show()
```
This gives the following output:

image


Or do I understand something wrong?

Balandat:
You're right, the docstring is wrong! Would you mind patching it as part of this PR?

jduerholt:
> You're right, the docstring is wrong! Would you mind patching it as part of this PR?

I just fixed it. Internally we call the inverse of the `eta` parameter `steepness`, which makes it at least for me clearer what it is doing.

botorch/utils/objective.py
jduerholt: I fixed it by catching it in `apply_constraints_nonnegative_soft`.

botorch/utils/objective.py
jduerholt: Fixed.

jduerholt:
I added the possibility for different `eta`s now to the `ConstrainedMCObjective` and into qehvi/qnehevi and the derived methods. 

In addition, I added tests, for qnehvi one test is failing, namely if one uses an eta of 0.1 or 1. Then it shows a strange behavior. For eta=1, it even results in values larger than with the unconstrained one. This should not be. Maybe one of you can have a look, as I do not understand what is going on. The test is failing here: https://github.com/jduerholt/botorch/blob/bb5a56088759302041e7c818f4efe33559906c36/test/acquisition/multi_objective/test_monte_carlo.py#L1301 

Any ideas?

botorch/acquisition/multi_objective/monte_carlo.py
```diff
@@ -85,6 +85,7 @@ def __init__(
         sampler: Optional[MCSampler] = None,
         objective: Optional[MCMultiOutputObjective] = None,
         constraints: Optional[List[Callable[[Tensor], Tensor]]] = None,
+        eta: Optional[Union[Tensor, float]] = 1e-3,
```
saitcakmak, line 88: Let's add a docstring for this as well.

botorch/utils/objective.py
```diff
@@ -78,14 +78,24 @@ def apply_constraints_nonnegative_soft(
...
+            then estimated with the i-th eta value.
 
     Returns:
         A `n_samples x b x q (x m')`-dim tensor of feasibility-weighted objectives.
     """
+    if type(eta) != Tensor:
+        eta = torch.full((len(constraints),), eta)
+    if len(eta) != len(constraints):
+        raise ValueError(
+            "Number of provided constraints and numer of provided etas does not match."
```
saitcakmak, line 94:
```suggestion
            "Number of provided constraints and number of provided etas do not match."
```

saitcakmak (commented):
I'll take a look at the test failure and follow up.

botorch/acquisition/multi_objective/multi_fidelity.py
```diff
@@ -46,9 +46,9 @@ def __init__(
         sampler: Optional[MCSampler] = None,
         objective: Optional[MCMultiOutputObjective] = None,
         constraints: Optional[List[Callable[[Tensor], Tensor]]] = None,
+        eta: Optional[Union[Tensor, float]] = 10e-3,
```
saitcakmak, line 49:
```suggestion
        eta: Optional[Union[Tensor, float]] = 1e-3,
```

test/acquisition/multi_objective/test_monte_carlo.py
```diff
@@ -527,6 +527,65 @@ def test_constrained_q_expected_hypervolume_improvement(self):
...
+            res = acqf(X)
+            self.assertAlmostEqual(
+                res.item(),
+                (
+                    torch.sigmoid(torch.as_tensor(-1.0))
+                    * torch.sigmoid(torch.as_tensor(-1.0))
+                    * 1.5
+                ).item(),
+                places=4,
+            )
```
saitcakmak, line 566: It'd be good to reduce the repetition in these test as well. Looks like only eta and the expected value changes. Maybe you can run them in a for loop?

test/acquisition/multi_objective/test_monte_carlo.py
```diff
@@ -1074,7 +1133,50 @@ def test_constrained_q_noisy_expected_hypervolume_improvement(self):
...
+                    lambda Z: torch.zeros_like(Z[..., -1]),
+                ],
+                eta=torch.tensor([1.0, 10.0]),
+                cache_root=False,
+            )
+            # set the MockPosterior to use samples over baseline points and new
+            # candidates
+            mm._posterior._samples = samples
+            res = acqf(X)
+            self.assertAlmostEqual(res.item(), 0.5 * 0.5 * 1.5, places=4)
```
saitcakmak, line 1178: looks like this could also go in the for loop.

test/acquisition/multi_objective/test_monte_carlo.py
```diff
@@ -1169,6 +1271,38 @@ def test_constrained_q_noisy_expected_hypervolume_improvement(self):
...
+                cache_root=False,
+            )
+            samples = torch.cat(
+                [
+                    baseline_samples.unsqueeze(0),
+                    torch.tensor([[[6.5, 4.5]]], **tkwargs),
+                ],
+                dim=1,
+            )
+            mm._posterior._samples = samples
```
saitcakmak, line 1298: Re test failure: It seems like you expect the infeasible acqf value to be the feasible acqf value times the sigmoid value of the constraints there. This does not work there, since the two acqfs compute the hypervolume improvement w.r.t two different partitionings. The feasible acqf has a partitioning that has a non-empty set of baseline points (you can print `acqf.partitioning.Y`) but the infeasible acqf has a partitioning that is empty. This difference happens because the constraints are taken into account while constructing the partitioning in `__init__ -> _set_cell_bounds`. 

If instead of constructing a new acqf in here, you just update the constraints and eta as below, the test will pass.
```suggestion
            acqf.constraints = [
                # lambda Z: torch.zeros_like(Z[..., -1]),
                lambda Z: torch.ones_like(Z[..., -1]),
            ]
            acqf.eta = torch.tensor([1.0])
```

botorch/acquisition/multi_objective/monte_carlo.py
jduerholt: Done.

test/acquisition/multi_objective/test_monte_carlo.py
jduerholt: Yeah, I introduced loops for this. In principle one could have it even more concise but then one would loose readability. Personally, I really like the `@pytest.mark.parametrize` decorator for these kind of tests, as they make the code very condense but still readable. But I think you are not using it in botorch.

test/acquisition/multi_objective/test_monte_carlo.py
jduerholt: Done.

test/acquisition/multi_objective/test_monte_carlo.py
jduerholt: Thank you very much for the explanation! I updated the tests accordingly.

jduerholt:
Hi @saitcakmak , thank you very much for your suggestions and especially your explanation regarding the base line partioning for qnehvi! I updated the tests accordingly, just tell me if more is needed.

saitcakmak (approved):
Thanks for making all the changes and thorough testing. LGTM!

test/acquisition/multi_objective/test_monte_carlo.py
```diff
@@ -1169,6 +1234,37 @@ def test_constrained_q_noisy_expected_hypervolume_improvement(self):
...
+            # this crashes for large etas, and I do not why
+            # set the MockPosterior to use samples over baseline points
+            etas = [torch.tensor([1.0]), torch.tensor([1.0, 10.0])]
+            constraints = [
+                [lambda Z: torch.ones_like(Z[..., -1])],
+                [
+                    lambda Z: torch.ones_like(Z[..., -1]),
+                    lambda Z: torch.ones_like(Z[..., -1]),
+                ],
+            ]
```
Balandat, line 1247: Could make this a bit cleaner, but not really necessary
```suggestion
            
            def con_func(Z):
                torch.ones_like(Z[..., -1])
            
            constraints = [[con_func], [con_func, con_func]]
```

jduerholt: Hi @Balandat , if you want I can change it accordingly, is this still possible as the PR is already imported to Phabricator?

Balandat: Yeah we can always reimport (not anymore for this one b/c it already got merged). So no worries for this one as this was really a minor nit.

PR #1535, opened 2022-12-01
Change defaults MC Samples to 512

jduerholt:
## Motivation

I just recognized that in `get_acquisition_function` a default of 500 mc samples is used, but when one initializes an MC acquisition function directly a default of 512 samples is used: https://github.com/pytorch/botorch/blob/8ecc903392b21ad9468c9b45f26b867d51378461/botorch/acquisition/acquisition.py#L149 I think it makes sense to keep things consistent. In addition, I think it is recommended for sobol sampling to use a sampling size which corresponds to basis 2. 

Yes.

## Test Plan

Only default value is changed. No tests necessary.


saitcakmak (approved):
Nice catch!

PR #1539, opened 2022-12-06
Suppress dtype warnings in unit tests

saitcakmak:
Summary: The dtype warnings were being raised a few too many times when running unit tests. This adds a filter to ignore them.

Differential Revision: D41757594


Balandat (approved):
great, thanks!

PR #1540, opened 2022-12-06
Put HOGP tutorial in ignorelist

saitcakmak:
This seems to be OOMing the nightly cron. Let's skip it until we get around to reducing its memory usage. 

Test plan:

https://github.com/pytorch/botorch/actions/runs/3631828439 should pass.


Balandat (approved):
Fair enough. We should probably try to get around to address some of these TODOs here at some point...

PR #1541, opened 2022-12-06
add eta to get_acquisition_function

jduerholt:
## Motivation

This PR adds the `eta` parameter for the constraints also to get_acquisition_function.

Yes.

## Test Plan

Unit tests.

### DIFF
--- a/botorch/acquisition/utils.py
+++ b/botorch/acquisition/utils.py
@@ -11,7 +11,7 @@ Utilities for acquisition functions.
 from __future__ import annotations
 
 import math
-from typing import Callable, Dict, List, Optional
+from typing import Callable, Dict, List, Optional, Union
 
 import torch
 from botorch.acquisition import analytic, monte_carlo, multi_objective  # noqa F401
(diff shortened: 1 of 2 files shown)


jduerholt:
I also fixed in this PR the linting issue from main.

botorch/acquisition/utils.py
```diff
@@ -61,8 +62,13 @@ def get_acquisition_function(
         constraints: A list of callables, each mapping a Tensor of dimension
             `sample_shape x batch-shape x q x m` to a Tensor of dimension
             `sample_shape x batch-shape x q`, where negative values imply
-            feasibility. Used when constraint_transforms are not passed
-            as part of the objective.
+            feasibility. Used only for qEHVI and qNEHVI.
```
Balandat, line 65: @sdaulton, @saitcakmak I guess this special handling is needed b/c we don't yet have posterior transforms implemented for MOO? Do we have any concrete plans to change this? It would be nice to make this interface more consistent.

Balandat (approved):
lgtm, thanks!

botorch/acquisition/utils.py
jduerholt: Hmm, I thought the reason is that in `qNEHVI` and `qEHVI` the constraints are defined directly in the acquisition function whereas in the other acqfs, we define the constraints as part of the `objective` and then give the objective to the constructor of the acquisition function. Or is there another reason?

saitcakmak: I think @jduerholt is correct here. IIRC, in q(N)EHVI we use constraints to weight the HVI, whereas in other acqfs, we feasibility weight the samples directly (via the objective).

PR #1546, opened 2022-12-07
Support `0d`-features in `FixedFeatureAcquisitionFunction`

SebastianAment:
Summary:
This commit adds support for `0d` value tensors in `FixedFeatureAcquisitionFunction`. In particular, this allows users to run `optimize_acqf` with the `fixed_features` keyword via the convenient syntax:
```
fixed_features=dict(zip(fixed_dims.tolist(), features[fixed_dims])),
```
which errors without this commit if `features` is a tensor.

Differential Revision: D41777838


Balandat (approved):
that looks reasonable to me!

PR #1551, opened 2022-12-09
Typo in mixed gp docstring

jduerholt:
## Motivation

This PR fixes a typo in the doctring of mixed GP.

Yes.

## Test Plan

Not needed for typo fix.


jduerholt:
For some reason, the tests are failing. For me it looks like some problem with pyro.

Balandat:
I think the issue is just that the pytorch nightlies are now 2.0+ and pyro explicitly checks for the version to be 1.x. 

We'll have to look into what the best way to deal with this is (pinning pytorch or bumping pyro / using the latest). Pytorch 2.0 is fully backward compatible so pyro really shouldn't be crying foul here.

PR #1554, opened 2022-12-10
Error out on NaN values in BoxDecomposition

saitcakmak:
Summary: In some cases, having NaN values can lead to partitioning evaluating to 0 hypervolume without any errors being raised. This is not a consistent behavior, and only happens with certain datasets and not others. Erroring out on NaN values seems like a safe solution.

Differential Revision: D41895761


Balandat (approved):
lgtm!

PR #1557, opened 2022-12-12
Fix cont_kernel_factory signature

jduerholt:
## Motivation

There was a bug in the signature and docstring for the `cont_kernel_factory` argument in `MixedSingleTaskGP`. The `batch_shape` argument was missing.

Yes.

## Test Plan

Not needed.

### DIFF
--- a/botorch/models/gp_regression_mixed.py
+++ b/botorch/models/gp_regression_mixed.py
@@ -62,7 +62,9 @@ class MixedSingleTaskGP(SingleTaskGP):
         train_X: Tensor,
         train_Y: Tensor,
         cat_dims: List[int],
-        cont_kernel_factory: Optional[Callable[[int, List[int]], Kernel]] = None,
+        cont_kernel_factory: Optional[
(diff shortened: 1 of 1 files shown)


SebastianAment:
Thank you for adding this @jduerholt! We'll wait until the internal tests finish, but this looks good to me.

jduerholt:
> Thank you for adding this @jduerholt! We'll wait until the internal tests finish, but this looks good to me.

Just out of curiosity: What are these internal additional tests? I was always wondering ... :D

saitcakmak:
> Just out of curiosity: What are these internal additional tests? I was always wondering ... :D

It includes test for some other OSS projects like Ax & AEPsych that depend on BoTorch and other internal projects that build on Ax & BoTorch. We also run BoTorch unit tests on both CPU and GPU (GitHub CI is CPU only) and also test some internal only BoTorch features (experimental or haven't made it to OSS due to limited bandwidth on our end).

PR #1559, opened 2022-12-12
BAxUS tutorial

LeoIV:
## Motivation

This pull request adds a tutorial for [BAxUS](https://openreview.net/pdf?id=e4Wf6112DI) (published at NeurIPS 2022), a novel BO method for high-dimensional problems. BAxUS allows for sample-efficient BO by using nested random embeddings that preserve observations. 

Including BAxUS as a tutorial in Botorch helps other researchers to build on our work and makes it accessible for practitioners.

Yes.

## Test Plan

This PR does not add any core functionality, therefore no tests were added.

## Related PRs

No new functionality.


dme65:
Hi Leonard, thank you for the PR! I'll review this in the next few days.

dme65 (commented):
Thank you for contributing this tutorial and sorry for the delayed review! This looks great to me overall.

On a high level there is plenty of overlap between this tutorial and the TuRBO and SCBO tutorials. This is outside the scope of this diff, but we should probably get some of the core trust region functionality checked into BoTorch so we can cut some of the code in these tutorials. cc @Balandat .

- Can you run isort and flake8 on the notebook? Some of the lines are quite long which makes the code hard to read.
- Why are you using a domain of [-5, 15]^500? This is a bit non-standard for Branin which is usually optimized over [-5, 10] x [0, 15]. You can do something similar to https://github.com/pytorch/botorch/blob/main/tutorials/saasbo.ipynb in which case you can avoid having to define DummyBranin.
- Why are you learning the GP hyperparameters using Adam instead of calling fit_gpytorch_mll which will use L-BFGS-B? We generally find that fit_gpytorch_mll works better and you can see how this is done in https://botorch.org/tutorials/turbo_1.
- Can you rename X_turbo, Y_turbo, etc. to X_baxus, Y_baxus, … ?
- Dummy Branin -> Embedded Branin

LeoIV:
Thank you for the review! I agree that there's a lot of overlap, but I think it's better to use similar code than to implement it differently if we cannot share code between the tutorials.

* I switched to `fit_gpytorch_mll`. I changed it originally because I ran into errors with that before. I guess the reason is that when I increase the target space, all observations are constrained to a linear subspace. But I switched back and have yet to face problems. 
* I now use [-5, 10] and [0, 15] (same method as Saasbo)
* Renamed "Dummy" to ``Embedded'' and `X_turbo` to `X_baxus` etc.
* I ran isort and black on the notebook

Please let me know if I need to make additional changes!

tutorials/baxus.ipynb
```diff
@@ -0,0 +1,985 @@
...
+    "        n_splits = round(math.log(self.dim, self.new_bins_on_split + 1))\n",
+    "        self.d_init = (\n",
+    "            np.argsort(\n",
+    "                [\n",
+    "                    np.abs(i * (self.new_bins_on_split + 1) ** n_splits - self.dim)\n",
+    "                    for i in range(1, self.new_bins_on_split + 1)\n",
+    "                ]\n",
+    "            )[0]\n",
+    "            + 1\n",
+    "        )\n",
```
Balandat, line 165: why not use `argmin()` instead of `argsort()[0]`?

Also, looks like there is no need to do a list comprehension here this can be vectorized

```   
        nbos = self.new_bins_on_split
        self.d_init = 1 + np.argmin(
            np.abs((1 + np.arange(nbos)) * (1 + nbos) ** n_splits - self.dim)
        )
```

tutorials/baxus.ipynb
```diff
@@ -0,0 +1,985 @@
...
+     "output_type": "execute_result"
+    }
+   ],
+   "source": [
+    "def embedding_matrix(input_dim: int, target_dim: int) -> torch.Tensor:\n",
+    "\n",
+    "    if (\n",
+    "        target_dim >= input_dim\n",
+    "    ):  # return identity matrix if target size greater than input size\n",
+    "        return torch.eye(input_dim).to(device=device, dtype=dtype)\n",
```
Balandat, line 242: There are a number of other places in this PR where a `.to` isn't necessary.

```suggestion
    "        return torch.eye(input_dim, device=device, dtype=dtype)\n",
```

Balandat (commented):
thanks, there are a few places in which things can be cleaned up but overall this lgtm.

tutorials/baxus.ipynb
LeoIV: Thanks, I think I fixed this where possible

Balandat:
Thanks @LeoIV for making the edits. @dme65 is currently out - let's wait for him to get back to merge this in.

LeoIV:
Thanks, I fixed the seed. When the embedding is increased, the previous observations lie in a lower-dimensional subspace of that higher-dimensional embedding. This seems to cause problems with LBFGS in some cases. What I do in my work is to catch the `ModelFittingError` and revert to `Adam`-based MLE. I could add that but it would increase the notebook length by ~10-20 lines.

saitcakmak:
> What I do in my work is to catch the ModelFittingError and revert to Adam-based MLE. I could add that but it would increase the notebook length by ~10-20 lines.

I think that could be helpful for someone else that's using this tutorial to optimize their problem and runs into a similar issue. Let's add it with a simple comment explaining why this is needed.

LeoIV:
I added the fallback optimization and a comment about that. Also, I added a context manager to suppress the input data validation. The data is normalized but to [-1, 1] instead of [0, 1] due to the choice of the embedding.

saitcakmak (approved):

PR #1562, opened 2022-12-14
Add timeout ability to gen_* functions

Balandat:
Summary: Exposes the ability to pass a `timeout_sec` kwarg to `gen_candidates_scipy` and `gen_candidates_torch` that allows to stop the optimization after running for more than `timeout_sec` seconds.

Differential Revision: D42024857

### DIFF
--- a/botorch/generation/gen.py
+++ b/botorch/generation/gen.py
@@ -10,6 +10,7 @@ Candidate generation utilities.
 
 from __future__ import annotations
 
+import time
 import warnings
 from functools import partial
 from typing import Any, Callable, Dict, List, NoReturn, Optional, Tuple, Type, Union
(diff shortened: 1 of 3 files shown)


Balandat:
Ugh the tutorial failure here is kind of a pain - the tutorial is importing a context manager from ax, but that context manager mocks things deep in botorch that this PR changes, and the context manager itself is only updated as part of the diff that contains this PR. Such circular dependency issue here is not a great design and we generally shouldn't be importing things from Ax in botorch (the fact that we have Ax tutorials makes this hard but for reasonably high-level behavior this should be ok). 

I think what I might do is add a `minimize` function back in as an alias that raises a `DeprecationWarning` that we can remove after the next Ax release

Issue #1566, opened 2022-12-15
fit_gpytorch_mll fails on cuda device


Balandat:
```
from botorch.fit import fit_gpytorch_mll
fit_gpytorch_mll(mll);
---------------------------------------------------------------------------
MDNotImplementedError                     Traceback (most recent call last)
[/usr/local/lib/python3.8/dist-packages/botorch/utils/dispatcher.py](https://localhost:8080/#) in __call__(self, *args, **kwargs)
     92         try:
---> 93             return func(*args, **kwargs)
     94         except MDNotImplementedError:

8 frames
MDNotImplementedError: 

During handling of the above exception, another exception occurred:

TypeError                                 Traceback (most recent call last)
[/usr/local/lib/python3.8/dist-packages/torch/_tensor.py](https://localhost:8080/#) in __array__(self, dtype)
    955             return self.numpy()
    956         else:
--> 957             return self.numpy().astype(dtype, copy=False)
    958 
    959     # Wrap Numpy array again in a suitable tensor when done, to support e.g.

TypeError: can't convert cuda:0 device type tensor to numpy. Use Tensor.cpu() to copy the tensor to host memory first.
```

which disappears after I move mll to CPU:
`fit_gpytorch_mll(mll.cpu());`

The model is in GPU:
`print(next(model.parameters()).is_cuda)
True`

_Originally posted by @jackliu333 in https://github.com/pytorch/botorch/issues/1323#issuecomment-1352487057_

jackliu333:
Sure, please find below reproducible example:
```
import torch

device = torch.device("cuda" if torch.cuda.is_available() else "cpu")
dtype = torch.double

# unknown objective function
from botorch.test_functions import Hartmann
neg_hartmann6 = Hartmann(negate=True)

# generate initial dataset
NOISE_SE = 0.1

def generate_initial_data(n=10):
    # generate training data
    train_x = torch.rand(n, 6, device=device, dtype=dtype)
    exact_obj = neg_hartmann6(train_x).unsqueeze(-1)  # add output dimension
    train_y = exact_obj + NOISE_SE * torch.randn_like(exact_obj)
    best_observed_value = train_y.max().item()
    return train_x, train_y, best_observed_value

train_x, train_y, best_observed_value = generate_initial_data(n=5)

# initialize GP model
from botorch.models import SingleTaskGP
from gpytorch.mlls import ExactMarginalLogLikelihood

def initialize_model(train_x, train_y):
[...]

PR #1567, opened 2022-12-15
Make `get_bounds_as_ndarray` device safe

saitcakmak:
Summary:
Issue raised in https://github.com/pytorch/botorch/issues/1566.

The `bounds` can be a `Dict[str, Tuple[Tensor, Tensor]]` where the tensors live on GPU, which leads to a device error when trying to assign them to a numpy array.

Differential Revision: D42060124


Balandat (approved):
thanks for cleaning this up

Balandat:
Should we add a test for this on both CPU and GPU?

PR #1568, opened 2022-12-15
remove input transform checks

jduerholt:
## Motivation

As I am currently refactoring our internal codebase, I had a look at @sdaulton PR regarding probabilistic reparameterization.
From my understanding one has to use it by representing the categoricals by a one hot encoding for the reparmeterized ACQF and then eventually transforming the input to a numerical represenation via `OneHotToNumeric` especially when one wants to use it togehter with `MixedSingleTaskGP`. Currently MixedSingleTaskGP is very strict on which input transforms are allowed. This PR lifts the restrictions to make it usable with  OneHotToNumeric`.

Note that the transform also has to be instantiated with `transform_on_train = False` and `train_X` has to be transformed before it is passed to the constructor of `MixedSingleTaskGP`, else the indices for the different kernels are mixed up.

Yes.

## Test Plan

Unit tests.

## Related PRs

https://github.com/pytorch/botorch/pull/1534


Balandat:
@saitcakmak this is an interesting one - any thoughts how this would relate to / interact with the input transform refactor that has been on your mind?

saitcakmak:
@balandat This just removes some user facing checks. It'll require the users be more careful about what input transform they use but it shouldn't affect anything else otherwise.

jduerholt:
One other thing in the context of probabilistic reparameterization:

Based on @sdaulton PR on probabilistic reparameterization, I asked myself how one would set up parameterized categoricals via probabilistic reparameterization. 

An example would be to have a categorical variable named molecule with 10 different categories (molecules). I am able to parametrize every molecule by a descriptor vector of length three, meaning three numbers representing three properties of every molecule (I think you did something like this in the original paper). 

From what I understood from @sdaulton's implementation one would encode this parametrized categorical still as one-hot for the optimizer and the reparametrized acqfs, but would need an input transform `OneHotToDescriptor` which takes the one-hot encoded categorical and translates it to the vector representation in the spirit of `OneHotToNumeric`, or is this wrong?

jduerholt:
> @Balandat This just removes some user facing checks. It'll require the users be more careful about what input transform they use but it shouldn't affect anything else otherwise.

@Balandat @saitcakmak Is this fine for you or should I change something in addition before the merge?

Balandat:
I am fine with merging this in as is

PR #1573, opened 2022-12-15
fix docstring in ModelList

jduerholt:
## Motivation

This PR fixes an error in the docstring of `ModelList`. There an example is shown in which a non-existent method named `predict` is called. I assume, the original intention was to call `posterior` there.

Yes.

## Test Plan

Only fix of a doc string, no additional tests needed.


Balandat (approved):
Thanks for the fix!

PR #1574, opened 2022-12-16
Adding input constructor of `LogExpectedImprovement`

SebastianAment:
Summary: Registering an acquisition function with `acqf_input_constructor` is necessary to play nicely with Ax.

Differential Revision: D42099841


Balandat (approved):
gtmtl!

PR #1576, opened 2022-12-17
Some numerical tweaks to fully Bayesian models

Balandat:
Summary: Streamlines the code, makes some numerical improvments, and gets rid of redundant computations.

Reviewed By: dme65

Differential Revision: D42061081


Balandat:
This is missing an increment to the loop var `i`, but there seem to be some larger issues with the numerics here even prior to this PR, holding off with this PR until those are resolved.

PR #1578, opened 2022-12-19
Raise an error that Pyro can catch when numerical issues occur during sampling

dme65:
Summary:
We have been seeing psd errors show up in the Pyro sampling. D41405255 (https://github.com/pytorch/botorch/commit/c6595ed2c8624368d68281b59384ac87beb85ab4) added `_psd_safe_pyro_mvn_sample` which tries to adaptively add jitter when `pyro.sample` fails, but this has turned out to add really extreme amounts of jitter (like `99999.999...`).

It turns out that the reason this error started showing up in the first place is that we switched over to using linear operator which raises a `LinAlgError` that Pyro isn't familiar with. In particular, Pyro has an inbuilt mechanism for catching numerical issues because of extreme hyperparameters, but does so by looking for a `RuntimeError` where `is not positive-definite` or `singular U` is part of the error message.

The current workaround is to catch `LinalgError`'s from linear operator and raise a runtime error that matches what Pyro is looking for. This also seems to have gotten rid of some mysterious Cholesky errors in gen that were most likely caused by learning bad hyperparameters due to too much jitter being added.

Reviewed By: Balandat

Differential Revision: D42146265

### DIFF
--- a/botorch/models/fully_bayesian.py
+++ b/botorch/models/fully_bayesian.py
@@ -32,7 +32,6 @@ References:
 
 
 import math
-import warnings
 from abc import abstractmethod
 from typing import Any, Dict, List, Mapping, Optional, Tuple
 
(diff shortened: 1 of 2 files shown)


saitcakmak:
Closing as abandoned

PR #1588, opened 2022-12-23
Add Constraint Active Search tutorial to ignore list

esantorella:
## Motivation

The tutorials have been failing in the nightly cron for several days. [Binary search](https://github.com/pytorch/botorch/pull/1581) identifies the constraint active search tutorial as the culprit. I suspect this is an OOM. I plan to fix it after the holidays, but since that will be over a week (optimistically) let's ignore it for now.

## Test Plan

Manually trigger cron on this branch and make sure it passes.

## Related PRs

#1581  debugs this.

### DIFF
--- a/scripts/run_tutorials.py
+++ b/scripts/run_tutorials.py
@@ -23,6 +23,9 @@ IGNORE = {  # ignored in smoke tests and full runs
     "vae_mnist.ipynb",  # requires setting paths to local data
     "bope.ipynb",  # flaky, keeps failing the workflows
     "preference_bo.ipynb",  # failing. Fix planned
+    # Causing the tutorials to crash when run without smoke test. Likely OOM.
+    # Fix planned.
+    "constraint_active_search",
 }
 IGNORE_SMOKE_TEST_ONLY = {  # only used in smoke tests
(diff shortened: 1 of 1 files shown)


saitcakmak (approved):
Thanks! That must've been a good bit of work to isolate this.

esantorella:
> Thanks! That must've been a good bit of work to isolate this.

It took the expect `log2(n_tutorials)` tries, but perhaps we should have a way to find the source of tutorials crashes faster in the future, like running them in parallel or logging stuff where it won't be lost in the crash. Also obviously I want to reduce memory usage and track regressions better so they don't OOM.

saitcakmak:
> but perhaps we should have a way to find the source of tutorials crashes faster in the future, like running them in parallel

We could rewrite the tutorial script to run only a single tutorial at a time, update the workflow to discover all available tutorials and run them in a matrix. It'd be a decent bit of work to do that though & would increase the setup overhead.

Balandat:
So since we're running the tutorials in subprocesses anyway, one thing we could do is to set a hard limit on the memory the subprocess can request (i.e. currently available memory minus some buffer). Let me put up a PR to that end.

PR #1590, opened 2022-12-23
Fix typo in tutorials ignorelist; add SASBO and scalable_constrained_bo to ignorelist because they are timing out

esantorella:
## Motivation

- Typo in #1588 
- scalable_constrained_bo timing out and causing a failure since Nighly Cron 708 (2 weeks ago) or earlier
- saasbo timing out and causing a failure at least some of the time since Nighly Cron 706 (2 weeks ago) or earlier

Yes

## Test Plan

Temporarily trigger actions to run tutorials without smoke test on PR so that we can confirm they will pass now

## Related PRs

#1588

### DIFF
--- a/.github/workflows/tutorials.yml
+++ b/.github/workflows/tutorials.yml
@@ -10,7 +10,7 @@ on:
 
 jobs:
   run_tutorials_with_smoke_test:
-    name: Run tutorials with smoke test on latest PyTorch / GPyTorch / Ax
+    name: Run tutorials without smoke test on latest PyTorch / GPyTorch / Ax
     uses: ./.github/workflows/reusable_tutorials.yml
     with:
       smoke_test: false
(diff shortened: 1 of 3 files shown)


.github/workflows/tutorials.yml
```diff
@@ -10,9 +10,9 @@ on:
 
 jobs:
   run_tutorials_with_smoke_test:
-    name: Run tutorials with smoke test on latest PyTorch / GPyTorch / Ax
+    name: Run tutorials without smoke test on latest PyTorch / GPyTorch / Ax
     uses: ./.github/workflows/reusable_tutorials.yml
     with:
-      smoke_test: true
+      smoke_test: false
```
Balandat, line 16: Hmm wait what's happening here? Do we mean to run these without or with smoke test here? And why?  Either way there is an inconsistency, this job is called `run_tutorials_with_smoke_test` but the setting and name say otherwise.

esantorella: Oh this is just for testing and will be reverted. It's so the tutorials run with `smoke_test: false` on a PR. Otherwise I would have needed to test this by making a branch on the botorch repo instead of on my own fork and then triggering the nightly cron manually.

esantorella: I did it through Actions since the OOM doesn't happen on my machine. I also plan to print the max memory usage and runtime of each tutorial so that tutorials issues can be more easily debugged from any machine.

PR #1591, opened 2022-12-23
Make tutorials print starting and maximum memory usage

esantorella:
## Motivation

Tutorials have sometimes failed due to running out of memory, which is very annoying. Also, it's helpful to know if changes are making memory usage better or worse. With this change, we will now print the starting and maximum memory usage for each tutorial.

Before:
> Running tutorial multi_fidelity_bo.ipynb took 16.61 seconds.

After:
> Running tutorial multi_fidelity_bo.ipynb took 16.61 seconds. Memory usage started at 230.27734375 MB and the maximum was 732.27734375 MB.

Yes

## Test Plan

- Looked at the output of running the tutorials locally and saw that the results looked reasonable. Will check again on GH actions.


test/utils/test_profiling.py
```diff
@@ -0,0 +1,20 @@
+#!/usr/bin/env python3
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+#
+# This source code is licensed under the MIT license found in the
+# LICENSE file in the root directory of this source tree.
+from botorch.utils.profiling import get_memory_usage_preserving_output
```
Balandat, line 6:
```suggestion

from botorch.utils.profiling import get_memory_usage_preserving_output
```

Balandat (commented):
I very much like the general idea of this. A couple of questions:

Will `memory_profiler` also properly account for the memory of non-python objects (I'm thinking primarily of torch Tensors here)? 

I think the pattern of having the decorator change the return type of a function is not ideal - people will look at the typing information on the function and it's going to be quite opaque that the return type is actually a different one due to the decorator. Not sure what a good alternative would be here. One option would be to just have the decorator take a dict or some other mutable container that is a global variable and then write the result to that, sth like
```
@get_memory_usage_preserving_output(out_dict=some_module_level_result_dict)
def run_script(script: str, env: Optional[Dict[str, str]] = None) -> None:
...
```
I think an ideal setup would be to just have a context manager that would look like
```
with profile_memory() as pf:
   run_script(...)
   mem_usage = pf.get_mem_usage()
```
but I'm not sure how easy it would be to do this.

esantorella:
> Will `memory_profiler` also properly account for the memory of non-python objects (I'm thinking primarily of torch Tensors here)?

Yes, because it's using psutil as a backend. Results are in line with what I eyeball from OSX Activity Monitor.
>
> I think the pattern of having the decorator change the return type of a function is not ideal

It turns out there was a much simpler way of doing this with nothing fancy needed. Rewritten.

Balandat (approved):
lgtm!

esantorella:
I can't reject the null hypothesis that the profiling has no effect on runtime, but there's a lot of inconsistency between runs. That becomes even more apparent when looking at individual tutorials. It would be nice to make the tutorials more deterministic -- that would make them useful for checking for regressions. And possibly a change as simple as setting a seed might fix some of the timeouts.

image

saitcakmak (approved):
Thanks for checking @esantorella. I think it's ok to land this and address any time outs if they come up. Fixing the seeds also sounds reasonable to me. It'd eliminate one source of variation.

esantorella:
And here's the data on memory for the last 4 on-PR workflow runs on this branch. We'll start getting more representative data one day at a time once this starts running "without smoke test" in the nightly cron.
image
And runtime, which is much more variable within-workflow and also more skewed:
image

PR #1595, opened 2022-12-27
Rewrite installation readme for clarity; remove outdated references (e.g. CUDA on OSX)

esantorella:
## Test Plan

I tried out all 4 installation options in different conda environments on a Mac and they all worked. None resulted in errors that mentioned MKL. Read over the README and made sure the markdown rendered okay.

### DIFF
--- a/README.md
+++ b/README.md
@@ -108,7 +108,7 @@ pip install --upgrade git+https://github.com/pytorch/botorch.git
 ### Option 3: Editable/dev install
 
 If you want to [contribute](CONTRIBUTING.md) to BoTorch, you will want to install editably so that you can change files and have the
-changes reflected in your local install. 
+changes reflected in your local install.
 
 If you want to install the current `gpytorch` and `linear_operator` development versions, as in Option 2, do that
 before proceeding.
--- a/README.md
+++ b/README.md
@@ -68,25 +68,27 @@ Optimization simply use Ax.
 
 The latest release of BoTorch is easily installed either via
 [Anaconda](https://www.anaconda.com/distribution/#download-section) (recommended):
+
+First install PyTorch:
 ```bash
-conda install botorch -c pytorch -c gpytorch -c conda-forge
+# Needed for OSX only, see below.
+conda install pytorch torchvision -c pytorch 
+```
+
+_Recommendation for MacOS users:_ PyTorch is a required dependency of BoTorch, and can be automatically installed via `pip`. However, we recommend installing PyTorch manually (as above) so that PyTorch is properly linked against MKL (a library that optimizes mathematical computation for Intel processors). This will result in up to an order-of-magnitude speed-up for Bayesian optimization, as at the moment, installing PyTorch from pip does not link against MKL.
+
+If you need CUDA on MacOS, or want to customize your installation (i.e. CPU-only option), please follow the [PyTorch installation instructions](https://pytorch.org/get-started/locally/) to build from source.
+
+Then install BoTorch:
+```bash
+conda install botorch -c conda-forge
 ```
 or via `pip`:
 ```bash
 pip install botorch
 ```
 
-You can customize your PyTorch installation (i.e. CUDA version, CPU only option)
-by following the [PyTorch installation instructions](https://pytorch.org/get-started/locally/).
-
-***Important note for MacOS users:***
-* Make sure your PyTorch build is linked against MKL (the non-optimized version
-  of BoTorch can be up to an order of magnitude slower in some settings).
-  Setting this up manually on MacOS can be tricky - to ensure this works properly,
-  please follow the [PyTorch installation instructions](https://pytorch.org/get-started/locally/).
-* If you need CUDA on MacOS, you will need to build PyTorch from source. Please
-  consult the PyTorch installation instructions above.
-
+_Note_: Make sure the `pip` being used is actually the one from the newly created Conda environment. If you're using a Unix-based OS, you can use `which pip` to check.
 
 ##### Installing from latest main branch
 
--- a/README.md
+++ b/README.md
@@ -71,11 +71,11 @@ The latest release of BoTorch is easily installed either via
 
 First install PyTorch:
 ```bash
-# Needed for OSX only, see below.
+# Needed for OSX on Intel Macs only, see below.
 conda install pytorch torchvision -c pytorch 
 ```
 
-_Recommendation for MacOS users:_ PyTorch is a required dependency of BoTorch, and can be automatically installed via `pip`. However, we recommend installing PyTorch manually (as above) so that PyTorch is properly linked against MKL (a library that optimizes mathematical computation for Intel processors). This will result in up to an order-of-magnitude speed-up for Bayesian optimization, as at the moment, installing PyTorch from pip does not link against MKL.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -68,25 +68,27 @@ Optimization simply use Ax.
...
+
+First install PyTorch:
+```bash
+# Needed for OSX only, see below.
+conda install pytorch torchvision -c pytorch 
+```
+
+_Recommendation for MacOS users:_ PyTorch is a required dependency of BoTorch, and can be automatically installed via `pip`. However, we recommend installing PyTorch manually (as above) so that PyTorch is properly linked against MKL (a library that optimizes mathematical computation for Intel processors). This will result in up to an order-of-magnitude speed-up for Bayesian optimization, as at the moment, installing PyTorch from pip does not link against MKL.
+
+If you need CUDA on MacOS, or want to customize your installation (i.e. CPU-only option), please follow the [PyTorch installation instructions](https://pytorch.org/get-started/locally/) to build from source.
```
Balandat, line 80: So in more recent pytorch versions CUDA doesn't seem to be supported anymore on MacOS, but MPS (for Apple silicon) is: https://pytorch.org/get-started/locally/

We should update the readme here accordingly.

README.md
```diff
@@ -68,25 +68,27 @@ Optimization simply use Ax.
...
+
+Then install BoTorch:
 ```bash
-conda install botorch -c pytorch -c gpytorch -c conda-forge
+conda install botorch -c conda-forge
```
Balandat, line 84: At the time the `conda-forge` channel was a fix to make sure `pyro-ppl` can be resolved (see https://github.com/pytorch/botorch/pull/1207). In general we have recommended people install gpytorch from the `gpytorch` and `botorch` from the `pytorch` channels, respectively, rather than relying on `conda-forge` (which may not have the latest packages).

However, if this copying of the package into `conda-forge` has been working reliably and most people use `conda-forge` anyway then we could consider just pointing to `conda-forge` to simplify the installation? 

```suggestion
conda install botorch -c pytorch -c gpytorch -c conda-forge
```

README.md
```diff
@@ -68,25 +68,27 @@ Optimization simply use Ax.
 
 The latest release of BoTorch is easily installed either via
 [Anaconda](https://www.anaconda.com/distribution/#download-section) (recommended):
+
+First install PyTorch:
+```bash
+# Needed for OSX only, see below.
```
Balandat, line 74:
```suggestion
# Needed for OSX on Intel Macs only, see below.
```

README.md
```diff
@@ -68,25 +68,27 @@ Optimization simply use Ax.
...
+conda install pytorch torchvision -c pytorch 
+```
+
+_Recommendation for MacOS users:_ PyTorch is a required dependency of BoTorch, and can be automatically installed via `pip`. However, we recommend installing PyTorch manually (as above) so that PyTorch is properly linked against MKL (a library that optimizes mathematical computation for Intel processors). This will result in up to an order-of-magnitude speed-up for Bayesian optimization, as at the moment, installing PyTorch from pip does not link against MKL.
```
Balandat, line 78:
```suggestion
_Recommendation for MacOS users:_ PyTorch is a required dependency of BoTorch, and can be automatically installed via `pip`. However, if you are using an Intel Mac we recommend installing PyTorch manually (as above) so that PyTorch is properly linked against MKL (a library that optimizes mathematical computation for Intel processors). This will result in up to an order-of-magnitude speed-up for Bayesian optimization, as at the moment, installing PyTorch from pip does not link against MKL.
```

README.md
```diff
@@ -68,25 +68,27 @@ Optimization simply use Ax.
...
+
+If you need CUDA on MacOS, or want to customize your installation (i.e. CPU-only option), please follow the [PyTorch installation instructions](https://pytorch.org/get-started/locally/) to build from source.
+
+Then install BoTorch:
 ```bash
-conda install botorch -c pytorch -c gpytorch -c conda-forge
+conda install botorch -c conda-forge
 ```
 or via `pip`:
```
Balandat, line 86: This "or" is now pretty far away from the "either"
```suggestion
Alternatively, you can install BoTorch via `pip`:
```

README.md
esantorella: Out of curiosity, has anyone tested BoTorch with MPS?

README.md
esantorella: yeah that seems right -- I thought that installing BoTorch from the pytorch channel rather than conda-forge fixed the MKL warning I was getting, but on second check I'm not sure that's right. I'll look into that more carefully and make sure I land on some install instructions that are actually working.

README.md
Balandat: I believe @sdaulton has. From browsing through its docs, MPS current still seems to be in its infancy in terms of supporting accelerated linear algebra, but it looks like they're working on closing that gap.

README.md
```diff
@@ -63,60 +63,77 @@ Optimization simply use Ax.
 - scipy
 - multiple-dispatch
 
+### Prerequisite only for MacOS users:
+Before installing BoTorch, we recommend first manually installing PyTorch, a required dependency of BoTorch. Installing it according to the [PyTorch installation instructions](https://pytorch.org/get-started/locally/) ensures that it is properly linked against MKL, a library that optimizes mathematical computation for Intel processors. This will result in up to an order-of-magnitude speed-up for Bayesian optimization, as at the moment, installing PyTorch from pip does not link against MKL.
 
-##### Installing the latest release
+The PyTorch installation instructions currently recommend:
+1. Install [Anaconda](https://www.anaconda.com/distribution/#download-section). Note that there are different installers for Intel and M1 Macs.
+2. Install PyTorch with `conda install pytorch torchvision -c pytorch`.
```
Balandat, line 71: Do we want to add this here? This means that unless we monitor the PyTorch installation instructions this could easily go stale. Ideally we can avoid that by just pointing to the installation instructions. Wdyt?

README.md
```diff
@@ -63,60 +63,77 @@ Optimization simply use Ax.
 - scipy
 - multiple-dispatch
 
+### Prerequisite only for MacOS users:
```
Balandat, line 66: There shouldn't be a need to do this for Apple Silicon (unless the user wants to use MPS, but that's a separate story)
```suggestion
### Prerequisite only for MacOS users with Intel processors:
```

README.md
```diff
@@ -63,60 +63,77 @@ Optimization simply use Ax.
...
+Before installing BoTorch, we recommend first manually installing PyTorch, a required dependency of BoTorch. Installing it according to the [PyTorch installation instructions](https://pytorch.org/get-started/locally/) ensures that it is properly linked against MKL, a library that optimizes mathematical computation for Intel processors. This will result in up to an order-of-magnitude speed-up for Bayesian optimization, as at the moment, installing PyTorch from pip does not link against MKL.
```
Balandat, line 67: We've generally tried to limit line lengths to around 100ish - mind doing this as well and adding a few breaks (here and below)? It's a lot easier to read in basic text editors/viewers that don't auto-wrap lines.

Balandat (approved):
Thanks for these updates! Overall this lgtm, just some nits and one point about just pointing to the pytorch installation instructions to avoid a stale readme.

README.md
esantorella, line 71:
```suggestion
2. Install PyTorch following the [PyTorch installation instructions](https://pytorch.org/get-started/locally/).
Currently, this suggests running `conda install pytorch torchvision -c pytorch`.
```

README.md
esantorella, line 67:
```suggestion
Before installing BoTorch, we recommend first manually installing PyTorch, a required dependency of
BoTorch. Installing it according to the [PyTorch installation instructions](https://pytorch.org/get-started/locally/)
ensures that it is properly linked against MKL, a library that optimizes mathematical computation for Intel processors.
This will result in up to an order-of-magnitude speed-up for Bayesian optimization, as at the moment,
installing PyTorch from pip does not link against MKL.
```

PR #1596, opened 2022-12-27
[DRAFT] Add memory limit to tutorial evaluation

Balandat:
Add memory limit to tutorial evaluation in `run_tutorials`. 

Context: https://github.com/pytorch/botorch/pull/1588#issuecomment-1366194396

Note: This is causing weird errors on my M1 Mac, putting this up as a PR to test on the CI machines.

### DIFF
--- a/scripts/run_tutorials.py
+++ b/scripts/run_tutorials.py
@@ -8,7 +8,6 @@ from __future__ import annotations
 
 import argparse
 import os
-import resource
 import subprocess
 import sys
 import tempfile
@@ -42,8 +41,13 @@ IGNORE_SMOKE_TEST_ONLY = {  # only used in smoke tests
 }
 
 
-def _limit_memory(soft: int, hard: int = resource.RLIM_INFINITY) -> None:
+def _limit_memory(soft: int, hard: Optional[int] = None) -> None:
     """Limit the memory usage of a running python process"""
+    import resource
+    
+    if hard is None:
+        hard = resource.RLIM_INFINITY
+    
     resource.setrlimit(resource.RLIMIT_AS, (soft, hard))
 
 
--- a/scripts/run_tutorials.py
+++ b/scripts/run_tutorials.py
@@ -8,9 +8,12 @@ from __future__ import annotations
 
 import argparse
 import os
+import resource
 import subprocess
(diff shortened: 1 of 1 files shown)


saitcakmak:
Looks like this is causing some weird failures: `/opt/hostedtoolcache/Python/3.8.15/x64/bin/python: error while loading shared libraries: libc.so.6: failed to map segment from shared object`

Balandat:
Seems like this is a KP: https://stackoverflow.com/questions/69365918/limiting-the-memory-used-by-a-task-in-subprocess-in-linux

I wonder if it has to do with accessing the module-level `resource` import, let me try something.

scripts/run_tutorials.py
```diff
@@ -56,12 +68,18 @@ def run_script(script: str, env: Optional[Dict[str, str]] = None) -> None:
             tmp_script.write(script)
     if env is not None:
         env = {**os.environ, **env}
+    if sys.platform != "darwin":
+        # limiting memory usage fails on Mac (at least Apple Silicon)
```
esantorella, line 72: Fails how? Is there any GH issue or something for this?

scripts/run_tutorials.py
```diff
@@ -56,12 +68,18 @@ def run_script(script: str, env: Optional[Dict[str, str]] = None) -> None:
...
+        preexec_fn = partial(_limit_memory, 4 * 1024 * 1024)  # 4GB memory limit
```
esantorella, line 73: Supposedly these machines should have at least 7 GB https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners

esantorella: Also I'm not sure this is in bytes? When I run

```bash
python -c "import resource; print(resource.getrlimit(resource.RLIMIT_AS))"
```

then I get `(9223372036854775807, 9223372036854775807)`.

When I run

```bash
python -c "import resource; resource.setrlimit(resource.RLIMIT_AS, (1024 * 1024 * 4, resource.RLIM_INFINITY))"
```

I get an error:

```bash
  File "<string>", line 1, in <module>
ValueError: current limit exceeds maximum limit
```

PR #1597, opened 2022-12-27
Add workflow_dispatch event to reusable tutorials workflow

Balandat:
This will allow us to manually run the reusable tutorial workflow against selected branches.

### DIFF
--- a/.github/workflows/reusable_tutorials.yml
+++ b/.github/workflows/reusable_tutorials.yml
@@ -1,6 +1,20 @@
 name: Reusable Tutorials Workflow
 
 on:
+  workflow_dispatch:
+    inputs:
+      smoke_test:
+        required: false
+        type: boolean
+        default: true
+      use_stable_pytorch_gpytorch:
+        required: false
+        type: boolean
+        default: false
+      use_stable_ax:
+        required: false
+        type: boolean
+        default: false
   workflow_call:
     inputs:
       smoke_test:
--- a/.github/workflows/reusable_tutorials.yml
+++ b/.github/workflows/reusable_tutorials.yml
@@ -4,17 +4,14 @@ on:
   workflow_dispatch:
     inputs:
       smoke_test:
-        required: false
+        required: true
         type: boolean
-        default: true
       use_stable_pytorch_gpytorch:
-        required: false
+        required: true
         type: boolean
-        default: false
       use_stable_ax:
-        required: false
+        required: true
         type: boolean
-        default: false
   workflow_call:
     inputs:
       smoke_test:


.github/workflows/reusable_tutorials.yml
```diff
@@ -1,6 +1,20 @@
 name: Reusable Tutorials Workflow
 
 on:
+  workflow_dispatch:
+    inputs:
+      smoke_test:
+        required: false
+        type: boolean
+        default: true
```
saitcakmak, line 9: Should we remove the defaults since this will be user triggered anyways? It's likely that we'll want to use non-default values with this most of the time.

saitcakmak (approved):
Cool, I've been thinking that this could be a useful addition.

.github/workflows/reusable_tutorials.yml
Balandat: sure I'll just make those required then

esantorella:
I'm not sure I understand the use case since the `tutorials.yml` workflow already can be run manually. That reproduces what runs on PR. Is the idea to allow for reproducing what runs in the nightly cron (no smoke test)? Or is it to provide more flexibility?

saitcakmak:
> Is the idea to allow for reproducing what runs in the nightly cron (no smoke test)?

This would let you run the full tutorials without having to run the whole cron (and crowd out cron actions tab), and also enables more combinations than what's available as part of cron (such as testing against stable, which is also available as part of "test against stable" -- or running smoke test against stable dependencies, which is not available anywhere).

esantorella:
> This would let you run the full tutorials without having to run the whole cron (and crowd out cron actions tab)

This makes sense to me, but reproducing the nightly cron seems a little tricky since it will take a couple extra clicks to figure out which inputs to run it with. Someone debugging a tutorials failure might not be super familiar with these yml files. Maybe there should be an additional workflow dispatch for running the tutorials the way they are in the nightly cron?

> and also enables more combinations than what's available as part of cron (such as testing against stable, which is also available as part of "test against stable" -- or running smoke test against stable dependencies, which is not available anywhere).

Yep this makes sense to me

saitcakmak:
> reproducing the nightly cron seems a little tricky since it will take a couple extra clicks to figure out which inputs to run it with.

That's fair, but it is actually pretty straightforward. The name of the workflow run from nightly cron gives you all the options you need:
Screen Shot 2022-12-27 at 3 16 54 PM
No smoketest, latest, latest

I believe I made the names explicit in all workflows that use the reusable tutorials. 

Another way to figure out is to look at what steps the workflow actually runs:
Screen Shot 2022-12-27 at 3 19 52 PM
Runs install latest & skips stable installs, skips smoke test runs and runs the tutorials without smoke test.

> Maybe there should be an additional workflow dispatch for running the tutorials the way they are in the nightly cron?

I think this'd require making another `.yml` file just for this purpose. I personally prefer to avoid duplication.

PR #1610, opened 2023-01-04
Fix nightly PyPI upload test

saitcakmak:
This was using the pinned version of GPyTorch & linear operator, leading to failures.

Test plan: PyPI upload test now passes
https://github.com/pytorch/botorch/actions/runs/3842548266/jobs/6543945107


Balandat (approved):
thanks!

PR #1611, opened 2023-01-05
Make "device" attribute of contextual models a property instead

esantorella:
## Motivation

As of https://github.com/cornellius-gp/gpytorch/pull/2234, the parent class of BoTorch kernels now has a property "device." This means that if a subclass tries to set `self.device`, it will error. This is why the BoTorch CI is currently breaking: https://github.com/pytorch/botorch/actions/runs/3841992968/jobs/6542850176

## Test Plan

Tests should pass

### DIFF
--- a/botorch/models/kernels/contextual_lcea.py
+++ b/botorch/models/kernels/contextual_lcea.py
@@ -61,7 +61,7 @@ class LCEAKernel(Kernel):
         self.decomposition = decomposition
         self.batch_shape = batch_shape
         self.train_embedding = train_embedding
-        self.device = device
+        self._device = device
 
         num_param = len(next(iter(decomposition.values())))
         self.context_list = list(decomposition.keys())
@@ -127,6 +127,10 @@ class LCEAKernel(Kernel):
             lambda m, v: m._set_outputscale_list(v),
         )
         self.register_constraint("raw_outputscale_list", Positive())
+
+    @property
+    def device(self) -> Optional[torch.device]:
+        return self._device
 
     @property
     def outputscale_list(self) -> Tensor:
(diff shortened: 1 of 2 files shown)


botorch/models/kernels/contextual_lcea.py
```diff
@@ -61,7 +61,7 @@ def __init__(
         self.decomposition = decomposition
         self.batch_shape = batch_shape
         self.train_embedding = train_embedding
-        self.device = device
+        self._device = device
```
Balandat, line 64: Hmm I don't think we need to assign this anywhere, actually. If we just use the `device` kwarg to set things here in `__init__` then the parameter `raw_outputscale_list` will be set with the proper `device` here: https://github.com/pytorch/botorch/pull/1611/files#diff-3bc5cfbe1a19e8fd5c96a54e56c367f98e78171fe26b408794a67d641d27fc57R111-R122

If we then remove the `device` property and just use that of the base class, then the standard functionality 
should return the device correctly (note that `has_lenghtscale` is `False` here): https://github.com/cornellius-gp/gpytorch/blob/master/gpytorch/kernels/kernel.py#L265-L273

Balandat: That said, we'll have to do this for the `SACKernel` below, so feel free to keep the current version if you prefer.

Balandat (approved):

botorch/models/kernels/contextual_lcea.py
esantorella: I went with this design for backwards compatibility: historically `self.device` has been `None` when the `device` argument isn't specified. That's probably not the ideal way for `self.device` to behave, but I figure we can reconsider that in a future PR.

PR #1613, opened 2023-01-05
Deduplicate getting started

esantorella:
## Motivation

* We shouldn't recommend installing from pip without a lot of caveats since this won't work well on OSX, as explained in the README
* It's best to not repeat information on version numbers, etc. in too many places; we should treat the README as the source of truth on installation instructions. Any install instructions other than the README should be brief and not misleading.

## Test Plan

* Looked at the changed files in a Markdown viewer
* Built the website and looked at the changed pages:
image
image

### DIFF
--- a/docs/getting_started.md
+++ b/docs/getting_started.md
@@ -14,13 +14,21 @@ Before jumping the gun, we recommend you start with the high-level
 #### Installation Requirements:
 
 BoTorch is easily installed via
-[Anaconda](https://www.anaconda.com/distribution/#download-section):
+[Anaconda](https://www.anaconda.com/distribution/#download-section) (strongly recommended for OSX)
+or `pip`:
 
+<!--DOCUSAURUS_CODE_TABS-->
+<!--Conda-->
 ```bash
 conda install botorch -c pytorch -c gpytorch -c conda-forge
 ```
+<!--pip-->
+```bash
+pip install botorch
+```
+<!--END_DOCUSAURUS_CODE_TABS-->
 
-For more installation options and more detailed instructions, please see the
+For more installation options and detailed instructions, please see the
 [Project Readme](https://github.com/pytorch/botorch/blob/main/README.md)
 on GitHub.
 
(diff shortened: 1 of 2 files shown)


docs/getting_started.md
```diff
@@ -13,30 +13,14 @@ Before jumping the gun, we recommend you start with the high-level
...
-<!--DOCUSAURUS_CODE_TABS-->
-<!--conda-->
 ```bash
 conda install botorch -c pytorch -c gpytorch -c conda-forge
 ```
-<!--pip-->
-```bash
-pip install botorch
-```
-<!--END_DOCUSAURUS_CODE_TABS-->
```
Balandat, line 37: Hmm there are probably a good amount of folks that don't use anaconda (and maybe aren't able to), and so I'd be hesitant to remove the `pip install` from the getting started page here.

website/pages/en/index.js
```diff
@@ -164,10 +164,7 @@ candidate  # tensor([0.4887, 0.5063])
           <ol>
             <li>
               <h4>Install BoTorch:</h4>
-              <a>via conda (recommended):</a>
               <MarkdownBlock>{bash`conda install botorch -c pytorch -c gpytorch -c conda-forge`}</MarkdownBlock>
-              <a>via pip:</a>
-              <MarkdownBlock>{bash`pip install botorch`}</MarkdownBlock>
```
Balandat, line 170: similarly here, I think we should keep this.

docs/getting_started.md
```diff
@@ -13,30 +13,14 @@ Before jumping the gun, we recommend you start with the high-level
...
 #### Installation Requirements:
 
-- Python >= 3.8
-- PyTorch >= 1.11
-- gpytorch == 1.9.1
-- linear_operator == 0.3.0
-- scipy
-- multiple-dispatch
-- pyro-ppl >= 1.8.4
-
```
Balandat, line 23: Great, yeah let's kill this here.

Balandat (approved):
thanks!

Issue #1623, opened 2023-01-09
[Bug] Batch shape missing in TransformedPosterior


roussel-ryan:
# 🐛 Bug

I'm getting an error when using Bilog, Power transformations (which return TransformedPosterior objects) with qNoisyExpectedHypervolumeImprovement. Works fine with transforms that return GPyTorchPosterior objects.

## To reproduce

** Code snippet to reproduce **
```python
import torch
from botorch.acquisition.multi_objective import qNoisyExpectedHypervolumeImprovement
from botorch.models import SingleTaskGP, ModelList
from botorch.fit import fit_gpytorch_mll
from botorch.utils import standardize
from gpytorch.mlls import ExactMarginalLogLikelihood
from botorch.models.transforms.outcome import Bilog, Standardize, Log

train_X = torch.rand(10, 2)
Y = 1 - torch.norm(train_X - 0.5, dim=-1, keepdim=True)
Y = Y + 0.1 * torch.randn_like(Y)  # add some noise
train_Y = standardize(Y)

standard_transform = Standardize(m=1)
bilog_transform = Bilog()

# try each transform
for transform in [standard_transform, bilog_transform]:
    gp = SingleTaskGP(train_X, train_Y, outcome_transform=transform)
    mll = ExactMarginalLogLikelihood(gp.likelihood, gp)
    fit_gpytorch_mll(mll)

    ml = ModelList(gp,gp)

    acq = qNoisyExpectedHypervolumeImprovement(
        model=ml,
        X_baseline=train_X,
        cache_root=False,
        ref_point=torch.zeros(2)
    )

    test_x = torch.rand(10,1,2)
    print(f"{transform}: {acq(test_x)}")
```

** Stack trace/error message **
```
Traceback (most recent call last):
  File "C:\Users\Ryan Roussel\AppData\Roaming\JetBrains\PyCharm2022.3\scratches\scratch_19.py", line 26, in <module>
    acq = qNoisyExpectedHypervolumeImprovement(
  File "C:\ProgramData\Miniconda3\envs\xopt\lib\site-packages\botorch\acquisition\multi_objective\monte_carlo.py", line 515, in __init__
    self._set_cell_bounds(num_new_points=X_baseline.shape[0])
  File "C:\ProgramData\Miniconda3\envs\xopt\lib\site-packages\botorch\acquisition\multi_objective\monte_carlo.py", line 572, in _set_cell_bounds
[...]

PR #1624, opened 2023-01-09
Add sample_shape property to ListSampler

saitcakmak:
Summary:
In some cases, we access sampler.sample_shape, which errors out with ListSampler. This adds a property that returns the sample shape of the underlying samplers.

See https://github.com/pytorch/botorch/issues/1623#issuecomment-1376306859

Differential Revision: D42421646


Balandat (approved):
lgtm, thanks!

PR #1625, opened 2023-01-09
Fix TransformedPosterior missing batch shape error in _update_base_samples

saitcakmak:
Summary: See https://github.com/pytorch/botorch/issues/1623. This is the only use case for `posterior.batch_shape`, so fixing it locally makes sense to me.

Differential Revision: D42421494

### DIFF
--- a/test/acquisition/multi_objective/test_monte_carlo.py
+++ b/test/acquisition/multi_objective/test_monte_carlo.py
@@ -37,6 +37,9 @@ from botorch.models import (
 from botorch.models.gp_regression import SingleTaskGP
 from botorch.models.transforms.input import InputPerturbation
 from botorch.models.transforms.outcome import Standardize
+from botorch.posteriors.posterior_list import PosteriorList
+from botorch.posteriors.transformed import TransformedPosterior
+from botorch.sampling.list_sampler import ListSampler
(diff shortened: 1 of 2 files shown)


test/acquisition/multi_objective/test_monte_carlo.py
```diff
@@ -1624,3 +1627,28 @@ def get_acqf(model, matheron):
...
+                TransformedPosterior(
+                    MockPosterior(samples=torch.rand(2, 3, 1)), lambda X: X
+                ),
+            )
+        )
+        sampler = ListSampler(
+            IIDNormalSampler(sample_shape=torch.Size([2])),
+            IIDNormalSampler(sample_shape=torch.Size([2])),
+        )
+        # This calls _set_sampler which would previously fail.
```
Balandat, line 1647: previously w.r.t to what? This PR?

Balandat (approved):
thanks for the fix

test/acquisition/multi_objective/test_monte_carlo.py
saitcakmak: Yep

Balandat: Maybe let's update the comment with what the test does specifically rather than saying "previously"? The context will get lost.

Issue #1064, opened 2022-01-19
Non GP model types in Botorch


jduerholt:
Hi,

I was thinking about the possibility to use non GP models within botorch. For example to use a GP for one objective and a neural network (ensemble) for another one. Using just a neural network should already be possible via the `GenericDeterministcModel` https://github.com/pytorch/botorch/blob/f8da711049161f3dc238ada2890963b6c6dbb8ee/botorch/models/deterministic.py#L83 via just hooking in a neural network written in torch as callable `f`. 

In this case, uncertainty estimates from a NN ensemble could not be used. My idea was to implement a new type of `Posterior` that takes also the variance from an NN ensemble and return it as `variance` of the posterior. https://github.com/pytorch/botorch/blob/f8da711049161f3dc238ada2890963b6c6dbb8ee/botorch/posteriors/posterior.py#L56 

This should it already allow to use the whole botorch machinery of analytical acquisition functions. Of course this assumes that the posterior is normally distributed. If one then also implements the `rsample` method of the posterior, then one should also be able to use the MC acquisiton functions.

Do you see any obstacles in this?

I see the benefit in the possibility of using other model types in situations in which they perform better than GPs and do not have to reimplement the great machinere of acquisition functions and so forth, which are already available in botorch.

Best,

Johannes

wjmaddox:
Yes, this is pretty possible (I have some research code doing exactly deep ensemble posteriors). In general, the mean and variance can be calculated using sample mean and variance of the network's output producing a normal approximation to the "posterior", while sampling can be done by selecting a random item or items in the list of networks. 

The only gotcha is in dealing with multi-batched data as some classes of NNs don't like that in torch (I'm thinking things like RNNs and LSTMs on string inputs for example).

Balandat:
Yeah I did think about this when designing the APIs, so this should be possible without too much trouble (if not we should fix that). Basically, as long as you can have a posterior object that implements `rsample` and allows back-propagating gradients through samples that should work. Would love to see more use cases of this, actually.

jduerholt:
Thanks for the info. I will try to setup a simple MWE for this in the next month. Maybe I will come back with some questions then ;)

jduerholt:
Hi,

[...]

PR #1636, opened 2023-01-20
Ensemble Posterior

jduerholt:
## Motivation

As discussed in https://github.com/pytorch/botorch/issues/1064, this is an attempt to add a `EnsemblePosterior` to botorch, that could be used for example by NN ensembles.

I have problems with implementing `rsample` properly. I think my current implementation is not correct, it is based on `DeterministicPosterior`, but I think we should sample directly solutions from the individual predictions of the ensemble. But I do not know how to interprete `sample_shape` in this context.

As sampler, I registered the `StochasticSampler` for the new posterior class. But also, there I am not sure if this is correct. Furthermore, I have another question regarding `StochasticSampler`. It is stated in the docstring of `StochasticSampler` that it should not be used in combination with `optimize_acqf`. But `StochasticSampler` is assigned to the `DeterministicPosterior`. Does it mean that one cannot use a `ModelList` consisting of a `DeterministicModel` and GPs in combination with `optimize_acqf`? 

@Balandat: any suggestions on this?

Yes.

## Test Plan

Unit tests. Not yet implemented/finished as it is still WIP.

### DIFF
--- a/botorch/models/deterministic.py
+++ b/botorch/models/deterministic.py
@@ -56,10 +56,8 @@ class DeterministicModel(EnsembleModel):
         pass  # pragma: no cover
 
     def forward_(self, X: Tensor) -> Tensor:
-        # in case of the deterministic model we have to append the sample
-        # dimension for backward compatibility
-        values = self.forward(X=X)
-        return values.unsqueeze(-1)
+        r"""Compatilizes the `DeterministicModel` with `EnsemblePosterior`"""
+        return self.forward(X=X).unsqueeze(-3)
 
 
 class GenericDeterministicModel(DeterministicModel):
--- a/botorch/models/deterministic.py
+++ b/botorch/models/deterministic.py
@@ -27,13 +27,11 @@ functions or in other places where a `Model` is expected.
 from __future__ import annotations
 
 from abc import abstractmethod
-from typing import Any, Callable, List, Optional, Union
+from typing import Callable, List, Optional, Union
 
 import torch
-from botorch.acquisition.objective import PosteriorTransform
-from botorch.exceptions.errors import UnsupportedError
 from botorch.models.ensemble import EnsembleModel
-from botorch.posteriors.deterministic import DeterministicPosterior
+from botorch.models.model import Model
 from torch import Tensor
 
 
--- a/botorch/models/deterministic.py
+++ b/botorch/models/deterministic.py
@@ -56,7 +56,7 @@ class DeterministicModel(EnsembleModel):
         pass  # pragma: no cover
 
     def forward_(self, X: Tensor) -> Tensor:
-        # in case of the determinist model we have to append the sample
+        # in case of the deterministic model we have to append the sample
         # dimension for backward compatibility
         values = self.forward(X=X)
         return values.unsqueeze(-1)
--- a/botorch/models/deterministic.py
+++ b/botorch/models/deterministic.py
@@ -55,7 +55,7 @@ class DeterministicModel(EnsembleModel):
         """
         pass  # pragma: no cover
 
-    def forward_(self, X: Tensor) -> Tensor:
+    def _forward(self, X: Tensor) -> Tensor:
         r"""Compatibilizes the `DeterministicModel` with `EnsemblePosterior`"""
         return self.forward(X=X).unsqueeze(-3)
 
--- a/botorch/models/deterministic.py
+++ b/botorch/models/deterministic.py
@@ -25,19 +25,18 @@ functions or in other places where a `Model` is expected.
 """
 
 from __future__ import annotations
-
-from abc import ABC, abstractmethod
+from abc import abstractmethod
 from typing import Any, Callable, List, Optional, Union
 
 import torch
 from botorch.acquisition.objective import PosteriorTransform
 from botorch.exceptions.errors import UnsupportedError
-from botorch.models.model import Model
+from botorch.models.ensemble import EnsembleModel
 from botorch.posteriors.deterministic import DeterministicPosterior
 from torch import Tensor
 
 
-class DeterministicModel(Model, ABC):
+class DeterministicModel(EnsembleModel):
     r"""
     Abstract base class for deterministic models.
 
@@ -57,55 +56,11 @@ class DeterministicModel(Model, ABC):
         """
         pass  # pragma: no cover
 
-    @property
-    def num_outputs(self) -> int:
-        r"""The number of outputs of the model."""
-        return self._num_outputs
-
-    def posterior(
-        self,
-        X: Tensor,
-        output_indices: Optional[List[int]] = None,
-        posterior_transform: Optional[PosteriorTransform] = None,
-        **kwargs: Any,
-    ) -> DeterministicPosterior:
-        r"""Compute the (deterministic) posterior at X.
-
-        Args:
-            X: A `batch_shape x n x d`-dim input tensor `X`.
-            output_indices: A list of indices, corresponding to the outputs over
-                which to compute the posterior. If omitted, computes the posterior
-                over all model outputs.
-            posterior_transform: An optional PosteriorTransform.
-
-        Returns:
-            A `DeterministicPosterior` object, representing `batch_shape` joint
-            posteriors over `n` points and the outputs selected by `output_indices`.
-        """
-        # Apply the input transforms in `eval` mode.
-        self.eval()
-        X = self.transform_inputs(X)
-        # Note: we use a Tensor instance check so that `observation_noise = True`
-        # just gets ignored. This avoids having to do a bunch of case distinctions
-        # when using a ModelList.
-        if isinstance(kwargs.get("observation_noise"), Tensor):
-            # TODO: Consider returning an MVN here instead
-            raise UnsupportedError(
-                "Deterministic models do not support observation noise."
-            )
-        values = self.forward(X)
-        # NOTE: The `outcome_transform` `untransform`s the predictions rather than the
-        # `posterior` (as is done in GP models). This is more general since it works
-        # even if the transform doesn't support `untransform_posterior`.
-        if hasattr(self, "outcome_transform"):
-            values, _ = self.outcome_transform.untransform(values)
-        if output_indices is not None:
-            values = values[..., output_indices]
-        posterior = DeterministicPosterior(values=values)
-        if posterior_transform is not None:
-            return posterior_transform(posterior)
-        else:
-            return posterior
+    def forward_(self, X: Tensor) -> Tensor:
+        # in case of the determinist model we have to append the sample
+        # dimension for backward compatibility
+        values = self.forward(X=X)
+        return values.unsqueeze(-1)
 
 
 class GenericDeterministicModel(DeterministicModel):
--- a/botorch/models/deterministic.py
+++ b/botorch/models/deterministic.py
@@ -56,7 +56,7 @@ class DeterministicModel(EnsembleModel):
         pass  # pragma: no cover
 
     def forward_(self, X: Tensor) -> Tensor:
-        r"""Compatilizes the `DeterministicModel` with `EnsemblePosterior`"""
+        r"""Compatibilizes the `DeterministicModel` with `EnsemblePosterior`"""
         return self.forward(X=X).unsqueeze(-3)
 
 
--- a/botorch/models/deterministic.py
+++ b/botorch/models/deterministic.py
@@ -25,6 +25,7 @@ functions or in other places where a `Model` is expected.
 """
 
 from __future__ import annotations
+
 from abc import abstractmethod
 from typing import Any, Callable, List, Optional, Union
 
--- a/botorch/models/ensemble.py
+++ b/botorch/models/ensemble.py
@@ -1,0 +1,98 @@
+#!/usr/bin/env python3
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+#
+# This source code is licensed under the MIT license found in the
+# LICENSE file in the root directory of this source tree.
+
+r"""
+Ensemble Models: Simple wrappers that allow the usage of ensembles
+via the BoTorch Model and Posterior APIs.
+"""
+
+from __future__ import annotations
+
+from abc import ABC, abstractmethod
+from typing import Any, List, Optional
+
+from botorch.acquisition.objective import PosteriorTransform
+from botorch.exceptions.errors import UnsupportedError
+from botorch.models.model import Model
+from botorch.posteriors.ensemble import EnsemblePosterior
+from torch import Tensor
+
+
+class EnsembleModel(Model, ABC):
+    r"""
+    Abstract base class for ensemble models.
+
+    :meta private:
+    """
+
+    @abstractmethod
+    def forward(self, X: Tensor) -> Tensor:
+        r"""Compute the (ensemble) model output at X.
+
+        Args:
+            X: A `batch_shape x n x d`-dim input tensor `X`.
+
+        Returns:
+            A `batch_shape x n x m x s`-dimensional output tensor where
+            s is the size of the ensemble.
+        """
+        pass  # pragma: no cover
+
+    @property
+    def num_outputs(self) -> int:
+        r"""The number of outputs of the model."""
+        return self._num_outputs
+
+    @abstractmethod
+    @property
+    def size(self) -> int:
+        r"""The size of the ensemble model"""
+        pass
+
+    def posterior(
+        self,
+        X: Tensor,
+        output_indices: Optional[List[int]] = None,
+        posterior_transform: Optional[PosteriorTransform] = None,
+        **kwargs: Any,
+    ) -> EnsemblePosterior:
+        r"""Compute the ensemble posterior at X.
+
+        Args:
+            X: A `batch_shape x n x d`-dim input tensor `X`.
+            output_indices: A list of indices, corresponding to the outputs over
+                which to compute the posterior. If omitted, computes the posterior
+                over all model outputs.
+            posterior_transform: An optional PosteriorTransform.
+
+        Returns:
+            A `EnsemblePosterior` object, representing `batch_shape` joint
+            posteriors over `n` points and the outputs selected by `output_indices`.
+        """
+        # Apply the input transforms in `eval` mode.
+        self.eval()
+        X = self.transform_inputs(X)
+        # Note: we use a Tensor instance check so that `observation_noise = True`
+        # just gets ignored. This avoids having to do a bunch of case distinctions
+        # when using a ModelList.
+        if isinstance(kwargs.get("observation_noise"), Tensor):
+            # TODO: Consider returning an MVN here instead
+            raise UnsupportedError(
+                "Deterministic models do not support observation noise."
+            )
+        values = self.forward(X)
+        # NOTE: The `outcome_transform` `untransform`s the predictions rather than the
+        # `posterior` (as is done in GP models). This is more general since it works
+        # even if the transform doesn't support `untransform_posterior`.
+        if hasattr(self, "outcome_transform"):
+            values, _ = self.outcome_transform.untransform(values)
+        if output_indices is not None:
+            values = values[..., output_indices, :]
+        posterior = EnsemblePosterior(values=values)
+        if posterior_transform is not None:
+            return posterior_transform(posterior)
+        else:
+            return posterior
--- a/botorch/models/ensemble.py
+++ b/botorch/models/ensemble.py
@@ -78,7 +78,7 @@ class EnsembleModel(Model, ABC):
         if isinstance(kwargs.get("observation_noise"), Tensor):
             # TODO: Consider returning an MVN here instead
             raise UnsupportedError(
-                "Deterministic models do not support observation noise."
+                "Ensemble models do not support observation noise."
             )
         values = self._forward(X)
         # NOTE: The `outcome_transform` `untransform`s the predictions rather than the
--- a/botorch/models/ensemble.py
+++ b/botorch/models/ensemble.py
@@ -77,9 +77,7 @@ class EnsembleModel(Model, ABC):
         # when using a ModelList.
         if isinstance(kwargs.get("observation_noise"), Tensor):
             # TODO: Consider returning an MVN here instead
-            raise UnsupportedError(
-                "Ensemble models do not support observation noise."
-            )
+            raise UnsupportedError("Ensemble models do not support observation noise.")
         values = self._forward(X)
         # NOTE: The `outcome_transform` `untransform`s the predictions rather than the
         # `posterior` (as is done in GP models). This is more general since it works
--- a/botorch/models/ensemble.py
+++ b/botorch/models/ensemble.py
@@ -41,6 +41,9 @@ class EnsembleModel(Model, ABC):
         """
         pass  # pragma: no cover
 
+    def forward_(self, X: Tensor) -> Tensor:
+        return self.forward(X=X)
+
     @property
     def num_outputs(self) -> int:
         r"""The number of outputs of the model."""
@@ -77,7 +80,7 @@ class EnsembleModel(Model, ABC):
             raise UnsupportedError(
                 "Deterministic models do not support observation noise."
             )
-        values = self.forward(X)
+        values = self.forward_(X)
         # NOTE: The `outcome_transform` `untransform`s the predictions rather than the
         # `posterior` (as is done in GP models). This is more general since it works
         # even if the transform doesn't support `untransform_posterior`.
--- a/botorch/models/ensemble.py
+++ b/botorch/models/ensemble.py
@@ -59,7 +59,7 @@ class EnsembleModel(Model, ABC):
         r"""Compute the ensemble posterior at X.
 
         Args:
-            X: A `batch_shape x n x d`-dim input tensor `X`.
+            X: A `batch_shape x q x d`-dim input tensor `X`.
             output_indices: A list of indices, corresponding to the outputs over
                 which to compute the posterior. If omitted, computes the posterior
                 over all model outputs.
--- a/botorch/models/ensemble.py
+++ b/botorch/models/ensemble.py
@@ -37,7 +37,7 @@ class EnsembleModel(Model, ABC):
 
         Returns:
             A `batch_shape x n x m x s`-dimensional output tensor where
-            s is the size of the ensemble.
+            `s` is the size of the ensemble.
         """
         pass  # pragma: no cover
 
--- a/botorch/models/ensemble.py
+++ b/botorch/models/ensemble.py
@@ -36,7 +36,7 @@ class EnsembleModel(Model, ABC):
             X: A `batch_shape x n x d`-dim input tensor `X`.
 
         Returns:
-            A `batch_shape x n x m x s`-dimensional output tensor where
+            A `batch_shape x s x n x m`-dimensional output tensor where
             `s` is the size of the ensemble.
         """
         pass  # pragma: no cover
@@ -87,7 +87,7 @@ class EnsembleModel(Model, ABC):
         if hasattr(self, "outcome_transform"):
             values, _ = self.outcome_transform.untransform(values)
         if output_indices is not None:
-            values = values[..., output_indices, :]
+            values = values[..., output_indices]
         posterior = EnsemblePosterior(values=values)
         if posterior_transform is not None:
             return posterior_transform(posterior)
--- a/botorch/models/ensemble.py
+++ b/botorch/models/ensemble.py
@@ -46,12 +46,6 @@ class EnsembleModel(Model, ABC):
         r"""The number of outputs of the model."""
         return self._num_outputs
 
-    @abstractmethod
-    @property
-    def size(self) -> int:
-        r"""The size of the ensemble model"""
-        pass
-
     def posterior(
         self,
         X: Tensor,
--- a/botorch/models/ensemble.py
+++ b/botorch/models/ensemble.py
@@ -41,7 +41,7 @@ class EnsembleModel(Model, ABC):
         """
         pass  # pragma: no cover
 
-    def forward_(self, X: Tensor) -> Tensor:
+    def _forward(self, X: Tensor) -> Tensor:
         return self.forward(X=X)
 
     @property
@@ -80,7 +80,7 @@ class EnsembleModel(Model, ABC):
             raise UnsupportedError(
                 "Deterministic models do not support observation noise."
             )
-        values = self.forward_(X)
+        values = self._forward(X)
         # NOTE: The `outcome_transform` `untransform`s the predictions rather than the
         # `posterior` (as is done in GP models). This is more general since it works
         # even if the transform doesn't support `untransform_posterior`.
--- a/botorch/models/ensemble.py
+++ b/botorch/models/ensemble.py
@@ -69,7 +69,7 @@ class EnsembleModel(Model, ABC):
             posterior_transform: An optional PosteriorTransform.
 
         Returns:
-            A `EnsemblePosterior` object, representing `batch_shape` joint
+            An `EnsemblePosterior` object, representing `batch_shape` joint
             posteriors over `n` points and the outputs selected by `output_indices`.
         """
         # Apply the input transforms in `eval` mode.
--- a/botorch/posteriors/ensemble.py
+++ b/botorch/posteriors/ensemble.py
@@ -111,4 +111,4 @@ class EnsemblePosterior(Posterior):
     ) -> Tensor:
         if base_samples.shape != sample_shape:
             raise ValueError("Base samples to not match sample shape.")
-        return self.values[..., base_samples].reshape(self._extended_shape)
+        return self.values[..., base_samples].reshape(self._extended_shape())
--- a/botorch/posteriors/ensemble.py
+++ b/botorch/posteriors/ensemble.py
@@ -30,7 +30,7 @@ class EnsemblePosterior(Posterior):
         r"""
         Args:
             values: Values of the samples produced by this posterior as
-                a `(b) x q x m x s` tensor where `m` is the output size of the
+                a `(b) x s x q x m` tensor where `m` is the output size of the
                 model and `s` is the ensemble size.
         """
         if values.ndim < 3:
@@ -40,7 +40,7 @@ class EnsemblePosterior(Posterior):
     @property
     def size(self) -> int:
         r"""The size of the ensemble"""
-        return self.values.shape[-1]
+        return self.values.shape[-3]
 
     @property
     def weights(self) -> Tensor:
@@ -61,7 +61,7 @@ class EnsemblePosterior(Posterior):
     @property
     def mean(self) -> Tensor:
         r"""The mean of the posterior as a `(b) x n x m`-dim Tensor."""
-        return self.values.mean(dim=-1)
+        return self.values.mean(dim=-3)
 
     @property
     def variance(self) -> Tensor:
@@ -70,8 +70,8 @@ class EnsemblePosterior(Posterior):
         Computed as the sample variance across the ensemble outputs.
         """
         if self.size == 1:
-            return torch.zeros_like(self.values[..., -1])
-        return self.values.var(dim=-1)
+            return torch.zeros_like(self.values.squeeze(-3))
+        return self.values.var(dim=-3)
 
     def _extended_shape(
         self, sample_shape: torch.Size = torch.Size()  # noqa: B008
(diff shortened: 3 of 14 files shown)


Balandat:
Hmm so one high level question I have here is why the `EnsemblePosterior` operates of a finite set of values rather than keeping a reference to the model and then evaluating that in the way @wjmaddox's code did here: https://github.com/samuelstanton/lambo/blob/dec2f303d847efefc4392ebfd2d78226d543388c/lambo/models/deep_ensemble.py#L177

Is the idea to just predict the output for all models in the model ensemble and then randomly sample from that (rather than re-evaluating a subset of the ensemble models)? I guess that seems reasonable if the ensemble models are explicitly defined and we don't have any scalability issues. But if the ensemble models are lazily defined by some parameters and only actually built upon evaluation this could cause issues. 

Assuming we're in the case where we can easily do this, then I think there may be some confusion about `sample_shape`. Basically this being an ensemble of (I take it determinstic) models, the number of different samples that can possibly be drawn here is finite (`s` in your case). So it seems to me that `rsample(sample_shape)` should essentially grab `sample_shape.numel()` elements from your `values` tensor and then arrange it appropriately, sth like this: 

```
indcs = torch.randint(values.shape[-1], sample_shape)
samples = values[..., indcs]
samples = samples.permute(
    *range(-1, -len(sample_shape) - 1, -1),
    *range(len(values) - 1)
)
```

> It is stated in the docstring of StochasticSampler that it should not be used in combination with optimize_acqf. But StochasticSampler is assigned to the DeterministicPosterior. Does it mean that one cannot use a ModelList consisting of a DeterministicModel and GPs in combination with optimize_acqf?

So `DeterministicPosterior` isn't really doing anything, it's mostly just a crutch to make sure that deterministic "models" can be used in the same way as other probabilistic models. So using `StochasticSampler` is just out of convenience here. I guess the proper thing would be to have a `DeterminsticSampler` that doesn't do anything either, but I guess we wanted to avoid inflation of dummy classes here? cc @saitcakmak. 

The point about not using a stochastic sampler with `optimize_acqf` is that by default we use optimization algorithms that assume a deterministic function and break in bad ways if that's not the case. But in this special case of a `DeterministicPosterior` there is actually no randomness and the used `StochasticSampler` does indeed produce deterministic values so all is good.  

For this ensemble posterior I do think it makes sense to use `StochasticSampler`. You'll probably want to  add the `EnsemblePosterior` to this check here though: https://github.com/pytorch/botorch/blob/main/botorch/sampling/stochastic_samplers.py#L80

Now if you do want to use deterministic optimization algorithms the idea would essentially be to fix the models that are used to generate the "samples". This would mean that the "base samples" are essentially the indices that you'd use to index into the `values`- these would stay fixed on the sampler, but a new `model.posterior(X)` call would generate a new `values` at a new `X` ( and so if the indices are fixed across evaluations and the model is differentiable then the samples will be differentiable w.r.t. `X`). 

Does this make sense?

jduerholt:
Hi @Balandat,

thanks for your detailed answer. This is just my first attempt to integrate an `EnsemblePosterior` into botorch, so I am very happy about your comments and open to changes of the concept.

To your high level question:

I saw @wjmaddox code, but I programmed it with the intention that we compute the output of all models first and then sample from this values as you described, as for example also a Random Forest is doing it. I have not thought about what you described about lazily defined models and/or scalability issues. 

So what is your recommendation here? Which solution would you prefer? What I like about my proposal is the similarity to the other models and posteriors in botorch (I also added an abstract `EnsembleModel` class.). A model has a `posterior` method which calls the `forward` method in which inference is performed.

Thanks for your explanation for the `StochasticSampler` and the `DeterministicModel`. I was just wondering about the doctstring there. Maybe we should add there just a sentence regarding the `DeterministicModel` as exception.

My intention for the `EnsemblePosterior` is to be able to use it with deterministic optimization algorithms. So which sampler should one register for it to have the `base_samples` as indices that are fixed for consecutive calls? In case one wants to use the `EnsemblePosterior` with stochastic optimizer, one could then register another sampler.

If we would follow my initial idea, the abstract `EnsembleModel` class would be very similar to the `DeterministicModel` class. Only difference is in principle just the different posterior class. So one should thing about introducing a new base class from both `DeterministicModel` and `EnsembleModel` are inheriting from that implements an abstract method `_posterior` which returns the respective posterior objects. 

Best,

Johannes

Balandat:
> So what is your recommendation here? Which solution would you prefer? What I like about my proposal is the similarity to the other models and posteriors in botorch (I also added an abstract EnsembleModel class.). A model has a posterior method which calls the forward method in which inference is performed.

Yes we definitely want to retain this API pattern. But one could do this in the two different ways I mentioned above. 

Your version makes sense if there is a full ensemble of models of reasonably small ensemble size so that computing the values across all models is not a big deal and we can always do that. Then we could have the sampler just pick out a random set of ensemble items. Of course if they're all computed already and most of the work has been done it stands to reason why we would actually sample in the first place rather than just use all computed outputs. 

The other idea would be to have the ensemble model be a model where the `posterior` call just produces an object that keeps a reference back to the model, similar to Wesley's implementation. Then sampling would mean selecting a subset of models from the ensemble to compute the output when sampling from the posterior. That is, models would be evaluated in a lazy fashion as requested by the sampler. This would have the benefit of avoiding wasting computation on evaluating all models in the ensemble model even if their outputs are never used.

> Thanks for your explanation for the StochasticSampler and the DeterministicModel. I was just wondering about the doctstring there. Maybe we should add there just a sentence regarding the DeterministicModel as exception.

I wonder if we should just be very clear here by implementing a dummy sampler that just evaluates `DeterministicModel`s rather than overloading the meaning of `StochasticSampler`. Curious about @saitcakmak's thoughts on this. 

> My intention for the EnsemblePosterior is to be able to use it with deterministic optimization algorithms. So which sampler should one register for it to have the base_samples as indices that are fixed for consecutive calls? In case one wants to use the EnsemblePosterior with stochastic optimizer, one could then register another sampler.

For deterministic optimization we'd probably want to implement something like an `EnsembleSampler` or `IndexSampler`. This would store the base samples as indices into the list of available samples and then index into the posterior whenever called on a `Posterior` object.

jduerholt:
So one Todo would be definitely to implement this `IndexSampler`. 

Concerning the two possible design patterns you outlined above, I am open to both, but the more I think about Wesleys implementation, the more I see its benefits. So I will change the PR accordingly and we decide to follow this design pattern. 

Fine for you?

Balandat:
I don't think it necessarily has to be either / or. One could have two different types of posterior, one that just contains a set of samples, and one that keeps the reference to the model and performs lazy evaluation. So we could start with whatever is easier / the more immediate need.

jduerholt:
Ok, I will then just proceed with what I have currently and focus on implementing the `IndexSampler` and `rsample` correctly. Both is also needed for the other version. I will also change the default shape of `values` from batch_shape x q x  m x sample_shape to 
sample_shape x batch_shape x q x m, so that it is consistent with the rest of the code, as for example here: https://github.com/pytorch/botorch/blob/3596b125ffad5482eb113ef55f32375ec66ff0ec/botorch/acquisition/objective.py#L280

It could be that there will be more questions over the course of the week ;-)

jduerholt:
Here comes already the first question:

In the docstring of the abstract `rsample` method it is defined that `sample_shape` is a `torch.Size` object which specifies the number of samples and the number of batches to sample. Here it has to be different, or?

The `IndexSampler` would not pass a `torch.Size` object to `rsample` but a `torch.Tensor` of `dtype=int64`, holding the indices to sample for each batch. Is this new meaning of `sample_shape` in `rsample` ok?

Or should we implement `rsample` as you described above and in the usual way and implement a method `rsample_by_indices` for the `EnsemblePosterior` which is then called by the `IndexSampler` instead of the conventional `rsample`?

Best,

Johannes

Balandat:
> The IndexSampler would not pass a torch.Size object to rsample but a torch.Tensor of dtype=int64, holding the indices to sample for each batch. Is this new meaning of sample_shape in rsample ok?

We can consider the index samples produced by the sampler as "base samples". Check out how the Normal samplers do this where the base samples are iid Normal, they call a `rsample_from_base_samples` method defined on the posterior here: https://github.com/pytorch/botorch/blob/main/botorch/sampling/normal.py#L43-L49. You'd implement a similar method on the `EnsemblePosterior`. In `EnsemblePosterior.rsample` you can then just generate those samples (if not using a `Sampler`) and call `rsample_from_base_samples`, similar to what the `GPyTorchPosterior` does here: https://github.com/pytorch/botorch/blob/main/botorch/posteriors/gpytorch.py#L180-L182

saitcakmak:
Sorry about the confusion around `StochasticSampler` and `DeterministicModel`. I was just avoiding creating a dummy class that's identical to the `StochasticSampler`. I'll add one to avoid any confusion going forward. Otherwise, @Balandat's explanation above is on point.

Regarding an `IndexSampler`, I agree that it'd be best to just use the indices as the `base_samples` and keep the API consistent by defining a `rsample_from_base_samples` that uses these index `base_samples`. The `rsample` can then simply sample `sample_shape` indices and pass these to `rsample_from_base_samples` as the `base_samples`. 

The motivation behind `rsample` and `rsample_from_base_samples` is to keep the `rsample` API consistent with the `rsample` of `torch.Distribution`, which produces differentiable but not necessarily deterministic samples; and use `rsample_from_base_samples` to ensure that the samples are deterministic and can be optimized with a gradient based deterministic optimizer, such as the L-BFGS-B used in `optimize_acqf`.

jduerholt:
Hi @saitcakmak and @Balandat,

I just added the `rsample_from_base_samples` (including an update of `rsample`) and the `IndexSampler`. Perhaps, you could have a look and tell me if this looks as you were describing it above.

Note, I kept the shape of `values` as `batch_shape x q x m x posterior_size`. For this reason, the sampling is conducted over the last dimension, followed by a reshape.

Best,

Johannes

botorch/sampling/index_sampler.py
```diff
@@ -0,0 +1,35 @@
...
+
+class IndexSampler(MCSampler):
+    def forward(self, posterior: Posterior) -> Tensor:
+        self._construct_base_samples(posterior=posterior)
+        samples = posterior.rsample_from_base_samples(
+            sample_shape=self.sample_shape, base_samples=self.base_samples
+        )
+        return samples
+
+    def _construct_base_samples(self, posterior: Posterior):
```
Balandat, line 23:
```suggestion
    def _construct_base_samples(self, posterior: Posterior) -> None:
```

Balandat (commented):
Yup, this lgtm! I think one thing we could consider is have the ensemble dimension be between the batch shape and the q dimension - that feels a bit more natural to me as usually trailing dimensions are outcome dimensions throughout the code base. But really I'm not sure if it makes a difference.

jduerholt:
> I think one thing we could consider is have the ensemble dimension be between the batch shape and the q dimension - that feels a bit more natural to me as usually trailing dimensions are outcome dimensions throughout the code base. But really I'm not sure if it makes a difference.

I am also not sure there, so far I have not changed it as no solution appears really preferable to me here. But if you want I can changed it. The tests for `IndexSampler` and `EnsemblePosterior` are now finished. Open question for me is the `EnsembleModel`, which you also touch above. Currently `EnsembleModel` and `DeterministicModel` is the same despite one line, namely when the `Posterior` object is created. I would advocate for having a shared base model from which both `EnsembleModel` and `DeterministicModel` inherit from. This base model would have an abstract method `_posterior` which is called during the actual call to `posterior` and in which the `Posterior` object is created depending if is an `EnsembleModel` or an `DeterministicModel`. 

If you are fine with this proposal, I can implement it, but I would need a name for this BaseModel. I am not very creative at this. We could also postpone this discussion and I take the `EnsembleModel` out of the PR and we merge first the `EnsemblePosterior` and the the `IndexSampler`. Both are now ready for review.

botorch/models/ensemble.py
```diff
@@ -0,0 +1,98 @@
...
+        self.eval()
+        X = self.transform_inputs(X)
+        # Note: we use a Tensor instance check so that `observation_noise = True`
+        # just gets ignored. This avoids having to do a bunch of case distinctions
+        # when using a ModelList.
+        if isinstance(kwargs.get("observation_noise"), Tensor):
+            # TODO: Consider returning an MVN here instead
+            raise UnsupportedError(
+                "Deterministic models do not support observation noise."
+            )
```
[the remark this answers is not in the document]
jduerholt: I do not feel competent enough to answer this question. I commented below regarding the similarity between the current version of `EnsembleModel` and `DeterministicModel`.

jduerholt:
An alternative would be to morphe the `DeterministicModel` into the `EnsembleModel`, as the deterministic model is just an `EnsembleModel` of ensemble size one. We could assign in both cases the `EnsemblePosterior` and just have one model for both use cases. What do you think?

jduerholt:
Ok, then I will do this. But it could be that I do not find time to finish it before the weekend. A few questions:

- I would propose to have `EnsembleModel` and `EnsemblePosterior`. And then implement for backwards compatibility a `DeterministicModel` which just inherits from `EnsembleModel` and always appends the ensemble dimension to its `values`. Do we then still need a `DeterministicPosterior`? It is just an `EnsemblePosterior` and we use as sampler just the `IndexSampler` and remove the `DeterministicSampler` again. Where do we need there backwards compatibility?
- Concerning lazy evaluation: I do not have currently a use case for it. My use cases are the following:
    - Small NN ensembles up to 10 MLPs as done here https://www.biorxiv.org/content/10.1101/2022.08.05.502972v2.full
    - Being able to compute ACQFs over random forests (I know that this then not anymore optimizable in a gradient based fashion). But there are alternatives like just random sampling, GAs or this approach https://arxiv.org/abs/1705.10883
    - Maybe fully bayesian models where I have a posterior distribution over parameters based on MCMC sampling, and I preselect a subset of the orginal MCMC samples and continue with this ensemble
    - In all cases, I do not see a real need for lazy evaluation as, I would like to always compute all model outputs to have my output distribution as representative as possible. Looking at the code from @samuelstanton  and his `EnsemblePosterior`, I have the feeling that the same also applies to him.

What do you think?

Balandat:
That makes sense to me! As to the backward compatibility of the posterior, I’d want to take a look if this is used in other work downstream. I think for now we can probably operate under the assumption that we don’t need it. 

The fully Bayesian model case is interesting - we have the FullyBayesianPosterior that does exactly that. I guess one could have an abstract EnaemblePosterior that just represents a mixture distribution from which we could derive that, but maybe that also seems a bit much to implement at this point. 

So I’m happy with your proposal.

jduerholt:
Great, I will implement as soon as I find time for it. Regarding the `FullyBayesianPosterior`: It does not implement `rsample` etc., from my perspective, one could also use the `EnsemblePosterior` as `FullyBayesianPosterior`. For example if I have a GLM and the `forward` function of my model gives me the predictive distribution (ensemble of predictions), I can put it into the `EnsemblePosterior` and use it for deterministic optimization with any MCMC sampled model.

jduerholt:
> That's a larger change though and I suggest we don't consider this as part of this PR.

Ok, I understand and fully agree. I will first finish this one as discussed and we can continue with this.

jduerholt:
I have now rewritten it in a way that `DeterministicModel` inherits from `EnsembleModel` and delivers also a `EnsemblePosterior`. To maintain backward compatibility, I added therefor a `forward_` method that just appends the sample dimension in the `deterministic` case. This works.

But a new problem occured. For deterministic models the `IndexSampler` is  now used and this sampler implements `rsample_from_base_samples`, we also need a `_update_base_samples` method in `IndexSampler`. What is the purpose of this method and how should it look like for `IndexSampler`? I am quite unsure here.

Balandat:
> But a new problem occured. For deterministic models the IndexSampler is now used and this sampler implements rsample_from_base_samples, we also need a _update_base_samples method in IndexSampler. What is the purpose of this method and how should it look like for IndexSampler? I am quite unsure here.

This is a bit in the weeds of the implementation. This method is used for Normal samplers, when we are caching the decomposition of the posterior MVN in the `CachedCholeskyAcquisitionFunciton`, where we need to ensure that we are reusing the right base samples. An implementation is here: https://github.com/pytorch/botorch/blob/b3d3074e7f0a8142a86ec5406b4f6553aba4e769/botorch/sampling/base.py#L160. I'm not particularly happy with this kind of special handling making it into all other samplers where this isn't really needed, maybe there is a better way to do this (cc @sdaulton)? 

I think you should be able to just leave it as a nullop for the `IndexSampler`.

jduerholt:
> I think you should be able to just leave it as a nullop for the `IndexSampler`.

Thx for the hint. I just implemented as now as nullop and it works. If the implementation is fine for you as it is now, I would then add the missing tests. Could you have a look especially at `EnsembleModel` and `DeterministicModel`?

Should I then also remove `DeterministicPosterior` and `DeterministicSampler` as they are not anymore needed, or do you want to keep them?

botorch/sampling/index_sampler.py
```diff
@@ -0,0 +1,62 @@
...
+            posterior (Posterior): The ensemble posterior to construct the base samples
+                for.
+        """
+        if self.base_samples is None or self.base_samples.shape != self.sample_shape:
+            with manual_seed(seed=self.seed):
+                base_samples = torch.multinomial(
+                    posterior.weights,
+                    num_samples=self.sample_shape.numel(),
+                    replacement=True,
+                ).reshape(self.sample_shape)
```
Balandat, line 52: I think the `manual_seed` setup is probably a bit outdated and could use an overhaul, let's do this instead
```suggestion
            with torch.random.fork_rng():
                torch.manual_seed(self.seed)
                base_samples = torch.multinomial(
                    posterior.weights,
                    num_samples=self.sample_shape.numel(),
                    replacement=True,
                ).reshape(self.sample_shape)
```

botorch/sampling/index_sampler.py
```diff
@@ -0,0 +1,62 @@
...
+            self.register_buffer("base_samples", base_samples)
+        if self.base_samples.device != posterior.device:
+            self.to(device=posterior.device)  # pragma: nocover
+
+    def _update_base_samples(
+        self, posterior: Posterior, base_sampler: MCSampler
+    ) -> None:
+        r"""Null operation just needed for compatibility with
+        `CachedCholeskyAcquisitionFunction`."""
+        pass
```
Balandat, line 62: I don't know if you actually need to implement this method, e.g. `StochasticSampler` does not.

Balandat (commented):
Thanks, overall this loos great to me. Curious about @saitcakmak's and @sdaulton's take. 

> Should I then also remove DeterministicPosterior and DeterministicSampler as they are not anymore needed, or do you want to keep them?

Let's keep those around for backward-compatibility reasons (e.g. we have some. other code that is not open source that uses these, and I wouldn't be surprised if other people used these directly as well somehow). We should issue a `DeprecationWarning` upon instantiation of either of those.

botorch/models/deterministic.py
```diff
@@ -57,55 +55,11 @@ def forward(self, X: Tensor) -> Tensor:
...
-            values = values[..., output_indices]
-        posterior = DeterministicPosterior(values=values)
-        if posterior_transform is not None:
-            return posterior_transform(posterior)
-        else:
-            return posterior
+    def forward_(self, X: Tensor) -> Tensor:
+        # in case of the determinist model we have to append the sample
+        # dimension for backward compatibility
+        values = self.forward(X=X)
```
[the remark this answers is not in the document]
jduerholt: We need it here to append in case of the `DeterministicModel` the sampling dimension to keep backwards compatibility. You see it in the next line.

botorch/sampling/index_sampler.py
jduerholt: We need it, as discussed in this comment: https://github.com/pytorch/botorch/pull/1636#issuecomment-1414022584. Else the `CachedCholeskyAcquisitionFunciton`s will fail.

botorch/posteriors/ensemble.py
```diff
@@ -0,0 +1,143 @@
...
+from botorch.posteriors.posterior import Posterior
+from torch import Tensor
+
+r"""
+Posterior module to be used with ensemble models.
+"""
+
+
+class EnsemblePosterior(Posterior):
+    r"""Ensemble posterior."""
```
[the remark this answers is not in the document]
jduerholt: Done.

jduerholt:
> I think one thing we could consider is have the ensemble dimension be between the batch shape and the q dimension - that feels a bit more natural to me as usually trailing dimensions are outcome dimensions throughout the code base. But really I'm not sure if it makes a difference.

I thought again about this old comment from you, and I think it should be done as you said, especially as we push the values in the `posterior` call in `EnsembleModel` though the `outcome_transform` before instantiating the `EnsemblePosterior`. I can expect problems there if the last dimension is then not the output dimension. For this reason I will change it from `batch_shape x q x m x s` to `barch_shape x s x q x m` as you suggested.

jduerholt:
Okay, I think it is done ;) I added the missing tests. It is ready for review now.

botorch/posteriors/ensemble.py
```diff
@@ -0,0 +1,147 @@
...
+
+    def _extended_shape(
+        self, sample_shape: torch.Size = torch.Size()  # noqa: B008
+    ) -> torch.Size:
+        r"""Returns the shape of the samples produced by the posterior with
+        the given `sample_shape`.
+        """
+        mask = self.values.ndim * [True]
+        mask[-3] = False
+        return sample_shape + tuple(torch.tensor(self.values.shape)[mask].tolist())
```
Balandat, line 84:
```suggestion
        return sample_shape + self.values.shape[:1] + self.values.shape[-2:]
```

botorch/posteriors/ensemble.py
```diff
@@ -0,0 +1,147 @@
...
+            values: Values of the samples produced by this posterior as
+                a `(b) x s x q x m` tensor where `m` is the output size of the
+                model and `s` is the ensemble size.
+        """
+        if values.ndim < 3:
+            raise ValueError("Values has to be at least three-dimensional.")
+        self.values = values
+
+    @property
+    def size(self) -> int:
```
Balandat, line 41: let's call this sth like `ensemble_size` to be unambiguous; `size()` is typically a alias for `shape`

botorch/models/deterministic.py
```diff
@@ -57,55 +55,9 @@ def forward(self, X: Tensor) -> Tensor:
...
-            values, _ = self.outcome_transform.untransform(values)
-        if output_indices is not None:
-            values = values[..., output_indices]
-        posterior = DeterministicPosterior(values=values)
-        if posterior_transform is not None:
-            return posterior_transform(posterior)
-        else:
-            return posterior
+    def forward_(self, X: Tensor) -> Tensor:
+        r"""Compatilizes the `DeterministicModel` with `EnsemblePosterior`"""
```
Balandat, line 59:
```suggestion
        r"""Compatibilizes the `DeterministicModel` with `EnsemblePosterior`"""
```

botorch/models/ensemble.py
```diff
@@ -0,0 +1,95 @@
...
+            X: A `batch_shape x n x d`-dim input tensor `X`.
+
+        Returns:
+            A `batch_shape x s x n x m`-dimensional output tensor where
+            `s` is the size of the ensemble.
+        """
+        pass  # pragma: no cover
+
+    def forward_(self, X: Tensor) -> Tensor:
+        return self.forward(X=X)
```
Balandat, line 45: Let's maybe call this `_forward` to keep it private. Also, if we agree that we want to phase out `DeterminsticModel` and instead just always use `EnsembleModel`s with ensemble size 1 (a question we still need to answer - @saitcakmak, @dme65, @sdaulton any thoughts here), then we should probably add a TODO to remove this method once `DeterminsticModel` is gone.

Balandat (commented):
Many thanks for the continued work on this. A few more minor points, o/w this looks great. I'd like to get some more input from other folks who have been working on the transforms more than I have more recently to make sure these changes cause any unforeseen issues.

botorch/models/ensemble.py
jduerholt: Done, I changed it according to your suggestion.

test/models/test_ensemble.py
```diff
@@ -0,0 +1,37 @@
...
+
+
+class DummyEnsembleModel(EnsembleModel):
+    r"""A dummy ensemble model."""
+
+    def __init__(self):
+        r"""Init model."""
+        super().__init__()
+        self._num_outputs = 2
+        self.a = torch.rand(16, 3, 2)
```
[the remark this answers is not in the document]
jduerholt: Done ;)

botorch/posteriors/ensemble.py
jduerholt: Done.

botorch/posteriors/ensemble.py
jduerholt: This will not work, as the batch dim is optional. Consider first this case:
```
self.values = torch.randn(4,16,5,2)
```
Here the batch size is 4 and the ensemble size is 16. In the second part of the sum above we want to filter out the ensemble dimension and `self.values.shape[:1] + self.values.shape[-2:]` would correctly return `torch.Size([4, 5, 2])`.

But now consider the case that no batch dimension is provided which is allowed:
```
self.values = torch.randn(16,5,2)
```
Here again the ensemble size is 16, but the expression from above would return `torch.Size([16, 5, 2])` which is wrong, it should return `torch.Size([5, 2])`. For this reason, I went for the rather complicated solution from above and also originally preferred having the ensemble size as last dimension, but this is for other reasons not preferable.

jduerholt:
> A few more minor points, o/w this looks great. I'd like to get some more input from other folks who have been working on the transforms more than I have more recently to make sure these changes cause any unforeseen issues.

I changed it according to your points. The one thing, I could not change accordingly is explained above. I am looking forward to the comments of the other folks and hope to get it merged soon ;)

botorch/posteriors/ensemble.py
Balandat: Got it, but `self.values.shape[:-3] + self.values.shape[-2:]` should work, right?

jduerholt: Yes, thx for the hint. "Manchmal sieht man den Wald vor lauter Bäumen nicht" ;)

Balandat: Yeah the classic "sometimes see man the forest before noisy trees not"

botorch/posteriors/ensemble.py
```diff
@@ -0,0 +1,145 @@
...
+                `IndexSampler`. This is used for deterministic optimization. The
+                predictions of the ensemble corresponding to the indices are then
+                sampled.
+
+        Returns:
+            Samples from the posterior, a tensor of shape
+            `self._extended_shape(sample_shape=sample_shape)`.
+        """
+        if base_samples.shape != sample_shape:
+            raise ValueError("Base samples to not match sample shape.")
```
saitcakmak, line 141:
```suggestion
            raise ValueError("Base samples do not match sample shape.")
```

botorch/posteriors/ensemble.py
```diff
@@ -0,0 +1,145 @@
...
+
+        This is intended to be used with a sampler that produces the corresponding base
+        samples, and enables acquisition optimization via Sample Average Approximation.
+
+        Args:
+            sample_shape: A `torch.Size` object specifying the sample shape. To
+                draw `n` samples, set to `torch.Size([n])`. To draw `b` batches
+                of `n` samples each, set to `torch.Size([b, n])`.
+            base_samples: A Tensor of indices as base samples of shape
+                `sample_shape x base_sample_shape`, typically obtained from
```
saitcakmak, line 131: `base_sample_shape` is not defined.

botorch/sampling/index_sampler.py
```diff
@@ -0,0 +1,62 @@
...
+        """
+        self._construct_base_samples(posterior=posterior)
+        samples = posterior.rsample_from_base_samples(
+            sample_shape=self.sample_shape, base_samples=self.base_samples
+        )
+        return samples
+
+    def _construct_base_samples(self, posterior: Posterior) -> None:
+        r"""Constructs base samples as indices to sample with them from
+        the Posterior
```
saitcakmak, line 39:
```suggestion
        the Posterior.
```

botorch/sampling/index_sampler.py
```diff
@@ -0,0 +1,62 @@
...
+from botorch.posteriors import Posterior
+from botorch.sampling.base import MCSampler
+from torch import Tensor
+
+
+class IndexSampler(MCSampler):
+    r"""A sampler that calls `posterior.rsample_from_base_samples` to
+    generate the samples via index base samples."""
+
+    def forward(self, posterior: Posterior) -> Tensor:
```
saitcakmak, line 22: Could type this as `EnsemblePosterior`.
```suggestion
    def forward(self, posterior: EnsemblePosterior) -> Tensor:
```

botorch/sampling/index_sampler.py
```diff
@@ -0,0 +1,62 @@
...
+        Returns:
+            The samples drawn from the posterior.
+        """
+        self._construct_base_samples(posterior=posterior)
+        samples = posterior.rsample_from_base_samples(
+            sample_shape=self.sample_shape, base_samples=self.base_samples
+        )
+        return samples
+
+    def _construct_base_samples(self, posterior: Posterior) -> None:
```
saitcakmak, line 37:
```suggestion
    def _construct_base_samples(self, posterior: EnsemblePosterior) -> None:
```

botorch/sampling/index_sampler.py
```diff
@@ -0,0 +1,62 @@
...
+# This source code is licensed under the MIT license found in the
+# LICENSE file in the root directory of this source tree.
+
+r"""
+Sampler to be used with `EnsemblePosteriors` to enable
+deterministic optimization of acquisition functions with ensemble models.
+"""
+
+import torch
+from botorch.posteriors import Posterior
```
saitcakmak, line 13:
```suggestion
from botorch.posteriors.ensemble import EnsemblePosterior
```

botorch/sampling/index_sampler.py
```diff
@@ -0,0 +1,62 @@
...
+                    posterior.weights,
+                    num_samples=self.sample_shape.numel(),
+                    replacement=True,
+                ).reshape(self.sample_shape)
+            self.register_buffer("base_samples", base_samples)
+        if self.base_samples.device != posterior.device:
+            self.to(device=posterior.device)  # pragma: nocover
+
+    def _update_base_samples(
+        self, posterior: Posterior, base_sampler: MCSampler
```
saitcakmak, line 58:
```suggestion
        self, posterior: EnsemblePosterior, base_sampler: IndexSampler
```

botorch/sampling/index_sampler.py
```diff
@@ -0,0 +1,62 @@
...
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+#
+# This source code is licensed under the MIT license found in the
+# LICENSE file in the root directory of this source tree.
+
+r"""
+Sampler to be used with `EnsemblePosteriors` to enable
+deterministic optimization of acquisition functions with ensemble models.
+"""
+
```
saitcakmak, line 11:
```suggestion

from __future__ import annotations

```

botorch/sampling/index_sampler.py
saitcakmak: `StochasticSampler` will error out on `_update_base_samples`. The reasoning is that it should never be used with `CachedCholeskyAcquisitionFunction` when `cache_root=True`.

botorch/sampling/get_sampler.py
```diff
@@ -115,6 +117,14 @@ def _get_sampler_deterministic(
     return DeterministicSampler(sample_shape=sample_shape, **kwargs)
```
saitcakmak, line 117: If the `DeterministicPosterior` is deprecated, should we deprecate the `DeterministicSampler` as well?

botorch/models/ensemble.py
```diff
@@ -0,0 +1,95 @@
...
+        self,
+        X: Tensor,
+        output_indices: Optional[List[int]] = None,
+        posterior_transform: Optional[PosteriorTransform] = None,
+        **kwargs: Any,
+    ) -> EnsemblePosterior:
+        r"""Compute the ensemble posterior at X.
+
+        Args:
+            X: A `batch_shape x n x d`-dim input tensor `X`.
```
saitcakmak, line 62: For consistency, we refer to -2 dim of the model input as q-batch dim.
```suggestion
            X: A `batch_shape x q x d`-dim input tensor `X`.
```

botorch/models/ensemble.py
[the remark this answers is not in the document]
saitcakmak: Technically, everything is an ensemble of a single model, unless it is an ensemble of multiple models. I don't see any immediate benefit from such a refactor, so I'd say lets revisit this if we end up with more ensemble models in the future.

botorch/models/ensemble.py
```diff
@@ -0,0 +1,95 @@
...
+        # Apply the input transforms in `eval` mode.
+        self.eval()
+        X = self.transform_inputs(X)
+        # Note: we use a Tensor instance check so that `observation_noise = True`
+        # just gets ignored. This avoids having to do a bunch of case distinctions
+        # when using a ModelList.
+        if isinstance(kwargs.get("observation_noise"), Tensor):
+            # TODO: Consider returning an MVN here instead
+            raise UnsupportedError(
+                "Deterministic models do not support observation noise."
```
saitcakmak, line 81:
```suggestion
                "Ensemble models do not support observation noise."
```

saitcakmak (approved):
Thanks for implementing this. A bunch of nits in line, lgtm otherwise.

botorch/sampling/get_sampler.py
saitcakmak: I guess it hasn't even been around long enough to be deprecated. We could just delete it if it is not needed anymore.

jduerholt: So, I remove it and register again the `StochasticSampler` for the deprecated `DeterministicPosterior` or what do you recommend?

botorch/posteriors/ensemble.py
jduerholt: fixed it.

jduerholt:
@saitcakmak : thanks for your review. I changed the PR according to your suggestions. I only let you a question for the `DeterministicSampler`.

botorch/sampling/get_sampler.py
saitcakmak: Since `DeterministicModel` is now an `EnsembleModel`, wouldn't the `IndexSampler` be sufficient there (already dispatched for `EnsembleModel`). If that works, you can just delete `DeterministicSampler` and call it a day. If it leads to any issues, it's ok to just leave it as is.

saitcakmak:
Hmm, codecov shows missing coverage for a line that's just a comment (https://app.codecov.io/gh/pytorch/botorch/pull/1636#597cdf37e0736ca95369f0ff0f74eff4-R244). I'll re-run the tests to see if it disappears.

saitcakmak:
I think something is wrong with codecov on that one. It's a file not touched by PR, the line is a comment and this diff doesn't remove any tests. I wouldn't worry about it

botorch/sampling/get_sampler.py
jduerholt: Yes, but the `DeterministicPosterior` is still around (but not used anymore in `DeterministicModel`), we just marked it as deprecated, but it should still have a default sampler or? The question is, which one? For this reason, I proposed to revert back to `StochasticSampler`.

jduerholt: @Balandat @saitcakmak: What do we do here? Just leave it in or should I remove it? Then the question is still what will be the default sampler of the deprecated `DeterministicPosterior`? For me it is ok, to just keep it until the `DeterministicPosterior` will be removed.

saitcakmak: What happens if you use the `IndexSampler` with `DeterministicPosterior`? If that works without any issues, then we don't need the `DeterministicSampler` and we can just delete it (and the dispatch rule associated with it). If it leads to any issues, let's keep it for now.

jduerholt: This does not work as the `IndexSampler` works with `rsample_from_base_samples` which is not implemented in the case of the `DeterministicPosterior`. So I would recommend to keep it and remove it togehter with the `DeterministicPosterior` in one of the next releases.

saitcakmak: Sounds good, let's do that

jduerholt: Then it should be done now ;)

Balandat: Cool. I'll just add some deprecation notes to the docstrings then before merging this in.

PR #1641, opened 2023-01-24
Introduce a `DeterministicSampler`

saitcakmak:
Summary: ... to sanction its use with `optimize_acqf`.

Differential Revision: D42700043


Balandat (approved):
It seems a bit weird to derive a `DeterministicSampler` from a `StochasticSampler` and call a method `rsample` that doesn't actually sample. But I guess that's fine.

Issue #1635, opened 2023-01-20
[Bug] AffineInputTransform causes retain_graph errors when previous transforms have parameters


mrcslws:
# 🐛 Bug

If you use an AffineInputTransform (like `Normalize` or `InputStandardize`) after an input transform with learned parameters, bad things happen. The AffineInputTransform will unintentionally store values that have autograd compute graphs, because this class was not written to expect `X` to have gradient information during training.

This has two effects:
1. During inference, when calling `loss.backward()` to optimize X, there are retain_graph errors.
2. It needlessly slows the model down, as pytorch will compute gradients for AffineInputTransform's buffers during `optimize_acqf`

This bug will occur any time `X` has gradient information during training, so there may be other user scenarios where this happens.

## To reproduce

```python
import botorch
import gpytorch
import torch

print(f"botorch: {botorch.__version__}")
print(f"gpytorch: {gpytorch.__version__}")
print(f"torch: {torch.__version__}")

train_X = torch.tensor([[0., 0.],
                        [0.1, 0.1],
                        [0.5, 0.5]])
train_Y = torch.tensor([[0.1],
                        [0.5],
                        [1.0]])
test_X = torch.tensor([[0.3, 0.3]])


# Workaround: change this to True
use_workaround = False
if use_workaround:
    class PatchedNormalize(botorch.models.transforms.Normalize):
        """
        Fix a bug in botorch's Normalize
        """
        def train(self, mode=True):
            if not mode:
                # During training, previous transforms have caused X to have a
                # backprop compute graph, which is then propagated into the
                # coefficient and offset.
                self.coefficient = self.coefficient.detach()
                self.offset = self.offset.detach()

            super().train(mode)

    normalize = PatchedNormalize(2)
else:
    normalize = botorch.models.transforms.Normalize(2)


input_transform = botorch.models.transforms.ChainedInputTransform(
    # Example input transform with learned parameters
    warp=botorch.models.transforms.Warp(indices=[0, 1]),
    normalize=normalize,
)

model = botorch.models.SingleTaskGP(
    train_X, train_Y,
    input_transform=input_transform
)

botorch.fit_gpytorch_mll(
    gpytorch.mlls.ExactMarginalLogLikelihood(model.likelihood, model)
)

posterior = model.posterior(test_X)

# Silly placeholder loss function; bug occurs with any loss function
loss = -(posterior.mean - posterior.variance).sum()
loss.backward()
```

**Error message:**

[...]

PR #1642, opened 2023-01-24
Detach coefficient and offset in AffineTransform

saitcakmak:
Summary: See https://github.com/pytorch/botorch/issues/1635. These should only retain grad if learning the coefficient and while in train mode.

Differential Revision: D42700421


mrcslws:
See https://github.com/pytorch/botorch/issues/1635#issuecomment-1402078557 for an explanation of why I think gradient information *should* flow through `coefficient` and `offset` during training.

PR #1644, opened 2023-01-25
Fix pickle error in TorchPosterior

### LINKED ISSUE #1639: [Bug] posteriors.TorchPosterior can't be torch.load-ed
# 🐛 Bug

`botorch.posteriors.TorchPosterior` doesn't support `torch.load`. Loading it throws an error.

Use case: Previously (a few months ago), during cross validation it was possible to `torch.save` the results and analyze them later. Now an error occurs when you try to `torch.load` a saved result.

## To reproduce

```python
import botorch
import torch

posterior = botorch.posteriors.TorchPosterior(
    torch.distributions.Normal(0.0, 1.0)
)

torch.save(posterior, "my-posterior.pt")
torch.load("my-posterior.pt")
```

```
  File "/Users/mrcslws/miniforge3/lib/python3.9/site-packages/torch/serialization.py", line 789, in load
    return _load(opened_zipfile, map_location, pickle_module, **pickle_load_args)
  File "/Users/mrcslws/miniforge3/lib/python3.9/site-packages/torch/serialization.py", line 1131, in _load
    result = unpickler.load()
  File "/Users/mrcslws/dev/src/botorch/botorch/posteriors/torch.py", line 79, in __getattr__
    return getattr(self.distribution, name)
  File "/Users/mrcslws/dev/src/botorch/botorch/posteriors/torch.py", line 79, in __getattr__
    return getattr(self.distribution, name)
  File "/Users/mrcslws/dev/src/botorch/botorch/posteriors/torch.py", line 79, in __getattr__
    return getattr(self.distribution, name)
  [Previous line repeated 993 more times]
RecursionError: maximum recursion depth exceeded
```

## System information

botorch: 0.7.3.dev22+g208470e7.d20221102
gpytorch: 1.9.1.dev32+g23b068b5
torch: 1.13.1
Mac OS 13.1 M1

saitcakmak:
## Motivation

Fixes #1639. The issue was that pickle uses `__getstate__` and `__setstate__` to save / load the objects. In `TorchPosterior`, `__getattr__` was taking precedence, so instead of calling `TorchPosterior.__set/getstate__` it was calling `self.distribution.__set/getstate__`. While loading the posterior, we start with a `TorchPosterior` that does not have any attributes --including `distribution`--, so calling `self.distribution` was leading to an infinite loop of calling `TorchPosterior` with `name=distribution` (since it calls `getattr(self.distribution, name)`).

## Test Plan

Units.

### DIFF
--- a/botorch/posteriors/torch.py
+++ b/botorch/posteriors/torch.py
@@ -84,7 +84,7 @@ class TorchPosterior(Posterior):
         Pickle uses `__get/setstate__` to serialize / deserialize the objects.
         Since we define `__getattr__` above, it takes precedence over these
         methods, and we end up in an infinite loop unless we also define
-        `__getattr__` and `__setattr__`.
+        `__getstate__` and `__setstate__`.
         """
         return self.__dict__
 
--- a/botorch/posteriors/torch.py
+++ b/botorch/posteriors/torch.py
@@ -79,7 +79,13 @@ class TorchPosterior(Posterior):
         return getattr(self.distribution, name)
 
     def __getstate__(self) -> Dict[str, Any]:
-        r"""A minimal utility to support pickle protocol."""
+        r"""A minimal utility to support pickle protocol.
+
+        Pickle uses `__get/setstate__` to serialize / deserialize the objects.
+        Since we define `__getattr__` above, it takes precedence over these
+        methods, and we end up in an infinite loop unless we also define
+        `__getattr__` and `__setattr__`.
+        """
         return self.__dict__
 
     def __setstate__(self, d: Dict[str, Any]) -> None:
--- a/botorch/posteriors/torch.py
+++ b/botorch/posteriors/torch.py
@@ -10,7 +10,7 @@ Posterior module to be used with PyTorch distributions.
 
 from __future__ import annotations
 
-from typing import Any, Optional
+from typing import Any, Dict, Optional
 
 import torch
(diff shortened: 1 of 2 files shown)


botorch/posteriors/torch.py
```diff
@@ -78,6 +78,14 @@ def __getattr__(self, name: str) -> Any:
         """
         return getattr(self.distribution, name)
 
+    def __getstate__(self) -> Dict[str, Any]:
+        r"""A minimal utility to support pickle protocol."""
```
Balandat, line 82: My be worth explaining in a comment in code why we need this (b/c of the definition of `__getattr__` on this class)

Balandat (approved):
Thanks for the fix!

botorch/posteriors/torch.py
```diff
@@ -78,6 +78,20 @@ def __getattr__(self, name: str) -> Any:
         """
         return getattr(self.distribution, name)
 
+    def __getstate__(self) -> Dict[str, Any]:
+        r"""A minimal utility to support pickle protocol.
+
+        Pickle uses `__get/setstate__` to serialize / deserialize the objects.
+        Since we define `__getattr__` above, it takes precedence over these
+        methods, and we end up in an infinite loop unless we also define
+        `__getattr__` and `__setattr__`.
```
saitcakmak, line 87: typo
```suggestion
        `__getstate__` and `__setstate__`.
```

PR #1652, opened 2023-02-03
Inducing Point Allocators for Sparse GPs

henrymoss:
## Motivation

As requested by @eytan, I have had a go at implementing BO-specific inducing point allocation (IPA) strategies from my new paper (https://arxiv.org/abs/2301.10123). These allow you to build sparse Gaussian processes that are somewhat customized to the particular task at hand, be it BO, active learning e.t.c. 

1. I have rewritten part of your approximate GP code to allow the specification of custom IPA strategies. 
2.  BoTorch's existing behaviour (using the greedy variance reduction of Burt et al (2020)) is a special case of the new functionality and remains the default.
3. I have included one of the new IPAs from my paper to help demonstrate how custom IPAs can be defined. The one I chose is quite complicated (requiring access to model from the previous BO step) and so being able to implement it was a good test case to check that this code for custom IPA initialisation isnt stupid.
4. There ended up being a lot of IPA related code, so I moved it all into a utility file, with its own testing file.

## Test Plan

I have added loads of tests in test_inducing_point_allocators.py. I also added some extra tests higher up in the code in test_approximate_gp.py. There is perhaps a little bit of redundancy between the two sets of tests, but they are testing different interfaces.

## Related PRs

I will write a follow-up PR with a notebook demonstrating this new inducing point allocation functionality and showing how a user can define their own IPA.


botorch/acquisition/max_value_entropy_search.py
```diff
@@ -13,8 +13,8 @@
 
 .. [Moss2021gibbon]
     Moss, H. B., et al.,
-    GIBBON: General-purpose Information-Based Bayesian OptimisatioN
-    arXiv:2102.03324, 2021
+    GIBBON: General-purpose Information-Based Bayesian OptimisatioN.
+    Journal of Machine Learning Research, 2021.
```
henrymoss, line 17: Took the opportunity to update this ;)

botorch/models/approximate_gp.py
```diff
@@ -13,29 +13,36 @@
...
     Artificial Intelligence, 2013, https://arxiv.org/abs/1309.6835.
 
+.. [moss2023ipa]
+    Henry B. Moss and Sebastian W. Ober and Victor Picheny,
+    Inducing Point Allocation for Sparse Gaussian Processes
+    in High-Throughput Bayesian Optimization,Proceedings of
+    the 25th International Conference on Artificial Intelligence
+    and Statistics, 2023, https://arxiv.org/pdf/2301.10123.pdf.
+
 """
```
henrymoss, line 28: I moved the Chen reference to the new file that deals with all the DPP sampling

esantorella:
Thanks so much for the PR! I've imported it to Meta-internal and someone should be reviewing it soon.

To fix the Sphinx documentation error, could you list the new module [here](https://github.com/pytorch/botorch/blob/main/sphinx/source/models.rst)?

botorch/models/utils/inducing_point_allocators.py
```diff
@@ -0,0 +1,388 @@
...
+            A tensor of quality scores for each input, of shape [n]
+        """
+
+        posterior = self._model.posterior(inputs)
+        mean = posterior.mean.squeeze(-2).squeeze(-1)  # removing redundant dimensions
+        sigma = posterior.variance.clamp_min(1e-12).sqrt().view(mean.shape)
+
+        best_f = torch.max(mean) if self._maximize else torch.min(mean)
+        u = (mean - best_f) / sigma if self._maximize else -(mean - best_f) / sigma
+        return sigma * (phi(u) + u * Phi(u))
```
Balandat, line 234: I wonder if we can reuse some of the numerics from @SebastianAment's recent work here (see e.g. here: https://github.com/pytorch/botorch/blob/main/botorch/acquisition/analytic.py#L174-L176). I guess this isn't logEI here and gradients aren't really of interest here, but could still make sense to leverage the EI computation from somewhere rather than re-implementing it here.

botorch/models/utils/inducing_point_allocators.py
```diff
@@ -0,0 +1,388 @@
...
+
+
+class QualityFunction(ABC):
+    """
+    A function that scores inputs with respect
+    to a specific criterion.
+    """
+
+    @abstractmethod
+    def __call__(inputs: Tensor) -> Tensor:  # [n, d] -> [n]
```
Balandat, line 166: is there a `self` missing here?

botorch/models/utils/inducing_point_allocators.py
```diff
@@ -0,0 +1,388 @@
...
+        self,
+        inputs: Tensor,
+        covar_module: Module,
+        num_inducing: int,
+        input_batch_shape: torch.Size,
+        quality_function: Callable[[Tensor], Tensor],
+    ) -> Tensor:
+        r"""
+        Private method to allow inducing point allocators to support
+        multi-task models and models with batched inputs.
```
Balandat, line 86: It looks like the only thing that happens in the `allocate_inducing_points` functions of the various classes is to instantiate the right quality function. Howe about just making an abstract `_get_quality_function` method that does what its name suggests and just call this as the first thing in `allocate_inducing_points` (which then doesn't need to be abstract anymore)?

botorch/models/utils/inducing_point_allocators.py
```diff
@@ -0,0 +1,388 @@
...
+
+
+class InducingPointAllocator(ABC):
+    r"""
+    This class provides functionality to initialize the inducing point locations
+    of an inducing point-based model, e.g. a `SingleTaskVariationalGP`.
+    """
+
+    @abstractmethod
+    def allocate_inducing_points(
```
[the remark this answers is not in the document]
Balandat: also, is there a `self` missing here or did you mean for this to be a static method or class method?

botorch/models/utils/inducing_point_allocators.py
```diff
@@ -0,0 +1,388 @@
...
+        return self._allocate_inducing_points(
+            inputs,
+            covar_module,
+            num_inducing,
+            input_batch_shape,
+            ExpectedImprovementQualityFunction(self._model, self._maximize),
+        )
+
+
+def _pivoted_cholesky_init(
```
Balandat, line 317: This is essentially a clone of https://github.com/pytorch/botorch/blob/d026e805f34bcffc9003720f1822e06f92ec08ab/botorch/models/approximate_gp.py#L493. Would you mind
1. changing the function here so that `quality_scores` is an optional arg (in which case they effectively would be a unit vector but we don't really need to do that in the implementation)
2. Modifying the original function to just call this one while raising a `DeprecationWarning` pointing to this function instead?

FYI: @j-wilson has implemented some pivoted cholesky functionality as well (e.g. here https://github.com/pytorch/botorch/blob/d026e805f34bcffc9003720f1822e06f92ec08ab/botorch/utils/probability/linalg.py#L74) - it may be worth thinking how that could be used here (though I'm not suggesting to make this part of this PR, that should stay separate).

botorch/models/utils/inducing_point_allocators.py
```diff
@@ -0,0 +1,388 @@
...
+    set of points that are highly diverse (according to the provided kernel_matrix)
+    and have high quality (according to the provided quality_scores).
+
+    Args:
+        train_inputs: training inputs (of shape n x d)
+        kernel_matrix: kernel matrix on the training
+            inputs
+        max_length: number of inducing points to initialize
+        quality_scores: scores representing the quality of each candidate
+            input (of shape [n])
```
Balandat, line 341: more nits
```suggestion
        train_inputs: training inputs (of shape n x d).
        kernel_matrix: kernel matrix on the training inputs.
        max_length: number of inducing points to initialize.
        quality_scores: scores representing the quality of each candidate
            input (of shape [n]).
```

test/models/test_approximate_gp.py
```diff
@@ -181,6 +185,25 @@ def test_initializations(self):
...
+        # test that can specify an inducing point allocator
+        for ipa in [
+            GreedyVarianceReduction(),
+            GreedyImprovementReduction(model, maximize=True),
+        ]:
+            model = SingleTaskVariationalGP(train_X, inducing_point_allocator=ipa)
+            self.assertTrue(type(model._inducing_point_allocator), type(ipa))
+
+        # test warning when learning on and custom IPA provided
+        with self.assertWarns(UserWarning):
```
Balandat, line 200:
```suggestion
        with self.assertWarnsRegex(UserWarning, r"set `learn_inducing_points` to False"):
```

test/models/test_approximate_gp.py
```diff
@@ -193,6 +216,8 @@ def test_inducing_point_init(self):
         model_2 = SingleTaskVariationalGP(train_X=train_X_2, inducing_points=5)
         model_2_inducing = model_2.model.variational_strategy.inducing_points
 
+        self.assertTrue(model_1_inducing.shape == (5, 1))
+        self.assertTrue(model_2_inducing.shape == (5, 1))
```
Balandat, line 220: can use equal here (and below)
```suggestion
        self.assertEqual(model_1_inducing.shape, (5, 1))
        self.assertEqual(model_2_inducing.shape, (5, 1))
```

Balandat (commented):
Thanks for the contribution @henrymoss, this is excellent. 

At a high level this all looks good to me, I have two main points, one is about the design of the instanstiation of the quality function for `_allocate_inducing_points` handled, the other one about avoiding code duplication in `_pivoted_cholesky_init`. The rest are mostly nits.

botorch/models/utils/inducing_point_allocators.py
henrymoss: So I actually move the old _pivoted_cholesky_init from approximate_gp.py into this file and deleted the old code. This new one is used for both the old (greedy variance reduction) and the new IPAs.

Do you want me to add the old one back into approximate_gp.py?

botorch/models/utils/inducing_point_allocators.py
```diff
@@ -0,0 +1,388 @@
...
+        raise ValueError(
+            "_pivoted_cholesky_init requires a quality score for each of train_inputs"
+        )
+
+    item_size = kernel_matrix.shape[-2]
+    cis = torch.zeros(
+        (max_length, item_size), device=kernel_matrix.device, dtype=kernel_matrix.dtype
+    )
+    di2s = kernel_matrix.diag()
+    scores = di2s * (quality_scores**2)
```
esantorella, line 366: When any of the tests [here](https://github.com/pytorch/botorch/blob/d026e805f34bcffc9003720f1822e06f92ec08ab/test/sampling/pathwise/test_posterior_samplers.py#L32) is run on a GPU, this generates 
```
RuntimeError: Expected all tensors to be on the same device, but found at least two devices, cuda:0 and cpu!
```

botorch/models/approximate_gp.py
```diff
@@ -417,131 +448,3 @@ def init_inducing_points(
             var_strat.variational_params_initialized.fill_(0)
 
         return inducing_points
-
-
-def _select_inducing_points(
```
esantorella, line 422: Deleting `_select_inducing_points` is creating an import error for a downstream library ([aepsych](https://github.com/facebookresearch/aepsych)). Can `_select_inducing_points` either call one of the new functions or raise a deprecation warning?

botorch/models/utils/inducing_point_allocators.py
Balandat: Oh, I must have missed that somehow, nvm.

botorch/models/approximate_gp.py
Balandat: I think we probably don't have to get too crazy in terms of deprecations here since this is a "private" method - @crasanders could we just update aepsych to use the new and shiny initialization strategies here?

botorch/models/utils/inducing_point_allocators.py
```diff
@@ -0,0 +1,388 @@
...
+
+import torch
+from botorch.models.model import Model
+
+from botorch.utils.probability.utils import ndtr as Phi, phi
+from gpytorch.module import Module
+from linear_operator.operators import LinearOperator
+from torch import Tensor
+
+NEG_INF = -(torch.tensor(float("inf")))
```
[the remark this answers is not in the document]
henrymoss: Why is this?

botorch/models/utils/inducing_point_allocators.py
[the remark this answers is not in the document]
henrymoss: Allocate gave a better acronym (IPA), but also means more sense it the setting where you have a single model that you update (and reallocate inducing points for) during each BO step, i.e. not in the BoTorch setting where you seem to always build a new model from scratch each BO step

botorch/models/utils/inducing_point_allocators.py
henrymoss: I had lots of circular import issues. Do you have a specific request for the function you want me to use?

botorch/models/approximate_gp.py
henrymoss: What did you two decide?

botorch/models/utils/inducing_point_allocators.py
henrymoss: how do I fix this?

botorch/models/utils/inducing_point_allocators.py
henrymoss: Great idea! I've updated it and its much much neater

botorch/models/utils/inducing_point_allocators.py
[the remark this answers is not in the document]
SebastianAment: A `float` just requires 32-64 bits in stack-allocated memory, so even `torch.tensor(-float("inf"))` would 1. store the float corresponding to `inf`, 2. negate it in place, since the variable is not used anywhere else, 3. allocate a `Tensor` object, which is generally heap-allocated and keeps track of a lot more information than a `float`: `requires_grad`, `grad`, `device`, `dtype`, etc.

`-(torch.tensor(float("inf")))` on the other hand, allocates two `Tensor` objects: First by calling `torch.tensor(float("inf"))` and subsequently by negating the first `Tensor`, which allocates a second `Tensor` object with all the extraneous variables. Clearly this would never be a bottleneck here, but in similar places with larger tensors, this could halve the memory footprint.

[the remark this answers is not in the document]
henrymoss: got it! Thanks for explaining

henrymoss:
> Thanks so much for the PR! I've imported it to Meta-internal and someone should be reviewing it soon.
> 
> To fix the Sphinx documentation error, could you list the new module [here](https://github.com/pytorch/botorch/blob/main/sphinx/source/models.rst)?

Hi @esantorella, I tried to do this but made things worse

botorch/models/utils/inducing_point_allocators.py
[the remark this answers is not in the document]
Balandat: It's always about the acronym, isn't it? This one I can get behind though, cheers.

botorch/models/utils/inducing_point_allocators.py
SebastianAment: There's the [`_log_ei_helper`](https://github.com/pytorch/botorch/blob/a8efd76b43c19982d5d0b79f14912b27094b5d93/botorch/acquisition/analytic.py#L836), which computes `log(phi(u) + u * Phi(u))` in a numerically accurate and stable fashion. Even if you don't need gradients, it could be worth returning `LogEI` here instead because it will be able to differentiate (no pun intended) between points whose naïve `EI` is numerically - but not mathematically - zero.

There's also [`_scaled_improvement`](https://github.com/pytorch/botorch/blob/a8efd76b43c19982d5d0b79f14912b27094b5d93/botorch/acquisition/analytic.py#L821), which computes `u`, and [`_ei_helper`](https://github.com/pytorch/botorch/blob/a8efd76b43c19982d5d0b79f14912b27094b5d93/botorch/acquisition/analytic.py#L829), which computes `phi(u) + u * Phi(u)` naïvely. While it wouldn't change the functionality, using the existing helpers would allow future improvements to translate automatically to this code.

Balandat: Hmm I think it's ok for now; this isn't that crazy. Once we merge all of @SebastianAment's stuff in we can probably take another more global look at how to reuse those in other places.

SebastianAment: If you get import errors for the helpers, it could make sense to move `_log_ei_helper` / `_ei_helper` to the utils. I can do that and adjust the other dependencies, just let me know!

botorch/models/utils/inducing_point_allocators.py
Balandat: You'll need to make sure the tensors here are on the same device; you could (and probably should) do this here with a `to()` call. But ideally this could already happen upstream where `_pivoted_cholesky_init` is called so that the inputs are already on the same device.

I think the reason this is happening here is that in `_allocate_inducing_points` the `covar_module` and `inputs` don't live on the device in the tests.

sphinx/source/models.rst
```diff
@@ -165,6 +165,11 @@ Model Conversion
 .. automodule:: botorch.models.converter
     :members:
 
+Inducing Point Allocators
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+.. automodule:: botorch.models.utils.InducingPointAllocator
```
Balandat, line 170: This needs to be the module name not the class name
```suggestion
.. automodule:: botorch.models.utils.inducing_point_allocators
```

henrymoss: sorry, stupid me

botorch/models/utils/inducing_point_allocators.py
henrymoss: yeah that would be great pls @SebastianAment

botorch/models/utils/inducing_point_allocators.py
henrymoss: so is it a quirk of the tests or an error in my code? Is it not fair to assume that `covar_module` and `inputs` are on device?

botorch/models/approximate_gp.py
```diff
@@ -179,6 +185,9 @@ def __init__(
                 VariationalStrategy). The default setting uses "whitening" of the
                 variational distribution to make training easier.
             inducing_points: The number or specific locations of the inducing points.
+            inducing_point_allocator: The `InducingPointAllocator` used to
+            initialize the inducing point locations. If omitted,
+            uses `GreedyVarianceReduction`.
```
Balandat, line 190:
```suggestion
                initialize the inducing point locations. If omitted,
                uses `GreedyVarianceReduction`.
```

test/models/test_approximate_gp.py
```diff
@@ -223,12 +252,60 @@ def test_inducing_point_init(self):
...
+    def test_custom_inducing_point_init(self):
+        train_X_0 = torch.rand(15, 1, device=self.device)
+        train_X_1 = torch.rand(15, 1, device=self.device)
+        train_X_2 = torch.rand(15, 1, device=self.device)
+        train_X_3 = torch.rand(15, 1, device=self.device)
+
+        # check error if multi-output
+        # check error if batched inputs
+
+        # todo
```
Balandat, line 273: just flagging this here

henrymoss: woops. Will delete

sphinx/source/models.rst
```diff
@@ -165,6 +165,11 @@ Model Conversion
 .. automodule:: botorch.models.converter
     :members:
 
+Inducing Point Allocators
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+.. automodule:: botorch.models.utils.inducing_point_allocator
```
Balandat, line 170:
```suggestion
.. automodule:: botorch.models.utils.inducing_point_allocators
```

esantorella:
> > Thanks so much for the PR! I've imported it to Meta-internal and someone should be reviewing it soon.
> > To fix the Sphinx documentation error, could you list the new module [here](https://github.com/pytorch/botorch/blob/main/sphinx/source/models.rst)?
> 
> Hi @esantorella, I tried to do this but made things worse

Thanks! Looks like there are additionally some citation-related issues. I'll clean those up and make a PR into this branch.

botorch/models/utils/inducing_point_allocators.py
esantorella: The test seems to be passing on a GPU now.

henrymoss:
I think we are ready @Balandat !

botorch/models/approximate_gp.py
Balandat: We'll fix the aepsych issue ourselves and then will merge those changes together. @esantorella could you take care of this, please?

Balandat:
Most excellent, thanks for all the cleaning up. We'll just fix the aepsych failure on our end and then we can merge this in!

botorch/models/approximate_gp.py
esantorella: Yup I'll put in a PR to aepsych

PR #1658, opened 2023-02-07
testing

esantorella:
## Motivation

We are recording whether the tutorials failed, how long they took, and how much memory they used in the logs. However, we are not using this data effectively because it's hard to access systematically. With this PR, we will
- write output from each tutorial that runs in the nightly cron to a csv
- upload the csv to a folder in the repo

## Test Plan

Use the artifacts to prototype a dashboard


esantorella:
Closing in favor of #1659

Issue #1673, opened 2023-02-14
Raise torch lower version bound to `>=1.13.1` (CVE-2022-45907)


LGro:
There is a known security vulnerability (CVE-2022-45907) in `torch <= 1.13.0` which is patched since `1.13.1` (pytorch/pytorch/issues/89855). However, the torch lower version bound in botorch's [current requirements](https://github.com/pytorch/botorch/blob/main/requirements.txt#L3) still seems to be `torch>=1.11`.

What are your thoughts about raising this lower version bound to `torch>=1.13.1` to promote using only dependencies without known security vulnerabilities with the current `botorch` releases?

Balandat:
Hmm `1.13.1` is the most recent release and I'm somewhat hesitant to seize supporting at least the `1.12` series. But wer could use a version exclusion of the form `torch != 1.13.0` to make sure we disallow that specific version.

Balandat:
Oh I misread the vulnerability, this is actually present in all versions, not just 1.13.0, my bad.

saitcakmak:
This seems like a relatively minor security vulnerability. From the original [pytorch issue](https://github.com/pytorch/pytorch/issues/88868#issuecomment-1338252251) on this:

> I believe the vulnerability is, if someone crafts a malicious Python file, and then you compile it TorchScript, it can trigger arbitrary code execution. That being said, I'm not really sure what your threat model is, since you probably already have problems if you're compiling arbitrary malicious Python code with TorchScript?

LGro:
I trust your judgement about the trade off between a broader torch compatibility and the security benefits that the fix in 1.13.0 brings, just wanted to make sure it's a conscious one instead of an oversight :relaxed:

PR #1675, opened 2023-02-14
Disallow torch 1.13.0 (due to security vulnerability)

Balandat:
Addresses #1673


Balandat:
Oh I misread the vulnerability, this is actually present in all versions, not just 1.13.0.

PR #1684, opened 2023-02-16
approximate qPI using MVNXPB

danielrjiang:
Summary: This is work by jiayuewan during his internship. I am simply moving it to OSS.

Differential Revision: D43337388

### DIFF
--- a/botorch/acquisition/analytic.py
+++ b/botorch/acquisition/analytic.py
@@ -17,6 +17,7 @@ from abc import ABC
 
 from contextlib import nullcontext
 from copy import deepcopy
+
 from typing import Dict, Optional, Tuple, Union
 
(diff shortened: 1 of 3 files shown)


botorch/acquisition/analytic.py
```diff
@@ -231,6 +234,70 @@ def forward(self, X: Tensor) -> Tensor:
         return Phi(u)
 
 
+class ApproximateProbabilityOfImprovement(AnalyticAcquisitionFunction):
```
Balandat, line 237: Feels like this should have a notion of batch support in the name? Maybe sth like `qAnalyticProbabilityOfImprovement` ? Or `BatchProbabilityOfImprovement`?

PR #1689, opened 2023-02-17
[WIP] Upload tutorials artifiact

esantorella:
## Motivation

- write the runtime and memory info we are already storing for the tutorials to a csv
- write that info to a different branch since we can't write to main. similar to publishing website with gh-pages

## Test Plan

see what gets uploaded


esantorella:
Closing -- testing only "on push" flows, not "on PR"

PR #1696, opened 2023-02-21
Add option to disable retrying on optimization warning

saitcakmak:
Summary:
When `gen_candidates` exists with an optimization warning, we retry with a new set of initial conditions. In certain settings, `gen_candidates_scipy` is expected to exit with an optimization warning. This allows turning off this behavior in such settings.

Example use case: When using straight through estimators for optimizing in mixed discrete search spaces, we often get `ABNORMAL_TERMINATION_IN_LNSRCH` since these gradients are a bit ill-behaved (due to function evaluations happening after rounding).

Reviewed By: esantorella

Differential Revision: D43478712

### DIFF
--- a/botorch/optim/optimize.py
+++ b/botorch/optim/optimize.py
@@ -80,6 +80,7 @@ class OptimizeAcqfInputs:
     ic_generator: Optional[TGenInitialConditions] = None
     timeout_sec: Optional[float] = None
     return_full_tree: bool = False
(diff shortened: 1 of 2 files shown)


saitcakmak:
This was merged in 8db9c7bb038ad1caecb67398ff2766052b5132b6. Not sure why the PR got disconnected.

PR #1701, opened 2023-02-23
Fix error about tensors on different devices

esantorella:
Summary:
- Unit tests for inducing point allocators are passing in a kernel on CPU and `train_X` inputs on GPU
- This causes an error about tensors on different devices starting in `allocate_inducing_points`
- I fixed this by putting the kernel on the same device as `train_X`, but equally this could have been the other way around, or we could say it's the user's responsibility to put everything on the same device -- thoughts?

Differential Revision: D43549039


esantorella:
Closing; absorbed into #1652

PR #1703, opened 2023-02-25
Actions job to concatenate tutorials data to one CSV and run analysis notebook

esantorella:
After tutorials performance data is uploaded to artifacts branch, a job runs to concatenate tutorials data to one CSV and run analysis notebook. The notebook, minus cell outputs, and the script that runs it will be in the 'main' branch, while the data and notebook outputs stay in the 'analytics' branch, which continues to not need to stay up-to-date with 'main' since it doesn't have its own Python code.

Yes

## Test Plan

[x] Ran with manual actions run "on push" (smoke test): [here](https://github.com/pytorch/botorch/actions/runs/4286198759/jobs/7465442757), more recent actions run [here](https://github.com/pytorch/botorch/actions/runs/4388959014/jobs/7686024723).
[x] Auto-generated notebook should have nifty visualizations [here](https://github.com/pytorch/botorch/blob/artifacts/notebooks/tutorials_performance_tracking.ipynb)
[x] Up-to-date dataset of all tutorials runs exists and looks OK [here](https://github.com/pytorch/botorch/blob/artifacts/tutorial_performance_data/all_data.csv)


.github/workflows/reusable_tutorials.yml
```diff
@@ -90,12 +90,23 @@ jobs:
...
+      # 5) Commit updated notebook and new CSVs to artifacts branch
       run: |
         git config user.email "github-actions@github.com"
         git config user.name "github-actions"
+        cp scripts/analyze_tutorials_performance.py analyze_tutorials_performance.py
+        cp notebooks/tutorials_performance_tracking.ipynb tutorials_performance_tracking.ipynb
         git switch artifacts
+        mkdir scripts/
+        mv analyze_tutorials_performance.py scripts/
         mv *.csv tutorial_performance_data/
```
saitcakmak, line 107: We don't really need a scripts directory since we're not committing this file. It's just temporary.
```suggestion
        mv *.csv tutorial_performance_data/
```

.github/workflows/reusable_tutorials.yml
```diff
@@ -90,12 +90,23 @@ jobs:
...
-        git add tutorial_performance_data/*.csv
-        git commit tutorial_performance_data/*.csv -m "Adding most recent tutorials output"
+        mv tutorials_performance_tracking.ipynb notebooks/
+        python scripts/analyze_tutorials_performance.py
```
saitcakmak, line 109:
```suggestion
        python analyze_tutorials_performance.py
```

saitcakmak (approved):
One concern I have is that by repeatedly committing updated plots to the repo, the repo size will increase substantially over time (due to commit history). I know that this is an issue with plotly plots in Ax (which is why we force-push the gh-pages branch with no history) but I don't know how much of an issue this is with seaborn or matplotlib plots. If the NB size with the output plots is small, then I think this is ok. If it is something we'd measure in MBs, then this might become an issue over time.

.github/workflows/reusable_tutorials.yml
esantorella, line 109: We don't need to add Seaborn as a dependency if we're only using it here, although it might be nice to use it more widely.

```suggestion
        mv tutorials_performance_tracking.ipynb notebooks/
        pip install seaborn
        python scripts/analyze_tutorials_performance.py
```

setup.py
```diff
@@ -25,6 +25,7 @@
     "memory_profiler",
     "papermill",
     "pykeops",
+    "seaborn",
```
esantorella, line 28:
```suggestion
```

.github/workflows/reusable_tutorials.yml
```diff
@@ -90,12 +90,23 @@ jobs:
...
         git config user.name "github-actions"
+        cp scripts/analyze_tutorials_performance.py analyze_tutorials_performance.py
+        cp notebooks/tutorials_performance_tracking.ipynb tutorials_performance_tracking.ipynb
         git switch artifacts
+        mkdir scripts/
+        mv analyze_tutorials_performance.py scripts/
         mv *.csv tutorial_performance_data/
-        git add tutorial_performance_data/*.csv
-        git commit tutorial_performance_data/*.csv -m "Adding most recent tutorials output"
+        mv tutorials_performance_tracking.ipynb notebooks/
```
esantorella, line 108:
```suggestion
        mv tutorials_performance_tracking.ipynb notebooks/
        pip install seaborn
```
We don't need Seaborn as a dependency everywhere if we're only using it here

esantorella:
> One concern I have is that by repeatedly committing updated plots to the repo, the repo size will increase substantially over time (due to commit history). I know that this is an issue with plotly plots in Ax (which is why we force-push the gh-pages branch with no history) but I don't know how much of an issue this is with seaborn or matplotlib plots. If the NB size with the output plots is small, then I think this is ok. If it is something we'd measure in MBs, then this might become an issue over time.

Great point; these were pretty big (~6 MB). I cut it down to ~1 MB by switching to SVG graphics and rearranging the plots, which hopefully makes them easier to read too.

Balandat:
So does this mean that this adds about ~1MB to the repo on every tutorials run?

esantorella:
> So does this mean that this adds about ~1MB to the repo on every tutorials run?

Yeah. I'm not thrilled with this setup, but I'm reluctant to abandon it since it seems to have already sown some interesting patterns, like a sudden ~13% increase in how much memory the average tutorial uses in smoke test mode:

scripts/analyze_tutorials_performance.py
```diff
@@ -0,0 +1,65 @@
...
+    """
+    mode, commit_hash, date_time = fname[:-4].split("_")
+    df = pd.read_csv(os.path.join(data_dir, fname)).assign(
+        mode=mode,
+        commit_hash=commit_hash,
+        datetime=pd.to_datetime(date_time),
+        fname=fname,
+    )
+    # clean out '.ipynb' if it is present
+    df["name"] = df["name"].apply(lambda x: x[: -len(".ipynb")])
```
esantorella, line 38:
```suggestion
    df["name"] = df["name"].apply(lambda x: x.replace(".ipynb", ""))
```

PR #1707, opened 2023-02-27
[Easy] 5-minute timeout for smoke test tutorials

esantorella:
## Motivation

Currently, all tutorials time out at 30 minutes. We want smoke test tutorials to take much less time than that, and currently they all run in under 3.5 minutes, and most run much faster than that (see [here](https://github.com/pytorch/botorch/blob/4f80c9b8f97244a01f9b138c2180e41c54ac65f3/notebooks/tutorials_performance_tracking.ipynb)). As we  work to get more tutorials online, it would be good to a have a clear idea of what is fast enough for a smoke-test tutorial run.

## Test Plan

Checked that it ran locally; unit tests

### DIFF
--- a/scripts/run_tutorials.py
+++ b/scripts/run_tutorials.py
@@ -116,7 +116,7 @@ def run_tutorial(
     except subprocess.TimeoutExpired:
         error = (
             f"Tutorial {tutorial.name} exceeded the maximum runtime of "
-            "{timeout_minutes} minutes."
+            f"{timeout_minutes} minutes."
         )
         return error, {}
 
--- a/scripts/run_tutorials.py
+++ b/scripts/run_tutorials.py
@@ -74,7 +74,9 @@ def parse_ipynb(file: Path) -> str:
     return script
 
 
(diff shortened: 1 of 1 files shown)


scripts/run_tutorials.py
```diff
@@ -100,16 +102,22 @@ def run_tutorial(
...
-            (run_script, (script,), {"env": env}), retval=True, include_children=True
+            (run_script, (script, timeout_minutes), {"env": env}),
+            retval=True,
+            include_children=True,
         )
     except subprocess.TimeoutExpired:
-        error = f"Tutorial {tutorial.name} exceeded the maximum runtime of 30 minutes."
+        error = (
+            f"Tutorial {tutorial.name} exceeded the maximum runtime of "
+            "{timeout_minutes} minutes."
```
saitcakmak, line 119:
```suggestion
            f"{timeout_minutes} minutes."
```

saitcakmak (approved):
A small fix in-line, LGTM otherwise!

scripts/run_tutorials.py
esantorella: 😱

PR #1708, opened 2023-02-27
[Easy] Run some tutorials only in smoke test mode

esantorella:
## Motivation
Some tutorials were being ignored both in "smoke test" and "standard" runs because they OOM or time out in standard runs, but they actually run fine in "smoke test" mode. I moved those three tutorials to the list now called `RUN_IF_SMOKE_TEST_IGNORE_IF_STANDARD` (renamed for clarity).

## Test Plan

These are all running in smoke test mode now (https://github.com/pytorch/botorch/actions/runs/4287842169/jobs/7469160586) and all take less than a minute and an unexceptional amount of memory.

## Related PRs

Depends on #1707 (add five-minute timeout for smoke test tutorials)

### DIFF
--- a/scripts/run_tutorials.py
+++ b/scripts/run_tutorials.py
@@ -26,18 +26,19 @@ IGNORE = {  # ignored in smoke tests and full runs
     "vae_mnist.ipynb",  # requires setting paths to local data
     "bope.ipynb",  # flaky, keeps failing the workflows
     "preference_bo.ipynb",  # failing. Fix planned
-    # Causing the tutorials to crash when run without smoke test. Likely OOM.
-    # Fix planned.
-    "constraint_active_search.ipynb",
-    # Timing out
-    "saasbo.ipynb",
-    # Timing out
-    "scalable_constrained_bo.ipynb",
 }
 IGNORE_SMOKE_TEST_ONLY = {  # only used in smoke tests
     "thompson_sampling.ipynb",  # very slow without KeOps + GPU
     "composite_mtbo.ipynb",  # TODO: very slow, figure out if we can make it faster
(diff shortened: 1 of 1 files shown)


scripts/run_tutorials.py
```diff
@@ -26,19 +26,20 @@
     "vae_mnist.ipynb",  # requires setting paths to local data
     "bope.ipynb",  # flaky, keeps failing the workflows
     "preference_bo.ipynb",  # failing. Fix planned
+}
+IGNORE_SMOKE_TEST_ONLY = {  # only used in smoke tests
+    "thompson_sampling.ipynb",  # very slow without KeOps + GPU
+    "composite_mtbo.ipynb",  # TODO: very slow, figure out if we can make it faster
+    "Multi_objective_multi_fidelity_BO.ipynb",  # TODO: very slow, speed up
+}
+IGNORE_STANDARD_ONLY = {
```
saitcakmak, line 35: I find this naming very confusing. `IGNORE_SMOKE_TEST_ONLY` means to run it in smoke test only -- maybe we should just call it that `RUN_SMOKE_TEST_ONLY`?

For `IGNORE_STANDARD_ONLY`, the comment says "# Causing the tutorials to crash when run without smoke test. Likely OOM.". Does that mean we also run these only in smoke test, in which case they should be in the previous list?

esantorella: OK, this should be clearer now. There actually were never any tutorials we ran only in standard mode. I was confused about this and probably so was whoever put those tutorials on the ignore list in the first place, since they actually run fine in smoke test mode. I moved those tutorials to the list now called "RUN_IF_SMOKE_TEST_IGNORE_IF_STANDARD"

saitcakmak (approved):
Thanks!

PR #1727, opened 2023-03-06
fix docstring

jduerholt:
## Motivation

Docstring was listing a non-existing keyword. Looks like a copy-paste error.

Yes.

## Test Plan

Not necessary, since only a docstring is affected.


saitcakmak (approved):
Thanks!

PR #1734, opened 2023-03-08
Strip Meta-specific info from baxus tutorial

esantorella:
Reviewed By: saitcakmak

Differential Revision: D43913205


esantorella:
I'm not sure why this hasn't been auto-closed but it was committed.

PR #1738, opened 2023-03-09
Added single-objective JES with additional conditioning, removal of MO-specific code / additional info-theoretic utils

hvarfner:
variable names consistent with MES/GIBBON. Input constructors for JES & JES-exploit added, as well as optimizer of pathwise posterior samples & utility method for retrieving optima.

#### Refactoring of single-objective JES - larger codebase, but slightly faster, better support and more intuitive (?)
There is a substantial amount of multi-objective specific computation in single-objective JES. At its core, it's relatively similar to MES/GIBBON, and this suggested change tries to reflect this. It runs faster than previously (due to the removal of MO-specific stuff), and is (arguably) a bit more intuitive. Moreover, there is the option of running the two different conditioning variants (noisy or noiseless optimal values). The performance diff, however, is minor.

Also added the exploit-variant of JES for added robustness in cases of model misspecification.

#### Optimization of samples from the posterior
Added batch optimization of pathwise samples and utility method in acquisition/utils. Hyperparameters are set so as to (anecdotally/empirically) not run into memory or runtime issues on laptop, but still be peak performant.

Yes!

## Test Plan
Have tested the related notebooks and Ax interfaces & performance of the two different JES methods.

## Related PRs
Yes, proposed fix to FBGPs. Coming soon.


saitcakmak:
Hi @hvarfner. Thanks for the PR! I'll try to get this reviewed within the next week. In the mean time: i) We require 100% unit test coverage for BoTorch, so you can work on those. ii) The `.py` file under `tutorials` is not needed. We auto generate a `.py` file from the notebook while uploading the website.

hvarfner:
@saitcakmak Thanks! The .py file in tutorials wasn't supposed to be in there. I'll work on the tests in the coming days!

botorch/utils/sampling.py
```diff
@@ -873,3 +874,79 @@ def sparse_to_dense_constraints(
...
+                                     :].requires_grad_(requires_grad=True)
+
+    for iter_ in range(maxiter):
+        # Decaying the learing rate - empirically, this has been well-calibrated
+        lr = lr_init / (iter_ + 1)
+        per_sample_outputs = path_func(X_top_candidates)
+        grads = torch.autograd.grad(
+            per_sample_outputs, X_top_candidates, grad_outputs=torch.ones_like(per_sample_outputs))[0]
+        X_top_candidates = torch.clamp(
+            X_top_candidates + lr * grads, bounds[0], bounds[1])  # TODO fix bounds here
```
Balandat, line 936: The sample paths here are deterministic (and smooth) functions - rather using a (manual) first-order gradient scheme here, we could probably get much faster convergence by using a (quasi)-second order method such as L-BFGS here.

hvarfner: Sure, will fix!

hvarfner: @Balandat After giving this some more thought, it seems pretty tricky to implement this in batch with the set of tools available. Would a for-loop solution suffice?

botorch/acquisition/joint_entropy_search.py
```diff
@@ -62,50 +69,345 @@ def __init__(
                 sample only contains one optimal set of inputs.
             optimal_outputs: A `num_samples x 1`-dim Tensor containing the optimal
                 set of objectives of dimension `1`.
+            condition_noiseless:
+            posterior_transform:
+            estimation_type: estimation_type: A string to determine which entropy
+                estimate is computed: Lower bound" (LB) or "Monte Carlo" (MC).
```
Balandat, line 75:
```suggestion
                estimate is computed: "Lower bound" ("LB") or "Monte Carlo" ("MC"). 
```

botorch/acquisition/joint_entropy_search.py
```diff
@@ -62,50 +69,345 @@ def __init__(
...
+        self.set_X_pending(X_pending)
+
+    @concatenate_pending_points
+    @t_batch_mode_transform()
+    def forward(self, X: Tensor) -> Tensor:
+        r"""Evaluates qLowerBoundJointEntropySearch at the design points `X`.
+        Args:
+            X: A `batch_shape x q x d`-dim Tensor of `batch_shape` t-batches with `q`
+            `d`-dim design points each.
+        Returns:
```
Balandat, line 153: formatting nit
```suggestion
        
        Args:
            X: A `batch_shape x q x d`-dim Tensor of `batch_shape` t-batches with `q`
            `d`-dim design points each.
        
        Returns:
```

test/acquisition/test_joint_entropy_search.py
```diff
@@ -69,8 +69,8 @@ def test_lower_bound_joint_entropy_search(self):
 
             num_samples = 20
 
-            optimal_inputs = torch.rand(num_samples, input_dim, **tkwargs)
-            optimal_outputs = torch.rand(num_samples, num_objectives, **tkwargs)
+            optimal_inputs = torch.rand(num_samples, input_dim, **tkwargs).unsqueeze(-1)
+            optimal_outputs = torch.rand(num_samples, num_objectives, **tkwargs).unsqueeze(-1)
```
Balandat, line 73:
```suggestion
            optimal_inputs = torch.rand(num_samples, input_dim, 1, **tkwargs)
            optimal_outputs = torch.rand(num_samples, num_objectives, 1, **tkwargs)
```

botorch/utils/sampling.py
```diff
@@ -873,3 +874,79 @@ def sparse_to_dense_constraints(
...
+    X_top_candidates_flat = X_top_candidates.reshape(
+        num_optima * batch_size, num_restarts, -1)
+
+    X_opt = X_top_candidates_flat[row_indexer,
+                                  sample_argmax].reshape(num_optima, batch_size, -1)
+    f_opt = per_sample_outputs.max(axis=-1).values.reshape(num_optima, batch_size, 1)
+
+    # if we minimize, then we have now found the negative minimum, needs to be flipped.
+    if not maximize:
+        f_opt = (-1) * f_opt
```
Balandat, line 950:
```suggestion
        f_opt = -f_opt
```

botorch/utils/sampling.py
```diff
@@ -873,3 +874,79 @@ def sparse_to_dense_constraints(
...
+        lr = lr_init / (iter_ + 1)
+        per_sample_outputs = path_func(X_top_candidates)
+        grads = torch.autograd.grad(
+            per_sample_outputs, X_top_candidates, grad_outputs=torch.ones_like(per_sample_outputs))[0]
+        X_top_candidates = torch.clamp(
+            X_top_candidates + lr * grads, bounds[0], bounds[1])  # TODO fix bounds here
+
+    per_sample_outputs = path_func(X_top_candidates).reshape(
+        num_optima * batch_size, num_restarts)
+    sample_argmax = torch.max(per_sample_outputs, dim=-1, keepdims=True)[1].flatten()
```
Balandat, line 940:
```suggestion
    sample_argmax = torch.max(per_sample_outputs, dim=-1, keepdims=True).indices.flatten()
```

botorch/utils/sampling.py
```diff
@@ -873,3 +874,79 @@ def sparse_to_dense_constraints(
...
+    gradient descent on the best num_restarts points.
+
+    Args:
+        paths: Random Fourier Feature-based sample paths from the GP  
+        bounds: The bounds on the search space.
+        candidates: A priori good candidates (typically previous design points)
+            which acts as extra initial guesses for the optimization routine.
+        raw_samples: The number of samples with which to query the samples initially.
+        num_restarts The number of gradient descent steps to use on each of the best 
+        found initial points.
```
Balandat, line 899:
```suggestion
            found initial points.
```

botorch/utils/sampling.py
```diff
@@ -873,3 +874,79 @@ def sparse_to_dense_constraints(
...
+
+    candidate_set = unnormalize(SobolEngine(
+        dimension=bounds.shape[1], scramble=True).draw(raw_samples), bounds)
+    # TODO add spray points
+    # queries all samples on all candidates - output raw_samples * num_objectives * num_optima
+    candidate_queries = path_func(candidate_set)
+    num_optima = candidate_queries.shape[0]
+    batch_size = candidate_queries.shape[1] if candidate_queries.ndim == 3 else 1
+    argtop_candidates = candidate_queries.argsort(dim=-1, descending=True)[
+        ..., 0:num_restarts]
```
Balandat, line 921: why not use `topk`?

botorch/utils/sampling.py
```diff
@@ -873,3 +874,79 @@ def sparse_to_dense_constraints(
...
+
+    # These are used as masks when retrieving the argmaxes
+    row_indexer = torch.arange(num_optima * batch_size).to(torch.long)
```
Balandat, line 924: should be a nullop
```suggestion
    row_indexer = torch.arange(num_optima * batch_size)
```

botorch/utils/sampling.py
Balandat: I feel like a for loop across each sample would be quite costly. Another option here would be to just define a function that sums the `per_sample_outputs` and then call L-BFGS-B on that (with the domain being `k` copies of the original domain, with `k` the number of sample paths). This is actually what we do for performing acquisition function optimization across random restarts: https://github.com/pytorch/botorch/blob/9302055f36981e3413e8dce08ae050bd701f5e7a/botorch/generation/gen.py#L175-L203

Balandat (commented):
Not a super thorough review, my main comment is about using a new acquisition function class to implement something generic like epsilon-greedy. 

Can you also make sure to run `black` on this?

botorch/utils/sampling.py
hvarfner: Okay, I did try it some time ago with the torch L-BFGS but I must have done something wrong if it indeed works here. I'll explore it!

botorch/acquisition/joint_entropy_search.py
```diff
@@ -62,50 +69,345 @@ def __init__(
...
+        sampler = SobolQMCNormalSampler(sample_shape=torch.Size([num_samples]))
+        MCSamplerMixin.__init__(self, sampler=sampler)
+        # To enable fully bayesian GP conditioning, we need to unsqueeze
+        # to get num_optima x num_gps unique GPs
+
+        # inputs come as num_optima_per_model x num_models x d
+        # but we want it four-dimensional in the Fully bayesian case,
+        # and three-dimensional otherwise (optionally remove, then re-add)
+        self.optimal_inputs = optimal_inputs.squeeze(1).unsqueeze(-2)
+        self.optimal_outputs = optimal_outputs.squeeze(1).unsqueeze(-2)
```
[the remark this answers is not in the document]
hvarfner: Sure! Some squeezes/unsqueezes cancelled out, so this should not be as confusing anymore.

botorch/acquisition/joint_entropy_search.py
```diff
@@ -62,42 +84,198 @@ def __init__(
...
+            "dtype": self.optimal_outputs.dtype,
+            "device": self.optimal_outputs.device,
+        }
+        initial_posterior = self.initial_model.posterior(X, observation_noise=True)
+        # need to check if there is a two-dimensional batch shape -
+        # the sampled optima appear in the dimension right after
+        batch_shape = X.shape[:-2]
+        sample_dim = len(batch_shape)
+        # We DISREGARD the additional constant term.
+        initial_entropy = 0.5 * torch.logdet(initial_posterior.mvn.covariance_matrix)
```
Balandat, line 206: The nice thing with this is that if the covariance matrix has special structure, thanks to linear_opeator's support for `__torch_function__` (see [here](https://github.com/cornellius-gp/linear_operator/blob/main/linear_operator/operators/_linear_operator.py#L1695-L1701) for the registration of the method)t his call will exploit that structure. 
```suggestion
        initial_entropy = 0.5 * torch.logdet(initial_posterior.mvn.lazy_covariance_matrix)
```

cc @gpleiss, @jacobrgardner

botorch/acquisition/joint_entropy_search.py
```diff
@@ -107,5 +285,64 @@ def forward(self, X: Tensor) -> Tensor:
...
+            "device": self.optimal_outputs.device,
+        }
+
+        initial_posterior = self.initial_model.posterior(X, observation_noise=True)
+
+        add_term = 0.5 * (1 + torch.log(torch.ones(1, **tkwargs) * 2 * pi))
+        batch_shape = X.shape[:-2]
+        sample_dim = len(batch_shape)
+        # We DISREGARD the additional constant term.
+        initial_entropy = 0.5 * torch.logdet(initial_posterior.mvn.covariance_matrix)
```
Balandat, line 299:
```suggestion
        initial_entropy = 0.5 * torch.logdet(initial_posterior.mvn.lazy_covariance_matrix)
```

botorch/acquisition/joint_entropy_search.py
```diff
@@ -107,5 +285,64 @@ def forward(self, X: Tensor) -> Tensor:
...
+
+        # initial_entropy of shape batch_size or batch_size x num_models if FBGP
+        # first need to unsqueeze the sample dim (first after batch dim) and then the two last
+        initial_entropy = (
+            initial_entropy.unsqueeze(sample_dim).unsqueeze(-1).unsqueeze(-1)
+        )
+
+        CLAMP_LB = torch.finfo(tkwargs["dtype"]).eps
```
Balandat, line 307: IIRC at east for numpy calling `finfo` has nontrivial overhead - how about defining these bounds also once at the module level (similarly for when this is used above)?

botorch/acquisition/joint_entropy_search.py
```diff
@@ -107,5 +285,64 @@ def forward(self, X: Tensor) -> Tensor:
...
+        # Compute the mixture mean and variance
+        posterior_m = self.conditional_model.posterior(
+            X.unsqueeze(MCMC_DIM), observation_noise=True
+        )
+        noiseless_var = self.conditional_model.posterior(
+            X.unsqueeze(MCMC_DIM), observation_noise=False
```
Balandat, line 313: In general this will do a lot of repeated work - for GP models the `observation_noise=True` arg will just add the likelihood noise to the posterior.

botorch/acquisition/joint_entropy_search.py
```diff
@@ -107,5 +285,64 @@ def forward(self, X: Tensor) -> Tensor:
...
+            torch.ones(1, device=X.device, dtype=X.dtype),
+        )
+        # prepare max value quantities and re-scale as required
+        normalized_mvs = (self.optimal_outputs - mean_m) / noiseless_var.sqrt()
+        mvs_rescaled_mc = (normalized_mvs - rho * normalized_samples) / (1 - rho**2)
+        cdf_mvs = normal.cdf(normalized_mvs).clamp_min(CLAMP_LB)
+        cdf_rescaled_mvs = normal.cdf(mvs_rescaled_mc).clamp_min(CLAMP_LB)
+        mv_ratio = cdf_rescaled_mvs / cdf_mvs
+
+        log_term = torch.log(mv_ratio) + conditional_logprobs
```
Balandat, line 339: Hmm it would be good to just work with logcdfs from the get go here for numerical stability, but unfortunately I don't think torch has great support for those

Balandat: Looks like my past self noticed that as well: https://github.com/pytorch/pytorch/issues/52973

botorch/utils/sampling.py
```diff
@@ -873,3 +874,70 @@ def sparse_to_dense_constraints(
...
+
+    Args:
+        paths: Random Fourier Feature-based sample paths from the GP
+        bounds: The bounds on the search space.
+        candidates: A priori good candidates (typically previous design points)
+            which acts as extra initial guesses for the optimization routine.
+        raw_samples: The number of samples with which to query the samples initially.
+        num_restarts: The number of points selected for gradient-based optimization.
+        maximize: Boolean indicating whether to maimize or minimize
+    Returns:
```
Balandat, line 899:
```suggestion
    
    Returns:
```

test/acquisition/test_utils.py
```diff
@@ -766,3 +768,35 @@ def test_project_to_sample_points(self):
...
+
+class TestGetOptimalSamples(BotorchTestCase):
+    def test_get_optimal_samples(self):
+        dims = 3
+        dtype = torch.float64
+        for_testing_speed_kwargs = {"raw_samples": 50, "num_restarts": 3}
+        num_optima = 7
+        batch_shape = (3,)
+
+        bounds = torch.Tensor([[0, 1]] * dims).T.to(dtype)
```
Balandat, line 781:
```suggestion
        bounds = torch.tensor([[0, 1]] * dims, dtype=dtype).T
```

test/acquisition/test_utils.py
```diff
@@ -766,3 +768,35 @@ def test_project_to_sample_points(self):
...
+        X = torch.rand(*batch_shape, 4, dims).to(dtype)
+        Y = torch.sin(X).sum(dim=-1, keepdim=True).to(dtype)
```
Balandat, line 783:
```suggestion
        X = torch.rand(*batch_shape, 4, dims, dtype=dtype)
        Y = torch.sin(X).sum(dim=-1, keepdim=True).to(dtype)
```

test/utils/test_sampling.py
```diff
@@ -528,3 +528,40 @@ def test_sample_polytope_unbounded(self):
...
+        dims = 2
+        dtype = torch.float64
+        eps = 1e-6
+        for_testing_speed_kwargs = {"raw_samples": 250, "num_restarts": 3}
+        nums_optima = (1, 7)
+        batch_shapes = ((), (3,), (5, 2))
+        for num_optima, batch_shape in itertools.product(nums_optima, batch_shapes):
+            bounds = torch.Tensor([[0, 1]] * dims).T.to(dtype)
+            X = torch.rand(*batch_shape, 52, dims).to(dtype)
+            Y = torch.pow(X - 0.5, 2).sum(dim=-1, keepdim=True).to(dtype)
```
Balandat, line 545:
```suggestion
            bounds = torch.Tensor([[0, 1]] * dims, dtype=dtype).T
            X = torch.rand(*batch_shape, 52, dims, dtype=dtype)
            Y = torch.pow(X - 0.5, 2).sum(dim=-1, keepdim=True)
```

Balandat (commented):
Thanks! A number of largely nits. I'll take another look at the gpytorch PR for allowing conditioning on samples also in the batched model (fully Bayesian) case in the next few days.

Balandat:
>  I'll take another look at the gpytorch PR for allowing conditioning on samples also in the batched model (fully Bayesian) case in the next few days. 

For the time being, does your code just throw an error? If so, maybe you can add an informative `NotImplementedError` pointing to the discussion where we're trying to resolve this? That way we can potentially get this merged in before having everything figured out.

hvarfner:
That makes sense, I added the error as well. I realize that I forgot to mention a specific technical detail previously:

For each of M models, we'd want to condition on an additional N optima (each of which yields a new GP). As such, the (entire) conditional model would have M x N outputs. This works fine for a regular `SingleTaskGP`, as a batch dimension is simply added when `optimal_inputs` are of shape `N x 1 x D` (adds a batch dim, whereas shape `N x D` just adds N observations to one GP. 

The  expected behavior in the FB case would then be to have `optimal_inputs` of shape `M x N x 1 x D`. I'm not sure if this would work if you want to preserve one copy of the training data only.

Happy to contribute any which way to the FB issue.

hvarfner:
After benchmarking a bit more, the method for obtaining posterior samples (which now uses `gen_candidates_scipy`) the quadratic scaling in the number of samples incurred by optimizing their sum r

botorch/acquisition/input_constructors.py
```diff
@@ -1239,3 +1241,39 @@ def optimize_objective(
         return_best_only=True,
         sequential=sequential,
     )
+
+
+# TODO make single-objective with pairwise and multi-objective with pareto
```
esantorella, line 1246: Is this something you plan to do, or something you think would be good for someone else to do? (If the latter, a GH issue would be better than a comment)

esantorella:
Hi, thanks for the pull request! A couple things:
* Could you run `black` and `ufmt` as described [here](https://github.com/pytorch/botorch/blob/main/CONTRIBUTING.md) to fix the errors the linter is throwing?
* Could you take a look at the documentation error raised [here](https://github.com/pytorch/botorch/actions/runs/4596270450/jobs/8128806009)? It looks like there's an indentation issue. Making the documentation work is finicky, so let me know if you get stuck on that and I'll fix it myself.

botorch/acquisition/input_constructors.py
hvarfner: Hmm. I actually cannot recall putting that comment in (but I probably did). 

Otherwise, but it could refer to this:
https://github.com/pytorch/botorch/issues/1763

botorch/utils/sampling.py
```diff
@@ -18,13 +18,14 @@
...
 
 import numpy as np
 import scipy
 import torch
 from botorch.exceptions.errors import BotorchError
 from botorch.sampling.qmc import NormalQMCEngine
+from botorch.utils.transforms import unnormalize
 from scipy.spatial import Delaunay, HalfspaceIntersection
 from torch import LongTensor, Tensor
 from torch.quasirandom import SobolEngine
```
Balandat, line 31:
```suggestion
from torch.quasirandom import SobolEngine


if TYPE_CHECKING:
    from botorch.sampling.pathwise.path import SamplePath
```

Balandat (commented):
> importing SamplePath causes a circular import. As such, SamplePath remains undefined which flake8 doesn't like.

You can get around this with the `TYPE_CHECKING` var from the `typing` module, made comments in the code.

botorch/acquisition/joint_entropy_search.py
```diff
@@ -12,44 +12,75 @@
...
+ESTIMATION_TYPES = ["MC", "LB"]
+MC_ADD_TERM = 0.5 * (1 + log(2 * pi))
+
+# The CDF query cannot be strictly zero in the division
+# and this clamping helps assure that it is always positive.
+CLAMP_LB = torch.finfo(torch.float32).eps
+FULLY_BAYESIAN_ERROR_MSG = """JES is not yet available with Fully Bayesian GPs.
+Track the issue, which regards conditioning on a number of optima on a collecion
+ of models, in detail at https://github.com/pytorch/botorch/issues/1680.
+"""
```
Balandat, line 46: This seems to throw off sphinx for some reason

```suggestion
FULLY_BAYESIAN_ERROR_MSG = (
    "JES is not yet available with Fully Bayesian GPs. Track the issue, "
    "which regards conditioning on a number of optima on a collection "
    "of models, in detail at https://github.com/pytorch/botorch/issues/1680".
)
```

Balandat (approved):
looks like there is still a minor docs issue, o/w this lgtm

botorch/acquisition/joint_entropy_search.py
```diff
@@ -12,44 +12,76 @@
...
+    Joint Entropy Search for Maximally-informed Bayesian Optimization.
+    In Proceedings of the Annual Conference on Neural Information
+    Processing Systems (NeurIPS), 2022.
+
+.. [Tu2022joint]
+    B. Tu, A. Gandy, N. Kantas, B. Shafei,
+    Joint Entropy Search for Multi-objective Bayesian Optimization.
+    In Proceedings of the Annual Conference on Neural Information
+    Processing Systems (NeurIPS), 2022.
+"""
```
Balandat, line 62: This is what trips sphinx up. Can you just move these references in the module-level docstring at the top (l.7) instead?

tutorials/information_theoretic_acquisition_functions.ipynb
```diff
@@ -424,9 +426,8 @@
     "\n",
     "jes_lb = qLowerBoundJointEntropySearch(\n",
```
esantorella, line 427:
```suggestion
    "jes_lb = qJointEntropySearch(\n",
```
I think a bunch of these need to be changed to fix the tutorial failure

hvarfner: @esantorella Oh, right. Thanks, I'll go through that, too!

hvarfner: @esantorella Done! Hopefully it's all good now =)

Balandat:
@hvarfner looks like there are a couple of outstanding docs / lint issues. Do you think you can clean that up soon so that we can merge this in?

hvarfner:
@Balandat @esantorella Sorry, I thought I had it figured out last time around. I don't quite understand how I address the last little bit, so I'd appreciate any pointers you may have.

tutorials/information_theoretic_acquisition_functions.ipynb
```diff
@@ -264,16 +267,8 @@
     "from botorch.fit import fit_gpytorch_mll\n",
     "\n",
     "tkwargs = {\"dtype\": torch.double, \"device\": \"cpu\"}\n",
-    "SMOKE_TEST = os.environ.get(\"SMOKE_TEST\")"
```
Balandat, line 267: You need to define that var; looks like the tutorial fails since you removed this.

Balandat (commented):
Left some comments that should fix the outstanding issues. There are also a couple of lines that aren't covered by tests, please make sure to get those as well. Thanks!

botorch/acquisition/joint_entropy_search.py
```diff
@@ -62,42 +90,205 @@ def __init__(
                 sample only contains one optimal set of inputs.
             optimal_outputs: A `num_samples x 1`-dim Tensor containing the optimal
                 set of objectives of dimension `1`.
+            condition_noiseless: Whether to condition on noiseless optimal observations
+                f* [Hvarfner et. al.]or noisy optimal observations y* [Tu et. al,].
+                These are sampled identically, so this only controls the fashion in
+                which the GP is reshaped as a result of conditioning on the optimum.
```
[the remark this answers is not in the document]
Balandat: (this will also avoid a sphinx warning that the `Hvarfner2022joint` reference was not referenced)

Balandat:
Looks like we're all good now. @esantorella could you merge this in, please?

hvarfner:
@Balandat @esantorella My pleasure, thanks for your patience! =)

PR #1739, opened 2023-03-14
Lower memory usage in CAS tutorial

dme65:
The CAS tutorial has been running into OOMs in the CI. This lowers the memory usage by using fewer samples in the ECI acquisition function.


Balandat (approved):
lgtm - can you also add the main changes to the summary? Kind of hard to spot given that this is edited in the json.

PR #1740, opened 2023-03-15
Speed up the SAASBO tutorial

dme65:
Speed up the SAASBO tutorial by using 256 warmup steps and 128 samples (with a comment that we still recommend 512 warmup steps and 256 samples). I also dropped the dimensionality of the problem from 50 to 30 and increased the batch size from 4 to 5. The tutorial now runs in ~12 minutes on Github Actions.


Balandat (approved):
lgtm thanks for making this change!

PR #1742, opened 2023-03-15
Speed up scbo tutorial

dme65:
Use 2000 instead of 4000 candidates in order to lower the memory usage and also make the tutorial run faster


Balandat (approved):
lgtm, thanks

Issue #1737, opened 2023-03-09
[Feature Request] Linear constraints between candidates in a q-batch in optimize_acqf


jduerholt:
# 🚀 Feature Request

## Motivation

Imagine a system in which you can run several experiments in parallel, but all experiments have to have to share the same value for at least one input parameter. Let's make it more concrete:
 
A system can run several chemical reactions in parallel, but one can only set one temperature in the system which then holds for all minireactors in the system, whereas all the other degrees of freedom can be set independently between the different reactors/candidates. It is related to a fixed feature but it is not the same as a fixed feature. The temperature can be proposed by the BO, but vor every candidate in the proposed q-batch it has to be the same temperature. How to find under this constraint the maximum acquisition function? 

## Pitch

From my understanding, this could be implemented as a linear equality constraint between the different candidates in the q-batch and the SLSQP should be able to handle it. If you request for example 2 candidates, it should hold `T_1 == T_2` which is the same as `T_1 - T_2 ==0`, where `T` is the temperature.

Any ideas, how to add these kind of inter-candidate linear constraints to `optimize_acqf`?

**Are you willing to open a pull request?** Yes.

saitcakmak:
That's an interesting application. I think you should be able to do this in a similar way to how we construct the constraints for scipy optimizer. I am not super familiar with that part of the code, but the existing setup duplicates the simple constraints for each batch dimension, so that they're applied to all candidates. You'd instead want to construct the constraints to span the q-batch dimension and only duplicate them for the t-batch dimensions. I think this should be relatively straightforward as long as you don't use `sequential=True` for optimization.

This is where the constraints are constructed: https://github.com/pytorch/botorch/blob/main/botorch/optim/parameter_constraints.py#L68

cc @esantorella who might be more familiar with how this all works

saitcakmak:
Actually, it looks like there's some support for constraints across q-batch dimensions: https://github.com/pytorch/botorch/blob/main/botorch/optim/parameter_constraints.py#L290-L301

Balandat:
Indeed this should already be supported that way (lmk if you're running into any problems). 

cc @Ryan-Rhys @qingfeng10 - this is the mechanism to enforce e.g. a budget constraint a batch of `q` candidates.

jduerholt:
Wow, I am very impressed. It works out of the box. Here is an example snippet:

```python
import torch
from botorch.models import SingleTaskGP
from botorch.fit import fit_gpytorch_mll
[...]

PR #1757, opened 2023-03-21
Polytope sampling for linear constraints along the q-dimension

jduerholt:
## Motivation

This PR adds polytope sampling for linear constraints along the q-dimension for acqf optimization in this case without the need for providing batch initial conditions explicitly. The provided solution is discussed in issue https://github.com/pytorch/botorch/issues/1737.

The PR is to 95% complete, but I am not sure with some naming things, so I named the two types of constraints in the method names often `horizontal` and `vertical`. I feel that there is better naming. So it would be nice if @Balandat or @saitcakmak  could just have a look and give some hints, so that I can finish it. From the functionality perspective, it should be complete.

Yes.

## Test Plan

Unit tests.

### DIFF
--- a/botorch/optim/initializers.py
+++ b/botorch/optim/initializers.py
@@ -70,10 +70,13 @@ def transform_constraints(
     d dimensional state.
 
     Args:
-        constraints (Union[List[Tuple[Tensor, Tensor, float]], None]): Constraints
-            to transform.
-        q (int): Size of the q-batch.
-        d (int): Dimensionality of the problem.
+        constraints: A list of tuples (indices, coefficients, rhs), with each tuple encoding an
+            (in-)equality constraint of the form `\sum_i (X[indices[i]] * coefficients[i]) (>)= rhs`.
+            If `indices` is a 2-d Tensor, this supports specifying constraints across the points
+            in the `q`-batch (inter-point constraints). If `None`, this function is a nullop and
+            simply returns `None`.
+        q: Size of the `q`-batch.
+        d: Dimensionality of the problem.
 
     Returns:
         List[Tuple[Tensor, Tensor, float]]: List of transformed constraints.
--- a/botorch/optim/initializers.py
+++ b/botorch/optim/initializers.py
@@ -196,17 +196,11 @@ def sample_q_batches_from_polytope(
     """
 
     # check if inter-point constraints are present
-    inter_point = False
-    if inequality_constraints is not None:
-        for c in inequality_constraints:
-            if len(c[0].shape) > 1:
-                inter_point = True
-                break
-    if inter_point is False and equality_constraints is not None:
-        for c in equality_constraints:
-            if len(c[0].shape) > 1:
-                inter_point = True
-                break
+    inter_point = any(
+        indices.shape > 1
+        for constraints in (inequality_constraints or [], equality_constraints or [])
+        for indices, _, _ in constraints
+    )
 
     if inter_point:
         return (
--- a/botorch/optim/initializers.py
+++ b/botorch/optim/initializers.py
@@ -182,7 +182,8 @@ def sample_q_batches_from_polytope(
         q: Number of samples per q-batch
         bounds: A `2 x d` tensor of lower and upper bounds for each column of `X`.
         n_burnin: The number of burn-in samples for the Markov chain sampler.
-        thinning: The amount of thinning (number of steps to take between returning samples).
+            thinning: The amount of thinning (number of steps to take between
+            returning samples).
         seed: The random seed.
         inequality constraints: A list of tuples (indices, coefficients, rhs),
             with each tuple encoding an inequality constraint of the form
@@ -197,7 +198,7 @@ def sample_q_batches_from_polytope(
 
     # check if inter-point constraints are present
     inter_point = any(
-        indices.shape > 1
+        len(indices.shape) > 1
         for constraints in (inequality_constraints or [], equality_constraints or [])
         for indices, _, _ in constraints
     )
--- a/botorch/optim/initializers.py
+++ b/botorch/optim/initializers.py
@@ -114,7 +114,7 @@ def transform_intra_point_constraint(
     """
     indices, coefficients, rhs = constraint
     if indices.max() >= d:
-        raise ValueError("d has to be larger than the largest index in the constraint.")
+        raise ValueError(f"Constraint indices cannot exceed the problem dimension {d=}.")
     return [
         (
             torch.tensor([i * d + j for j in indices], dtype=torch.int64),
--- a/botorch/optim/initializers.py
+++ b/botorch/optim/initializers.py
@@ -66,6 +66,18 @@ TGenInitialConditions = Callable[
 def transform_constraints(
     constraints: Union[List[Tuple[Tensor, Tensor, float]], None], q: int, d: int
 ) -> List[Tuple[Tensor, Tensor, float]]:
+    """Transform constraints to sample from a d*q dimensional space instead of a
+    d dimensional state.
+
+    Args:
+        constraints (Union[List[Tuple[Tensor, Tensor, float]], None]): Constraints
+            to transform.
+        q (int): Size of the q-batch.
+        d (int): Dimensionality of the problem.
+
+    Returns:
+        List[Tuple[Tensor, Tensor, float]]: List of transformed constraints.
+    """
     if constraints is None:
         return None
     transformed = []
@@ -80,22 +92,25 @@ def transform_constraints(
 def transform_intra_point_constraint(
     constraint: Tuple[Tensor, Tensor, float], d: int, q: int
 ) -> List[Tuple[Tensor, Tensor, float]]:
-    """_summary_
+    """Transforms an intra-point/pointwise constraint from
+    d-dimensional space to a d*q dimesional space.
 
     Args:
-        constraint (Tuple[Tensor, Tensor, float]): _description_
-        d (int): _description_
-        q (int): _description_
+        constraint (Tuple[Tensor, Tensor, float]): Constraints
+            to transform.
+        d (int): Size of the q-batch.
+        q (int): Dimensionality of the problem.
 
     Raises:
-        ValueError: _description_
+        ValueError: If indices in the constraints are larger than the
+            dimensionality d of the problem.
 
     Returns:
-        List[Tuple[Tensor, Tensor, float]]: _description_
+        List[Tuple[Tensor, Tensor, float]]: List of transformed constraints.
     """
     indices, coefficients, rhs = constraint
     if indices.max() >= d:
-        raise ValueError()
+        raise ValueError("d has to be larger than the largest index in the constraint.")
     return [
         (
             torch.tensor([i * d + j for j in indices], dtype=torch.int64),
@@ -109,22 +124,25 @@ def transform_intra_point_constraint(
 def transform_inter_point_constraint(
     constraint: Tuple[Tensor, Tensor, float], d: int
 ) -> Tuple[Tensor, Tensor, float]:
-    """_summary_
+    """Transforms an inter-point constraint from
+    d-dimensional space to a d*q dimesional space.
 
     Args:
-        constraint (Tuple[Tensor, Tensor, float]): _description_
-        d (int): _description_
+        constraint (Tuple[Tensor, Tensor, float]): Constraint
+            to transform.
+        d (int): Size of the q-batch.
+        q (int): Dimensionality of the problem.
 
     Raises:
-        ValueError: _description_
-        ValueError: _description_
+        ValueError: If indices in the constraints are larger than the
+            dimensionality d of the problem.
 
     Returns:
-        Tuple[Tensor, Tensor, float]: _description_
+        List[Tuple[Tensor, Tensor, float]]: Transformed constraint.
     """
     indices, coefficients, rhs = constraint
     if indices[:, 1].max() >= d:
-        raise ValueError()
+        raise ValueError("d has to be larger than the largest index in the constraint.")
     return (
         torch.tensor([r[0] * d + r[1] for r in indices], dtype=torch.int64),
         coefficients,
--- a/botorch/optim/initializers.py
+++ b/botorch/optim/initializers.py
@@ -203,25 +203,21 @@ def sample_q_batches_from_polytope(
     )
 
     if inter_point:
-        return (
-            get_polytope_samples(
-                n=n,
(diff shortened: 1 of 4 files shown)


botorch/optim/initializers.py
```diff
@@ -63,6 +63,73 @@
 ]
 
 
+def transform_constraints(
+    constraints: List[Tuple[Tensor, Tensor, float]], q: int, d: int
+) -> List[Tuple[Tensor, Tensor, float]]:
```
Balandat, line 68: docstring please :)

botorch/optim/initializers.py
```diff
@@ -63,6 +63,73 @@
...
+    transformed = []
+    for constraint in constraints:
+        if len(constraint[0].shape) == 1:
+            transformed += transform_horizontal_constraint(constraint, d, q)
+        else:
+            transformed.append(transform_vertical_constraint(constraint, d))
+    return transformed
+
+
+def transform_horizontal_constraint(
```
Balandat, line 78: Maybe call this "intra-point" constraint instead? Or "pointwise"?

botorch/optim/initializers.py
```diff
@@ -63,6 +63,73 @@
...
+        (
+            torch.tensor([i * d + j for j in indices], dtype=torch.int64),
+            coefficients,
+            rhs,
+        )
+        for i in range(q)
+    ]
+
+
+def transform_vertical_constraint(
```
Balandat, line 107: Maybe call this "inter-point" or "cross-batch" constraint instead?

botorch/optim/initializers.py
```diff
@@ -169,10 +236,18 @@ def gen_batch_initial_conditions(
             else:
                 X_rnd = (
                     get_polytope_samples(
-                        n=n * q,
-                        bounds=bounds,
-                        inequality_constraints=inequality_constraints,
-                        equality_constraints=equality_constraints,
+                        n=n,
+                        bounds=torch.hstack([bounds for _ in range(q)]),
```
Balandat, line 240: So the whole point of this change is to draw a `n` samples from a `q*d`-dim space, rather than `n*q` samples from a `d`-dim space. I don't have a great sense of this, but one thing that may happen by doing os is that you may need to increase the burnin and thinning options here to get comparable mixing behavior for the chain. I haven't really looked into this much, but it would be worth trying to do a basic numerical study that compares this (without actually imposing the cross-batch constraints) just to see how much of an issue this is.

test/optim/test_initializers.py
```diff
@@ -270,6 +273,115 @@ def test_gen_batch_initial_conditions_warning(self):
...
+                    torch.tensor([-1, -1]).to(dtype=dtype), constraint[1]
+                )
+                self.assertEqual(constraint[2], -1.0)
+            # test failure on invalid d
+            constraint = (
+                torch.tensor([[0, 3]]),
+                torch.tensor([-1.0, -1.0], dtype=dtype),
+                0,
+            )
+            with self.assertRaises(ValueError):
```
Balandat, line 310: suggest using `assertRaisesRegex` here and check for the error message (once you've added that)

test/optim/test_initializers.py
```diff
@@ -270,6 +273,115 @@ def test_gen_batch_initial_conditions_warning(self):
...
+                transform_vertical_constraint(constraint=constraint, d=3)
+
+    def test_gen_batch_initial_conditions_transform_constraints(self):
+        for dtype in (torch.float, torch.double):
+            constraints = [
+                (
+                    torch.tensor([0, 1], dtype=torch.int64, device=self.device),
+                    torch.tensor([-1.0, -1.0]).to(dtype=dtype, device=self.device),
```
Balandat, line 344:
```suggestion
                    torch.tensor([-1.0, -1.0], dtype=dtype, device=self.device),
```

Balandat (commented):
Thanks! I think the horizontal / vertical constraint naming is not really clear until you already understand the setup very well. How about naming those something like intra/inter point constraints? Or pointwise / cross-batch? Open to other naming suggestions as well. 

Other than that this lgtm apart from docstrings.

botorch/optim/initializers.py
```diff
@@ -169,10 +236,18 @@ def gen_batch_initial_conditions(
...
+                        inequality_constraints=transform_constraints(
+                            constraints=inequality_constraints, q=q, d=bounds.shape[1]
+                        )
+                        if inequality_constraints is not None
+                        else None,
+                        equality_constraints=transform_constraints(
+                            constraints=equality_constraints, q=q, d=bounds.shape[1]
+                        )
+                        if equality_constraints is not None
+                        else None,
```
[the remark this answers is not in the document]
jduerholt: Done.

botorch/optim/initializers.py
jduerholt: I decided for intra-point.

test/optim/test_initializers.py
jduerholt: DOne.

botorch/optim/initializers.py
jduerholt: Good point, I did some easy test and looked at the autocorrelation of the samples and indeed it has an influence:

In the old way, I get for a scenario with a q-batch of 4 and 1024 raw samples the following autocorrealtion:

``` python
import numpy as np
import torch
from botorch.utils.sampling import get_polytope_samples
from botorch.optim.initializers import transform_constraints
import matplotlib.pyplot as plt


tkwargs = {'dtype': torch.float64, 'device': 'cpu'}

bounds = torch.tensor([[0,0,0,0,0], [1,1,1,1,1]]).to(**tkwargs)

equalities = [
    (torch.tensor([0,1,2,3], dtype=torch.int64), torch.tensor([1,1,1,1]).to(**tkwargs), 1.)
]

inequalities = [
    (torch.tensor([0,1], dtype=torch.int64), torch.tensor([1,1]).to(**tkwargs), 0.3)
]

q = 4
raw_samples=1024

samples = get_polytope_samples(
    n=q*raw_samples,
    bounds=bounds,
    inequality_constraints=inequalities,
    equality_constraints=equalities,
    n_burnin=10000,
    thinning=32
)

samples = samples.detach().numpy()

plt.acorr(samples[:,0], maxlags=24,normed=True)
plt.show()
```

image

Doing it the new way via sampling from a q*d space, I get the following:

``` python
from botorch.optim.initializers import transform_constraints

samples2 = get_polytope_samples(
    n=raw_samples,
    bounds=torch.hstack([bounds for _ in range(q)]),
    inequality_constraints=transform_constraints(inequalities,q=q,d=5),
    equality_constraints=transform_constraints(equalities,q=q,d=5),
    n_burnin=10000,
    thinning=32
)

samples2 = samples2.reshape(torch.Size((q*raw_samples, 5)))

samples2 = samples2.detach().numpy()

plt.acorr(samples2[:,0], maxlags=24,normed=True)
plt.show()
```
image

We see higher correlation for every fourth sample, which makes sense as four samples here correspond to one polytope sample in the new setup. Setting `thinning` to `32*q` cures the problem:

image

So I changed the default to always 32*q.

What do you think? Is this sufficient?

botorch/optim/initializers.py
```diff
@@ -63,6 +63,93 @@
...
+    constraint: Tuple[Tensor, Tensor, float], d: int, q: int
+) -> List[Tuple[Tensor, Tensor, float]]:
+    """Transforms an intra-point/pointwise constraint from
+    d-dimensional space to a d*q dimesional space.
+
+    Args:
+        constraint (Tuple[Tensor, Tensor, float]): Constraints
+            to transform.
+        d (int): Size of the q-batch.
+        q (int): Dimensionality of the problem.
```
Balandat, line 102:
```suggestion
        constraints: A list of tuples (indices, coefficients, rhs), with each tuple encoding an
            (in-)equality constraint of the form `\sum_i (X[indices[i]] * coefficients[i]) (>)= rhs`.
            Here `indices` must be one-dimensional, and the constraint is applied to all points
            within the `q`-batch.
        d: Dimensionality of the problem.
```

botorch/optim/initializers.py
```diff
@@ -63,6 +63,93 @@
...
+    Raises:
+        ValueError: If indices in the constraints are larger than the
+            dimensionality d of the problem.
+
+    Returns:
+        List[Tuple[Tensor, Tensor, float]]: List of transformed constraints.
+    """
+    indices, coefficients, rhs = constraint
+    if indices.max() >= d:
+        raise ValueError("d has to be larger than the largest index in the constraint.")
```
Balandat, line 113: Kind of a weird way to put this, how about
```suggestion
        raise ValueError(f"Constraint indices cannot exceed the problem dimension {d=}.")
```

botorch/optim/initializers.py
```diff
@@ -63,6 +63,93 @@
...
+    constraint: Tuple[Tensor, Tensor, float], d: int
+) -> Tuple[Tensor, Tensor, float]:
+    """Transforms an inter-point constraint from
+    d-dimensional space to a d*q dimesional space.
+
+    Args:
+        constraint (Tuple[Tensor, Tensor, float]): Constraint
+            to transform.
+        d (int): Size of the q-batch.
+        q (int): Dimensionality of the problem.
```
Balandat, line 134:
```suggestion
    Args:
        constraints: A list of tuples (indices, coefficients, rhs), with each tuple encoding an
            (in-)equality constraint of the form `\sum_i (X[indices[i]] * coefficients[i]) (>)= rhs`.
            `indices` must be a 2-d Tensor, where in each row `indices[i] = (k_i, l_i)` the first index
            `k_i` corresponds to the `k_i`-th element of the `q`-batch and the second index `l_i`
            corresponds to the `l_i`-th feature of that element.
```

botorch/optim/initializers.py
```diff
@@ -63,6 +63,93 @@
...
+    Raises:
+        ValueError: If indices in the constraints are larger than the
+            dimensionality d of the problem.
+
+    Returns:
+        List[Tuple[Tensor, Tensor, float]]: Transformed constraint.
+    """
+    indices, coefficients, rhs = constraint
+    if indices[:, 1].max() >= d:
+        raise ValueError("d has to be larger than the largest index in the constraint.")
```
Balandat, line 145:
```suggestion
        raise ValueError(f"Constraint indices cannot exceed the problem dimension {d=}.")
```

botorch/optim/initializers.py
Balandat: Actually, one concern here is that you are transforming these constraints regardless of whether there are any inter-point / cross-batch constraints present. That means that in that case we're running the hit-and-run sampler in an unnecessarily high-dimensional space, and could instead just run it for `n*q` samples as we did before. Could you please modify this logic here to do this branching? Would be good to move this whole branch starting in l 257 out into a helper function to keep things tidy.

Balandat (changes requested):
Thanks for iterating on this. I do have one request regarding not unnecessarily increasing the dimensionality for the sampler if there aren't any inter-point / cross-batch constraints present.

jduerholt:
> Thanks for iterating on this. I do have one request regarding not unnecessarily increasing the dimensionality for the sampler if there aren't any inter-point / cross-batch constraints present.

Done ;)

botorch/optim/initializers.py
```diff
@@ -63,6 +63,184 @@
...
+    equality_constraints: Optional[List[Tuple[Tensor, Tensor, float]]] = None,
+) -> Tensor:
+    r"""Samples `n` q-baches from a polytope of dimension `d`.
+
+    Args:
+        n: Number of q-batches to sample.
+        q: Number of samples per q-batch
+        bounds: A `2 x d` tensor of lower and upper bounds for each column of `X`.
+        n_burnin: The number of burn-in samples for the Markov chain sampler.
+        thinning: The amount of thinning.
```
Balandat, line 185:
```suggestion
        thinning: The amount of thinning (number of steps to take between returning samples).
```

botorch/optim/initializers.py
```diff
@@ -63,6 +63,184 @@
...
+    if inequality_constraints is not None:
+        for c in inequality_constraints:
+            if len(c[0].shape) > 1:
+                inter_point = True
+                break
+    if inter_point is False and equality_constraints is not None:
+        for c in equality_constraints:
+            if len(c[0].shape) > 1:
+                inter_point = True
+                break
```
Balandat, line 209: Really more a stylistic preference than a necessity.
```suggestion
    inter_point = any(
        indices.shape > 1
        for constraints in (inequality_constraints or [], equality_constraints or [])
        for indices, _, _ in constraints
    )
```

botorch/optim/initializers.py
```diff
@@ -63,6 +63,184 @@
...
+            bounds=bounds,
+            inequality_constraints=inequality_constraints,
+            equality_constraints=equality_constraints,
+            seed=seed,
+            n_burnin=n_burnin,
+            thinning=thinning,
+        )
+        .view(n, q, -1)
+        .cpu()
+    )
```
Balandat, line 241: Again a stylistic / conciseness nit

```suggestion
    if inter_point:
        samples = get_polytope_samples(
            n=n,
            bounds=torch.hstack([bounds for _ in range(q)]),
            inequality_constraints=transform_constraints(
                constraints=inequality_constraints, q=q, d=bounds.shape[1]
            ),
            equality_constraints=transform_constraints(
                constraints=equality_constraints, q=q, d=bounds.shape[1]
            ),
            seed=seed,
            n_burnin=n_burnin,
            thinning=thinning * q,
        )
    else:
        samples = get_polytope_samples(
            n=n * q,
            bounds=bounds,
            inequality_constraints=inequality_constraints,
            equality_constraints=equality_constraints,
            seed=seed,
            n_burnin=n_burnin,
            thinning=thinning,
        )
    return samples.view(n, q, -1).cpu()
```

Balandat (approved):
Most excellent, this looks great. A couple of tiny nits, but o/w ready to merge in!

jduerholt:
@Balandat: I just incorporated your suggestions. Should be final now ;)

Balandat (commented):
Almost - actually, this throws a test failure on the GPU due to device mismatch, see inline comment for the fix.

jduerholt:
> Looks like one more `ufmt` pass is needed. Home stretch.

Should be fine now ;) I want to have a button for this directly in github ;)

Balandat:
>  I want to have a button for this directly in github ;) 

GitHub may accept PRs as well...

Balandat:
Found another cuda issue, but fixed that internally. All good from your end. Thanks!

Issue #1759, opened 2023-03-23
[Bug] Weird `bounds` handling in `Ackley` (and potentially others)


saitcakmak:
# 🐛 Bug

This is a fun one. Let's initialize `Ackley(bounds=[(0.0, 1.0), (0.0, 1.0)])`. We can expect both `self.bounds` and `self._bounds` to return `[0, 1]` bounds, right? Turns out it is a bit more complicated than that.

In `Ackley.__init__`, we first ignore the `bounds` kwarg and set `self._bounds = [(-32.768, 32.768) for _ in range(self.dim)]`. Since we ignored the bounds, then both `self.bounds` and `self._bounds` should be `[-32.768, 32.768]`? Nope...

We call `super().__init__` with `bounds=bounds`. The `super` is `SyntheticTestFunction`, which, after some checks, sets `self._bounds` to tensorized `bounds`. So, `[0, 1]` again?

But before doing this `SyntheticTestFunction.__init__` also calls `super().__init__` (no bounds passed down this time). `super` is `BaseTestProblem`, which registers `self._bounds` as the `bounds` buffer. Since this happened between `Ackley.__init__` setting `self._bounds` to `[-32.768, 32.768]` and `SyntheticTestFunction` setting `self._bounds` to `[0, 1]`, we ended up with `self.bounds` as `[-32.768, 32.768]` (tensorized) but `self._bounds` as `[0, 1]` (list of tuples). 

## To reproduce

** Code snippet to reproduce **
```python
func = Ackley(bounds=[(0.0, 1.0), (0.0, 1.0)])
func.bounds, func._bounds
```

## Expected Behavior

Both bounds are set to `[0, 1]`.

PR #1760, opened 2023-03-23
Fix custom bounds handling in test problems

saitcakmak:
Summary: Fixes https://github.com/pytorch/botorch/issues/1759

Differential Revision: D44326798

### DIFF
--- a/test/test_functions/test_synthetic.py
+++ b/test/test_functions/test_synthetic.py
@@ -4,6 +4,8 @@
 # This source code is licensed under the MIT license found in the
 # LICENSE file in the root directory of this source tree.
 
+import torch
+from botorch.exceptions.errors import InputDataError
 from botorch.test_functions.synthetic import (
     Ackley,
     Beale,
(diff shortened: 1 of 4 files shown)


test/test_functions/test_synthetic.py
```diff
@@ -45,8 +47,34 @@ class DummySyntheticTestFunctionWithOptimizers(DummySyntheticTestFunction):
 
 
 class TestSyntheticTestFunction(BotorchTestCase):
+    functions_with_custom_bounds = [
```
Balandat, line 50: maybe add a comment what the int value here is

Balandat (commented):
Some tests are still failing, o/w this generally lgtm, thanks.

test/test_functions/test_synthetic.py
```diff
@@ -57,6 +85,19 @@ def test_custom_bounds(self):
...
+            )
+        )
+
+        # Test each function with custom bounds.
+        for func_class, dim in self.functions_with_custom_bounds:
+            bounds = [(-1e5, 1e5) for _ in range(dim)]
+            bounds_tensor = torch.tensor(bounds).T
+            func = func_class(bounds=bounds)
+            self.assertEqual(func._bounds, bounds)
+            self.assertTrue(torch.allclose(func.bounds, bounds_tensor))
```
esantorella, line 100: Thank you for this!

saitcakmak:
> Do these problems need to have a _bounds attribute at all? It seems like self._bounds is never accessed after it’s set. ([code search](https://github.com/search?q=repo%3Apytorch%2Fbotorch+self._bounds&type=code)) I feel a lot better about having both now that we have the unit test that they contain the same values though!

Do they need to? Probably not. They've had it for a long time, so it makes sense to keep it around. I generally wouldn't rely on code search to find usage of test problems since they'll mostly be used for benchmarking by us or external users. I guess this makes it a bit harder to catch non-breaking bugs with them as well.

PR #1762, opened 2023-03-25
Exit publish_site.sh script on error

Balandat:
Adds `set -e` to the publish_site.sh script to exit the script if any command has non-zero exit code. This avoids silent failures in the automated website deployment on GHA.


esantorella (approved):
Thanks! Seems like we should use this in most of the scripts.

Balandat:
> Thanks! Seems like we should use this in most of the scripts.

Yeah we should follow up on that.

PR #1765, opened 2023-03-26
Fix minimize test due to issue with rounding np array element

Balandat:
When running tests locally I ran into the following error: 

self = <test.optim.utils.test_timeout.TestMinimizeWithTimeout testMethod=test_minimize_with_timeout>

    def test_minimize_with_timeout(self):
        def f_and_g(x: np.ndarray, sleep_sec: float = 0.0):
            time.sleep(sleep_sec)
            return x**2, 2 * x

        base_kwargs = {
            "fun": f_and_g,
            "x0": np.array([1.0]),
            "method": "L-BFGS-B",
            "jac": True,
            "bounds": [(-2.0, 2.0)],
        }

        with self.subTest("test w/o timeout"):
            res = minimize_with_timeout(**base_kwargs)
            self.assertTrue(res.success)
            self.assertAlmostEqual(res.fun, 0.0)
            self.assertAlmostEqual(res.x, 0.0)
            TypeError: type numpy.ndarray doesn't define __round__ method
            

This adds an `item()` to the single-element arrays to return a python `float` instead. Not sure why this doesn't fail in the CI, presumably it's a scipy/numpy versioning issue? Probably not worth digging deep here if this works both locally and on the CI.

### DIFF
--- a/test/optim/utils/test_timeout.py
+++ b/test/optim/utils/test_timeout.py
@@ -30,14 +30,14 @@ class TestMinimizeWithTimeout(BotorchTestCase):
             res = minimize_with_timeout(**base_kwargs)
             self.assertTrue(res.success)
             self.assertAlmostEqual(res.fun, 0.0)
-            self.assertAlmostEqual(res.x, 0.0)
+            self.assertAlmostEqual(res.x.item(), 0.0)
             self.assertEqual(res.nit, 2)  # quadratic approx. is exact
 
         with self.subTest("test w/ non-binding timeout"):
             res = minimize_with_timeout(**base_kwargs, timeout_sec=1.0)
             self.assertTrue(res.success)
             self.assertAlmostEqual(res.fun, 0.0)
-            self.assertAlmostEqual(res.x, 0.0)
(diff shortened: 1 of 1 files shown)


esantorella:
I wonder if this is a flaky or machine-dependent numerical error. `assertAlmostEqual` first checks for equality then only tries rounding if the arguments aren't equal. At least on my laptop,`self.assertAlmostEqual(res.x, 0.0)` passes but `self.assertAlmostEqual(res.x, 1e-10)` raises the error about "doesn't define `__round__`."

I think we shouldn't be passing tensors or numpy arrays to `assertAlmostEqual` at all, which we are doing in a few other places. One solution would be to modify [`BotorchTestCase.assertAllClose`](https://github.com/pytorch/botorch/blob/9302055f36981e3413e8dce08ae050bd701f5e7a/botorch/utils/testing.py#L65) for all checks for near-equality and have it dispatch to `torch.testing.assert_close`, `np.testing.assert_allclose`, or `unittest.assertAlmostEqual` as appropriate. But low priority because this hasn't been failing.

esantorella (approved):

PR #1774, opened 2023-04-04
Change batch range for FullyBayesianPosterior

Balandat:
Summary: Changes the batch range of FullyBayesianPosterior, which effectively means that we will regard the MC batch as a t-batch for the purposes of sampling.

Reviewed By: dme65

Differential Revision: D44599898


Balandat:
Merged as part of #1773

Issue #1776, opened 2023-04-04
[Feature Request] Make `gen_batch_initial_conditions` more flexible


jduerholt:
# 🚀 Feature Request

## Motivation

When optimizing an acqf it could be possible that the default starting point sampler is not sufficient (for example when dealing with non-linear constraints or NChooseK constraints). In these case one can provide a initializer method via the `ic_generator` argument or samples directly  via the `batch_initial_conditions` keyword. When looking at the default generator `gen_batch_initial_conditions` one sees that there is lot of logic besides simple sampling included, namely boltzmann weighting of the samples, catching of zero acqf values etc. 

I propose to make this logic more broadly available to have a easy way to use it in situations where the default intialization does not work and custom samplers has to be used.

## Pitch

For this purpose, I propose to add a new optional argument to `gen_batch_initial_conditions` which accepts a callable of type `Callable[[int, int], Tensor]` that just returns samples that are then further processed (boltzmann weighted based on the acqf values etc.) The two arguments of the function would be `n` and `q` and the returned Tensor would be of shape `n x q x d`. 

What do you think?

**Are you willing to open a pull request?** Yes.

Balandat:
Sounds like a great improvement to me! Happy to review a PR for this.

PR #1777, opened 2023-04-04
Fixed slashes in ////zeta in BOPE tutorial

esantorella:
## Motivation

A pandas update broke the CI. This fixes it.

### Have you read the [Contributing Guidelines on pull requests](https://github.com/pytorch/botorch/blob/main/CONTRIBUTING.md#pull-requests)?

Yes

## Test Plan

Repro'd the error locally and confirmed that this fixes it. The CI should pass now.

PR #1779, opened 2023-04-05
Make gen_batch_initial_conditions more flexible

jduerholt:
## Motivation

This PR adds the feature regarding additional flexibility of `gen_batch_initial_conditions` as discussed in issue https://github.com/pytorch/botorch/issues/1776.

Yes.

## Test Plan

Unit tests.


jduerholt:
Hmm, is the error in the tutorial pipeline really caused by this PR, no or?

One other thing: 

I just saw that `optimize_acqf_mixed` do not have the `ic_generator` and `ic_gen_kwargs` arguments. From my perspective, they should be added also there, or?

Balandat:
> Hmm, is the error in the tutorial pipeline really caused by this PR, no or?

No you're all good, this was an issue due to some pandas upgrade, this was fixed in #1777.

botorch/optim/initializers.py
```diff
@@ -274,6 +275,9 @@ def gen_batch_initial_conditions(
         equality constraints: A list of tuples (indices, coefficients, rhs),
             with each tuple encoding an inequality constraint of the form
             `\sum_i (X[indices[i]] * coefficients[i]) = rhs`.
+        generator: Callable for generating samples that are then further
+            processed. It receives `n`, `q` and `seed` as arguments and  returns
```
Balandat, line 279: should we make `seed` optional here?

botorch/optim/initializers.py
```diff
@@ -297,6 +301,11 @@ def gen_batch_initial_conditions(
             "Option 'sample_around_best' is not supported when equality"
             "constraints are present."
         )
+    if sample_around_best and generator:
+        raise UnsupportedError(
+            "Option 'sample_around_best' is not supported when custom"
+            "generator should be used."
```
Balandat, line 307:
```suggestion
            "Option 'sample_around_best' is not supported when custom "
            "generator is be used."
```

botorch/optim/initializers.py
```diff
@@ -327,7 +336,9 @@ def gen_batch_initial_conditions(
     while factor < max_factor:
         with warnings.catch_warnings(record=True) as ws:
             n = raw_samples * factor
-            if inequality_constraints is None and equality_constraints is None:
+            if generator:
```
Balandat, line 339:
```suggestion
            if generator is not None:
```

test/optim/test_initializers.py
```diff
@@ -595,6 +595,80 @@ def test_gen_batch_initial_conditions_interpoint_constraints(self):
...
+            gen_batch_initial_conditions(
+                MockAcquisitionFunction(),
+                bounds=torch.tensor([[0, 0], [1, 1]], **tkwargs),
+                q=1,
+                num_restarts=1,
+                raw_samples=1,
+                generator=generator,
+                options={"sample_around_best": True},
+            )
+            self.assertTrue(
```
Balandat, line 667: could instead use `assertRaisesRegex`, which makes this easier / cleaner.

Balandat (approved):
a few nits and and there, but otherwise this looks great. Thanks!

test/optim/test_initializers.py
jduerholt: Done.

botorch/optim/initializers.py
jduerholt: I would keep it mandatory, as a seed is always used in `gen_batch_initial_conditions` and the seed number is increased in every iteration.

Balandat:
Thanks @esantorella! Should be an easy fix; once that's in this should be good to land!

jduerholt:
Thanks @esantorella for fixing it. Is there anything to do for me or is now everything fixed?

esantorella:
> Thanks @esantorella for fixing it. Is there anything to do for me or is now everything fixed?

Looks to me like everything is fixed now! Nothing more should be needed on your end.

PR #1784, opened 2023-04-06
Fix typo in saasbo tutorial

jduerholt:
## Motivation

In the saasbo tutorial it was stated that the posterior is made up of 16 MC samples, which is wrong. There are only 8 samples, since 128 MC steps are performed in combination with a thinning of 8. This amounts to 8 posterior samples.

Furthermore, I noticed that the link to the "custom acquistion tutorial" in the introduction of the notebook is not working, but I do not know where it should link to originally.

Yes.


Balandat (approved):
thanks for fixing this!

PR #1791, opened 2023-04-14
SampleReducingMCAcquisitionFunction

SebastianAment:
Summary:
This primary purpose of this commit is to introduce `SampleReducingMCAcquisitionFunction`, which enables incorporating constraints to acquisition functions in a general and principled manner.
[See this overleaf doc](https://www.overleaf.com/read/zmkgyzcyffjs
) for context on how this differs to the current approach using `SampleReducingMCAcquisitionFunction` and why it could matter for BO performance.

In addition, the new approach leads to increased code sharing of existing `MCAcquisitionFunctions`.

Differential Revision: D43990659

### DIFF
--- a/botorch/acquisition/acquisition.py
+++ b/botorch/acquisition/acquisition.py
@@ -140,6 +140,14 @@ class MCSamplerMixin(ABC):
             )
         return self.sampler(posterior=posterior)
 
+    @property
+    def sample_shape(self) -> torch.Size:
+        return (
+            self.sampler.sample_shape
+            if self.sampler is not None
+            else self._default_sample_shape
+        )
+
 
 class MultiModelAcquisitionFunction(AcquisitionFunction, ABC):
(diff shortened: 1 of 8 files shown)


jduerholt:
Hi Sebastian, this looks interesting. Unfortunately, I have no permission to open the overleaf document. Could you share a pdf version? 

Best,

Johannes

SebastianAment:
@jduerholt Hi Johannes, I updated the overleaf link with a sharable one.

jduerholt:
> @jduerholt Hi Johannes, I updated the overleaf link with a sharable one.

Thanks! Makes definitely sense for me. In a next step one could then also think about updating https://github.com/pytorch/botorch/blob/646b6c1ac3d5c8c7c4e9af2c835be23478db84e6/botorch/acquisition/utils.py#L38

If constraints are provided then one would returned the constrained acqf based on the requested acqf.

Best, 

Johannes

PR #1792, opened 2023-04-14
Fix warnings from acquisition function tests

esantorella:
## Motivation

Lots of warnings are raised by tests. This is annoying. Also, sometimes these indicate a problem that needs to be fixed. For example, BoTorch is using some deprecated functionality.

This isn't all the warnings by any means; I fixed almost all of those stemming from test/acquisition/multi_objective and a few in test/acquisition.

* I replaced deprecated functionality with appropriate replacement
* I silenced some warnings when the test has intentionally done the thing the warning is complaining about. For example, using single precision
* Didn't do anything about numerical warnings

## Test Plan

Mostly changing tests.

### DIFF
--- a/test/acquisition/test_proximal.py
+++ b/test/acquisition/test_proximal.py
@@ -5,6 +5,7 @@
 # LICENSE file in the root directory of this source tree.
 
 from typing import List
+import warnings
 
 import torch
 from botorch.acquisition import LinearMCObjective, ScalarizedPosteriorTransform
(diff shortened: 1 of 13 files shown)


test/acquisition/test_proximal.py
```diff
@@ -56,11 +58,25 @@ def test_proximal(self):
...
+                            message=_get_single_precision_warning(torch.float32),
+                        )
+                        warnings.filterwarnings(
+                            "ignore",
+                            message=(
+                                "Input data is not contained to the unit cube. "
+                                "Please consider min-max scaling the input "
+                                "data.",
+                            ),
+                        )
```
Balandat, line 73: Could we instead of filtering the warnings just standardize the data (for the inputs we can do it manually when not using a `Normalize` transform)?

Balandat (approved):
This is awesome, down with warnings!

test/acquisition/test_proximal.py
esantorella: For this one, yeah. I think there were some tests where it looked like the data was intentionally not standardized.

esantorella: Yep. It turns out that the single-precision warning is already suppressed by default by `BotorchTestCase.setUp`, but that didn't happen here since `setUp` was overridden. So I added a call to `super().setUp()`.

PR #1793, opened 2023-04-16
Feature/non linear q

jduerholt:
## Motivation

This PR adds the possibilities of q>1 for optimizing acqfs with nonlinear constraints. I just removed the blocking of q>1 in `gen_candidates_scipy` and tested it with NChooseK constraints. It worked. So I am wondering, why it was blocked originally. Any idea?

I can also provide an example showing that it works.

Yes.

## Test Plan

Unit tests.


Balandat:
Nice. 

> I just removed the blocking of q>1 in gen_candidates_scipy and tested it with NChooseK constraints. I can also provide an example showing that it works.

An example would be great. I wonder if a short-but-sweet mini-tutorial for parameter constraints (incl. a simple linear constraint and e.g. more complicated NChooseK constraints) would make sense? 

> So I am wondering, why it was blocked originally. Any idea?

I think it may just have been something we didn't test / weren't sure about when we first added the experimental support for nonlinear constraints.

jduerholt:
> An example would be great. I wonder if a short-but-sweet mini-tutorial for parameter constraints (incl. a simple linear constraint and e.g. more complicated NChooseK constraints) would make sense?

So you recommend a small jupyter notebook within `tutorials`?

Balandat:
> So you recommend a small jupyter notebook within `tutorials`?

Yes, I think that would be great. Maybe a simple synthetic function and then show a few different ways to optimize an acquisition function subject to parameter constraints (both linear and nonlinear). No need to make a full Bayesian Optimization loop out of it.

jduerholt:
Ok, sounds good. Do you want to have it as part of this PR, or specific one?

jduerholt:
Small update on this PR, we should not yet merge it. It looks that it was  somehow by coincidence that the non-linear constraint was fulfilled for q>1. In more elaborate experiments, I found that this is not always the case. I have to dig there a bit deeper ;)

jduerholt:
The point is the following:

Just commenting out applies the constraint over the flattened q-batch, which means the constraint is applied inter-point wise, but we want to have it always as intra-point, or do we want to have options for both? I would say, always go for intra-point is sufficient.

Balandat:
> Just commenting out applies the constraint over the flattened q-batch, which means the constraint is applied inter-point wise, but we want to have it always as intra-point, or do we want to have options for both? I would say, always go for intra-point is sufficient.

Hmm and the additional complexity compared to the linear constraints is that there is no standardized interface for passing in intra- or inter-point constraints, right? And we'd like to make it easier for folks to apply a nonlinear constraint that's defined for a single element of the domain to all elements of a q-batch, right? I think then restricting to intra-point should be ok, so long as we make clear in the docstring that that's the restriction.

jduerholt:
Other option would be to change the signature of the nonlinear constraints from `Optional[List[Callable]]` to `Optional[List[Tuple[Callable, bool]]]` where the bool indicates if it is intra or interpoint ...

Balandat:
> Other option would be to change the signature of the nonlinear constraints from `Optional[List[Callable]]` to `Optional[List[Tuple[Callable, bool]]]` where the bool indicates if it is intra or interpoint ...

Hmm that's an interesting idea as well. I guess this is a rather advanced feature so I wouldn't be too worried about the interface being a bit clunkier, so that sounds good to me if we think that there will be use cases for non-linear inter-point constraints.

jduerholt:
One can come up with use cases, the question is how realistically they are. But, in general I think it is good to keep it as flexible as possible, I will try to implement it as described above.

jduerholt:
I found some time to implement the prototype of what I had in mind.

This script tests it on the 6dim Hartmann, with NChooseK. Most of the times it seems to work but sometimes it returns candidates where the constraint eval leads to values that are substantially smaller than zero. Not sure what is going on there:

```
import torch
from botorch.acquisition import ExpectedImprovement, qExpectedImprovement
from botorch.fit import fit_gpytorch_model
from botorch.models.transforms import Standardize
from botorch.test_functions import Hartmann

def narrow_gaussian(x, ell):
    return torch.exp(-0.5 * (x / ell) ** 2)


def ineq_constraint(x, ell=1e-3):
    # Approximation of || x ||_0 <= 3. The constraint is >= 0 to conform with SLSQP
    return narrow_gaussian(x, ell).sum(dim=-1) - 3


def get_feasible_sobol_points(n):
    """Sobol sequence where we randomly set three of the parameters to zero to satisfy the constraint"""
    X = SobolEngine(dimension=6, scramble=True).draw(n).to(torch.double)
    inds = torch.argsort(torch.rand(n, 6), dim=-1)[:, :3]
    X[torch.arange(X.shape[0]).unsqueeze(-1), inds] = 0
    return X


def get_batch_initial_conditions(num_restarts, raw_samples, q, acqf):
    X = get_feasible_sobol_points(raw_samples*q).unsqueeze(1)
    X = X.reshape((torch.Size((raw_samples,q,6))))
    acq_vals = acqf(X)
    return X[acq_vals.topk(num_restarts).indices]


hartmann = Hartmann(dim=6)

X = get_feasible_sobol_points(10)
Y = hartmann(X).unsqueeze(-1)
print(f"Best initial point: {Y.min().item():.3f}")

gp = SingleTaskGP(X, Y, outcome_transform=Standardize(m=1))
fit_gpytorch_model(mll)

EI = qExpectedImprovement(model=gp, best_f=Y.min(), maximize=False)

batch_initial_conditions = get_batch_initial_conditions(num_restarts=1, raw_samples=512, acqf=EI, q=3)

print(batch_initial_conditions)

    EI,
    bounds=torch.cat((torch.zeros(1, 6), torch.ones(1, 6))),
    q=3,
    nonlinear_inequality_constraints=[ineq_constraint],
    batch_initial_conditions=batch_initial_conditions,
    num_restarts=20,
)

print(candidate)

print(ineq_constraint(candidate))
```

Balandat:
> sometimes it returns candidates where the constraint eval leads to values that are substantially smaller than zero. Not sure what is going on there:

Interesting; I'll take a look.

Balandat:
I looked into this a bit, and every time the constraints are infeasible it's b/c `scipy.minimize` failed. Reasons in the warnings include  

> OptimizationWarning('Optimization failed within `scipy.optimize.minimize` with status 8 and message Positive directional derivative for linesearch.')]

or 

> OptimizationWarning('Optimization failed within `scipy.optimize.minimize` with status 5 and message Singular matrix E in LSQ subproblem.')

There also seem to be some failures that don't result in infeasible candidates, but those are rare. haven't figured out what's going on exactly, but this simply seems to be due to the optimization not terminating properly.

Out of 100: 
No warnings, feasible: 91
No warnings, infeasible: 0
1 warning, feasible: 1
1 warning, infeasible: 8

jduerholt:
Hi Max, 

thanks for investigating this. 

So how to proceed? Is the proposed solution fine for you? 

Should we remove the old solution which was somehow able for interpoint constraints or keep it and give the user the possibility to toggle between them?

Best,

Johannes

jduerholt:
Hi Max, 

any opinion to the question from above?

Best,

Johannes

Balandat:
Sorry, I missed this. Yeah, let's give the user the possibility to toggle between inter-point and intra-point constraints.

Balandat:
@jduerholt checking in on the status of this - are you still planning on merging this in? Anything needed from our end at this point?

jduerholt:
Hi Max, it is still on my menu. I hope to get it finalized in November. There were a lot of other things to do, so I had to reprioritize ...

Balandat:
No worries, looking forward to it :)

botorch/generation/gen.py
```diff
@@ -47,7 +47,7 @@ def gen_candidates_scipy(
     upper_bounds: Optional[Union[float, Tensor]] = None,
     inequality_constraints: Optional[List[Tuple[Tensor, Tensor, float]]] = None,
     equality_constraints: Optional[List[Tuple[Tensor, Tensor, float]]] = None,
-    nonlinear_inequality_constraints: Optional[List[Callable]] = None,
+    nonlinear_inequality_constraints: Optional[List[Tuple[Callable, bool]]] = None,
```
Balandat, line 50: Thinking about this a bit more, I am somewhat concerned that this backward-incompatible change may cause a bunch of breakages for users. One thing we could do is both accept the old and new format of the constraints (defaulting to the previous intrapoint behavior) and emit a `DeprecationWarning` if the old one is used (we'd keep the type as only the new one to indicate that that is the correct one to use). You could write a little helper function for this that checks the format, converts to the new one and raises a warning if it's the old one - you can then use that helper in different places to avoid code duplication.

That said, this is pretty advanced functionality so I don't expect a ton of people to actually use this, and so we might just be ok hard breaking this...

@saitcakmak any thoughts on this?

botorch/generation/gen.py
```diff
@@ -69,11 +69,14 @@ def gen_candidates_scipy(
...
-            non-linear inequality constraints of the form `callable(x) >= 0`. Each
+        nonlinear_inequality_constraints: A list of tuples, one tuple per
+            constraint, first element of the tuple is a callable that represents
+            a non-linear inequality constraint of the form `callable(x) >= 0`. Each
             callable is expected to take a `(num_restarts) x q x d`-dim tensor as
             an input and return a `(num_restarts) x q`-dim tensor with the
-            constraint values. The constraints will later be passed to SLSQP.
+            constraint values. The second element is a boolean indicating it is
+            an intrapoint or interpoint constraint. `True` for intrapoint,
+            `False` for interpoint. The constraints will later be passed to SLSQP.
```
Balandat, line 79: We should explain what the difference between interpoint and intrapoint constraint are, both here and in other places. We could refer to some other docstring if this would get to long as part of this one.

botorch/generation/gen.py
```diff
@@ -254,8 +252,10 @@ def f(x):
     # SLSQP sometimes fails in the line search or may just fail to find a feasible
     # candidate in which case we just return the starting point. This happens rarely,
     # so it shouldn't be an issue given enough restarts.
+    # TODO: this needs to be updated with the two options
```
Balandat, line 255: reminder

botorch/optim/parameter_constraints.py
```diff
@@ -312,8 +312,35 @@ def _make_linear_constraints(
...
+            f_obj, f_grad = _make_f_and_grad_nonlinear_inequality_constraints(
+                f_np_wrapper=f_np_wrapper,
+                nlc=get_intrapoint_constraint(b=i, q=j, nlc=nlc),
+            )
+            constraints.append(
+                {
+                    "type": "ineq",
+                    "fun": f_obj,
+                    "jac": f_grad,
+                }
```
Balandat, line 336:
```suggestion
                {"type": "ineq", "fun": f_obj, "jac": f_grad}
```

botorch/optim/parameter_constraints.py
```diff
@@ -479,20 +508,31 @@ def make_scipy_nonlinear_inequality_constraints(
         )
 
     scipy_nonlinear_inequality_constraints = []
-    for nlc in nonlinear_inequality_constraints:
+    for constraint in nonlinear_inequality_constraints:
+        nlc = constraint[0]
```
Balandat, line 512:
```suggestion
    for nlc, is_intrapoint in nonlinear_inequality_constraints:
```

botorch/optim/parameter_constraints.py
```diff
@@ -479,20 +508,31 @@ def make_scipy_nonlinear_inequality_constraints(
...
-        )
-        scipy_nonlinear_inequality_constraints.append(
-            {
-                "type": "ineq",
-                "fun": f_obj,
-                "jac": f_grad,
-            }
-        )
+
+        if constraint[1] is True:
```
Balandat, line 521:
```suggestion
        if is_intrapoint:
```

botorch/generation/gen.py
jduerholt: Good point, so we accept the old format and translate it to Tuple[Callable, bool], where the bool is set to `True` indicating an intrapoint constraint. Only question would be up to which level, we are supporting this. I would recommend only up to `optimize_acqf` and `gen_candidates_scipy`. Not deeper. Ok?

jduerholt: It is also a good idea, because than I do not break BoFire, where I currently need only the intra ones ;)

Balandat: >  would recommend only up to optimize_acqf and gen_candidates_scipy. Not deeper. Ok?

Makes sense to me!

botorch/generation/gen.py
jduerholt: Done

botorch/optim/parameter_constraints.py
jduerholt: This is outdated now.

botorch/generation/gen.py
jduerholt: Fully agree, still thinking where to do this in the best way ...

botorch/generation/gen.py
jduerholt: Done.

jduerholt:
Hi @Balandat,

the current status is how I would do it. Note that this introduces some substantial changes in the handling of nonlinear constraints:
- In the original version, only `q=1` and a `batch_limit` of one was allowed for optimizations with nonlinear constraints (https://github.com/pytorch/botorch/blob/d0d3ad8ff47f23c39976ab39e198f7de4d29ce67/botorch/optim/optimize.py#L291). 
- For this reason, in contrast to the linear constraint no broadcasting at all was necessary for the nonlinear ones as it was evaluated always on a tensor of shape ``1 x 1 x d`. In this context also the original docstring is a bit misleading from my perspective.
- In this PR, I introduce now broadcasting, both along the batch and q-dimension. In case of an intrapoint point constraint, this constraint will be broadcasted `q x b` times and will be evaluated just on a one-dim tensor of length `d`. An inter-point constraint will be broadcased `b` times and will be always evaluated on a 2-dimensional tensor of shape `q x d`. 
- All this means, that one could in principle also lift the `batch_limit` of one for optimizations with nonlinear constraints.

Before updating the docstrings, it would be nice if you could have a look at the PR and tell me if it is fine for you as I have implemented it, and if I overlooked some pitfalls or necessary tests ;)

Best,

Johannes

botorch/generation/gen.py
```diff
@@ -254,14 +260,17 @@ def f(x):
...
-    ):
-        candidates = torch.from_numpy(x0).to(candidates).reshape(shapeX)
-        warnings.warn(
-            "SLSQP failed to converge to a solution the satisfies the non-linear "
-            "constraints. Returning the feasible starting point."
-        )
+    if nonlinear_inequality_constraints:
+        for constraint in nonlinear_inequality_constraints:
+            if not nonlinear_constraint_is_fulfilled(
+                constraint[0], constraint[1], candidates
```
Balandat, line 266:
```suggestion
        for con, is_intrapoint in nonlinear_inequality_constraints:
            if not nonlinear_constraint_is_fulfilled(
                con, is_intrapoint=is_intrapoint, x=candidates
```

botorch/optim/parameter_constraints.py
```diff
@@ -449,10 +482,40 @@ def f_grad(X):
...
+    b, q, _ = x.shape
+    if not intra:  # this is an interpoint constraint
+        for i in range(b):
+            if _arrayify(nonlinear_inequality_constraint(x[i])).item() < NLC_TOL:
+                return False
+    else:  # this is an intrapoint constraint
+        for i in range(b):
+            for j in range(q):
+                if _arrayify(nonlinear_inequality_constraint(x[i, j])).item() < NLC_TOL:
+                    return False
```
Balandat, line 510:
```suggestion
    def check_x(x: Tensor) -> bool:
        return _arrayify(nonlinear_inequality_constraint(x)).item() < NLC_TOL
    
    for x_ in x:
        if is_intrapoint:
            return all(check_x(x__) for for x__ in x_)
         else:
            return check_x(x_)
```

botorch/optim/parameter_constraints.py
```diff
@@ -449,10 +482,40 @@ def f_grad(X):
...
+    return True
```
Balandat, line 511:
```suggestion
```

botorch/optim/parameter_constraints.py
```diff
@@ -472,27 +535,26 @@ def make_scipy_nonlinear_inequality_constraints(
...
+            raise ValueError(
+                f"A nonlinear constraint has to be a tuple, got {type(constraint)}."
+            )
+        if len(constraint) != 2:
+            raise ValueError(
+                "A nonlinear constraint has to be a tuple of length 2, "
+                f"got length {len(constraint)}."
+            )
+        nlc, is_intra = constraint
+        if not nonlinear_constraint_is_fulfilled(nlc, is_intra, x0.reshape(shapeX)):
```
Balandat, line 551: just some formatting nits; let's call most functions with kwargs (except where things are very obvious)
```suggestion
        nlc, is_intrapoint = constraint
        if not nonlinear_constraint_is_fulfilled(nlc, is_intrapoint=is_intrapoint, x= x0.reshape(shapeX)):
```

test/optim/test_optimize.py
```diff
@@ -930,25 +950,25 @@ def nlc4(x):
...
-                optimize_acqf(
-                    acq_function=mock_acq_function,
-                    bounds=bounds,
-                    q=1,
-                    nonlinear_inequality_constraints=[nlc1],
-                    batch_initial_conditions=torch.rand(5, 1, 3, **tkwargs),
-                    num_restarts=5,
-                    options={"batch_limit": 5},
-                )
+            # with self.assertRaisesRegex(
```
Balandat, line 958: lets add this and the restriction back in for now given my comment above

Balandat (commented):
Thanks for the updates! Overall lgtm - @dme65 any concerns about this? 

> For this reason, in contrast to the linear constraint no broadcasting at all was necessary for the nonlinear ones as it was evaluated always on a tensor of shape ``1 x 1 x d`. In this context also the original docstring is a bit misleading from my perspective.

Yeah I guess the term broadcasting there is used in the sense of how the constraint is interpreted, not necessarily that there is (torch-level) broadcasting happening. Functionally though from the outside this is effectively "broadcasting" as in applying the constraint to all batch elements.

> All this means, that one could in principle also lift the batch_limit of one for optimizations with nonlinear constraints.

That is true. The main challenge is that SLSQP (the default method used here if there are nonlinear constraints) really sucks and becomes very slow if there are a large number of variables and constraints. So this limitation is mostly there to protect against that. In some previous tests I have seen that chunking up a batch of say 20 restarts into 20 individual optimizations problems when using SLSQP was a lot faster than using batches of size >1. So I'd prefer not to lift this restriction on the batch limit unless we run some benchmarks to re-evaluate whether this is still the case. In the future we can hopefully plug in better general purpose nonlinear solvers in which case it makes sense to remove this limitation.

jduerholt:
Thanks for the review. I will update the PR over the course of the week ;)

botorch/generation/utils.py
```diff
@@ -18,6 +20,36 @@
...
+    for nlc in nonlinear_inequality_constraints:
+        if callable(nlc):
+            # old style --> convert
+            nlcs.append((nlc, True))
+            warnings.warn(
+                "The `nonlinear_inequality_constraints` argument is expected "
+                "take a list of tuples. Passing a list of callables "
+                "will result in an error in future versions.",
+                DeprecationWarning,
+            )
```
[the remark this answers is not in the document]
jduerholt: Done.

test/optim/test_optimize.py
jduerholt: Done.

botorch/generation/utils.py
```diff
@@ -18,6 +20,36 @@
...
 
 
+def _convert_nonlinear_inequality_constraints(
+    nonlinear_inequality_constraints: List[Union[Callable, Tuple[Callable, bool]]]
+) -> List[Tuple[Callable, bool]]:
+    """Convert legacy defintions of nonlinear inequality constraints into the new
+    format. Assumes intra-point constraints.
+    """
+    nlcs = []
+    if not isinstance(nonlinear_inequality_constraints, list):
```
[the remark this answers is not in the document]
jduerholt: I moved it two levels up into `gen_candidates_scipy` in `gen.py.`

botorch/generation/gen.py
jduerholt: I also updated the doc strings.

jduerholt:
Hi @Balandat,

I think I am done with all your suggestions. You can have a look again.

Best,

Johannes

test/generation/test_utils.py
```diff
@@ -9,15 +9,40 @@
...
+    _convert_nonlinear_inequality_constraints,
     _flip_sub_unique,
     _remove_fixed_features_from_optimization,
 )
 from botorch.utils.testing import BotorchTestCase, MockAcquisitionFunction
 
 
 class TestGenerationUtils(BotorchTestCase):
+    def test_convert_nonlinear_inequality_constraints(self):
+        for dtype in (torch.float, torch.double):
```
Balandat, line 24: Don't think you need to test for different `dtype`s here? Not used right now...

Balandat (commented):
Thanks. Mostly cosmetic changes, once those are in I'm happy to merge this

botorch/generation/gen.py
```diff
@@ -125,6 +133,16 @@ def gen_candidates_scipy(
         else:
             reduced_domain = None not in fixed_features.values()
 
+    if nonlinear_inequality_constraints:
+        if not isinstance(nonlinear_inequality_constraints, list):
+            raise ValueError(
+                "`nonlinear_inequality_constraints` must be a list of tuples, "
+                f"got {type(nonlinear_inequality_constraints)}."
+            )
```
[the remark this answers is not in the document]
jduerholt: The legacy format is supported here (have a look at the following lines), here it is only checked that a list is provided.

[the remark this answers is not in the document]
Balandat: my bad

botorch/optim/optimize.py
```diff
@@ -450,24 +449,31 @@ def optimize_acqf(
...
             `make_scipy_linear_constraints` for an example.
-        nonlinear_inequality_constraints: A list of callables with that represent
-            non-linear inequality constraints of the form `callable(x) >= 0`. Each
-            callable is expected to take a `(num_restarts) x q x d`-dim tensor as an
-            input and return a `(num_restarts) x q`-dim tensor with the constraint
-            values. The constraints will later be passed to SLSQP. You need to pass in
-            `batch_initial_conditions` in this case. Using non-linear inequality
-            constraints also requires that `batch_limit` is set to 1, which will be
-            done automatically if not specified in `options`.
+        nonlinear_inequality_constraints: A list of tuples representing the nonlinear
```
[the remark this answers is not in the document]
jduerholt: Done.

botorch/optim/optimize.py
```diff
@@ -744,14 +749,20 @@ def optimize_acqf_list(
...
             `\sum_i (X[indices[i]] * coefficients[i]) = rhs`
-        nonlinear_inequality_constraints: A list of callables with that represent
-            non-linear inequality constraints of the form `callable(x) >= 0`. Each
-            callable is expected to take a `(num_restarts) x q x d`-dim tensor as an
-            input and return a `(num_restarts) x q`-dim tensor with the constraint
-            values. The constraints will later be passed to SLSQP. You need to pass in
-            `batch_initial_conditions` in this case. Using non-linear inequality
-            constraints also requires that `batch_limit` is set to 1, which will be
-            done automatically if not specified in `options`.
+        nonlinear_inequality_constraints: A list of tuples representing the nonlinear
```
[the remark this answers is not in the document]
jduerholt: Done.

botorch/optim/parameter_constraints.py
```diff
@@ -449,50 +506,91 @@ def f_grad(X):
...
+    shapeX: torch.Size,
 ) -> List[Dict]:
     r"""Generate Scipy nonlinear inequality constraints from callables.
 
     Args:
-        nonlinear_inequality_constraints: List of callables for the nonlinear
-            inequality constraints. Each callable represents a constraint of the
-            form >= 0 and takes a torch tensor of size (p x q x dim) and returns a
-            torch tensor of size (p x q).
+        nonlinear_inequality_constraints: A list of tuples representing the nonlinear
```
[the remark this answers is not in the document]
jduerholt: Done.

botorch/optim/parameter_constraints.py
```diff
@@ -449,50 +506,91 @@ def f_grad(X):
...
+            of an inter-point constraint, it takes a two dimensional torch tensor of
+            type (q x d) and returns again a scalar. The second element is a boolean,
+            indicating if it is an intra-point or inter-point constraint. `True` for
+            intra-point. `False` for inter-point.
         f_np_wrapper: A wrapper function that given a constraint evaluates the value
              and gradient (using autograd) of a numpy input and returns both the
              objective and the gradient.
         x0: The starting point for SLSQP. We return this starting point in (rare)
             cases where SLSQP fails and thus require it to be feasible.
+        shapeX: Shape of the theedimensional batch X, that should be optimized.
```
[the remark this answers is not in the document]
jduerholt: Ooopsie, I forgot an "r".

botorch/optim/parameter_constraints.py
```diff
@@ -312,8 +312,64 @@ def _make_linear_constraints(
...
+            the value and gradient (using autograd) of a numpy input and returns both
+            the objective and the gradient.
+        nlc: Callable representing a constraint of the form >= 0 and takes
+            in case of an intra-point constraint an one-dimensional torch tensor of
+            length `d` and returns a scalar. In case of an inter-point constraint, it
+            takes a two dimensional torch tensor of type (q x d) and returns again a
+            scalar.
+        is_intrapoint: A Boolean indicating if a constraint is an intra-point or
+            inter-point constraint.
+        shapeX: Shape of the theedimensional batch X, that should be optimized.
```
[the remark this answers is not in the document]
jduerholt: Done.

test/generation/test_utils.py
jduerholt: I think it is not needed, as I am not doing anything with tensors here, so I removed it.

Balandat (approved):
Many thanks!

Issue #1733, opened 2023-03-08
[Feature Request] Acquisition Functions for Top-K Estimation


Ryan-Rhys:
# 🚀 Feature Request

Acquisition function implementation for the task of top-k estimation with diversity as per the following paper:

https://openreview.net/pdf?id=coQhmtxr5SN

## Motivation

Top-k estimation (estimating a set of k optimal designs under a penalty that encourages diversity) is an important task in applications such as virtual screening for drug discovery and materials design.

## Additional context

@sangttruong is prepared to open a PR for his implementation.

eytan:
Awesome! Thank you so much for your contributions in advance, Sang!

Ryan, FYI https://arxiv.org/pdf/2210.01383.pdf is probably a more canonical
reference since that is the published NeurIPS 2022 version.

Since many acquisition functions can be built off of HES/EHIG, if it’s not
too hard, I wonder if it’s worth implementing something other than top-k at
the same time, just to set up a good separation of concerns and class
structure. For example, a generic EHIG class that was an abstract base
class with a (differentiable) loss function method (from which top-k is a
subclass of), or some kind of wrapper that returns a new AF with a given
loss function (IIRC we do something similar to this with probabilistic
reparameterization).

To test the separation of concerns, it could be helpful to try implementing
another simple EHIG AF, such as k-guesses from that paper.

E

Balandat:
This would be great to have. More than happy to provide guidance on the implementation. 

As Eytan said, a generic setup for this would be nice. Maybe there are also opportunities to share some of the generic components with the entropy-based acquisition functions that were added recently.

sangttruong:
Thank you for your comment.  I agree it would be good to have an EHIG abstract based class with a differentiable loss function. I will also write up a few test AFs from the paper, such as k-guess and top-k, to test out the separation of concerns and class structure.

eytan:
Hi @sangttruong, just checking in on this. It would be really exciting to get this framework and AF into BoTorch.  Let us know if there is any way we can help here, and we are happy to review intermediate diffs to provide feedback.

sangttruong:
Dear @eytan, I apologize for the delayed response -- it has been quite hectic starting the quarter here at Stanford. The good news is that I have the code almost ready with various loss functions, and I am trying to complete a few details before submitting a pull request. I also have a corresponding tutorial notebook for the sanity check. I am eager to receive your feedback.

eytan:
No worries!  Looking forward to it, thanks again!

PR #1794, opened 2023-04-16
Adding H-Entropy Search and a corresponding tutorial

sangttruong:
## Motivation

This pull request introduces an implementation of the H-Entropy Search [1] procedure, along with a tutorial covering two tasks: Top-K Search and MinMax Search. This PR addresses issue #1733, which requested a new acquisition function for top-K search.
[1] W. Neiswanger, L. Yu, S. Zhao, C. Meng, S. Ermon. Generalizing Bayesian Optimization with Decision-theoretic Entropies. In Proceedings of the 36th Conference on Neural Information Processing Systems (NeurIPS 2022).

Yes, I have read the Contributing Guidelines on pull requests. 

## Test Plan

In the associated tutorial, two tests can be executed using the built-in 2D Ackley function with an input range of [-1, 1]. Within this range, there are four maxima located at each corner and one minimum at the center. The H-Entropy Search (HES) procedure, employing a Top-K (K = 2) loss function, is expected to identify two of the four maxima, with any combination being acceptable. Meanwhile, the H-Entropy Search procedure utilizing the MinMax loss function should locate one maximum and the central minimum.

Below are the results after running 10 steps of HES with MinMax:
image

And here are the results after running 10 steps of HES with TopK:
image

Both images can be reproduced by following the respective tutorial.

## Related PRs

Not applicable. 


botorch/acquisition/hentropy_search.py
```diff
@@ -0,0 +1,275 @@
...
+
+        Returns:
+            A Tensor of shape `n_fantasy_at_action_pts x batch`.
+        """
+
+        assert A.shape[-2] == 2
+
+        Y[:, :, :, 0] = -1 * Y[:, :, :, 0]
+
+        q_hes = Y.sum(dim=-1).mean(dim=0)
```
Balandat, line 272: This is problematic as you're changing the input tensor in-place. That is, the `Y` that you passed to `forward` outsize will change once you call the objective on it. How about this?
```suggestion
        q_hes = (Y[..., 1:].sum(dim=-1) - Y[..., 0]).mean(dim=0)
```

botorch/acquisition/hentropy_search.py
```diff
@@ -0,0 +1,275 @@
...
+from botorch.sampling.normal import SobolQMCNormalSampler
+from botorch.utils.transforms import t_batch_mode_transform
+from torch import Tensor
+
+
+class qHEntropySearch(MCAcquisitionFunction, OneShotAcquisitionFunction):
+    def __init__(
+        self,
+        model: Model,
+        loss_function_class: Type[MCAcquisitionObjective],
```
Balandat, line 45: Hmm why is this `Type[MCAcquisitionObjective]`? Looks like the loss functions you're defining below are just `nn.Module`s?

Balandat (commented):
Thanks a lot for contributing this! I made a first pass over the `hentropy_search.py` module, but did not review the tutorial yet.

Right now it doesn't look like the tutorial actually contains any plots. It would be great if you could modify things so that the plots show inline so people who view the tutorial can see the results even before running the tutorial themselves.

esantorella:
Thanks a lot for the PR! A couple things:

- BoTorch requires 100% test coverage. Could you add unit tests for the H-entropy search functionality? To check test coverage, you can run `pytest -ra --cov=. --cov-report term-missing` as described in [CONTRIBUTING.md](https://github.com/pytorch/botorch/blob/main/CONTRIBUTING.md), or just wait for the Codecov analysis on GitHub.
- Could you run `black .` and `ufmt format .` as described in CONTRIBUTING.md? This will get the linter passing.

botorch/acquisition/hentropy_search.py
```diff
@@ -0,0 +1,275 @@
...
+        loss_function_hyperparameters: Dict[str, Any],
+        n_fantasy_at_design_pts: Optional[int] = 64,
+        n_fantasy_at_action_pts: Optional[int] = 64,
+        design_sampler: Optional[MCSampler] = None,
+        action_sampler: Optional[MCSampler] = None,
```
esantorella, line 50: This has a very different signature than `MCAcquisitionFunction.__init__`. This will make it harder to compare this acquisition function against other acquisition functions or use it within a "glue" framework like Ax that allows for using different acquisition functions with a common interface. It will also make static type checkers like Mypy or Pyre fail.

We have not been requiring type checkers to pass on BoTorch, and we already have a lot of acquisition functions that don't adhere to the common interface defined by `MCAcquisitionFunction` and `AcquisitionFunction`, so this signature isn't doing anything unusual. However, it would be helpful if the signature were as similar as possible to the signature for [`MCAcquisitionFunction.__init__`](https://github.com/pytorch/botorch/blob/646b6c1ac3d5c8c7c4e9af2c835be23478db84e6/botorch/acquisition/monte_carlo.py#L56). Is that possible?

botorch/acquisition/hentropy_search.py
```diff
@@ -0,0 +1,275 @@
...
+            X: Design tensor of shape `batch x q x num_dim_design`.
+            A: Action tensor of shape `batch x n_fantasy_at_design_pts
+                x num_actions x num_dim_action`.
+
+        Returns:
+            A Tensor of shape `batch`.
+        """
+
+        # construct the fantasy model of shape `n_fantasy_at_design_pts x b`
+        model = copy.deepcopy(self.model)
```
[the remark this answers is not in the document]
sangttruong: Thank you for the question. It appears that the issue is related to the gradient of X being retained by self.model.fantasize even after calling optimizer.zero_grad(), leading to an error later on ("Trying to backward through the graph a second time..."). To address this issue, we have two potential solutions: either create a deep copy of self.model using model = copy.deepcopy(self.model), or set retain_graph=True when calling loss.backward(). I believe the latter approach may have a higher memory consumption. Do you happen to know of any alternative approaches that can avoid the need for both deep copy and saving the graph?

botorch/acquisition/hentropy_search.py
sangttruong: The H-Entropy Search acquisition function needs a loss function that is tailored to the problem at hand. However, incorporating this function into the initialization process may potentially disrupt the signature of MCAcquisitionFunction.init. If you have any insights on how to circumvent this issue, please let me know. I would really like to keep the signature of H-Entropy Search consistent with MCAcquisitionFunction as well.

botorch/acquisition/hentropy_search.py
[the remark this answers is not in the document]
sangttruong: Hi @Balandat! I hope you're having a great day. If you have a moment, could you take a look at my question posted above? I appreciate any suggestions you may have. Thank you!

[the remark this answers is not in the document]
Balandat: Sorry for the late response, I was out last week.

What exactly is the code you're running when you're observing this behavior? Do you have a repro?

Note that in other acquisition functions that call `fantasize` we don't need to deepcopy the model or set `retain_graph=True`, see e.g. for the Knowledge gradient here: https://github.com/pytorch/botorch/blob/main/botorch/acquisition/knowledge_gradient.py#L194-L197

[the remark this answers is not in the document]
sangttruong: Thank you for the reply. This behavior is observed when running the [corresponding tutorial of H-Entropy Search](https://github.com/sangttruong/botorch/blob/sang-1733-adding-hentropy-search/tutorials/hentropy_search.ipynb). I'm aware that Knowledge Gradient doesn't need a deep copy or retain graph, but I wasn't sure why this was possible in KG and not in HES.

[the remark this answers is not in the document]
sangttruong: I talked to @Ryan-Rhys today about this and he suggests that I post the error message here. I include below the output when turning off both deep copy and retain_graph when running the corresponding tutorial. It's just a generic autograd error that isn't very helpful when debugging the acquisition function. I remember talking with @willieneis about a year ago about this error, but I don't think we have a better way to address this, other than either deep copy or setting retain_graph=True during backward, as the error message suggested. 

```
BO round: 0, Optim round: 0, Loss: -1311.93
  File "hentropy_search.py", line 256, in <module>
    from botorch.acquisition.hentropy_search import (
  File "/dfs/user/sttruong/botorch/botorch/acquisition/hentropy_search.py", line 486, in <module>
    loss.backward()
  File "/dfs/user/sttruong/env/geometric_ampere/lib/python3.8/site-packages/torch/_tensor.py", line 488, in backward
    torch.autograd.backward(
  File "/dfs/user/sttruong/env/geometric_ampere/lib/python3.8/site-packages/torch/autograd/__init__.py", line 197, in backward
    Variable._execution_engine.run_backward(  # Calls into the C++ engine to run the backward pass
RuntimeError: Trying to backward through the graph a second time (or directly access saved tensors after they have already been freed). Saved intermediate values of the graph are freed when you call .backward() or autograd.grad(). Specify retain_graph=True if you need to backward through the graph a second time or if you need to access saved tensors after calling backward.

```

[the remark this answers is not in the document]
Ryan-Rhys: So I guess `torch.optim` is the issue 🤔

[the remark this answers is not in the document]
sangttruong: @Balandat @esantorella  Hi Max and Elizabeth! Have you had a chance to look further into this? It seems like we always have this issue when using SGD `torch.optim` with acquisition functions that call `fantasize`, including the KG acquisition function. The reason KG works fine on the tutorial on the BoTorch page is that it uses a built-in optimizer in BoTorch. Please let me know if we should keep either the `deepcopy` call or using `retain_graph = True` when calling the optimizer. Thank you so much!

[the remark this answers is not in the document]
Balandat: @sangttruong I took a look at this - I believe the reason why autograd is complaining is that we are backpropagating through the train-train covariance cache of the base (non-fantasized) model when calling `backward()`. In particular, in the very first `backward()` call this  doesn't happen due to the fact that the caches haven't been filled. The reason this doesn't cause any issues with the botorch optimizers is b/c under the hood we're calling `torch.autograd.grad(loss, X)` rather than `loss.backward()`, so we aren't backpropagating through the train-train cache of the model (since they do not depend on `X`). I am not an expert on the `torch.optim` module internals, but it seems somewhat wasteful to compute gradients for variables that aren't optimization variables in the optimizer. I wonder if there is a way to do this?

In the meantime setting `retain_graph=True` should "fix" the problem - though it seems kind of wasteful since we don't actually need to backprop twice through anything. But it seems a lot less wasteful than deepcopying the whole model. You can also just benchmark the runtime of the two versions. 

Let me know how ii goes!

Balandat:
Sorry this docs failure isn't exactly actionable - you'll have to add an entry to this json file to link it on the website: https://github.com/pytorch/botorch/blob/main/website/tutorials.json

Speaking of the tutorial - can you push a version that has the outputs included as well? Could you please also make the saving the results (plots) to disk optional and not do that by default?

Balandat:
Thanks for tying up all the loose ends. One more outstanding issue: The tutorial is pretty heavyweight, so it'll cause some issues for the CI. We usually use a `SMOKE_TEST` variable in the notebook to select parameters to make it much cheaper tu run (e.g. number of iterations, data size, number of mc samples, etc). See e.g. how this is used in https://github.com/pytorch/botorch/blob/main/tutorials/closed_loop_botorch_only.ipynb. Could you please add this to the notebook to make sure that it runs quickly? It doesn't have to produce interesting outputs, this is mostly to test in an automated fashion that it's still functional and didn't break due to other code changes. Thanks!

Balandat (approved):
Thanks, this lgtm!

sangttruong:
@esantorella @Ryan-Rhys Do you have any other comments or feedback on this PR?

botorch/acquisition/hentropy_search.py
```diff
@@ -0,0 +1,282 @@
...
+
+        Args:
+            batch_as_full: A `b x (q + num_fantasies) x d`-dim Tensor with `b`
+                t-batches of `q + num_fantasies` design points each.
+
+        Returns:
+            A `b x q x d`-dim Tensor with `b` t-batches of `q` design points each.
+        """
+
+        assert len(batch_as_full.shape) == 2
```
esantorella, line 184: Is the docstring wrong? The docstring makes it sound like this would be 3d.

Also, we don't usually use asserts -- something like a `ValueError` would be better. If you were to add a ValueError here, you'd need to add a unit test for it. You could also just remove this line.

botorch/acquisition/hentropy_search.py
```diff
@@ -0,0 +1,282 @@
...
+from botorch.acquisition.monte_carlo import MCAcquisitionFunction
+from botorch.models.model import Model
+from botorch.sampling.base import MCSampler
+from botorch.sampling.normal import SobolQMCNormalSampler
+from botorch.utils.transforms import t_batch_mode_transform
+from torch import Tensor
+
+
+class qHEntropySearch(MCAcquisitionFunction, OneShotAcquisitionFunction):
+    r"""H-Entropy Search using one-shot optimization."""
```
esantorella, line 42: Nit: This doesn't actually need to be a raw string, and some linters will complain if a raw string is used unnecessarily. It's fine for BoTorch though.

esantorella:
> @esantorella @Ryan-Rhys Do you have any other comments or feedback on this PR?

Sorry for the slow response! It looks like this is still missing unit tests; in particular for `acquisition/hentropy_search.py`. See my comment above on this, and see [here](https://github.com/pytorch/botorch/tree/main/test/acquisition) for how other acquisition functions have been tested. Thanks for your patience!

botorch/acquisition/hentropy_search.py
```diff
@@ -0,0 +1,282 @@
...
+
+
+class qLossFunctionTopK(nn.Module):
+    r"""Batch loss function for the task of finding top-K."""
+
+    def __init__(self, dist_weight=1.0, dist_threshold=0.5) -> None:
+        r"""Batch loss function for the task of finding top-K.
+
+        Args:
+            loss_function_hyperparameters: hyperparameters for the loss function class.
```
esantorella, line 206: Docstring doesn't match arguments

botorch/acquisition/hentropy_search.py
```diff
@@ -0,0 +1,282 @@
...
+            A_distance = torch.cdist(A, A, p=1.0)
+            A_distance_triu = torch.triu(A_distance)
+            # >>> n_fantasy_at_design_pts x batch_size x num_actions x num_actions
+
+            A_distance_triu[A_distance_triu > self.dist_threshold] = self.dist_threshold
+            denominator = num_actions * (num_actions - 1) / 2.0
+            dist_reward = A_distance_triu.sum((-1, -2)) / denominator
+            # >>> n_fantasy_at_design_pts x batch_size
+
+        q_hes = Y + self.dist_weight * dist_reward
```
esantorella, line 246: As written, I think this will error if Y is on the GPU, since `self.dist_weight` is a CPU tensor. (Does `dist_weight` need to be a tensor at all?)

```suggestion
        q_hes = Y + self.dist_weight.to(Y.device) * dist_reward
```

esantorella (changes requested):
I left a few small comments, but unit tests are the most important thing.

botorch/acquisition/hentropy_search.py
```diff
@@ -0,0 +1,282 @@
...
+This file has an implementation of Batch H-Entropy Search (qHES) via
+one-shot optimization as introduced in [Neiswanger2022]_.
+
+The authors adopt a generalized definition of entropy from past work
+in Bayesian decision theory, which proposes a family of decision-theoretic
+entropies parameterized by a problem-specific loss function and
+action set. Therefore, the method allow the development of a common
+acquisition optimization procedure, which applies generically
+to many members of this family (where each member is induced by a
+specific loss function and action set).
```
Ryan-Rhys, line 17: Perhaps it would be nice to add a brief definition of what the action set is? From the user's perspective they will probably be content with knowing that an action represents some point in the input space!

Ryan-Rhys: or a set of points for tasks such as top-k estimation!

botorch/acquisition/hentropy_search.py
```diff
@@ -0,0 +1,282 @@
...
+        batch_xs = batch_xs.reshape(n_restarts, -1, self.cfg.n_dim_design)
+        batch_as_full = batch_as_full.reshape(
+            n_restarts, self.cfg.n_actions, self.cfg.n_dim_action
+        )
+
+        return batch_xs
+
+
+class qLossFunctionTopK(nn.Module):
+    r"""Batch loss function for the task of finding top-K."""
```
Ryan-Rhys, line 200: Might be worth specifying that top-K is relative to the values of the objective function?

botorch/acquisition/hentropy_search.py
```diff
@@ -0,0 +1,282 @@
...
+            # >>> n_fantasy_at_design_pts x batch_size
+
+        q_hes = Y + self.dist_weight * dist_reward
+        # >>> n_fantasy_at_design_pts x batch_size
+
+        return q_hes
+
+
+class qLossFunctionMinMax(nn.Module):
+    r"""Batch loss function for the task of finding min and max."""
```
Ryan-Rhys, line 253: As above, might be worth specifying that min and max are relative to the values of the objective function?

botorch/acquisition/hentropy_search.py
```diff
@@ -0,0 +1,282 @@
...
+        if num_actions >= 2:
+            A = A.contiguous()
+            A_distance = torch.cdist(A, A, p=1.0)
+            A_distance_triu = torch.triu(A_distance)
+            # >>> n_fantasy_at_design_pts x batch_size x num_actions x num_actions
+
+            A_distance_triu[A_distance_triu > self.dist_threshold] = self.dist_threshold
+            denominator = num_actions * (num_actions - 1) / 2.0
+            dist_reward = A_distance_triu.sum((-1, -2)) / denominator
+            # >>> n_fantasy_at_design_pts x batch_size
```
Ryan-Rhys, line 244: These notes are expected shapes I'm guessing?

botorch/acquisition/hentropy_search.py
```diff
@@ -0,0 +1,282 @@
...
+
+
+class qLossFunctionMinMax(nn.Module):
+    r"""Batch loss function for the task of finding min and max."""
+
+    def __init__(self) -> None:
+        r"""Loss function for task of finding min and max.
+
+        Args:
+            loss_function_hyperparameters: hyperparameters for the loss function class
```
Ryan-Rhys, line 259: Are these `loss_function_hyperparameters` used anywhere?

sangttruong: The min-max loss function doesn't have an argument, so I can remove line 259.

botorch/acquisition/hentropy_search.py
sangttruong: I added the note "relative to the values of the objective function" to the doc string.

botorch/acquisition/hentropy_search.py
sangttruong: That's a good idea. I add a brief description of the action set to the doc string.

botorch/acquisition/hentropy_search.py
```diff
@@ -0,0 +1,282 @@
...
+        # >>> n_fantasy_at_design_pts x batch_size
+
+        num_actions = A.shape[-2]
+
+        dist_reward = 0
+        if num_actions >= 2:
+            A = A.contiguous()
+            A_distance = torch.cdist(A, A, p=1.0)
+            A_distance_triu = torch.triu(A_distance)
+            # >>> n_fantasy_at_design_pts x batch_size x num_actions x num_actions
```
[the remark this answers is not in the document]
sangttruong: Yes, the shape provided in line 239 is for the A_distance_triu tensor. I added some more shape comments to other tensors to make it more clear.

botorch/acquisition/hentropy_search.py
sangttruong: The output shape should be 3D. I adopt the docstring from the MST file: https://github.com/pytorch/botorch/blob/main/botorch/acquisition/multi_step_lookahead.py#L283

sangttruong:
@Ryan-Rhys Ry @esantorella Thank you so much for the feedback. I added a unit test file and adjust the main class according to your comments above. Please let me know if there is any further adjustment I should make.

botorch/acquisition/hentropy_search.py
```diff
@@ -0,0 +1,275 @@
...
+        elif n_fantasy_at_design_pts is not None:
+            if design_sampler.sample_shape != torch.Size([n_fantasy_at_design_pts]):
+                raise ValueError((
+                    "The design_sampler shape must match n_fantasy_at_design_pts="
+                    f"{n_fantasy_at_design_pts}."
+                ))
+        else:
+            n_fantasy_at_design_pts = design_sampler.sample_shape[0]
+
+        if action_sampler is None:
```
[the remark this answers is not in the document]
Balandat: would still be nice to make that into a little helper to avoid code duplication

botorch/acquisition/hentropy_search.py
Balandat: looks like `dist_weight` is not a tensor anymore (though docstring needs updating). If that's always the case this comment should be resolved.

botorch/acquisition/hentropy_search.py
```diff
@@ -0,0 +1,277 @@
...
+
+        super().__init__()
+        self.dist_weight = dist_weight
+        self.dist_threshold = dist_threshold
+
+    def forward(self, A: Tensor, Y: Tensor) -> Tensor:
+        r"""Evaluate batch loss function on a tensor of actions.
+
+        Args:
+            A: Actor tensor with shape `n_fantasy_at_design_pts x batch_size
```
Balandat, line 209: Nit: Inconsistency: Used "Action" tensor above but "Actor" tensor here - let's standardize this throughout the code.

Balandat: Also used `num_dim_action` and `action_dim` for the same thing

tutorials/hentropy_search.ipynb
```diff
@@ -0,0 +1,454 @@
...
+    "    XY = torch.tensor(np.array([X, Y]))\n",
+    "    Z = func(XY.reshape(2, -1).T).reshape(X.shape)\n",
+    "    cs = ax.contourf(X, Y, Z, levels=30, cmap=\"bwr\", alpha=0.7)\n",
+    "    ax.set_aspect(aspect=\"equal\")\n",
+    "    cbar = fig.colorbar(cs)\n",
+    "    cbar.ax.set_ylabel(\"$f(x)$\", rotation=270, labelpad=20)\n",
+    "\n",
+    "    # Plot data, optimal actions, next query ###########################################\n",
+    "    ax.scatter(data_x[:, 0], data_x[:, 1], label=\"Data\")\n",
+    "    ax.scatter(optimal_action[:, 0], optimal_action[:, 1], label=\"Action\")\n",
```
Balandat, line 126: The default orange color is quite hard to see on the red background - could you change the color here to make it more prominent?

tutorials/hentropy_search.ipynb
```diff
@@ -0,0 +1,454 @@
...
...nFAgMD0aJFCxQQmubfffdddOvWDQMGDPA4/uyzz+LTTz/Fpk2bcMstt+CBBx7AmyIbJc6fPx9RUVHuV3x8vPwPpAfs2BoNydm1C/aOHTFy8mRZ5yUMGoSFy5Z5HLvjxhtxaMsWLYdHIcEXYoahfXvyeJrGvEt4Y6QRuphKS13TQc2rtNTVV3x8vMe9aP78+bzXPHv2LBwOB1q3bu1xvHXr1oL3z6NHj2LNmjVwOBzYsGEDnn76abz66qt4/vnnVX1+PfG5y+mJJ57Aiy++KNrmgAbR55cuXcLKlSvx9NNPe73HPtavXz9UVFTg5Zdfxj//+U/evmbNmoWsrCz3z2VlZfqLmnoHUJgHVJ4FwlsBrZOBAJ69oXRwQb37ySd4cMIEvPvppzhTWIg2nEkhh7DQUISFhmoyLvKLmix1myQgmCRdW24hPV+IGfa1T5wgj6chiaWh2yBYm0YoZrTm1KlTHns5hWiYZVtfX4+YmBi88847sNvtSE5OxunTp/Hyyy9jzpw5ml1HS3xuoXnkkUdw4MAB0VeHDh0QGxuLoqIij3Pr6upQUlJCFPuyZs0aVFZWYvz48ZJtU1NT8ffffwua70JCQhAZGenxkqS62rX9AftFyvFNwKfDgQ2Tga2Pub5+Otx1nA8NLTXlFRX4ZP16TBs3DiOvuw7vr1nj8f5X332Hq0aPRmjXrmh15ZW4+R//AAAMGTsWJ06fxsPPPQdbYiJsiYkA+F1Oiz/6CB2vvRbBXbqg6/XX48PPP/d435aYiP+tWoWb//EPhHfrhs7XXYcvNwl8dq2RuzmlGblwwbdihg2JpYZEpJG6nU6eJGtHMQ4mi0lvF9OhQ54vP4R7HxISNK1atYLdbkdhYaHH8cLCQsH7Z1xcHLp06QI7a1Plbt26oaCgADVM9XuT4XNBEx0djaSkJNFXcHAw0tLScOHCBeTl5bnP3bJlC+rr65Gamip5nXfffRc33XQTogkyKHbv3o3mzZsrV7vsSXT8OFB3eZfsJk08X1yBwydyjm8CNmcBFZ7/iKgoch3XWdR8un49kjp2RNeOHXH3mDFYtnq1O75o/ZYtuPn++3HDkCH4bf16bP7oI6RcLtj1+ZIluCIuDs8+/DDyc3ORL5AeuHbjRsx89lk8cu+9+D07G/+46y5MeuwxfM/x68594w3cPnIk9n7zDW4YMgTjHn4YJY05LZ
```
Balandat, line 332: is this needed?

Balandat, line 330: this is called twice here?

tutorials/hentropy_search.ipynb
```diff
@@ -0,0 +1,454 @@
...
+    "        if i % 200 == 0:\n",
+    "            print(f\"Eval optim round: {i}, Loss: {loss.item():.2f}\")\n",
+    "\n",
+    "    A = torch.sigmoid(A) * (ba_u - ba_l) + ba_l\n",
+    "    eval_metric = qhes.loss_function(A=A, Y=fantasized_outcome)\n",
+    "    eval_metric = eval_metric[0, 0].cpu().detach().numpy()\n",
+    "    optimal_action = A[0, 0].cpu().detach().numpy()\n",
+    "    value = qhes.loss_function(A=A, Y=func(A))[0, 0].cpu().detach().numpy()\n",
+    "    data_x = data.x.cpu().detach().numpy()\n",
+    "    next_x = next_x.cpu().detach().numpy()\n",
```
Balandat, line 103: can probably get rid of a bunch of the detaches here by using a `no_grad()` context

tutorials/hentropy_search.ipynb
Balandat, line 338: You can pass the `bounds` directly to the constructor of `Ackley`, no need to modify those in-place ex-post.

Balandat, line 301: It's not exactly clear in the output what the iterations are. Could you explain / delineate those better in the outputs?

Balandat (commented):
Left some more comments, I think once those are addressed we're ready to merge this in!

tutorials/hentropy_search.ipynb
sangttruong: Thank you for the comment. I wasn't able to see what some questions are about since the GitHub comment doesn't point to a specific line in the notebook. Could you please clarify this question, as well as the one below about something being called twice?

sangttruong: Similar to the two questions above, I wasn't able to see what lines in the code this question is about. Could you please help me to clarify this?

test/acquisition/test_hentropy_search.py
```diff
@@ -0,0 +1,221 @@
...
+NO = "botorch.utils.testing.MockModel.num_outputs"
+
+
+class TestQHEntropySearch(BotorchTestCase):
+    def test_initialize_q_hentropy_search(self):
+        for dtype in (torch.float, torch.double):
+            mean = torch.zeros(1, 1, device=self.device, dtype=dtype)
+            mm = MockModel(MockPosterior(mean=mean))
+            # test error when neither specifying neither sampler nor num_fantasies
+            with self.assertRaises(ValueError):
```
esantorella, line 27:
```suggestion
            with self.assertRaisesRegex(ValueError, "start of expected message here"):
```

Nit: It's helpful to check that the correct error message is raised

test/acquisition/test_hentropy_search.py
```diff
@@ -0,0 +1,221 @@
...
+                loss_function_class=qLossFunctionMinMax,
+                loss_function_hyperparameters={},
+            )
+            self.assertEqual(hes.n_fantasy_at_design_pts, 64)
+            self.assertEqual(hes.n_fantasy_at_action_pts, 64)
+            self.assertIsInstance(hes.design_sampler, SobolQMCNormalSampler)
+            self.assertIsInstance(hes.action_sampler, SobolQMCNormalSampler)
+            self.assertEqual(hes.design_sampler.sample_shape, torch.Size([64]))
+            self.assertEqual(hes.action_sampler.sample_shape, torch.Size([64]))
+            self.assertEqual(hes.get_augmented_q_batch_size(q=3), 64 + 3)
```
esantorella, line 61: Nice, thanks for all these checks!

test/acquisition/test_hentropy_search.py
```diff
@@ -0,0 +1,221 @@
...
+
+        # MinMax loss function
+        for dtype in (torch.float, torch.double):
+            bounds = torch.tensor([[0], [1]], device=self.device, dtype=dtype)
+            bounds = bounds.repeat(1, d)
+            train_X = torch.rand(num_data, d, device=self.device, dtype=dtype)
+            train_Y = torch.rand(num_data, 1, device=self.device, dtype=dtype)
+            model = SingleTaskGP(train_X, train_Y)
+
+            # default evaluation testsÎ
```
esantorella, line 111:
```suggestion
            # default evaluation tests
```

test/acquisition/test_hentropy_search.py
```diff
@@ -0,0 +1,221 @@
...
+
+        # TopK loss function
+        for dtype in (torch.float, torch.double):
+            bounds = torch.tensor([[0], [1]], device=self.device, dtype=dtype)
+            bounds = bounds.repeat(1, d)
+            train_X = torch.rand(num_data, d, device=self.device, dtype=dtype)
+            train_Y = torch.rand(num_data, 1, device=self.device, dtype=dtype)
+            model = SingleTaskGP(train_X, train_Y)
+
+            # default evaluation testsÎ
```
esantorella, line 168:
```suggestion
            # default evaluation tests
```

test/acquisition/test_hentropy_search.py
```diff
@@ -0,0 +1,221 @@
...
+            )
+            eval_A = torch.rand(
+                t_batch_size, num_fantasies, q, d, device=self.device, dtype=dtype
+            )
+            result = hes(X=eval_X, A=eval_A)
+            self.assertEqual(result.shape, torch.Size([t_batch_size]))
+
+            # extract candidates
+            cand = hes.extract_candidates(eval_X)
+            self.assertEqual(cand.shape, torch.Size([t_batch_size, q, d]))
```
esantorella, line 221:
```suggestion
            self.assertEqual(cand.shape, torch.Size([t_batch_size, q, d]))
            
```

botorch/acquisition/hentropy_search.py
```diff
@@ -0,0 +1,277 @@
...
+            A Tensor of shape `n_fantasy_at_action_pts x batch`.
+        """
+
+        if A.shape[-2] != 2:
+            raise RuntimeError("qLossFunctionMinMax only supports 2 actions.")
+
+        q_hes = (Y[..., 1:].sum(dim=-1) - Y[..., 0]).mean(dim=0)
+        # >>> n_fantasy_at_design_pts x batch_size
+
+        return q_hes
```
esantorella, line 277:
```suggestion
        return q_hes
        
```

A newline at the end of the file is required by black. (You should also be able to run `ufmt format .` to fix this.)

tutorials/hentropy_search.ipynb
Balandat: Oh I just mean that you're running "iterations" here in the tutorial but it's not exactly clear what these are and why they're being run. Can you preface the actual run with some more comments that explain what is happening? Just basically some more guiding the reader so they know what's going on at a higher level.

Balandat: Hmm not sure what happened to this comment - I can't actually find it in the diff view. I also don't really remember what exactly this was about.

saitcakmak:
Hi @sangttruong. Looks like this PR went stale but hopefully you're doing well and still interested in landing the contribution! We're trying out a new support model for community contributions: https://github.com/pytorch/botorch/blob/main/CONTRIBUTING.md#community-contributions-beta. Can you move the new additions to the botorch_community, notebooks_community & test_community directories and add the contributor info and other metadata to the files as outlined in the readme files? Thanks!

saitcakmak:
Hi @sangttruong. Just checking in to see if you're still planning to land this PR. Thanks!

sangttruong:
Hi @saitcakmak -- Sorry for the delayed reply. I totally forgot about this PR. I plan to update it in the next few days

saitcakmak:
I incorporated some of the suggestions from earlier reviews, rebased the PR to latest BoTorch and addressed some of the deprecation issues and other failures. This should be ready to land now. As this is a community contribution, you're the primary owner of these methods. Feel free to make any improvements in follow up PRs. Thanks again for the PR!

Balandat:
Great to get this in - thanks for your contribution @sangttruong !

PR #1797, opened 2023-04-17
[Easy] Add SMOKE_TEST mode to information-theoretic acquisition functions tutorial

esantorella:
## Motivation

The information-theoretic acquisition functions tutorial has been flakily timing out, since it takes about as long as the 5m limit in SMOKE_TEST mode. It was actually running the same code in both smoke-test and standard modes; this PR gives it a smoke-test mode.

The qPredictiveEntropySearch acquisition function becomes much more costly with q=4  than q=3; this PR changes q=4 to q=3, reducing the runtime of the most expensive cell from around two minutes to around 12 seconds on my machine.

## Test Plan

Ran the tutorial, looked at outputs


saitcakmak:
The only diff between smoke tests & non-smoke test is `q`! That's some impressive speed-up. Thanks!

saitcakmak (approved):

tutorials/information_theoretic_acquisition_functions.ipynb
```diff
@@ -251,8 +251,29 @@
...
+    },
+    {
+     "name": "stdout",
+     "output_type": "stream",
+     "text": [
+      "[KeOps] Warning : omp.h header is not in the path, disabling OpenMP.\n",
+      "[KeOps] Warning : Cuda libraries were not detected on the system ; using cpu only mode\n"
+     ]
+    }
+   ],
```
esantorella, line 273:
```suggestion
   "outputs": [],
```

PR #1798, opened 2023-04-18
Update CONTRIBUTING with latest formatting / linting practices.

Balandat:
Updated to simply run `ufmt` instead of `black` separately. Some other minor updates.


esantorella:
Thanks! I was thinking it would also be good to update the PR form to help people remember to run these checks. also, we can set up pre-commit so that everything that's needed runs when someone commits.

PR #1799, opened 2023-04-19
Add `batch_shape` property to `SingleTaskVariationalGP`

Balandat:
This enables the use of `SingleTaskVariationalGP` with certain botorch features (e.g. with entropy-based acquistion functions as requested in https://github.com/pytorch/botorch/discussions/1795).

This is a bit of a band-aid, the proper thing to do here is to fix up the PR upstreaming this to gpytorch (https://github.com/cornellius-gp/gpytorch/pull/2307) to enable support for `batch_shape` on all approximate gpytorch models, and then just call that on the `model` in `ApproximateGPyTorchModel`.

### DIFF
--- a/botorch/models/approximate_gp.py
+++ b/botorch/models/approximate_gp.py
@@ -426,6 +426,17 @@ class SingleTaskVariationalGP(ApproximateGPyTorchModel):
 
         self.to(train_X)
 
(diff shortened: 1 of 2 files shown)


Balandat:
cc @acb-code

PR #1814, opened 2023-05-04
Making GIBBON robust to numerical instability

SebastianAment:
Summary:
When using GIBBON for batch acquisition, numerical instabilities can be encountered that lead a quantity (`V_determinant`) that is mathematically positive to be numerically negative. Taking the logarithm of this quantity then leads to NaNs in the information gain tensor, which in turn leads to optimization failures.

This commit clamps a term in the expression for `V_determinant` so that the resultant quantity is positive, while ensuring that gradients continue to flow through the backward pass to all relevant variables.

Differential Revision: D45584690

### DIFF
--- a/botorch/acquisition/max_value_entropy_search.py
+++ b/botorch/acquisition/max_value_entropy_search.py
@@ -647,11 +647,19 @@ class qLowerBoundMaxValueEntropy(DiscreteMaxValueBase):
         # 1 x m x m
 
         # use determinant of block matrix formula
-        V_determinant = variance_m - inv_quad(B, A.transpose(1, 2)).unsqueeze(1)
+        inv_quad_term = inv_quad(B, A.transpose(1, 2)).unsqueeze(1)
+        # NOTE: Even when using Cholesky to compute inv_quad, `V_determinant` can be
+        # negative due to numerical issues. To avoid this, we clamp the variance
+        # so that `V_determinant` > 0, while still allowing gradients to be
+        # propagated through `inv_quad_term`, as well as through `variance_m`
+        # in the expression for `r` below.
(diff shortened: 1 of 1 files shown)


SebastianAment:
@henrymoss Hi Henry, Sebastian "from GitHub" here :) 

I encountered some NaNs while using GIBBON for batch optimization and  traced it to `V_determinant` being negative. Do you think clamping `variance_m` to enforce `V_determinant > 0` is a reasonable thing to do here?

henrymoss:
Hello Sebastian "from GitHub". I guess the issue came after a fair few optimisation steps? 

Once we are quite confident where the optimum is, then the batch gets very concentrated and we run into numerical precision issues with the ratio of the pdf/cdf in the GIBBON calculation (I think). 

I think clamping makes sense, but I would keep the lower limit as small as possible. Towards the end of BO the actual size of information gain can be pretty tiny (hence all the other numerical hacks in the code), so you might wanna try 1e-12 or something?

SebastianAment:
Thanks for the feedback Henry! I just updated the clamp to use `inv_quad_term * (1 + eps)`, where `eps = 1e-6` for floats and `eps = 1e-12` for doubles. If we go much lower than `1e-6` for floats, `1 + eps == 1` numerically, so the NaNs will appear. 

> numerical precision issues with the ratio of the pdf/cdf in the GIBBON calculation

As far as I can tell, `V_determinant` is not affected by the pdf-cdf `ratio`, which is already checked for NaNs, so this issue seems different.

This occurred mostly after a couple of iterations, but I've even observed it in the second batch on a 12-dimensional problem after initializing with 24 Sobol points and with a candidate set consisting of 1024 points. Maybe the dimensionality also has an effect on this. 

In any case, works like a charm now!

PR #1822, opened 2023-05-09
Fix missing secrets in website deployment

saitcakmak:
Website deployment didn't work when called from deploy on release or nightly workflows. This was due to missing secrets, which we now inherit from the parent workflow.

Website deployment works through nightly cron with this change: https://github.com/pytorch/botorch/actions/runs/4929215096/jobs/8808556629


esantorella (approved):
Thanks so much for fixing this!

Issue #1824, opened 2023-05-10
[Bug] SingleTaskVariationalGP raises a warning when using input_transform


crasanders:
# 🐛 Bug

Calling `SingleTaskVariationalGP.posterior` with an `input_transform` raises a warning, whereas the equivalent call with  `SingleTaskGP` does not. I'm not sure if `input_transform` works correctly with `SingleTaskVariationalGP` or if I can safely interpret the resulting posterior. It also seems a little odd to me that this would be a warning and not an exception.

## To reproduce

** Code snippet to reproduce **
```python
import torch
from botorch.models import SingleTaskGP, SingleTaskVariationalGP
from botorch.models.transforms import Normalize

X = torch.rand((20, 1))
y = torch.sin(X)

model = SingleTaskGP(train_X=X, train_Y=y, input_transform=Normalize(1))
post = model.posterior(X) # No warning

model = SingleTaskVariationalGP(train_X=X, train_Y=y, input_transform=Normalize(1))
post = model.posterior(X) # Warning
```

** Stack trace/error message **
```
RuntimeWarning: Could not update `train_inputs` with transformed inputs since SingleTaskVariationalGP does not have a `train_inputs` attribute. Make sure that the `input_transform` is applied to both the train inputs and test inputs.```
```

## System information

botorch version = 0.8.5
gpytorch version = 1.10
torch version = 1.13.1

Balandat:
cc @saitcakmak this is relevant for the proposed transforms refactor in https://github.com/cornellius-gp/gpytorch/pull/2114

crasanders:
Great, glad it's a false alarm. Thanks for the quick reply.

saitcakmak:
I vaguely recall some issue with inducing points sometimes getting transformed and sometimes not with ApproximateGP. Looking at my old notes, I found this:
> It does not play well with ApproximateGP, since inducing points sometimes get transformed (in train mode) and sometimes don’t (in posterior). This leads to bugs as we train them under one setup and evaluate them under another.

So, there might be some truth to the warning here. The proper solution would be to push through https://github.com/pytorch/botorch/pull/1372 and fix this for good. I'll leave this open just in case.

PR #1826, opened 2023-05-11
Don't raise specious warning about input transforms with approximate GPs

esantorella:
## Motivation

The BoTorch base `Model` class warns if an input transform has been provided, the `eval` method is called, and the object has no `train_inputs` attribute. This is not appropriate for `ApproximateGPyTorchModel`s; see #1824 . This PR gives `ApproximateGPyTorchModel` the `train` and `eval` modes from `torch.nn.Module`, which is the same as the methods it had been inheriting from `Model` but without the irrelevant input transform logic.

A nicer fix would be to remove the input transform logic from `Model` and have it only in subclasses that it applies to, so that subclasses like `ApproximateGPyTorchModel` would not need to do anything special to avoid inheriting that.

I think this all applies to `EnsembleModel`s as well as `ApproximateGPyTorchModel`s --looking into this now.

## Test Plan

Added a unit test to make sure transforms are applied appropriately

### DIFF
--- a/botorch/models/approximate_gp.py
+++ b/botorch/models/approximate_gp.py
@@ -32,7 +32,7 @@ from __future__ import annotations
 import copy
 import warnings
 
-from typing import Optional, Type, Union
+from typing import Optional, Type, TypeVar, Union
 
 import torch
 from botorch.models.gpytorch import GPyTorchModel
@@ -64,9 +64,13 @@ from gpytorch.variational import (
     VariationalStrategy,
 )
 from torch import Tensor
+from torch.nn import Module
 
 
 MIN_INFERRED_NOISE_LEVEL = 1e-4
+
+
+TApproxModel = TypeVar("TApproxModel", bound="ApproximateGPyTorchModel")
 
 
 class ApproximateGPyTorchModel(GPyTorchModel):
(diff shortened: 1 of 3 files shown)


Balandat:
I'm ok with this change but I second this point of yours:

> A nicer fix would be to remove the input transform logic from Model and have it only in subclasses that it applies to, so that subclasses like ApproximateGPyTorchModel would not need to do anything special to avoid inheriting that.

In the long term we can hopefully move the input transform logic out of BoTorch altogether and upstream it into gpytorch: https://github.com/cornellius-gp/gpytorch/issues/1652

esantorella:
The reason I didn't remove the input transform logic from model and put it in relevant subclasses is that it would break any classes outside BoTorch that are relying on that functionality. In my experience, breaking BC-compatibility in a minor way always turns out worse than I expected it to. I'd vote to note this as an issue to bookmark for a major BC-breaking release.

Similarly, it would be good to not have methods in model that [raise a NotImplementedError](https://github.com/pytorch/botorch/blob/1eed96abbd71bbfaef21b2271e00e8b0ee4d6d7f/botorch/models/model.py#L122); if subclasses really must implement these, those methods should be abstract, and if it's okay for subclasses to not have those, then the method shouldn't be in the base class. The fix for that would look something like #1462, which wound up hairier than anticipated since it broke some subclass checks and dispatchers in projects that use BoTorch.

Balandat:
> Similarly, it would be good to not have methods in model that [raise a NotImplementedError](https://github.com/pytorch/botorch/blob/1eed96abbd71bbfaef21b2271e00e8b0ee4d6d7f/botorch/models/model.py#L122); if subclasses really must implement these, those methods should be abstract, and if it's okay for subclasses to not have those, then the method shouldn't be in the base class. The fix for that would look something like https://github.com/pytorch/botorch/pull/1462, which wound up hairier than anticipated since it broke some subclass checks and dispatchers in projects that use BoTorch.

That generally makes sense. I guess the situation in which this pattern is justified is where we'd like to implement the method in the subclasses but they're not essential and we want to work incrementally. I guess the alternative would be to make the method on the base class abstract and then define an explicit override that raises a `NotImplementedError` in the subclass...

PR #1830, opened 2023-05-12
Update publish_site to gracefully exit if there are no changes

saitcakmak:
https://github.com/pytorch/botorch/actions/runs/4957898093/jobs/8870503024 failed with `nothing to commit, working tree clean` message, presumably due to `git commit ...` returning a non-zero exit code with this. This solution checks if there are changes to be committed and exists with 0 if there's nothing to commit.

The solution is adapted from https://stackoverflow.com/a/5139672

### DIFF
--- a/scripts/publish_site.sh
+++ b/scripts/publish_site.sh
@@ -144,7 +144,7 @@ if [[ $VERSION == false ]]; then
   cd botorch-gh-pages || exit
   git add .
   # Exit gracefully if there's nothing to commit.
-  [[ -z $(git diff --cached --exit-code) ]] && exit 0
+  [[ -z $(git diff --cached --exit-code) ]] && echo "Nothing to commit"; exit 0
   # Commit & push if there's something to commit.
   git commit -m 'Update latest version of site'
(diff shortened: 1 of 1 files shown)


scripts/publish_site.sh
```diff
@@ -143,6 +143,9 @@ if [[ $VERSION == false ]]; then
   # Push changes to gh-pages
   cd botorch-gh-pages || exit
   git add .
+  # Exit gracefully if there's nothing to commit.
+  [[ -z $(git diff --cached --exit-code) ]] && exit 0
```
esantorella, line 147: I'm not sure this is correct bash, but it would be good to print something letting us know it intentionally exited early

```suggestion
  [[ -z $(git diff --cached --exit-code) ]] && (echo "Nothing to commit"; exit 0)
```

saitcakmak:
https://github.com/pytorch/botorch/actions/runs/4960971968/jobs/8877170591 now succeeds!
Screenshot 2023-05-12 at 9 32 29 AM

PR #1831, opened 2023-05-12
Remove cp for missing files from build_docs

saitcakmak:
We're attempting to copy files that no longer exist (due to sphinx version changes), leading to error messages:
```
cp: cannot stat 'sphinx/build/html/_static/jquery.js': No such file or directory
cp: cannot stat 'sphinx/build/html/_static/underscore.js': No such file or directory
```
This removes those copy operations.

### DIFF
--- a/scripts/build_docs.sh
+++ b/scripts/build_docs.sh
@@ -70,8 +70,6 @@ if [[ $ONLY_DOCUSAURUS == false ]]; then
 
   # move JS files from /sphinx/build/html/_static/*:
   cp "${SPHINX_JS_DIR}documentation_options.js" "${DOCUSAURUS_JS_DIR}documentation_options.js"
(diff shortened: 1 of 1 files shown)


saitcakmak:
I tried to test this in https://github.com/pytorch/botorch/actions/runs/4960937108/jobs/8877092075 but it still produced the errors. My guess is that the workflow runs the scripts from a clone of the main branch, so these changes will not make it in before we commit them. I checked and there are no other cp operations involving these files in the repo.

esantorella (approved):
LGTM, thanks!

Balandat:
Are there other files that were added that we may want to copy instead? I guess we're probably ok so long as nothing is broken on the website?

saitcakmak:
I assume not, since everything works fine. These error messages have been around for a while.

PR #1833, opened 2023-05-12
Make SCBO tutorial deterministic

esantorella:
## Motivation

The SCBO tutorial is not deterministic and uses a while loop, so its runtime can vary quite a bit. I saw 50s and 100s in smoke test mode on my laptop. In the CI, it usually runs well under the 5-minute limit in smoke test mode, but recently timed out and caused a failure.

This change makes it deterministic by seeding the SobolEngine.

## Test Plan

Ran the tutorial to make sure it runs and is deterministic.


Balandat (approved):
lgtm, thanks!

PR #1839, opened 2023-05-17
Refactor test function test utils

Balandat:
Summary:
Refactors the test utils for botorch test functions to make things more modular.

Instead of subclassing the base test cases, we now have class Mixins for synthetic, constrained, and multi-objective problems. This means we can mix-and-match the test case as needed.

Further, `BaseTestProblemTestCase` now subclasses `BotorchTestCase` so that we don't need to subclass from `BotorchTestCase` separately for every test.

Finally, this flattens the `TestMultiObjectiveProblems` and `TestConstrainedMultiObjectiveProblems` test collections (the idea behind the test design is that we have separate test cases for each problem (potentially with different arguments) to better identify and group the tests, so this reestablishes that.

Differential Revision: D45969036

### DIFF
--- a/botorch/utils/testing.py
+++ b/botorch/utils/testing.py
@@ -8,6 +8,7 @@ from __future__ import annotations
 
 import math
 import warnings
+from abc import abstractproperty
 from collections import OrderedDict
 from typing import Any, List, Optional, Tuple, Union
 from unittest import TestCase
@@ -93,10 +94,7 @@ class BotorchTestCase(TestCase):
         )
 
 
-class BaseTestProblemBaseTestCase:
-
-    functions: List[BaseTestProblem]
-
+class BaseTestProblemTestCaseMixIn:
     def test_forward(self):
         for dtype in (torch.float, torch.double):
             for batch_shape in (torch.Size(), torch.Size([2]), torch.Size([2, 3])):
@@ -113,8 +111,14 @@ class BaseTestProblemBaseTestCase:
                     )
                     self.assertEqual(res.shape, batch_shape + tail_shape)
 
+    @abstractproperty
+    def functions(self) -> List[BaseTestProblem]:
+        # The functions that should be tested. Typically defined as a class
+        # attribute on the test case subclassing this class.
+        pass  # pragma: no cover
 
-class SyntheticTestFunctionBaseTestCase(BaseTestProblemBaseTestCase):
+
+class SyntheticTestFunctionTestCaseMixin:
     def test_optimal_value(self):
         for dtype in (torch.float, torch.double):
             for f in self.functions:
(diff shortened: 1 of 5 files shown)


botorch/utils/testing.py
```diff
@@ -93,7 +93,7 @@ def assertAllClose(
         )
 
 
-class BaseTestProblemBaseTestCase:
+class BaseTestProblemTestCaseMixIn:
 
     functions: List[BaseTestProblem]
```
esantorella, line 98: Pyre is unhappy about the uninitialized attribute. One way to deal with this would be to make `functions` an abstract property. Another way would be to just have `functions: List[BaseTestProblem] = []`. Either way, the easiest thing might be for me to get the type checker passing in a follow-up diff -- what do you think?

test/test_functions/test_multi_objective.py
```diff
@@ -304,27 +330,119 @@ def test_gen_pareto_front(self):
...
+    functions = [WeldedBeam()]
+
+
+class TestOSY(
+    BotorchTestCase,
+    BaseTestProblemTestCaseMixIn,
+    MultiObjectiveTestProblemTestCaseMixin,
+    ConstrainedTestProblemTestCaseMixin,
+):
+    functions = [OSY()]
```
esantorella, line 448: in conjunction with above suggestion
```suggestion
```

botorch/utils/testing.py
esantorella, line 98:
```suggestion
    functions: Sequence[BaseTestProblem] = []
```

esantorella (approved):
The refactoring makes a lot of sense, thanks! See comments.

botorch/utils/testing.py
Balandat: So the issue with just initializing it as an empty list is that we might end up generating a bunch of noop tests if we're not careful. So abstractproperty is probably a good call here!

test/test_functions/test_multi_objective.py
```diff
@@ -304,27 +330,119 @@ def test_gen_pareto_front(self):
...
+
+
+class TestCONSTR(
+    BotorchTestCase,
+    BaseTestProblemTestCaseMixIn,
+    MultiObjectiveTestProblemTestCaseMixin,
+    ConstrainedTestProblemTestCaseMixin,
+):
+    functions = [CONSTR()]
+
```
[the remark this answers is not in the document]
Balandat: I actually wanted to keep these separate to make the tests more granular and identifiable. It's a bit more boilerplate code but we have clear individual test names rather than one big blob test.  The idea of the `functions` was to group together different variants of test functions if they can be modified via init args.

PR #1844, opened 2023-05-24
Fix codecov coverage

saitcakmak:
We measure the test coverage based on a CPU-only run of the test suite. These two lines were only executed for GPU tensors. Removing the condition will allow them to be executed always.


Balandat (commented):
Why not just remove the `lower.device.type == "cuda"` condition? If the tensor is already on the CPU then `.cpu()` is a nullop. That would also cover the case if this were to run on some other (non-cuda) accelerator.

saitcakmak:
Yeah, that'd also work

PR #1848, opened 2023-05-31
Run the cleanup steps when publish site exits with no changes

saitcakmak:
This is a follow up to https://github.com/pytorch/botorch/pull/1830. The current setup doesn't perform the cleanup steps if we exit with nothing to commit.

Test: https://github.com/pytorch/botorch/actions/runs/5135512160/jobs/9241030759
Screenshot 2023-05-31 at 9 53 34 AM

### DIFF
--- a/scripts/publish_site.sh
+++ b/scripts/publish_site.sh
@@ -236,5 +236,6 @@ else
 fi
 
 # Clean up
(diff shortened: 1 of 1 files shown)


Balandat (approved):
lgtm, thanks!

PR #1856, opened 2023-06-04
Change the inheritance order of ModelListGPyTorchModel

saitcakmak:
A bunch of tests are currently failing as a result of https://github.com/cornellius-gp/gpytorch/pull/2360 removing `num_outputs` from `IndependentModelList`. This change lead to `ModelListGP.num_outputs` pointing to `GPyTorchModel.num_outputs`, which requires the model to define a `_num_outputs`. The correct behavior for `ModelListGP` is to use `ModelList.num_outputs`, which can be achieved by changing the inheritance order.

### DIFF
--- a/botorch/models/gpytorch.py
+++ b/botorch/models/gpytorch.py
@@ -527,7 +527,7 @@ class BatchedMultiOutputGPyTorchModel(GPyTorchModel):
         return new_model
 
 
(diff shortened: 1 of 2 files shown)


esantorella (approved):
Looks good, thanks!

Issue #1857, opened 2023-06-04
[Feature Request] [documentation] `SaasFullyBayesianMultiTaskGP` initializer docstring doesn't explain all parameters


esantorella:
# 🚀 Feature Request

The docstring for the init of  `SaasFullyBayesianMultiTaskGP` doesn't explain all parameters.  It currently reads

```python
    def __init__(
        self,
        train_X: Tensor,
        train_Y: Tensor,
        train_Yvar: Optional[Tensor],
        task_feature: int,
        output_tasks: Optional[List[int]] = None,
        rank: Optional[int] = None,
        outcome_transform: Optional[OutcomeTransform] = None,
        input_transform: Optional[InputTransform] = None,
        pyro_model: Optional[PyroModel] = None,
    ) -> None:
        r"""Initialize the fully Bayesian multi-task GP model.

        Args:
            train_X: Training inputs (n x (d + 1))
            train_Y: Training targets (n x 1)
            train_Yvar: Observed noise variance (n x 1). If None, we infer the noise.
                Note that the inferred noise is common across all tasks.
            task_feature: The index of the task feature (`-d <= task_feature <= d`).
            rank: The num of learned task embeddings to be used in the task kernel.
                If omitted, set it to be 1.
        """
```

## Motivation

I am looking at #2959 and I'm trying to understand what this code does.

## Pitch

Document all the arguments. This might be quick, since at least one of these features -- `output_tasks` -- is passed directly to the superclass, so the docstring can be copied from the superclass. Others are likely the same as in the superclass.

Even better would be to enable flake8 docstring lints so that the CI fails if there are any parameters that are not documented.

**Are you willing to open a pull request?** (See [CONTRIBUTING](../CONTRIBUTING.md))

Maybe, but it would be a better task for someone who is more familiar with this code.

Balandat:
@dme65

PR #1858, opened 2023-06-04
Update the docstring of SAAS MTGP

saitcakmak:
Resolves https://github.com/pytorch/botorch/issues/1857

### DIFF
--- a/botorch/models/gpytorch.py
+++ b/botorch/models/gpytorch.py
@@ -527,7 +527,7 @@ class BatchedMultiOutputGPyTorchModel(GPyTorchModel):
         return new_model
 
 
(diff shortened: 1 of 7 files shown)


botorch/models/gpytorch.py
```diff
@@ -527,7 +527,7 @@ def subset_output(self, idcs: List[int]) -> BatchedMultiOutputGPyTorchModel:
         return new_model
 
 
-class ModelListGPyTorchModel(GPyTorchModel, ModelList, ABC):
+class ModelListGPyTorchModel(ModelList, GPyTorchModel, ABC):
```
esantorella, line 530: looks like there was some kind of merge issue here

Issue #1862, opened 2023-06-06
[Bug?] qExpectedImprovement for minimization


MaGbrt:
# 🐛 Bug(?)
There might be a bug in the qExpectedImprovement implementation with the tag 'maximizing=False'. If I'm not missing anything, it performs worse than minimizing the opposite objective.

## To reproduce
I used some tutorials from the BOTorch documentation to create a simple example which seems to recreate my original bug.

** Code snippet to reproduce **
```python
import torch
from botorch.acquisition import qExpectedImprovement
from botorch.fit import fit_gpytorch_mll
from botorch.models import SingleTaskGP
from botorch.optim import optimize_acqf
from botorch.test_functions import Ackley
from botorch.utils.transforms import unnormalize
from torch.quasirandom import SobolEngine

from gpytorch.constraints import Interval
from gpytorch.likelihoods import GaussianLikelihood
from gpytorch.mlls import ExactMarginalLogLikelihood


device = torch.device("cpu")
dtype = torch.double
dim = 10
batch_size = 4
n_init = 10 * dim
max_cholesky_size = float("inf")  # Always use Cholesky
n_rep = 10
budget = 96

#%% Helper functions
def get_initial_points(dim, n_pts, seed=0):
    sobol = SobolEngine(dimension=dim, scramble=True, seed=seed)
    X_init = sobol.draw(n=n_pts).to(dtype=dtype, device=device)
    return X_init

def fill_target(target, Y_end, batch_size):
    budget = len(target) - 1
    n_init = len(Y_end) - budget*batch_size
    for k in range(0, budget+1):
        d = k * batch_size + n_init 
        target[k] = Y_end[:d].max()
        
#%% Maximization of the negative Ackley function
fun = Ackley(dim=dim, negate=True).to(dtype=dtype, device=device)
fun.bounds[0, :].fill_(-5)
fun.bounds[1, :].fill_(10)
lb, ub = fun.bounds


def eval_objective(x):
    """This is a helper function we use to unnormalize and evalaute a point"""
    return fun(unnormalize(x, fun.bounds))

target_ei = torch.zeros((n_rep, budget+1))
for k_rep in range(0, n_rep):    
    X_ei = get_initial_points(dim, n_init)
    Y_ei = torch.tensor(
        [eval_objective(x) for x in X_ei], dtype=dtype, device=device
    ).unsqueeze(-1)
    
    for i in range(0, budget):
[...]

PR #1863, opened 2023-06-06
Remove `**kwargs` from `qExpectedImprovement` constructor

Balandat:
Summary: Unnecessary and confused a user here: https://github.com/pytorch/botorch/issues/1862

Differential Revision: D46482468


saitcakmak (approved):
Thanks!

PR #1865, opened 2023-06-06
[WIP] Remove functionality deprecated in #1073 such as `ScalarizedObjective`

esantorella:
## Motivation

Removes functionality that was deprecated in #1073, several major versions ago, and makes 

* Remove `ScalarizedObjective`, `UnstandardizeAnalyticMultiOutputObjective`, `AcquisitionObjective`, `UnstandardizeAnalyticMultiOutputObjective`,
* Removed `kwargs` from various acquisition functions where  `kwargs` will now be unused (some of these used to accept an argument `objective` in the `kwargs`,  but that triggered a deprecation warning)
* Increase the numerical tolerance on a flaky test.

## Test Plan

I still need to add a unit test for a couple new UnsupportedErrors I added.

## Related PRs

#1073


botorch/acquisition/input_constructors.py
```diff
@@ -457,6 +411,44 @@ def construct_inputs_mc_base(
...
+    Returns:
+        A dict mapping kwarg names of the constructor to values.
+    """
+    return _construct_inputs_mc_base(
+        model=model,
+        objective=objective,
+        posterior_transform=posterior_transform,
+        X_pending=X_pending,
+        sampler=sampler,
+    )
```
saitcakmak, line 449: Is the idea here to not give `construct_inputs_mc_base` dual purpose? So we construct a public method that only serves `qSimpleRegret` and use the private method for the common functionality. If so, should we rename the public method to `construct_inputs_qSimpleRegret` or similar to avoid "This function has a misleading name."?

botorch/acquisition/input_constructors.py
```diff
@@ -692,6 +685,8 @@ def construct_inputs_EHVI(
     # This selects the objectives (a subset of the outcomes) and set each
     # objective threhsold to have the proper optimization direction.
     if objective is None:
+        # TODO: "IdentityAnalyticMultiOutputObjective is deprecated. "
+        # "Use IdentityPosteriorTransform instead."
```
saitcakmak, line 689: TODO :)

botorch/acquisition/multi_objective/objective.py
```diff
@@ -262,15 +257,23 @@ def forward(self, samples: Tensor, X: Optional[Tensor] = None) -> Tensor:
         return samples * self.Y_std + self.Y_mean
 
 
-class AnalyticMultiOutputObjective(AcquisitionObjective):
+# TODO: this can't trivially be deleted because it's still used in
+# `construct_inputs_EHVI`
```
saitcakmak, line 261: Do we have replacement classes for this and subclasses? I vaguely recall that @SebastianAment implemented those not long ago.

botorch/acquisition/multi_step_lookahead.py
```diff
@@ -527,11 +520,11 @@ def _construct_inner_samplers(
             respective stage.
         inner_mc_samples: A list `[n_0, ..., n_k]` containing the number of MC
             samples to be used for evaluating the stage value function. Ignored if
-            the objective is `None` or a `ScalarizedObjective`.
+            the objective is `None`.
         objective: The objective under which the output is evaluated. If `None`, use
             the model output (requires a single-output model or a posterior transform).
             Otherwise the objective is MC-evaluated (using `inner_sampler`).
-        posterior_transform: A PosteriorTransform (optional).
+        posterior_transform: Optional; not used.
```
saitcakmak, line 527: Should we remove it if it is not used?

saitcakmak (approved):
Thanks for putting this together!

botorch/acquisition/multi_objective/objective.py
esantorella: Hmm,`AcquisitionObjective` was deprecated a long time ago, but `AnalyticMultiOutputObjective` (a subclass of `AcquisitionObjective`) was only deprecated two months ago.  I'm not sure if `AnalyticMultiOutputObjective` should be considered ready for deletion then... it would have been triggering deprecation warnings for over a year in some `isinstance` checks, but not upon initialization.

PR #1867, opened 2023-06-06
Fix test_pyro_catch_error

Balandat:
Summary:
This test would fail when run through `test_cuda` since the exception
handler was not properly reset. This adds a `tearDown` method that takes care
of this.

Reviewed By: dme65

Differential Revision: D46486189


esantorella:
Thank you! I'd seen this failing and couldn't figure out what went wrong.

Issue #1880, opened 2023-06-09
[Bug] `qNoisyExpectedHypervolumeImprovement` fails at single objective problem with uninformative error message


kstoneriv3:
# 🐛 Bug

Hi, thanks for creating such a solid package for BO! It's really nice to be able to try various BO methods out of the box!! 

I have a minor bug report (or a request):

For m-objective Bayesian optimization, the `qNoisyExpectedHypervolumeImprovement` fails when m=1 while it works fine for m>1. 

## To reproduce

** Code snippet to reproduce **
```python
import torch
from botorch.acquisition.multi_objective import qNoisyExpectedHypervolumeImprovement
from botorch.models import SingleTaskGP

ref_point = [0.]
X_train = torch.randn(5, 2)
X_test = torch.randn(3, 2)
y_train = torch.randn(5, 1)  # 1-dimensional objective
model = SingleTaskGP(X_train, y_train)
qNEHVI = qNoisyExpectedHypervolumeImprovement(model, ref_point, X_test)
```

** Stack trace/error message **
```
(env_torch) kei@kei-Alienware-15-R4 ~/D/tmp (main)> python
Python 3.10.6 (main, May 29 2023, 11:10:38) [GCC 11.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import torch
>>> from botorch.acquisition.multi_objective import qNoisyExpectedHypervolumeImprovement
>>> from botorch.models import SingleTaskGP
>>> 
>>> ref_point = [0.]
>>> X_train = torch.randn(5, 2)
>>> X_test = torch.randn(3, 2)
>>> y_train = torch.randn(5, 1)  # 1-dimensional objective
>>> model = SingleTaskGP(X_train, y_train)
/home/kei/Desktop/env_torch/lib/python3.10/site-packages/botorch/models/gpytorch.py:129: UserWarning: The model inputs are of type torch.float32. It is strongly recommended to use double precision in BoTorch, as this improves both precision and stability and can help avoid numerical errors. See https://github.com/pytorch/botorch/discussions/1444
  warnings.warn(_get_single_precision_warning(X.dtype), UserWarning)
/home/kei/Desktop/env_torch/lib/python3.10/site-packages/botorch/models/utils/assorted.py:173: InputDataWarning: Input data is not contained to the unit cube. Please consider min-max scaling the input data.
  warnings.warn(msg, InputDataWarning)
[...]

PR #1881, opened 2023-06-10
Raise ValueError when m<2 in qNoisyExpectedHypervolumeImprovement

kstoneriv3:
Raise more informative message when qNoistyExpectedHypervolumeImprovement cannot handle single objective input as discussed in #1880

### DIFF
--- a/test/acquisition/multi_objective/test_monte_carlo.py
+++ b/test/acquisition/multi_objective/test_monte_carlo.py
@@ -624,7 +624,7 @@ class TestQNoisyExpectedHypervolumeImprovement(BotorchTestCase):
         tkwargs = {"device": self.device}
         for dtype, m in product(
             (torch.float, torch.double),
-            (2, 3),
+            (1, 2, 3),
         ):
             tkwargs["dtype"] = dtype
             ref_point = self.ref_point[:m]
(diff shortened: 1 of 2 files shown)


esantorella (changes requested):
Thank you for this! This looks like the right change to me. Just one more thing: To get the code coverage check passing, you'll need to add a unit test that covers the newly added lines. That test could go approximately [here](https://github.com/pytorch/botorch/blob/e2caef4439f0a145eea463e13f7aed8206fcb0a6/test/acquisition/multi_objective/test_monte_carlo.py#L623) and should look something like [this](https://github.com/pytorch/botorch/blob/e2caef4439f0a145eea463e13f7aed8206fcb0a6/test/utils/test_containers.py#L39).

kstoneriv3:
> Thank you for this! This looks like the right change to me. Just one more thing: To get the code coverage check passing, you'll need to add a unit test that covers the newly added lines. That test could go approximately [here](https://github.com/pytorch/botorch/blob/e2caef4439f0a145eea463e13f7aed8206fcb0a6/test/acquisition/multi_objective/test_monte_carlo.py#L623) and should look something like [this](https://github.com/pytorch/botorch/blob/e2caef4439f0a145eea463e13f7aed8206fcb0a6/test/utils/test_containers.py#L39).

Hi, thanks for the review! I added a test for it but I am not sure if it is in the right place. Feel free to ask me to move it to somewhere more appropriate.

test/acquisition/multi_objective/test_monte_carlo.py
```diff
@@ -641,6 +641,19 @@ def test_q_noisy_expected_hypervolume_improvement(self):
...
+            if m == 1:
+                with self.assertRaises(ValueError):
+                    acqf = qNoisyExpectedHypervolumeImprovement(
+                        model=mm,
+                        ref_point=ref_point,
+                        X_baseline=X_baseline,
+                        sampler=sampler,
+                        cache_root=False,
+                    )
+                continue
```
esantorella, line 655:
```suggestion
```

Is this needed?

esantorella (approved):

test/acquisition/multi_objective/test_monte_carlo.py
kstoneriv3: I am added `continue` to skip the test afterwards. Or do we need to run full test cases for`m=1`?

esantorella: Oh yeah, that makes sense :)

PR #1890, opened 2023-06-16
Fix memory leak in inducing point allocators

### LINKED ISSUE #1788: [Bug] Memory not freed after a variational GP model is discarded
# 🐛 Bug

Iteratively creating variational GP `SingleTaskVariationalGP` will result in out of memory. I find a similar problem in [#1585](https://github.com/pytorch/botorch/issues/1585) which uses exact GP, i.e., `SingleTaskGP`. Use `gc.collect()` will solve the problem in [#1585](https://github.com/pytorch/botorch/issues/1585) but is useless for my problem. 

I add `torch.cuda.empty_cache()` and `gc.collect()` in my code and the code only creates the `SingleTaskVariationalGP` model and doesn't do anything about forward. However, the memory still increases.

## To reproduce

```python
import torch
from botorch.models import SingleTaskVariationalGP
from botorch.models.utils.inducing_point_allocators import GreedyVarianceReduction 
from gpytorch.mlls import VariationalELBO
import time
import gc

def create_gp_forward(device):
    # create the training data
    train_X = torch.rand(4096, 5).to(device)
    train_Y = torch.sin(train_X).sum(dim=1, keepdim=True).to(device)

    # creating the gp model
    gp = SingleTaskVariationalGP(
        train_X, train_Y, 
        learn_inducing_points=False,
        inducing_point_allocator=GreedyVarianceReduction(),
    )
    mll = VariationalELBO(gp.likelihood, gp.model, num_data=len(train_X))

    # gp(train_X)

    print(torch.cuda.memory_summary(device=device).split("\n")[7])

if __name__ == "__main__":
    for _ in range(10000):
        create_gp_forward(device = "cuda")
        time.sleep(2)
        torch.cuda.empty_cache()
        gc.collect()
```

```
| Allocated memory      |  447602 KB |  509638 KB |     902 MB |  476950 KB |
| Allocated memory      |     870 MB |     930 MB |    1805 MB |     935 MB |
| Allocated memory      |    1303 MB |    1363 MB |    2708 MB |    1405 MB |
| Allocated memory      |    1736 MB |    1796 MB |    3611 MB |    1875 MB |
| Allocated memory      |    2169 MB |    2229 MB |    4514 MB |    2345 MB |
| Allocated memory      |    2602 MB |    2662 MB |    5417 MB |    2815 MB |
| Allocated memory      |    3035 MB |    3095 MB |    6320 MB |    3285 MB |
| Allocated memory      |    3468 MB |    3528 MB |    7223 MB |    3755 MB |
| Allocated memory      |    3901 MB |    3961 MB |    8125 MB |    4224 MB |
| Allocated memory      |    4333 MB |    4394 MB |    9028 MB |    4694 MB |
| Allocated memory      |    4766 MB |    4827 MB |    9931 MB |    5164 MB |
...
CUDA out of memory. Tried to allocate 64.00 MiB (GPU 0; 5.79 GiB total capacity; 4.78 GiB already allocated; 34.06 MiB free; 4.78 GiB reserved in total by PyTorch) If reserved memory is >> allocated memory try setting max_split_size_mb to avoid fragmentation.  See documentation for Memory Management and PYTORCH_CUDA_ALLOC_CONF
```

## Expected Behavior

The variables created in `create_gp_forward` are local variables, so when finishing the function, the memory should be released and the occupation should be constant during the loop. 

## System information

- botorch 0.8.2
- gpytorch 1.9.1
- torch 1.12.0
- ubuntu 18.04

esantorella:
Fixes #1788 .
## Motivation

`allocate_inducing_points` leaks memory when passed a `kernel_matrix` with `requires_grad=True`. The memory leak happens due to a specific pattern of in-place torch operations in `_pivoted_cholesky_init`;  see [this comment](https://github.com/pytorch/botorch/issues/1788#issuecomment-1593893838) for more explanation. There is no need for `allocate_inducing_points` to support a `kernel_matrix` with `requires_grad=True`, because the output of `allocate_inducing_points` is not differentiable anyway (thanks to in-place operations).

[x] make `_pivoted_cholesky_init` raise an `UnsupportedError` when passed a `kernel_matrix` with `requires_grad=True`. That is mildly BC-breaking, but I think that is okay since the alternative is a memory leak.
[x] Evaluate kernels with `torch.no_grad()` where they are only used to be passed to `_pivoted_cholesky_init`

Yes

## Test Plan

[x] Unit test for memory leak
[x] Unit test for UnsupportedError

### DIFF
--- a/botorch/models/approximate_gp.py
+++ b/botorch/models/approximate_gp.py
@@ -191,7 +191,7 @@ class _SingleTaskVariationalGP(ApproximateGP):
         Args:
             train_X: Training inputs (due to the ability of the SVGP to sub-sample
                 this does not have to be all of the training inputs).
-            train_Y: Training targets (optional).
+            train_Y: Not used.
             num_outputs: Number of output responses per input.
             covar_module: Kernel function. If omitted, uses a `MaternKernel`.
             mean_module: Mean of GP model. If omitted, uses a `ConstantMean`.
@@ -402,7 +402,6 @@ class SingleTaskVariationalGP(ApproximateGPyTorchModel):
 
         model = _SingleTaskVariationalGP(
             train_X=transformed_X,
-            train_Y=train_Y,
             num_outputs=num_outputs,
             learn_inducing_points=learn_inducing_points,
             covar_module=covar_module,
--- a/botorch/models/utils/inducing_point_allocators.py
+++ b/botorch/models/utils/inducing_point_allocators.py
@@ -24,6 +24,7 @@ from abc import ABC, abstractmethod
 from typing import Union
 
 import torch
+from botorch.exceptions.errors import UnsupportedError
 from botorch.models.model import Model
 
 from botorch.utils.probability.utils import ndtr as Phi, phi
@@ -74,10 +75,23 @@ class InducingPointAllocator(ABC):
         quality_function = self._get_quality_function()
         covar_module = covar_module.to(inputs.device)
 
-        train_train_kernel = covar_module(inputs).evaluate_kernel()
+        # We use 'no_grad' here because `inducing_points` are not
+        # auto-differentiable with respect to the kernel hyper-parameters,
+        # because `_pivoted_cholesky_init` does in-place operations.
+        with torch.no_grad():
+            # Evaluate lazily because this may only be needed to figure out what
+            # case we are in
+            possibly_lazy_kernel = covar_module(inputs)
 
-        # base case
-        if train_train_kernel.ndimension() == 2:
+        base_case = possibly_lazy_kernel.ndimension() == 2
+        multi_task_case = (
+            possibly_lazy_kernel.ndimension() == 3 and len(input_batch_shape) == 0
+        )
+
+        if base_case or multi_task_case:
+            train_train_kernel = possibly_lazy_kernel.evaluate_kernel()
+
+        if base_case:
             quality_scores = quality_function(inputs)
             inducing_points = _pivoted_cholesky_init(
                 train_inputs=inputs,
@@ -85,8 +99,9 @@ class InducingPointAllocator(ABC):
                 max_length=num_inducing,
                 quality_scores=quality_scores,
             )
-        # multi-task case
-        elif train_train_kernel.ndimension() == 3 and len(input_batch_shape) == 0:
+            return inducing_points
+
+        if multi_task_case:
             input_element = inputs[0] if inputs.ndimension() == 3 else inputs
             kernel_element = train_train_kernel[0]
             quality_scores = quality_function(input_element)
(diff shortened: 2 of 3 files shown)


botorch/models/approximate_gp.py
```diff
@@ -192,7 +192,7 @@ def __init__(
         Args:
             train_X: Training inputs (due to the ability of the SVGP to sub-sample
                 this does not have to be all of the training inputs).
-            train_Y: Training targets (optional).
+            train_Y: Not used.
```
Balandat, line 194: interesting - why did this have the arg in the first place? Any reason (other than linting) not to get rid of this?

botorch/models/utils/inducing_point_allocators.py
```diff
@@ -74,19 +75,33 @@ def allocate_inducing_points(
         quality_function = self._get_quality_function()
         covar_module = covar_module.to(inputs.device)
 
-        train_train_kernel = covar_module(inputs).evaluate_kernel()
+        # We use 'no_grad' here because inducing points are not differentiable
+        # anyway. Inducing points are not differentiable bcause
```
Balandat, line 79:
```suggestion
        # anyway. Inducing points are not differentiable because
```

Balandat (approved):
thanks for the memory leak fix!

test/models/utils/test_inducing_point_allocators.py
```diff
@@ -262,12 +288,29 @@ def test_inducing_points_different_when_minimizing(self):
...
-        with self.assertRaises(ValueError):
-            inputs = torch.rand(15, 1, device=self.device)
+        inputs = torch.rand(15, 1, device=self.device)
+        with torch.no_grad():
             train_train_kernel = (
                 MaternKernel().to(self.device)(inputs).evaluate_kernel()
             )
-            quality_scores = torch.ones([10, 1], device=self.device)
+        quality_scores = torch.ones([10, 1], device=self.device)
+        with self.assertRaises(ValueError):
```
saitcakmak, line 297: It'd be good explicitly check for the specific error using `assertRaisesRegex`.

saitcakmak (approved):
Thanks! This was a tricky one to debug. More evidence that in-place operations and autograd do not mix well.

botorch/models/approximate_gp.py
esantorella: Perhaps it had this so that `_SingleTaskVariationalGP` would have a similar signature to `SingleTaskVariationalGP`? In any case, I don't see a good reason to keep it, especially since they differ in other ways at this point.

test/models/utils/test_inducing_point_allocators.py
esantorella: definitely agree. I ignored it since this test was pre-existing (I just rearranged it). Changing a bunch of these to `assertRaisesRegex` could be a good starter task

botorch/models/approximate_gp.py
SebastianAment: I think we should keep it to support inducing point allocators that take into account the target values. To this end, we'd need to add `outputs` (in addition to `inputs`) to the signature of
```
inducing_point_allocator.allocate_inducing_points
```
This would be ignored for the default variance reduction strategy, but would be necessary for e.g. @henrymoss's [Inducing Point Allocation for Sparse Gaussian Processes in High-Throughput
Bayesian Optimisation](https://proceedings.mlr.press/v206/moss23a/moss23a.pdf). This would also require supporting `outputs` in the quality functions, but not a lot else as far as I can tell.

Balandat:
cc @henrymoss

PR #1891, opened 2023-06-17
[Easy] Stop `test_get_deterministic_model` from being flaky by slightly loosening numerical tolerances

esantorella:
## Motivation

`test_get_deterministic_model` was flaky; it sometimes failed due to small numerical discrepancies. I loosened the tolerances a little.

Yes

## Test Plan
 Ran it 1000 times

### DIFF
--- a/test/utils/test_gp_sampling.py
+++ b/test/utils/test_gp_sampling.py
@@ -375,6 +375,8 @@ class TestRandomFourierFeatures(BotorchTestCase):
 
     def test_get_deterministic_model(self):
         tkwargs = {"device": self.device}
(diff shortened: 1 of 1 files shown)


saitcakmak (approved):
Thanks! Do you think it makes sense to restrict this to run with only a small set of seeds to make it 100% non-flaky?

esantorella:
> Thanks! Do you think it makes sense to restrict this to run with only a small set of seeds to make it 100% non-flaky?

Yep that makes sense.

Balandat (approved):
lgtm

Issue #1889, opened 2023-06-15
[Bug] timeout_sec only runs a single scipy_minimize step


roussel-ryan:
# 🐛 Bug

In this location https://github.com/pytorch/botorch/blob/185e98fc6f3b7ec9e36648c58312bf7783e91d4b/botorch/optim/optimize.py#L226
the variable timeout_sec is subtracted by the current machine time. As a result `timeout_sec` is almost always negative, resulting in an immediate `OptimizationTimeoutError`  in https://github.com/pytorch/botorch/blob/185e98fc6f3b7ec9e36648c58312bf7783e91d4b/botorch/optim/utils/timeout.py#L50 which means that the optimizer only takes a single step.

## To reproduce

** Code snippet to reproduce **
```python
import torch
from botorch.models import SingleTaskGP
from botorch.fit import fit_gpytorch_mll
from botorch.utils import standardize
from gpytorch.mlls import ExactMarginalLogLikelihood

train_X = torch.rand(10, 2)
Y = 1 - torch.norm(train_X - 0.5, dim=-1, keepdim=True)
Y = Y + 0.1 * torch.randn_like(Y)  # add some noise
train_Y = standardize(Y)

gp = SingleTaskGP(train_X, train_Y)
from botorch.acquisition import UpperConfidenceBound

UCB = UpperConfidenceBound(gp, beta=0.1)
from botorch.optim import optimize_acqf

bounds = torch.stack([torch.zeros(2), torch.ones(2)])
candidate, acq_value = optimize_acqf(
    UCB, bounds=bounds, q=1, num_restarts=5, raw_samples=20, timeout_sec=10
)
candidate  # tensor([0.4887, 0.5063])
```

## Expected Behavior

Running multiple optimization iterations until runtime is greater than timeout_sec (which is positive and in seconds).
## System information

**Please complete the following information:**
- botorch version 0.8.5

Balandat:
Hmm yeah this looks odd, what we should be doing here is sth like `timeout_sec = timeout_sec / opt_inputs.q` instead, unless I'm missing something. We should also make sure that the unit tests check this behavior.  cc @esantorella who I believe did a more recent refactor of this.

esantorella:
Fixed by #1893

PR #1893, opened 2023-06-18
Fix 'timeout' arithmetic

esantorella:
Addresses #1889 

## Motivation

See #1889 for context, since @roussel-ryan 's diagnosis looks exactly right (thank you very much!). We should not be subtracting `start_time` (and actually I don't see why it exists at all, so I removed it). 

## Test Plan

I added a unit test checking that the correct values of `timeout_sec` are passed down to `minimize_with_timeout` for several different cases.


Balandat (approved):
Thanks, lgtm!

PR #1895, opened 2023-06-18
Remove some usages of `_filter_kwargs` and code that triggers warnings from `_filter_kwargs`

esantorella:
## Motivation

Unused keyword arguments are a perennial bugbear. All else equal, it's better for a function to error upon receiving an unused keyword argument rather than ignoring it. `_filter_kwargs` gives a warning rather than an error if unused keyword arguments are passed. The warnings were cluttering unit test output and annoying me, so I removed usage of `_filter_kwargs` where it was straightforward to do so. I left a couple calls that would have been tricky to change. I also updated a tutorial to not use the deprected `fit_gpytorch_torch` function, which tends to generate a lot of `_filter_kwargs` warnings.

Yes

## Test Plan

Units

### DIFF
--- a/botorch/optim/optimize.py
+++ b/botorch/optim/optimize.py
@@ -313,16 +313,15 @@ def _optimize_acqf_batch(
         if has_parameter_constraints:
             # only add parameter constraints to gen_kwargs if they are specified
             # to avoid unnecessary warnings in _filter_kwargs
-            gen_kwargs.update(
-                {
-                    "inequality_constraints": opt_inputs.inequality_constraints,
-                    "equality_constraints": opt_inputs.equality_constraints,
-                    # the line is too long
-                    "nonlinear_inequality_constraints": (
-                        opt_inputs.nonlinear_inequality_constraints
-                    ),
-                }
-            )
+            if opt_inputs.inequality_constraints is not None:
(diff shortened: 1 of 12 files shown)


botorch/optim/optimize.py
```diff
@@ -308,16 +303,17 @@ def _optimize_batch_candidates() -> Tuple[Tensor, Tensor, List[Warning]]:
...
+        # only add parameter constraints to gen_kwargs if they are specified
+        # to avoid unnecessary warnings in _filter_kwargs
+        if opt_inputs.inequality_constraints is not None:
+            gen_kwargs["inequality_constraints"] = opt_inputs.inequality_constraints
+        if opt_inputs.equality_constraints is not None:
+            gen_kwargs["equality_constraints"] = opt_inputs.equality_constraints
+        if opt_inputs.nonlinear_inequality_constraints is not None:
+            gen_kwargs[
+                "nonlinear_inequality_constraints"
+            ] = opt_inputs.nonlinear_inequality_constraints
```
Balandat, line 315:
```suggestion
        for ctype in ("inequality", "equality", "nonlinear_inequality"):
            attr = "_".join([ctype, constraints])
            val = getattr(opt_inputs, attr)
            if val is not None:
                gen_kwargs[attr] = val
```

Balandat (approved):

PR #1896, opened 2023-06-19
[Easy] Stop input transform tests from being flaky

esantorella:
## Motivation

Tests were flaky.
[x] Break one long test into two
[x] Loosen a tolerance
[x] Set seeds to be randomly chosen from the range (0, 1000), where I made sure this is not flaky

Yes

## Test Plan

Ran each test 1000 times with each seed in (0, 1000)

### DIFF
--- a/test/models/transforms/test_input.py
+++ b/test/models/transforms/test_input.py
@@ -7,6 +7,7 @@
 import itertools
 import warnings
 from copy import deepcopy
+from random import randint
 
 import torch
 from botorch import settings
@@ -142,7 +143,7 @@ class TestInputTransforms(BotorchTestCase):
 
     def test_normalize(self) -> None:
         # set seed to range where this is known to not be flaky
-        torch.manual_seed(torch.randint(1000, torch.Size([])).item())
+        torch.manual_seed(randint(0, 1000))
         for dtype in (torch.float, torch.double):
             # basic init, learned bounds
             nlz = Normalize(d=2)
@@ -531,7 +532,7 @@ class TestInputTransforms(BotorchTestCase):
         batch_shapes = (torch.Size(), torch.Size([2]))
         dtypes = (torch.float, torch.double)
         # set seed to range where this is known to not be flaky
-        torch.manual_seed(torch.randint(1000, torch.Size([])).item())
+        torch.manual_seed(randint(0, 1000))
 
         for d, batch_shape, dtype in itertools.product(ds, batch_shapes, dtypes):
             bounds = torch.tensor(
@@ -628,7 +629,7 @@ class TestInputTransforms(BotorchTestCase):
         int_idcs = [0, 4]
         categorical_feats = {2: 2, 5: 3}
         # set seed to range where this is known to not be flaky
-        torch.manual_seed(torch.randint(1000, torch.Size([])).item())
+        torch.manual_seed(randint(0, 1000))
         for dtype, batch_shape, approx, categorical_features in itertools.product(
             (torch.float, torch.double),
             (torch.Size(), torch.Size([3])),
@@ -777,7 +778,7 @@ class TestInputTransforms(BotorchTestCase):
 
     def test_log10_transform(self) -> None:
         # set seed to range where this is known to not be flaky
-        torch.manual_seed(torch.randint(1000, torch.Size([])).item())
+        torch.manual_seed(randint(0, 1000))
         for dtype in (torch.float, torch.double):
             # basic init
             indices = [0, 2]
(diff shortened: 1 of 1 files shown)


test/models/transforms/test_input.py
```diff
@@ -259,7 +261,9 @@ def test_normalize(self):
                     [X.min(dim=-2, keepdim=True)[0], X.max(dim=-2, keepdim=True)[0]],
                     dim=-2,
                 )
-                self.assertAllClose(nlz.bounds, expected_bounds)
+                atol = 1e-6 if dtype is torch.float32 else None
```
SebastianAment, line 264: What do you think of dividing the exponents here by two for double precision, rather than using the defaults implicitly through `None`?
```suggestion
                atol = 1e-6 if dtype is torch.float32 else 1e-12
```

test/models/transforms/test_input.py
```diff
@@ -259,7 +261,9 @@ def test_normalize(self):
...
+                rtol = 1e-4 if dtype is torch.float32 else None
```
SebastianAment, line 265:
```suggestion
                rtol = 1e-4 if dtype is torch.float32 else 1e-8
```

test/models/transforms/test_input.py
```diff
@@ -591,27 +597,25 @@ def test_chained_input_transform(self):
...
-            round_tf = Round(
-                integer_indices=int_idcs, categorical_features=categorical_feats
-            )
-            self.assertEqual(round_tf.integer_indices.tolist(), int_idcs)
-            self.assertEqual(round_tf.categorical_features, categorical_feats)
-            self.assertTrue(round_tf.training)
-            self.assertFalse(round_tf.approximate)
-            self.assertEqual(round_tf.tau, 1e-3)
-            self.assertTrue(round_tf.equals(Round(**round_tf.get_init_args())))
+    def test_round_transform_init(self):
```
SebastianAment, line 600: While we're at it ...
```suggestion
    def test_round_transform_init(self) -> None:
```

test/models/transforms/test_input.py
```diff
@@ -757,6 +776,8 @@ def test_round_transform(self):
                 )
 
     def test_log10_transform(self):
```
SebastianAment, line 778:
```suggestion
    def test_log10_transform(self) -> None:
```

test/models/transforms/test_input.py
```diff
@@ -811,6 +832,8 @@ def test_log10_transform(self):
                 self.assertTrue(torch.equal(log_tf.preprocess_transform(X), X_tf))
 
     def test_warp_transform(self):
```
SebastianAment, line 834:
```suggestion
    def test_warp_transform(self) -> None:
```

test/models/transforms/test_input.py
```diff
@@ -956,6 +979,8 @@ def test_warp_transform(self):
             self.assertTrue((warp_tf.concentration1 == 3.0).all())
 
     def test_one_hot_to_numeric(self):
```
SebastianAment, line 981:
```suggestion
    def test_one_hot_to_numeric(self) -> None:
```

test/models/transforms/test_input.py
```diff
@@ -1106,6 +1134,9 @@ def f2(x: Tensor, n_f: int = 1) -> Tensor:
             result = x[..., -2:].unsqueeze(-2)
             return result.expand(*result.shape[:-2], n_f, -1)
 
+        # set seed to range where this is known to not be flaky
+        torch.manual_seed(torch.randint(100, torch.Size([])).item())
```
SebastianAment, line 1138: Nit: Can add `from random import randint` at the top of the file and then use
```suggestion
        torch.manual_seed(randint(0, 100))
```
which does not allocate an intermediate tensor just to extract an integer.

test/models/transforms/test_input.py
```diff
@@ -1336,6 +1367,9 @@ def test_filter_features(self):
         with self.assertRaises(ValueError):
             FilterFeatures(torch.tensor([0, 1, 1], dtype=torch.long))
 
+        # set seed to range where this is known to not be flaky
+        torch.manual_seed(torch.randint(100, torch.Size([])).item())
```
SebastianAment, line 1371:
```suggestion
        torch.manual_seed(randint(0, 100))
```

test/models/transforms/test_input.py
```diff
@@ -1042,6 +1067,9 @@ def test_append_features(self):
         with self.assertRaises(ValueError):
             AppendFeatures(torch.ones(3, 4, 2))
 
+        # set seed to range where this is known to not be flaky
+        torch.manual_seed(torch.randint(100, torch.Size([])).item())
```
SebastianAment, line 1071: Same as on line 1138 (and there are a couple more).

test/models/transforms/test_input.py
esantorella: makes sense -- the defaults are `rtol=1e-5`, `atol=1e-8`, so that change is stricter

PR #1900, opened 2023-06-21
Point to example for constraints def in optimize_acqf docstring

Balandat:
Summary: In response to https://github.com/pytorch/botorch/discussions/1899

Differential Revision: D46900427

### DIFF
--- a/botorch/optim/optimize.py
+++ b/botorch/optim/optimize.py
@@ -446,7 +446,8 @@ def optimize_acqf(
         inequality_constraints: A list of tuples (indices, coefficients, rhs),
             with each tuple encoding an inequality constraint of the form
             `\sum_i (X[indices[i]] * coefficients[i]) >= rhs`. `indices` and
-            `coefficients` should be torch tensors. When q=1, or when
+            `coefficients` should be torch tensors. See the docstring of
+            `make_scipy_linear_constraints` for an example. When q=1, or when
             applying the same constraint to each candidate in the batch,
             `indices` should be a 1-d tensor. For inter-point constraints,
             `indices` must be a 2-d Tensor, where in each row `indices[i] =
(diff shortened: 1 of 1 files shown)


botorch/optim/optimize.py
```diff
@@ -446,7 +446,8 @@ def optimize_acqf(
         inequality_constraints: A list of tuples (indices, coefficients, rhs),
             with each tuple encoding an inequality constraint of the form
             `\sum_i (X[indices[i]] * coefficients[i]) >= rhs`. `indices` and
-            `coefficients` should be torch tensors. When q=1, or when
+            `coefficients` should be torch tensors. See the docstring of
+            `make_scipy_linear_constraints` for an example. When q=1, or when
```
saitcakmak, line 450: This is definitely an improvement but what is `make_scipy_linear_constraints` and where does one find it? It is not imported in this file, so the user would have to search the repo to find where it is defined.

DavidSiretMarques: google botorch make_scipy_linear_constraints. Also, look at [this post](https://github.com/pytorch/botorch/discussions/1899) I made when trying to understand how it works

Balandat: Fair point, this could definitely improved upon, but I wanted to get something out that's actionable (search/googleable) right away rather than putting this on a backlog list and then forget about it :)

DavidSiretMarques: Adding the 
>see the docstring of `make_scipy_linear_constraints` for an example 

allows for googling

I just searched for it and the 3rd option was the charm, giving me this [link](https://botorch.org/v/0.6.1/api/optim.html), where you can find, with a bit more search, [this part of the page](https://botorch.org/v/0.6.1/api/optim.html#botorch.optim.parameter_constraints.make_scipy_linear_constraints)
image


I can see it's not the best option, but at least it allows the user to understand what's going on behind the scenes.

DavidSiretMarques: You could also link to [this question I made in stackoverflow](https://stackoverflow.com/questions/76523226/botorch-equality-constraints-argument-example-in-the-optimize-acqf-function) in the docs. IT's just my rambling and understanding of how the process is working, but could be useful

Issue #1901, opened 2023-06-25
[Documentation][Easy fix] Docstring for `SobolQMCNormalSampler` illustrates incorrect usage


esantorella:
## Issue description

The docstring for `SobolQMCNormalSampler` gives example usage as  `SobolQMCNormalSampler(1024, seed=1234)`. However, this usage is deprecated. Correct usage would be `SobolQMCNormalSampler(torch.Size([1024]), seed=1234)`.

## Code example

```python
from botorch.sampling.normal import SobolQMCNormalSampler

SobolQMCNormalSampler(1024, seed=1024)
```

andrewkaldani:
Are you referring to the docstring for that class? 
```
    Example:
        >>> sampler = SobolQMCNormalSampler(1024, seed=1234)
        >>> posterior = model.posterior(test_X)
        >>> samples = sampler(posterior)
    """
```

esantorella:
> Are you referring to the docstring for that class?

Yep, that's the one. And thanks very much for the PR! If any further work is needed, I'll comment over there.

PR #1902, opened 2023-06-26
fixing issue #1901

andrewkaldani:
## Motivation

Supposed to fix issue #1901 . Let me know if this is not correct.

Yes

## Test Plan

No, test plan only docstring changed


esantorella (approved):
Thanks! I checked that this code doesn't result in a DeprecationWarning, whereas the old version did. (Checking that requires `warnings.simplefilter("always")` since DeprecationWarnings are suppressed by default.)

PR #1903, opened 2023-06-26
Stop round transform test from being flaky

esantorella:
## Motivation

A test for the round transform asserting that approximately-rounded inputs are closer to exactly-rounded values than the original inputs are fails flakily. This happens because an error is introduced due to the approximation used:

```python
x = torch.tensor(0., dtype=torch.float64)
# tensor(4.5398e-05, dtype=torch.float64)
# equivalent to (tanh(-5) + 1) / 2
print(approximate_round(x, tau=1e-1))
```

This PR sets the tolerance for the relevant test to the error induced by the approximation. I also added to a docstring.

Yes

## Test Plan

Is a test.

### DIFF
--- a/botorch/models/transforms/input.py
+++ b/botorch/models/transforms/input.py
@@ -695,6 +695,13 @@ class InputStandardize(AffineInputTransform):
 class Round(InputTransform, Module):
     r"""A discretization transformation for discrete inputs.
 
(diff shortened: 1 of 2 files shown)


saitcakmak (approved):
Thanks!

PR #1907, opened 2023-06-28
Fix escaped character error in BOPE tutorial and simplify plotting

esantorella:
## Motivation

* A Pandas upgrade changed the way that Greek letters needed to be escaped for plotting, so in #1777 we changed "\\\\zeta" to "\\zeta". However, this caused failures for those using an older version of Pandas. Therefore, we will simply represent it as "zeta" for plotting.
* Since only one replication is run, there is no need to compute an "sem", since it is always NaN over one data point. So I removed that part of the analysis and plotting logic.

Yes

## Test Plan

I ran the notebook.

## Related PRs

#1777


Balandat (approved):
Thanks!

Issue #1904, opened 2023-06-27
[Feature Request] Fixed features and nonlinear constraints


jduerholt:
# 🚀 Feature Request

I would like to use fixed features in combination with non-linear inequality constraints.

## Motivation

I have an optimization problem with 4 continuous variables and an NChooseK constraint on top of them modelled via continuous relaxation as a non-linear inequality constraint and a categorical variable. Running `optimize_acqf_mixed` on this problem leads to this `NotImplementedError`: https://github.com/pytorch/botorch/blob/main/botorch/generation/gen.py#L120. It is already stated in a comment that this should be in principle possible, especially in the case from above where the fixed feature is independent from the nonlinear constraint.

## Pitch

My idea would to just remove the if statement and the `NotImplementedError` and rely on the user that the callables for the nonlinear constraints are also valid for the reduced problem, but perhaps anybody has a better/safer solution. Any suggestions?

**Are you willing to open a pull request?**  Yes.

```[tasklist]
### Tasks
```

Balandat:
How about something like this?
```
if fixed_features:
    if nonlinear_inequality_constraints:
        new_callables = []
        for nic in nonlinear_inequality_constraints:

            def _wrap_nic(X: Tensor) -> Tensor:
                return nic(fix_features(X, fixed_features))

        new_callables.append(_wrap_nic)
    nonlinear_inequality_constraints = new_callables
```

jduerholt:
Looks perfect. Thanks. I will create a PR in the course of the next week.

PR #1912, opened 2023-07-03
Enabling candidate generation with both `non_linear_constraints` and `fixed_features`

jduerholt:
## Motivation

This PR adds the possibility to use `fixed_features` in combination with `nonlinear_inequality_constraints` as discussed in issue https://github.com/pytorch/botorch/issues/1904.

@Balandat: concerning the solution that you suggested: I think it is much easier as the `f_np_wrapper` is also used for the nonlinear constraints: https://github.com/pytorch/botorch/blob/38e20275188f4cb2b99c8631bb91450b59140308/botorch/generation/gen.py#L224 and `fix_features` is already applied there: https://github.com/pytorch/botorch/blob/38e20275188f4cb2b99c8631bb91450b59140308/botorch/generation/gen.py#L189

So no need for another wrapper, or?

A wrapper function would be only needed if one wants to have the possibility to use the nonlinear constraints also with the reduced domain, then one would need a method, which takes the reduced X as input and augments it with the fixed features. 

Here is some example code, that shows that it works:

``` python
import torch
from botorch.utils import standardize
from botorch.utils.sampling import get_polytope_samples

def constraint(X):
    return torch.sum(X[...,:3], dim=-1) -1.80

train_X = torch.rand(10, 3)
Y = 1 - torch.norm(train_X[:,:2] - 0.5, dim=-1, keepdim=True)
Y = Y + 0.1 * torch.randn_like(Y)  # add some noise
train_Y = standardize(Y)

gp = SingleTaskGP(train_X, train_Y)
fit_gpytorch_mll(mll)

from botorch.acquisition import UpperConfidenceBound

UCB = UpperConfidenceBound(gp, beta=0.1)

bounds = torch.stack([torch.zeros(3), torch.ones(3)])

batch_initial_conditions = get_polytope_samples(
    n=5,
    bounds=bounds,
    inequality_constraints=[(torch.tensor([0,1,2]),torch.tensor([1.,1.,1.]),1.8)],
    n_burnin=1000
).unsqueeze(-2)

    UCB, 
    bounds=bounds, 
    q=1, 
    num_restarts=5, 
    raw_samples=20,
    fixed_features={2:0.5}, 
    nonlinear_inequality_constraints = [constraint],
    batch_initial_conditions = batch_initial_conditions
)
print(candidate)  # tensor([0.4887, 0.5063])
print(candidate.sum())
```

Yes.

## Test Plan

Unit tests.

### DIFF
--- a/botorch/generation/utils.py
+++ b/botorch/generation/utils.py
@@ -66,7 +66,7 @@ class _NoFixedFeatures:
     upper_bounds: Optional[Union[float, Tensor]]
     inequality_constraints: Optional[List[Tuple[Tensor, Tensor, float]]]
     equality_constraints: Optional[List[Tuple[Tensor, Tensor, float]]]
-    nonlinear_inequality_constraints: Optional[List[Callable]]
+    nonlinear_inequality_constraints: Optional[List[Callable[[Tensor], Tensor]
 
 
 def _remove_fixed_features_from_optimization(
@@ -77,7 +77,7 @@ def _remove_fixed_features_from_optimization(
     upper_bounds: Optional[Union[float, Tensor]],
     inequality_constraints: Optional[List[Tuple[Tensor, Tensor, float]]],
     equality_constraints: Optional[List[Tuple[Tensor, Tensor, float]]],
-    nonlinear_inequality_constraints: Optional[List[Callable]],
+    nonlinear_inequality_constraints: Optional[List[Callable[[Tensor], Tensor],
 ) -> _NoFixedFeatures:
     """
     Given a set of non-empty fixed features, this function effectively reduces the
--- a/botorch/generation/utils.py
+++ b/botorch/generation/utils.py
@@ -7,11 +7,14 @@
 from __future__ import annotations
 
 from dataclasses import dataclass
-from typing import Dict, List, Optional, Tuple, Union
+from typing import Callable, Dict, List, Optional, Tuple, Union
 
 import torch
 from botorch.acquisition import AcquisitionFunction, FixedFeatureAcquisitionFunction
-from botorch.optim.parameter_constraints import _generate_unfixed_lin_constraints
+from botorch.optim.parameter_constraints import (
+    _generate_unfixed_lin_constraints,
+    _generate_unfixed_nonlin_constraints,
+)
 from torch import Tensor
 
 
@@ -63,6 +66,7 @@ class _NoFixedFeatures:
     upper_bounds: Optional[Union[float, Tensor]]
     inequality_constraints: Optional[List[Tuple[Tensor, Tensor, float]]]
     equality_constraints: Optional[List[Tuple[Tensor, Tensor, float]]]
+    nonlinear_inequality_constraints: Optional[List[Callable]]
 
 
 def _remove_fixed_features_from_optimization(
@@ -73,6 +77,7 @@ def _remove_fixed_features_from_optimization(
     upper_bounds: Optional[Union[float, Tensor]],
     inequality_constraints: Optional[List[Tuple[Tensor, Tensor, float]]],
     equality_constraints: Optional[List[Tuple[Tensor, Tensor, float]]],
+    nonlinear_inequality_constraints: Optional[List[Callable]],
 ) -> _NoFixedFeatures:
     """
     Given a set of non-empty fixed features, this function effectively reduces the
@@ -140,6 +145,11 @@ def _remove_fixed_features_from_optimization(
         dimension=d,
         eq=True,
     )
+    nonlinear_inequality_constraints = _generate_unfixed_nonlin_constraints(
+        constraints=nonlinear_inequality_constraints,
+        fixed_features=fixed_features,
+        dimension=d,
+    )
     return _NoFixedFeatures(
         acquisition_function=acquisition_function,
         initial_conditions=initial_conditions,
@@ -147,4 +157,5 @@ def _remove_fixed_features_from_optimization(
         upper_bounds=upper_bounds,
         inequality_constraints=inequality_constraints,
         equality_constraints=equality_constraints,
+        nonlinear_inequality_constraints=nonlinear_inequality_constraints,
     )
--- a/botorch/generation/utils.py
+++ b/botorch/generation/utils.py
@@ -66,7 +66,7 @@ class _NoFixedFeatures:
     upper_bounds: Optional[Union[float, Tensor]]
     inequality_constraints: Optional[List[Tuple[Tensor, Tensor, float]]]
     equality_constraints: Optional[List[Tuple[Tensor, Tensor, float]]]
-    nonlinear_inequality_constraints: Optional[List[Callable[[Tensor], Tensor]
+    nonlinear_inequality_constraints: Optional[List[Callable[[Tensor], Tensor]]]
 
 
 def _remove_fixed_features_from_optimization(
@@ -77,7 +77,7 @@ def _remove_fixed_features_from_optimization(
     upper_bounds: Optional[Union[float, Tensor]],
     inequality_constraints: Optional[List[Tuple[Tensor, Tensor, float]]],
     equality_constraints: Optional[List[Tuple[Tensor, Tensor, float]]],
-    nonlinear_inequality_constraints: Optional[List[Callable[[Tensor], Tensor],
+    nonlinear_inequality_constraints: Optional[List[Callable[[Tensor], Tensor]]],
 ) -> _NoFixedFeatures:
     """
     Given a set of non-empty fixed features, this function effectively reduces the
@@ -103,6 +103,11 @@ def _remove_fixed_features_from_optimization(
         equality constraints: A list of tuples (indices, coefficients, rhs),
             with each tuple encoding an inequality constraint of the form
             `sum_i (X[indices[i]] * coefficients[i]) = rhs`.
+        nonlinear_inequality_constraints: A list of callables with that represent
+            non-linear inequality constraints of the form `callable(x) >= 0`. Each
+            callable is expected to take a `(num_restarts) x q x d`-dim tensor as
+            an input and return a `(num_restarts) x q`-dim tensor with the
+            constraint values.
 
     Returns:
         _NoFixedFeatures dataclass object.
--- a/botorch/optim/parameter_constraints.py
+++ b/botorch/optim/parameter_constraints.py
@@ -333,8 +333,6 @@ def _generate_unfixed_nonlin_constraints(
 
     values = torch.tensor(list(fixed_features.values()), dtype=torch.double)
 
-    new_constraints = []
-
     def _wrap_nlc(nlc: Callable) -> Callable:
         def new_nlc(X: Tensor) -> Tensor:
             ivalues = values.to(X).expand(*X.shape[:-1], len(fixed_features))
--- a/botorch/optim/parameter_constraints.py
+++ b/botorch/optim/parameter_constraints.py
@@ -313,10 +313,10 @@ def _make_linear_constraints(
 
 
 def _generate_unfixed_nonlin_constraints(
-    constraints: Optional[List[Callable[[Tensor], Tensor]],
+    constraints: Optional[List[Callable[[Tensor], Tensor]]],
     fixed_features: Dict[int, float],
     dimension: int,
-) -> Optional[List[Callable[[Tensor], Tensor]]:
+) -> Optional[List[Callable[[Tensor], Tensor]]]:
     # If constraints is None or an empty list, then return itself
     if not constraints:
         return constraints
@@ -333,7 +333,9 @@ def _generate_unfixed_nonlin_constraints(
 
     values = torch.tensor(list(fixed_features.values()), dtype=torch.double)
 
-    def _wrap_nonlin_constraint(constraint: Callable[[Tensor], Tensor]) -> Callable[[Tensor], Tensor]:
+    def _wrap_nonlin_constraint(
+        constraint: Callable[[Tensor], Tensor]
+    ) -> Callable[[Tensor], Tensor]:
         def new_nonlin_constraint(X: Tensor) -> Tensor:
             ivalues = values.to(X).expand(*X.shape[:-1], len(fixed_features))
             X_perm = torch.cat([X, ivalues], dim=-1)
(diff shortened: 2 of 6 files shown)


test/optim/test_optimize.py
```diff
@@ -862,6 +862,26 @@ def nlc4(x):
...
+                q=1,
+                nonlinear_inequality_constraints=[nlc1, nlc2, nlc3, nlc4],
+                batch_initial_conditions=batch_initial_conditions,
+                num_restarts=num_restarts,
+                fixed_features={0: 2},
+            )
+            self.assertTrue(
+                torch.allclose(
+                    candidates,
+                    torch.tensor([[2, 1, 1]], **tkwargs),
```
Balandat, line 878: meganit - let's indicate directly that these are expected to be float (even if thats achieved via the kwargs)
```suggestion
                    torch.tensor([[2.0, 1.0, 1.0]], **tkwargs),
```

test/optim/test_optimize.py
```diff
@@ -862,6 +862,26 @@ def nlc4(x):
...
 
+            # test with fixed features
+            candidates, acq_value = optimize_acqf(
+                acq_function=mock_acq_function,
+                bounds=bounds,
+                q=1,
+                nonlinear_inequality_constraints=[nlc1, nlc2, nlc3, nlc4],
+                batch_initial_conditions=batch_initial_conditions,
+                num_restarts=num_restarts,
+                fixed_features={0: 2},
```
Balandat, line 873: Hmm so with this setup this test may pass even if fixing the features doesn't work (given that it only narrows this to a specific solution of the previous solution set). Can you use a different value here? I might also make sense to use a different mock acquisition function so that the optimizer is unique...

Balandat (commented):
Yeah I think this makes sense. If there are multiple fixed features it's going to be slower than optimizing over the reduced domain since the optimizer needs to (i) identify the fixed features and (ii) SLSQP scales pretty poorly with the problem dimension anyway.

But that's probably fine for now (maybe add this comment in code for future reference?

> tensor([0.4887, 0.5063])
 
This doesn't seem to be the right output? Should be 3dim with the third feature fixed to 0.5?

Balandat:
Test failure seems unrelated I'll take a look at that

jduerholt:
> Yeah I think this makes sense. If there are multiple fixed features it's going to be slower than optimizing over the reduced domain since the optimizer needs to (i) identify the fixed features and (ii) SLSQP scales pretty poorly with the problem dimension anyway.

I am already working on the proper solution with the reduced domain. I will give it a few more minutes and if I think it is feasible without taking too much time, I will add it to the PR.


> This doesn't seem to be the right output? Should be 3dim with the third feature fixed to 0.5?

You care correct, this is copied from the botorch website minimal example 
image
and I forgot to remove it :D

test/optim/test_optimize.py
jduerholt: I will have a deeper look to the test. All this mocking is sometimes a bit hard for me to understand ;)

Balandat: I don't think there is actually any mocking going on here - it's just that `mock_acq_function` is a `SquaredAcquisitionFunction` rather than a true acquisition function.

If you don't want to change that, one thing you could do would be to just replace one of the constraints to result in a different optimizer.

jduerholt:
Hi @Balandat,

I implemented now also the solution using the reduced domain. Which one do you prefer?

Best,

Johannes

botorch/optim/parameter_constraints.py
```diff
@@ -312,13 +312,48 @@ def _make_linear_constraints(
...
+    idx_X, idx_f = 0, dimension - len(fixed_features)
+    for i in range(dimension):
+        if i in fixed_features.keys():
+            selector.append(idx_f)
+            idx_f += 1
+        else:
+            selector.append(idx_X)
+            idx_X += 1
+
+    values = torch.tensor([v for v in fixed_features.values()], dtype=torch.double)
```
Balandat, line 334:
```suggestion
    values = torch.tensor(list(fixed_features.values()), dtype=torch.double)
```

botorch/optim/parameter_constraints.py
```diff
@@ -312,13 +312,48 @@ def _make_linear_constraints(
...
+        def new_nlc(X: Tensor) -> Tensor:
+            ivalues = values.to(X).expand(*X.shape[:-1], len(fixed_features))
+            X_perm = torch.cat([X, ivalues], dim=-1)
+            return nlc(X_perm[..., selector])
+
+        return new_nlc
+
+    for nlc in constraints:
+        new_constraints.append(_wrap_nlc(nlc=nlc))
+    return new_constraints
```
Balandat, line 348: Could just do this
```suggestion
    return [_wrap_nlc(nlc=nlc) for nlc in constraints]
```

test/optim/test_parameter_constraints.py
```diff
@@ -215,6 +216,42 @@ def test_make_scipy_linear_constraints_unsupported(self):
...
+            return x[..., 0] - 1
+
+        # first test with one constraint
+        new_constraints = _generate_unfixed_nonlin_constraints(
+            constraints=[nlc1], fixed_features={1: 2.0}, dimension=3
+        )
+        self.assertAllClose(
+            nlc1(torch.tensor([[4.0, 2.0, 2.0]], device=self.device)),
+            new_constraints[0](torch.tensor([[4.0, 2.0]], device=self.device)),
+        )
```
Balandat, line 233:
```suggestion
        new_nlc1,  = _generate_unfixed_nonlin_constraints(
            constraints=[nlc1], fixed_features={1: 2.0}, dimension=3
        )
        self.assertAllClose(
            nlc1(torch.tensor([[4.0, 2.0, 2.0]], device=self.device)),
            new_nlc1(torch.tensor([[4.0, 2.0]], device=self.device)),
        )
```

test/optim/test_parameter_constraints.py
```diff
@@ -215,6 +216,42 @@ def test_make_scipy_linear_constraints_unsupported(self):
...
+        # test with several constraints
+        constraints = [nlc1, nlc2]
+        new_constraints = _generate_unfixed_nonlin_constraints(
+            constraints=constraints, fixed_features={1: 2.0}, dimension=3
+        )
+        for i in range(2):
+            self.assertAllClose(
+                constraints[i](torch.tensor([[4.0, 2.0, 2.0]], device=self.device)),
+                new_constraints[i](torch.tensor([[4.0, 2.0]], device=self.device)),
+            )
```
Balandat, line 243:
```suggestion
        for nlc, new_nlc in zip(constraints, new_constraints):
            self.assertAllClose(
               nlc(torch.tensor([[4.0, 2.0, 2.0]], device=self.device)),
                new_nlc(torch.tensor([[4.0, 2.0]], device=self.device)),
            )
```

test/optim/test_parameter_constraints.py
```diff
@@ -215,6 +216,42 @@ def test_make_scipy_linear_constraints_unsupported(self):
...
+        constraints = [nlc1, nlc2]
+        new_constraints = _generate_unfixed_nonlin_constraints(
+            constraints=constraints, fixed_features={1: 2.0, 2: 1.0}, dimension=3
+        )
+        for i in range(2):
+            self.assertAllClose(
+                constraints[i](torch.tensor([[4.0, 2.0, 1.0]], device=self.device)),
+                new_constraints[i](torch.tensor([[4.0]], device=self.device)),
+            )
+
```
Balandat, line 254:
```suggestion
        for nlc, new_nlc in zip(constraints, new_constraints):
            self.assertAllClose(
                nlc(torch.tensor([[4.0, 2.0, 1.0]], device=self.device)),
                new_nlc(torch.tensor([[4.0]], device=self.device)),
            )

```

Balandat (approved):
Looks great, just a couple of minor style nits. Thanks for putting this up!

jduerholt:
@Balandat, I incorporated your changes and added another test which was still missing (from my perspective). From my side, this is now ready for merge.

Balandat:
Great, thanks you. I'll pull it in and merge it.

SebastianAment:
> Test failure seems unrelated I'll take a look at that

This was solved by #1919, should pass upon re-running the tests.

botorch/optim/parameter_constraints.py
```diff
@@ -312,13 +312,44 @@ def _make_linear_constraints(
     return constraints
 
 
+def _generate_unfixed_nonlin_constraints(
+    constraints: Optional[List[Callable]],
+    fixed_features: Dict[int, float],
+    dimension: int,
+) -> Optional[List[Callable]]:
```
SebastianAment, line 319: Two subjective suggestions to make this quicker to parse:
1) How about `_gen_nonlin_constraints_of_variable_features`? Could apply same convention to the linear case below. (cc @Balandat)
2) Could you add a doc-string here, something like "Given a dictionary of `fixed_features`, returns a list of non-linear constraints on the variable (non-fixed) features."

botorch/optim/parameter_constraints.py
```diff
@@ -312,13 +312,44 @@ def _make_linear_constraints(
...
+        if i in fixed_features.keys():
+            selector.append(idx_f)
+            idx_f += 1
+        else:
+            selector.append(idx_X)
+            idx_X += 1
+
+    values = torch.tensor(list(fixed_features.values()), dtype=torch.double)
+
+    def _wrap_nlc(nlc: Callable) -> Callable:
```
SebastianAment, line 336: Complete type signature and more explicit naming.
```suggestion
    def _wrap_nonlin_constraint(constraint: Callable[[Tensor], Tensor]) -> Callable[[Tensor], Tensor]:
```

botorch/optim/parameter_constraints.py
```diff
@@ -312,13 +312,44 @@ def _make_linear_constraints(
...
+        def new_nlc(X: Tensor) -> Tensor:
```
SebastianAment, line 337:
```suggestion
        def new_nonlin_constraint(X: Tensor) -> Tensor:
```

botorch/optim/parameter_constraints.py
```diff
@@ -312,13 +312,44 @@ def _make_linear_constraints(
...
+            ivalues = values.to(X).expand(*X.shape[:-1], len(fixed_features))
+            X_perm = torch.cat([X, ivalues], dim=-1)
+            return nlc(X_perm[..., selector])
```
SebastianAment, line 340:
```suggestion
            return constraint(X_perm[..., selector])
```

botorch/optim/parameter_constraints.py
```diff
@@ -312,13 +312,44 @@ def _make_linear_constraints(
...
+
+        return new_nlc
```
SebastianAment, line 342:
```suggestion
        return new_nonlin_constraint
```

botorch/optim/parameter_constraints.py
```diff
@@ -312,13 +312,44 @@ def _make_linear_constraints(
...
+
+    return [_wrap_nlc(nlc=nlc) for nlc in constraints]
```
SebastianAment, line 344:
```suggestion
    return [_wrap_nonlin_constraint(constraint=constraint) for constraint in constraints]
```

botorch/optim/parameter_constraints.py
SebastianAment, line 319:
```suggestion
) -> Optional[List[Callable[[Tensor], Tensor]]:
```

botorch/optim/parameter_constraints.py
```diff
@@ -312,13 +312,44 @@ def _make_linear_constraints(
     return constraints
 
 
+def _generate_unfixed_nonlin_constraints(
+    constraints: Optional[List[Callable]],
```
SebastianAment, line 316:
```suggestion
    constraints: Optional[List[Callable[[Tensor], Tensor]],
```

botorch/generation/utils.py
```diff
@@ -73,6 +77,7 @@ def _remove_fixed_features_from_optimization(
     upper_bounds: Optional[Union[float, Tensor]],
     inequality_constraints: Optional[List[Tuple[Tensor, Tensor, float]]],
     equality_constraints: Optional[List[Tuple[Tensor, Tensor, float]]],
+    nonlinear_inequality_constraints: Optional[List[Callable]],
```
SebastianAment, line 80:
```suggestion
    nonlinear_inequality_constraints: Optional[List[Callable[[Tensor], Tensor],
```

botorch/generation/utils.py
```diff
@@ -63,6 +66,7 @@ class _NoFixedFeatures:
     upper_bounds: Optional[Union[float, Tensor]]
     inequality_constraints: Optional[List[Tuple[Tensor, Tensor, float]]]
     equality_constraints: Optional[List[Tuple[Tensor, Tensor, float]]]
+    nonlinear_inequality_constraints: Optional[List[Callable]]
```
SebastianAment, line 69:
```suggestion
    nonlinear_inequality_constraints: Optional[List[Callable[[Tensor], Tensor]
```

SebastianAment (commented):
Thanks for adding this @jduerholt! Mainly left cosmetic suggestions that will aid readability, as well as completed type signatures. 

Last, could you rename the PR something a little more descriptive like "Enabling candidate generation with both `non_linear_constraints` and `fixed_features`"? This will make it easier to understand what's going on when browsing the commit history on a high level.

botorch/optim/parameter_constraints.py
SebastianAment: Thank you! If you're up for it, you can do both in this commit. The only mentions of `_generate_unfixed_lin_constraints` are in:
- `generation/utils.py`,
- `optim/parameter_constraints.py`, and
- `test/optim/parameter_constraints.py`.
A quick search and replace in these files should do it.

jduerholt: Hmm, but do you really prefer the new name? What do you do not like about `_generate_unfixed_lin_constraints`?

jduerholt: But ofc, I can do it ;)

Balandat (approved):

botorch/optim/parameter_constraints.py
SebastianAment: `_generate_unfixed_(non)lin_constraints` makes it sound like the constraints are "unfixed", whereas it is the features that are. But let's keep it then, especially since the docstring makes it clear now. Thanks for adding it!

PR #1923, opened 2023-07-10
Loopifying conditional kwarg processing in `_optimize_acqf_batch`

SebastianAment:
Summary: Small code simplification, pattern could be reused elsewhere for `kwarg` processing, particularly where a number of fields have to be check.

Differential Revision: D47297929

### DIFF
--- a/botorch/optim/optimize.py
+++ b/botorch/optim/optimize.py
@@ -305,14 +305,13 @@ def _optimize_acqf_batch(opt_inputs: OptimizeAcqfInputs) -> Tuple[Tensor, Tensor
 
         # only add parameter constraints to gen_kwargs if they are specified
         # to avoid unnecessary warnings in _filter_kwargs
-        if opt_inputs.inequality_constraints is not None:
-            gen_kwargs["inequality_constraints"] = opt_inputs.inequality_constraints
-        if opt_inputs.equality_constraints is not None:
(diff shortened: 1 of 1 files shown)


botorch/optim/optimize.py
```diff
@@ -305,14 +305,13 @@ def _optimize_batch_candidates() -> Tuple[Tensor, Tensor, List[Warning]]:
...
-            gen_kwargs[
-                "nonlinear_inequality_constraints"
-            ] = opt_inputs.nonlinear_inequality_constraints
+        for constraint_name in [
+            "inequality_constraints",
+            "equality_constraints",
+            "nonlinear_inequality_constraints",
+        ]:
+            if getattr(opt_inputs, constraint_name) is not None:
+                gen_kwargs[constraint_name] = getattr(opt_inputs, constraint_name)
```
esantorella, line 314: This is probably a bad idea since it doesn't work for Python < 3.8, but...

```suggestion
            if  (constraint := getattr(opt_inputs, constraint_name)) is not None:
                gen_kwargs[constraint_name] = constraint
```

esantorella (approved):
LGTM

botorch/optim/optimize.py
SebastianAment: Nice! The main botorch readme lists Python >= 3.8 as a requirement, should be fine :)

PR #1924, opened 2023-07-10
Require Python 3.9+, test for 3.9 and 3.11

esantorella:
## Motivation

Python 3.11 was released over 8 months ago, so we should be testing for it. We can also drop support for Python 3.8, which was supplanted by 3.9 nearly 3 years ago.

Yes

## Test Plan

Manually running nightly cron: https://github.com/pytorch/botorch/actions/runs/5508340274

### DIFF
--- a/.conda/meta.yaml
+++ b/.conda/meta.yaml
@@ -13,7 +13,7 @@ build:
 
 requirements:
   host:
-    - python>=3.8
+    - python>=3.9
     - setuptools
     - setuptools_scm
   run:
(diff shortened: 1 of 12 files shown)


Balandat:
Interesting, looks like py 3.11 is stricter on the dataclasses, I guess this will require a small patch.

saitcakmak (approved):
I think Python 3.8 was the latest version when I started using BoTorch. Python evolves fast!

PR #1927, opened 2023-07-10
Python 3.11 compliance for dataclasses: Don't use mutable defaults

esantorella:
## Motivation

See error here: https://github.com/pytorch/botorch/actions/runs/5508336054/jobs/10039481686?pr=1924
* Tested with a tensor instead of a `DenseContainer` since we needed something non-mutable
* Added a test that the tensor is properly cast to a `DenseContainer`
* made code more readable (for me anyway)

Yes

## Test Plan

Units

## Related PRs

Unblocks #1924

### DIFF
--- a/test/utils/test_datasets.py
+++ b/test/utils/test_datasets.py
@@ -30,14 +30,15 @@ class TestDatasets(BotorchTestCase):
     def test_supervised_meta(self):
         X = rand(3, 2)
         Y = rand(3, 1)
-        A = DenseContainer(rand(3, 5), event_shape=Size([5]))
+        t = rand(3, 5)
+        A = DenseContainer(t, event_shape=Size([5]))
         B = rand(2, 1)
 
         SupervisedDatasetWithDefaults = make_dataclass(
             cls_name="SupervisedDatasetWithDefaults",
             bases=(SupervisedDataset,),
             fields=[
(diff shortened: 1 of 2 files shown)


test/utils/test_datasets.py
```diff
@@ -30,14 +30,15 @@ def test_base(self):
...
+        t = rand(3, 5)
+        A = DenseContainer(t, event_shape=Size([5]))
         B = rand(2, 1)
 
         SupervisedDatasetWithDefaults = make_dataclass(
             cls_name="SupervisedDatasetWithDefaults",
             bases=(SupervisedDataset,),
             fields=[
-                ("default", DenseContainer, field(default=A)),
+                ("default", DenseContainer, field(default=t)),
```
saitcakmak, line 41: This was the part erroring out, right? I am not missing some other change?

saitcakmak (approved):
Thanks for the fix. A tangential question I have is: why do we need this meta-class in the first place? Couldn't the same goal be achieved using something like `__post_init__`?

esantorella:
> Thanks for the fix. A tangential question I have is: why do we need this meta-class in the first place? Couldn't the same goal be achieved using something like __post_init__?

Based on looking at the code, I think so. That would be less confusing.

PR #1933, opened 2023-07-11
Minor patch to MVNXPB

j-wilson:
### Overview
This PR introduces a minor fix to the MVNXPB algorithm, so that bounds are properly whitened when computing initial plug-in estimators (which can impact the pivot order).

### Test Plan
I have run the existing unit tests locally.

### DIFF
--- a/botorch/utils/probability/mvnxpb.py
+++ b/botorch/utils/probability/mvnxpb.py
@@ -183,11 +183,11 @@ class MVNXPB:
             inv_Lii = Lii.reciprocal()
             bounds_i = bounds[..., i, :].clone()
             if i != 0:
-                bounds_i = bounds_i - torch.einsum(
-                    "...i,...i", L[..., i, :i].clone(), y[..., :i].clone()
-                ).unsqueeze(-1)
+                bounds_i = bounds_i - torch.sum(
+                    L[..., i, :i].clone() * y[..., :i].clone(), dim=-1, keepdim=True
+                )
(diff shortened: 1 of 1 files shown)


botorch/utils/probability/mvnxpb.py
```diff
@@ -176,18 +176,20 @@ def solve(self, num_steps: Optional[int] = None, eps: float = 1e-10) -> Tensor:
...
-            else:
-                db = (L[..., i, :i].clone() * y[..., :i].clone()).sum(-1, keepdim=True)
-                lb, ub = (bounds[..., i, :].clone() - db).unbind(dim=-1)
-
-            Phi_i = Phi(inv_Lii * ub) - Phi(inv_Lii * lb)
+            bounds_i = bounds[..., i, :].clone()
+            if i != 0:
+                bounds_i = bounds_i - torch.einsum(
+                    "...i,...i", L[..., i, :i].clone(), y[..., :i].clone()
+                ).unsqueeze(-1)
```
Balandat, line 188: Why change to `einsum` here? The parsing of the formula actually has some overhead - I just tested this, for small dimensions the previous implementation is about 5X faster.

j-wilson: Ah, good catch. I switched to `einsum` to avoid allocating additional memory, thinking that it would run at roughly the same speed. Since it's significantly slower, we should probably revert.

esantorella (approved):
LGTM

PR #1946, opened 2023-07-25
Fixing issue with Windows timeout test

esantorella:
Context and explanation: Timing works differently on Windows, and its imprecision has caused test failures in the past. When we started testing against Python 3.11, Windows tests involving timing started failing flakily. Adding some print statements showed that when the tests failed, time elapsed was measured as 0.0 seconds when it should have been substantially more. This is odd, since timing (including `monotonic`) was supposed to be more precise in Python 3.11. I couldn't debug this too thoroughly since I don't have access to a Windows machine outside of GitHub Actions. 

The fix: Increased `sleep` from 1e-3s to 1e-2s in a test where we are making sure something times out within a limit of 1e-4s.

# Test plan:

I'm going to run units a few times to ensure that they pass consistently.

### DIFF
--- a/botorch/logging.py
+++ b/botorch/logging.py
@@ -6,6 +6,7 @@
 
 import logging
 
+import torch
 
 LOG_LEVEL_DEFAULT = logging.CRITICAL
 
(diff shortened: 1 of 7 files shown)


esantorella:
The bug is strange. Windows with Python 3.11 is measuring an elapsed time of 0.0 after sleeping for 1e-3s, but timing in Windows is supposed to be more accurate in 3.11.

saitcakmak (approved):
Thanks for the fix. I think it's ok to wait an extra 0.01 secs. Debugging a windows specific Python issue is probably not worth the extra time investment.

PR #1958, opened 2023-08-01
compare types with "is" so flake8 passes

esantorella:
## Motivation

Flake8 is failing in the CI because it now requires comparing types with "is" rather than "==".

Yes

## Test Plan

Units, flake8


saitcakmak (approved):
Thanks!

PR #1960, opened 2023-08-01
Bump required pytorch version to 1.13.1

Balandat:
This has been out for a while now, and we can include it as part of our "major" 0.9.0 version bump.


esantorella (approved):
LGTM... 2.0 has been out since March :)

PR #1963, opened 2023-08-01
Update min req linear_operator to 0.5.1

Balandat:
linear_operator 0.5.0 had a bug that causes some things in botorch to raise errors. This bumps things to the newly released 0.5.1 that fixes these issues.


saitcakmak (approved):
Thanks!

PR #1964, opened 2023-08-01
Release notes for 0.9.1

Balandat:
0.9.1 is the same as 0.9.0 apart from a version bump to linear_operator that was required since the pinned 0.5.0 version had a bug.


saitcakmak (approved):
Thanks

PR #1966, opened 2023-08-01
Hack around missing pytorch 1.13.1 binaries for py3.11 on mac

Balandat:
This fix is needed to avoid errors in the CI

### DIFF
--- a/.github/workflows/test_stable.yml
+++ b/.github/workflows/test_stable.yml
@@ -70,10 +70,14 @@ jobs:
       with:
         python-version: ${{ matrix.python-version }}
     - name: Install dependencies
+      env:
+        PYTHON_VERSION: ${{ matrix.python-version }}
       run: |
         python setup.py egg_info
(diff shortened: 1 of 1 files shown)


saitcakmak:
I think you need to do this for the deploy_on_release workflow as well: https://github.com/pytorch/botorch/blob/main/.github/workflows/deploy_on_release.yml

Balandat:
I don't think so, we don't pin to the specific torch version there, so it should just install 2.0.1 for that one

saitcakmak:
Yeah, that makes more sense :D

saitcakmak (approved):

Balandat:
Pushing the changes to https://github.com/pytorch/botorch/tree/no_mac_py311_ci to test it in the workflows. For some reason it's not picking up the new min_torch_version assignment - not sure what's going on I'll try a few things.

Balandat:
Updated, newest version tested here: https://github.com/pytorch/botorch/actions/runs/5733147831

PR #1968, opened 2023-08-02
Hotfix/polytopesampler seed

jduerholt:
## Motivation

During bugfixing, I came around the issue that setting a seed in `get_polytope_samples` is not leading to the same samples when called several times with the same seed. This PR fixes it. 

Yes.

## Test Plan

Unit tests.


Balandat (approved):
Thanks a lot for the fix!

PR #1980, opened 2023-08-10
Type signature fix for private methods of `qMultiStepLookahead`

SebastianAment:
Summary: `posterior_transform` is `Optional` in the constructor of the acquisition function, so it should be optional in the private methods as well.

Differential Revision: D48244486


saitcakmak:
committed

PR #1994, opened 2023-08-28
Remove long-deprecated functionality:  `weights` argument of `RiskMeasureMCObjective` and `squeeze_last_dim`

esantorella:
## Motivation

* `utils.transforms.squeeze_last_dim` was deprecated prior to 0.7.0 (#487 ), so since we are at 0.9.2 it can be deleted.
* The `weights` argument of `acquisition.risk_measures.RiskMeasureMCObjective` was deprecated in 0.7.2 (#1400 ). Technically, the deprecation message only says that `weights` should be None rather than that it should not be passed, but I think it's okay to just remove it. Where it was not the last argument, I added a `*` to require that subsequent arguments be keyword-only.
* 

Yes

## Test Plan

Existing units

## Related PRs

#487, #1400


Balandat (approved):
Thanks for cleaning this up! LGTM!

PR #1996, opened 2023-08-28
Store bounds in double precision in test functions

esantorella:
## Motivation

* BoTorch recommends double precision in optimization, and will warn on single-precision inputs.
* `torch.tensor(a_float)` will use single precision by default. This is how bounds are currently instantiated in test problems.
* Native floats are already double precision, so this is actually downcasting them.

## Test Plan

* Existing units. I'll take another look at tutorials once they run in the CI.
* This made numerical warnings disappear from the BOPE tutorial. (I didn't re-run the tutorial and update the outputs since the warnings were suppressed anyway.)


saitcakmak (approved):
Nice catch, thanks!

Issue #2001, opened 2023-09-01
Usage of deprecated torch.norm


kit1980:
`torch.norm` is deprecated and may be removed in a future PyTorch release. Its documentation and behavior may be incorrect, and it is no longer actively maintained.

Please follow https://pytorch.org/docs/stable/generated/torch.norm.html to update.

Issues found with https://github.com/pytorch/test-infra/tree/main/tools/torchfix:

./botorch/acquisition/penalized.py:47:14: TOR101 Use of deprecated function torch.norm
./botorch/acquisition/penalized.py:75:14: TOR101 Use of deprecated function torch.norm
./botorch/acquisition/penalized.py:104:20: TOR101 Use of deprecated function torch.norm
./botorch/acquisition/penalized.py:260:35: TOR101 Use of deprecated function torch.norm
./botorch/acquisition/penalized.py:292:17: TOR101 Use of deprecated function torch.norm
./botorch/models/pairwise_gp.py:674:21: TOR101 Use of deprecated function torch.norm
./botorch/test_functions/synthetic.py:142:60: TOR101 Use of deprecated function torch.norm
./botorch/test_functions/synthetic.py:234:17: TOR101 Use of deprecated function torch.norm
./botorch/test_functions/synthetic.py:458:31: TOR101 Use of deprecated function torch.norm
./botorch/utils/sampling.py:169:22: TOR101 Use of deprecated function torch.norm
./test/acquisition/test_penalized.py:44:18: TOR101 Use of deprecated function torch.norm
./test/acquisition/test_penalized.py:62:29: TOR101 Use of deprecated function torch.norm
./test/acquisition/test_penalized.py:80:18: TOR101 Use of deprecated function torch.norm
./test/acquisition/test_penalized.py:269:28: TOR101 Use of deprecated function torch.norm
./test/models/kernels/test_categorical.py:30:26: TOR101 Use of deprecated function torch.norm
./test/models/kernels/test_categorical.py:37:26: TOR101 Use of deprecated function torch.norm
./test/models/kernels/test_contextual.py:36:26: TOR101 Use of deprecated function torch.norm
./test/models/kernels/test_contextual.py:64:26: TOR101 Use of deprecated function torch.norm
./test/models/kernels/test_downsampling.py:51:26: TOR101 Use of deprecated function torch.norm
./test/models/kernels/test_downsampling.py:67:26: TOR101 Use of deprecated function torch.norm
[...]

PR #2002, opened 2023-09-01
Remove deprecated usage of `torch.norm`

Balandat:
See title

Addresses #2001

Differential Revision: D48923425


saitcakmak (approved):
Thanks!

PR #2006, opened 2023-09-04
LearnedObjective and PairwiseGP dtype fixes and cleanup

esantorella:
## Motivation

When `LearnedObjective.forward` is passed single-precision data, but it has a `pref_model` with double-precision data and an input transform, the coefficients of the input transform can be downcast from double-precision to single-precision. To prevent precision loss, and to ensure that the `LearnedObjective` gives the same result when called repeatedly, this PR up-casts the input data to double-precision in this situation.

## Changes:
* Cast single-precision  data to double in the situation describe above
* Single precision warning for `PairwiseGP`
* Cleaned up arguments to `PairwiseGP.__init__` so that they are passed explicitly rather than as part of `kwargs`
* 

Yes

## Test Plan

Added units:
* Tests for new warnings (and silenced warnings where we are intentionally passing in single-precision data for testing)
* Test that `LearnedObjective` gives the same result when called multiple times
* Minor clean-up in unit tests (e.g. `assertTrue(x is y)` -> `assertIs(x, y)`)


botorch/acquisition/objective.py
```diff
@@ -522,6 +523,13 @@ def forward(self, samples: Tensor, X: Optional[Tensor] = None) -> Tensor:
         )
 
 
+def _get_learned_objective_pref_model_mixed_dtype_warn() -> str:
+    return (
+        "pref_model has double-precision data, but single-precision data "
+        "was passed to the LearnedObjective. Upcasting to double."
+    )
```
Balandat, line 530: Why not make this a constant?

botorch/acquisition/objective.py
```diff
@@ -576,7 +584,17 @@ def forward(self, samples: Tensor, X: Optional[Tensor] = None) -> Tensor:
             A `(sample_size * num_samples) x batch_shape x N`-dim Tensor of
             objective values sampled from utility posterior using `pref_model`.
         """
-        post = self.pref_model.posterior(samples)
+        if samples.dtype == torch.float32 and any(
+            [d == torch.float64 for d in self.pref_model.dtypes_of_buffers]
```
Balandat, line 588: No need to create a list here
```suggestion
            d == torch.float64 for d in self.pref_model.dtypes_of_buffers
```

botorch/acquisition/objective.py
```diff
@@ -576,7 +584,17 @@ def forward(self, samples: Tensor, X: Optional[Tensor] = None) -> Tensor:
...
+        ):
+            warnings.warn(
+                _get_learned_objective_pref_model_mixed_dtype_warn(),
+                InputDataWarning,
+            )
+            samples = samples.to(torch.float64)
+
+        posterior_ = self.pref_model.posterior
+        post = posterior_(samples)
```
Balandat, line 597: Is there anything wrong with this? 
```suggestion
        post = self.pref_model.posterior(samples)
```

botorch/models/model.py
```diff
@@ -244,6 +245,10 @@ def train(self, mode: bool = True) -> Model:
             self._set_transformed_inputs()
         return super().train(mode=mode)
 
+    @property
+    def dtypes_of_buffers(self) -> Set[torch.dtype]:
+        return {t.dtype for t in self._buffers.values() if t is not None}
```
Balandat, line 250: Why not use the `.buffers()` iterator here? 
```suggestion
        return {t.dtype for t in self.buffers() if t is not None}
```

botorch/models/pairwise_gp.py
```diff
@@ -167,12 +169,17 @@ class PairwiseGP(Model, GP, FantasizeMixin):
 
     def __init__(
         self,
-        datapoints: Tensor,
-        comparisons: Tensor,
+        datapoints: Optional[Tensor],
+        comparisons: Optional[Tensor],
```
Balandat, line 173: Should we update the docstring with what happens when these are `None`?

botorch/models/pairwise_gp.py
```diff
@@ -184,22 +191,35 @@ def __init__(
...
+        # Input data validation
+        if datapoints is not None and datapoints.dtype == torch.float32:
+            warnings.warn(
+                _get_single_precision_warning(str(datapoints.dtype)),
+                category=InputDataWarning,
+            )
+        if comparisons is not None and comparisons.dtype.is_floating_point:
+            warnings.warn(
+                "An integer dtype is expected for `comparisons`.", InputDataWarning
+            )
```
Balandat, line 215: nice

Balandat (approved):
Many thanks!

PR #2009, opened 2023-09-06
Remove `train_Yvar` in `MixedSingleTaskGP.construct_inputs`

saitcakmak:
Summary: `MixedSingleTaskGP` only supports inferred noise at the moment. Due to recent dataset changes, we are passing in `train_Yvar`, which leads to errors when used from Ax. Originally raised in https://github.com/facebook/Ax/issues/1815

Differential Revision: D48987818


saitcakmak:
already committed

Issue #2032, opened 2023-10-03
[Bug] The definition of the center of the trust region is wrong


Luplz:
# 🐛 Bug

In [scalable_constrained_bo.ipynb](https://github.com/pytorch/botorch/blob/main/tutorials/scalable_constrained_bo.ipynb), **the center of the trust region** is wrong. It should be the best point (taking constraints into account) and not the maximum.

### Best point

Replace
```python
 # if the minimum voilation candidate is smaller than the violation of the incumbent 
 if min_violation < self.best_constraint_values.sum()  :
```
in the function `update_state` by
```python
 # if the minimum voilation candidate is smaller than the violation of the incumbent and the best constraint values is not valid (otherwise none of new candidates are not better)
 if min_violation < self.best_constraint_values.sum() and (not torch.all(self.best_constraint_values<=0)) :
```

### Center of the trust region 
Add 
```python
    best_x : Tensor = torch.ones(dim, ) * torch.inf
```
in the class`ScboState` and replace
```python
   x_center = X[Y.argmax(), :].clone()
```
in the function `generate_batch` by 
```python
   x_center = state.best_x.clone()
```

dme65:
Thanks for reporting! I will put up a fix for this as well as https://github.com/pytorch/botorch/issues/2031.

dme65:
Fixed in #2034.

PR #2033, opened 2023-10-03
[Bug] The definition of the center of the trust region is wrong #2032…

Luplz:
## Motivation

I have observed convergence errors by using [scalable_constrained_bo.ipynb](https://github.com/pytorch/botorch/blob/main/tutorials/scalable_constrained_bo.ipynb) for parametric shape optimisation.

Yes

## Test Plan

Resolution of 20-dimensional Ackley function with Two Simple Constraint Functions and resolution of (15) in [Bayesian optimization under mixed constraints with a slack-variable augmented Lagrangian](https://arxiv.org/pdf/1605.09466.pdf) (error in the paper, it's `-c1` and `-c2`). For the second problem, I have to add a bilog transformation of the constraint functions thus as

```python

class fun_test :

    def __init__(self, dim) :
        self.dim : int = dim
        self.bounds : Tensor = torch.zeros((2, dim))

    def __call__(self, x) :
        return -torch.sum(x)

def c1_(x) :
    return -0.5*np.sin(2*np.pi*(x[0]**2-2*x[1]))-x[0]-x[1]+1.5

def c2_(x) :
    return x[0]**2+x[1]**2-1.5

def c1(x) :
    c = c1_(x)
    return np.sign(c)*np.log(1+abs(c))

def c2(x) :
    c = c2_(x)
    return np.sign(c)*np.log(1+abs(c))
```

and take `btach_size=8`.


dme65:
Thanks for reporting the SCBO-related bugs. #2034 should hopefully fix everything, so I think we can close this!

Issue #2031, opened 2023-10-02
[Bug] The definition of total violation is wrong.


Jiangyan-Zhao:
# 🐛 Bug

In the modules below, the definition of **total violation** is wrong. 
It should be $\sum_{\ell=1}^m\max [c_{\ell}(x),0]$, not $\sum_{\ell=1}^m c_{\ell}(x)$.

- tutorial [scalable_constrained_bo.ipynb](https://github.com/pytorch/botorch/blob/main/tutorials/scalable_constrained_bo.ipynb) 
    - function `update_state()` 
- botorch.generation.sampling.ConstrainedMaxPosteriorSampling

## Example

Replace
```python
sum_violation = C_next.sum(dim=-1)
```
 in the function `update_state()` of the chunk [4] of the tutorial [scalable_constrained_bo.ipynb](https://github.com/pytorch/botorch/blob/main/tutorials/scalable_constrained_bo.ipynb) with
```python
violation = torch.where(C_next > 0, C_next, 0)
sum_violation = violation.sum(dim=-1)
```

saitcakmak:
Thanks for reporting! cc @nataliemaus

dme65:
Fixed in #2034.

PR #2034, opened 2023-10-04
Fix SCBO tutorial

dme65:
This fixes several constraint handling bugs in the SCBO tutorial as reported in #2031, #2032, and #2033. In addition, it also fixes a bug and cleans up some of the logic in `ConstrainedMaxPosteriorSampling`. 

The previous problem was too easy, so we never actually tested some of the aspects of the constraint handling. I switched to the same version of the Ackley problem as considered in https://arxiv.org/pdf/2002.08526.pdf and I'm observing comparable performance.


Balandat (approved):
lgtm pending the test TODO

botorch/generation/sampling.py
```diff
@@ -254,82 +252,101 @@ def __init__(
         objective: Optional[MCAcquisitionObjective] = None,
         posterior_transform: Optional[PosteriorTransform] = None,
         replacement: bool = True,
-        minimize_constraints_only: bool = False,
```
Balandat, line 257: Should we deprecate this arg so this doesn't cause hard failures in other code?

botorch/generation/sampling.py
```diff
@@ -254,82 +252,101 @@ def __init__(
...
             replacement=replacement,
         )
         self.constraint_model = constraint_model
-        self.minimize_constraints_only = minimize_constraints_only
+
+    def _convert_samples_to_scores(self, Y_samples, C_samples) -> Tensor:
+        r"""Convert the objective and constraint samples into a score.
+
+        The logic is as follows:
+            - If a realization has at least on feasible candidate we use the objective
```
Balandat, line 288:
```suggestion
            - If a realization has at least one feasible candidate we use the objective
```

botorch/generation/sampling.py
```diff
@@ -254,82 +252,101 @@ def __init__(
...
+        has_feasible_candidate = is_feasible.any(dim=-1)
+
+        scores = Y_samples.clone()
+        scores[~is_feasible] = -float("inf")
+        if (~has_feasible_candidate).any():
+            # Use negative total violation for samples where no candidate is feasible
+            total_violation = (
+                C_samples[~has_feasible_candidate]
+                .clamp(min=0)
+                .sum(dim=-1, keepdims=True)
```
Balandat, line 315: I thought it was `keepdim` not `keepdims`? Just checked though and it works. Seems like this is an  undocumented alias?

botorch/generation/sampling.py
```diff
@@ -254,82 +252,101 @@ def __init__(
...
+            A `num_samples x batch_shape x num_cand x 1`-dim Tensor of scores.
+        """
+        is_feasible = (C_samples <= 0).all(
+            dim=-1
+        )  # num_samples x batch_shape x num_cand
+        has_feasible_candidate = is_feasible.any(dim=-1)
+
+        scores = Y_samples.clone()
+        scores[~is_feasible] = -float("inf")
+        if (~has_feasible_candidate).any():
```
Balandat, line 310: This reads more intuitively. ALso it's a micro optimization that saves a few bit flips :)
```suggestion
        if not has_feasible_candidate.all():
```

Balandat: while we're at the topic of micro-optimziation, you may want to just flip the bits and  assign `~has_feasible_candidate` to a variable since you're using it repeatedly here

botorch/generation/sampling.py
dme65: I think this should be ok since this functionality probably has no users (that would have likely resulted in Github issues since it wasn't working as intended).

botorch/generation/sampling.py
dme65: Oh good catch, interesting that this worked!

PR #2037, opened 2023-10-06
Numpy deprecation warning

dme65:
This gets rid of the following warning: `DeprecationWarning: Conversion of an array with ndim > 0 to a scalar is deprecated, and will error in future.`


Balandat (approved):
lgtm

PR #2047, opened 2023-10-17
Fix handling of optional constraints arguments

### LINKED ISSUE #2042: [Bug] Optional constraints in `gen_batch_initial_conditions`
# 🐛 Bug

This is not a real bug but I suspect it's not desired behavior either. The function `gen_batch_initial_conditions` takes constraints in the following way:
```python
...
inequality_constraints: Optional[List[Tuple[Tensor, Tensor, float]]] = None,
equality_constraints: Optional[List[Tuple[Tensor, Tensor, float]]] = None,
...
```
and I guess most people would expect that the function should behave in the same way when passing empty constraints lists or when omitting the arguments / providing `None`s. 

However, further below in the function, there are explicit None-comparisons, i.e.
```python
...
elif inequality_constraints is None and equality_constraints is None:
...
```
which trigger different behavior when empty constraint lists or `None` are given as arguments. In particular, in the first case, `sample_q_batches_from_polytope` gets called while, in the second case, `draw_sobol_samples` is executed.

## To Reproduce

For example, use the following setup:
```python
train_X = torch.tensor([[0.0]])
train_Y = torch.tensor([[0.0]])
model = SingleTaskGP(train_X, train_Y)
qEI = qExpectedImprovement(model, best_f=0.0)
bounds = torch.tensor([[0.0], [1.0]])
```

The following calls then execute different code internally:
```python
Xinit = gen_batch_initial_conditions(
    qEI, bounds, q=3, num_restarts=25, raw_samples=500
)
Xinit = gen_batch_initial_conditions(
    qEI, bounds, q=3, num_restarts=25, raw_samples=500, equality_constraints=[]
)
```

## Expected Behavior

Both above calls of `gen_batch_initial_conditions` should trigger the same logic, namely `draw_sobol_samples` should be called, but they don't due to the above described issue.

## Potential Fix 
I guess the proper fix would be to explicitly convert the optional `None`s to empty lists at the function start and then test instead for:
```
elif not (inequality_constraints or equality_constraints):
```

AdrianSosic:
## Motivation

Fixes #2042.

Yes.

### DIFF
--- a/botorch/optim/initializers.py
+++ b/botorch/optim/initializers.py
@@ -338,7 +338,8 @@ def gen_batch_initial_conditions(
             n = raw_samples * factor
             if generator is not None:
                 X_rnd = generator(n, q, seed)
-            elif inequality_constraints is None and equality_constraints is None:
+            # check if no constraints are provided
+            elif not (inequality_constraints or equality_constraints):
                 if effective_dim <= SobolEngine.MAXDIM:
                     X_rnd = draw_sobol_samples(bounds=bounds_cpu, n=n, q=q, seed=seed)
                 else:


esantorella (approved):
LGTM, just one suggestion

PR #2049, opened 2023-10-17
Clean up usage of some deprecated functionality

saitcakmak:
Summary:
Updates some code & tests that lead to deprecation & other warnings.

See also https://github.com/cornellius-gp/gpytorch/pull/2423

Differential Revision: D50384526


esantorella (approved):
Wow thank you, this is a lot of different improvements! Especially nice to see we won't be in danger of using functionality from other libraries that won't disappear and to see a few usages of `fit_gpytorch_model` go.

PR #2077, opened 2023-10-30
Fix dataset equality checks

saitcakmak:
Summary: This was erroring out when self.Yvar was a tensor but other.Yvar was None. Updated to equality checks to handle this case.

Differential Revision: D50818635


esantorella (approved):
LGTM, thanks!

PR #2085, opened 2023-11-06
Fix PyTorch version in test-against-stable workflow

saitcakmak:
It seems that the conditional block was always evaluating to true, leading to PyTorch 2.0.1 being used across the board as the minimum version. With this change, all jobs except for Python 3.11 on macOS use PyTorch 1.31.1.

Note: I could not get the expression to work while keeping the `min_torch_version == '1.13.1'` check in place, so I removed it. It doesn't have any functional effect as long as we address the TODO when we bump the min version.

Old run that used PyTorch 2.0.1 for all jobs: https://github.com/pytorch/botorch/actions/runs/6739329549/job/18320600699
New run that uses correct versions: https://github.com/pytorch/botorch/actions/runs/6774951160/job/18413146849

1.13.1 with ubuntu jobs:
Screenshot 2023-11-06 at 10 25 48 AM
1.13.1 with Py 3.9 on MacOs:
Screenshot 2023-11-06 at 10 26 21 AM
2.0.1 with Py 3.11 on MacOS:
Screenshot 2023-11-06 at 10 31 10 AM


esantorella (approved):
Thanks!

PR #2086, opened 2023-11-06
Remove stable Ax from BoTorch CI, add smoke test tutorials to test-against-stable workflow

saitcakmak:
Stable Ax often lacks changes that would make it compatible with latest BoTorch. As a result, this was creating more noise than the signal it provided. Removing it to clean up signals.

Also adds a smoke-test tutorial run to the workflow, which should help provide faster signals when something fails.

Running in https://github.com/pytorch/botorch/actions/runs/6776353879
(there are a couple pre-existing failures, fixed in https://github.com/pytorch/botorch/pull/2083)


esantorella (approved):
Thanks!

PR #2091, opened 2023-11-08
Spelling fixes

esantorella:
## Motivation

I wanted to see if spelling issues could be fixed quickly and programmatically. I was able to identify misspellings pretty quickly with cspell and a custom dictionary, but fixing them did not prove to be time-efficient, so I only fixed the first few files alphabetically.

## Test Plan

Existing units (there is one minor code change; the rest are docstrings).


Balandat (approved):
spelling 🐝

PR #2106, opened 2023-11-14
pytorch/botorch import cleanup

azad-meta:
Differential Revision: D51316424


Balandat:
I don't think we want this. Note that this module is auto-generated by setuptools-scm during setup time as configured here: https://github.com/pytorch/botorch/blob/main/pyproject.toml#L7, so this is important for the installation to provide its version

PR #2136, opened 2023-12-03
Support different noise levels for different outputs in test functions 

### LINKED ISSUE #2135: [Feature Request] Support different noise levels for different outcomes in test functions
# 🚀 Feature Request

Add the ability for different outcomes of test functions to be subject to different levels of observation noise.

## Motivation

Currently, test functions with multiple outputs (`MultiObjectiveTestProblem` or `ConstrainedBaseTestProblem`) still only support a single `noise_std`. In practice we often see different noise levels on different outcomes, and we should allow to specify this in the test function setup.

seanohagan:
Adds support for different noise levels for different objectives in a multiobjective test function

Closes #2135 

Modifies `BaseTestProblem` class to allow for `noise_std` to be a list of floats, where the length should be the number of objectives in a `MultiObjectiveTestProblem`. This way each component of the objective function can be subject to a separate noise level.

### DIFF
--- a/botorch/test_functions/synthetic.py
+++ b/botorch/test_functions/synthetic.py
@@ -50,6 +50,7 @@ import math
 from typing import List, Optional, Tuple, Union
 
 import torch
+from botorch.exceptions.errors import InputDataError
 from botorch.test_functions.base import BaseTestProblem, ConstrainedBaseTestProblem
 from botorch.test_functions.utils import round_nearest
 from torch import Tensor
@@ -804,7 +805,57 @@ class ThreeHumpCamel(SyntheticTestFunction):
 #  ------------ Constrained synthetic test functions ----------- #
 
 
-class ConstrainedGramacy(ConstrainedBaseTestProblem, SyntheticTestFunction):
+class ConstrainedSyntheticTestFunction(
+    ConstrainedBaseTestProblem, SyntheticTestFunction
+):
+    r"""Base class for constrained synthetic test functions."""
+
+    def __init__(
+        self,
+        noise_std: Union[None, float, List[float]] = None,
+        constraint_noise_std: Union[None, float, List[float]] = None,
+        negate: bool = False,
+        bounds: Optional[List[Tuple[float, float]]] = None,
+    ) -> None:
+        r"""
+        Args:
+            noise_std: Standard deviation of the observation noise. If a list is
+                provided, specifies separate noise standard deviations for each
+                objective in a multiobjective problem.
+            constraint_noise_std: Standard deviation of the constraint noise.
+                If a list is provided, specifies separate noise standard
+                deviations for each constraint.
+            negate: If True, negate the function.
+            bounds: Custom bounds for the function specified as (lower, upper) pairs.
+        """
+        self.setup_constraint_noise(constraint_noise_std)
+        SyntheticTestFunction.__init__(
+            self, noise_std=noise_std, negate=negate, bounds=bounds
+        )
+
+    def setup_constraint_noise(self, constraint_noise_std):
+        """
+        Validates that constraint_noise_std has length equal to
+        the number of constraints, if given as a list
+
+        Args:
+            constraint_noise_std: Standard deviation of the constraint noise.
+                If a list is provided, specifies separate noise standard
+                deviations for each constraint.
+        """
+        if (
+            isinstance(constraint_noise_std, list)
+            and len(constraint_noise_std) != self.num_constraints
+        ):
+            raise InputDataError(
+                f"If specified as a list, length of constraint_noise_std "
+                f"({len(constraint_noise_std)}) must match the "
+                f"number of constraints ({self.num_constraints})"
+            )
+        self.constraint_noise_std = constraint_noise_std
+
+
+class ConstrainedGramacy(ConstrainedSyntheticTestFunction):
     r"""Constrained Gramacy test function.
 
     This problem comes from [Gramacy2016]_. The problem is defined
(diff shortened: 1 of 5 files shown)


Balandat (commented):
Thanks a lot for getting started on this! 

Overall this looks great, a couple of minor inline comments. Could you please also update the docstrings to explain how the new list arg works? 

The other thing that needs updating are the functions in https://github.com/pytorch/botorch/blob/main/botorch/test_functions/synthetic.py - this should largely be a search and replace exercise. There are also constrained problems (subclassing also from `ConstrainedBaseTestProblem`) that would currently not be handled by these changes - if you feel up for it you can think through what a good setup / interface for that could look like. But that's also something that you (or someone else) could do in a follow-up PR.

Balandat:
Thanks for the updates. 

> For the SyntheticTestFunction base class, I've updated the type hints and docstring, but it seems like all of its child classes (in test_functions/synthetic.py and test_functions/multi_fidelity.py are real-valued/single-objective examples. I wonder if the type hints (and docstrings) should remained unchanged as noise_std: Optional[float] for single-objective test problems (since a list should not be given for a problem with a single objective).

Yes, I think that makes a lot of sense. The constrained problems at the [end of synthetic.py](https://github.com/pytorch/botorch/blob/main/botorch/test_functions/synthetic.py#L802) technically do have multiple outputs, but this is exposed through the `evaluate_slack` function, so for these we do need to think of a separate way of defining the noise level - right now this is done [here](https://github.com/pytorch/botorch/blob/21cdb44b9ba4a7cfde46eeda1f98875fd3603995/botorch/test_functions/base.py#L104-L107) assuming that `noise_std` is a scalar. There currently isn't a separation of the noise level between objective(s) and constraints. I think the cleanest way to do this would be to give `ConstrainedBaseTestProblem` a separate list of constraint noise levels, say `constraint_noise_std: Union[None, float, List[float]]` that could be used in a similar way.  In order to expose this cleanly, we probably have a few options: 
1. Define a constructor for each synthetic constrained problem subclassing both `SyntheticTestFunction` and `ConstrainedBaseTestProblem` to ensure that `constraint_noise_std` can be set and assigned as an attribute
2. Define a `ConstrainedSyntheticTestFunction` that defines a generic constructor that basically just calls `SyntheticTestFunction.__init__()` and then assigns the `constraint_noise_std` attribute. Then all constrained synthetic problems can just subclass that and we don't have to write multiple constructors.

seanohagan:
Thanks for the suggestions. I liked the approach that you described

I gave the base class a `setup_constraint_noise` method and called it from the constructor instead of putting this logic in the constructor solely to reduce code duplication in `ConstrainedHartmann` and `ConstrainedHartmannSmooth`-- I figured I either had to write constructors for these two and do it like this, or basically just copy the Hartmann logic into them, and I preferred this.

botorch/test_functions/synthetic.py
```diff
@@ -802,7 +805,57 @@ def evaluate_true(self, X: Tensor) -> Tensor:
...
+        if (
+            isinstance(constraint_noise_std, list)
+            and len(constraint_noise_std) != self.num_constraints
+        ):
+            raise InputDataError(
+                f"If specified as a list, length of constraint_noise_std "
+                f"({len(constraint_noise_std)}) must match the "
+                f"number of constraints ({self.num_constraints})"
+            )
+        self.constraint_noise_std = constraint_noise_std
```
Balandat, line 855: Minor style suggestion: I prefer these kinds of helpers to primarily validate but do assignments that change class attributes in the constructor itself. Nothing wrong with your approach, but doing this makes it more clear where things are happening / the code easier to read IMO. 

```suggestion
        self.constraint_noise_std = self._validate_constraint_noise(constraint_noise_std)
        SyntheticTestFunction.__init__(
            self, noise_std=noise_std, negate=negate, bounds=bounds
        )

    def _validate_constraint_noise(self, constraint_noise_std) -> Union[None, float, List[float]]:
        """
        Validates that constraint_noise_std has length equal to
        the number of constraints, if given as a list

        Args:
            constraint_noise_std: Standard deviation of the constraint noise.
                If a list is provided, specifies separate noise standard
                deviations for each constraint.
        """
        if (
            isinstance(constraint_noise_std, list)
            and len(constraint_noise_std) != self.num_constraints
        ):
            raise InputDataError(
                "If specified as a list, length of constraint_noise_std "
                f"({len(constraint_noise_std)}) must match the "
                f"number of constraints ({self.num_constraints})"
            )
        return constraint_noise_std
```

test/test_functions/test_synthetic.py
```diff
@@ -111,6 +112,32 @@ def test_custom_bounds(self):
...
+class TestConstraintNoise(BotorchTestCase):
+
+    functions = [
+        DummyConstrainedSyntheticTestFunction(),
+        DummyConstrainedSyntheticTestFunction(constraint_noise_std=0.1),
+        DummyConstrainedSyntheticTestFunction(constraint_noise_std=[0.1]),
+    ]
+
+    def test_constraint_noise_length_validation(self):
+        with self.assertRaises(InputDataError):
```
Balandat, line 137:
```suggestion
        with self.assertRaisesRegex(InputDataError, "must match the number of constraints"):
```

test/test_functions/test_multi_objective.py
```diff
@@ -75,6 +75,8 @@ def test_base_mo_problem(self):
                     self.assertTrue(torch.equal(f_X, expected_f_X))
                 with self.assertRaises(NotImplementedError):
                     f.gen_pareto_front(1)
+            with self.assertRaises(InputDataError):
```
Balandat, line 78: Good practice to use `assertRaisesRegex` to make sure that the right error message is being raised (and that this doesn't come from some other similar error)

```suggestion
            with self.assertRaisesRegex(InputDataError, "must match the number of objectives"):
```

Balandat (approved):
Very nice, thanks for whipping this into shape so quickly. 

FYI there is a some internal code that we will have to update based on these changes, so while this is pretty much ready to go in of itself, it may be a few days until we can actually merge this in so as to not break anything downstream.

Thanks a lot for your contribution!

PR #2146, opened 2023-12-09
Circumvent conda install bug in >= 23.10.0

Balandat:
In newer conda versions, some packages end up not being installed correctly: https://github.com/conda/conda/issues/13412

Selecting the classic solver fixes this issue. This does this for all conda github workflows. Once this is fixed in conda itself we can remove this setting again.


dme65 (approved):
Thanks!

PR #2148, opened 2023-12-11
Make python a runtime dependency in conda's meta.yml

Balandat:
Analogue to https://github.com/cornellius-gp/gpytorch/pull/2457, necessary b/c of conda solver issues, see https://github.com/conda/conda/issues/13412#issuecomment-1849600501


esantorella (approved):
Thanks!

PR #2151, opened 2023-12-16
Fixed condition_on_observations in fully Bayesian models

hvarfner:
## Motivation

Conditioning on observations in fully bayesian models - enables fully Bayesian JES & KG(?).

Yes.

## Test Plan
Tests are written to ensure functionality for inferred and fixed noise.  __note that the `_aug_batch_shape` attribute assignment was removed in `condition_on_observations`.__ In `FullyBayesianGPs`, this argument could not be assigned (hence the removal). I could not find the use for this argument, and all tests passed when removing it. 

Other changes are commented throughout, and the changes were made so as to assure that FBGPs can have one set of training data throughout. Howver, conditioning on obervations adds a batch dim to the training data (which is necessary in GPyTorch [here](https://github.com/cornellius-gp/gpytorch/blob/58c033564d28a5537397bc464827783313534e56/gpytorch/models/exact_gp.py#L176)) to infer the correct batch dim.

### DIFF
--- a/botorch/models/fully_bayesian.py
+++ b/botorch/models/fully_bayesian.py
@@ -498,9 +498,8 @@ class SaasFullyBayesianSingleTaskGP(ExactGP, BatchedMultiOutputGPyTorchModel):
         rest of this method will not run.
         """
         self._check_if_fitted()
-        x = X.unsqueeze(MCMC_DIM)
-        mean_x = self.mean_module(x)
-        covar_x = self.covar_module(x)
+        mean_x = self.mean_module(X)
+        covar_x = self.covar_module(X)
         return MultivariateNormal(mean_x, covar_x)
 
     # pyre-ignore[14]: Inconsistent override
@@ -534,7 +533,7 @@ class SaasFullyBayesianSingleTaskGP(ExactGP, BatchedMultiOutputGPyTorchModel):
         """
         self._check_if_fitted()
         posterior = super().posterior(
-            X=X,
+            X=X.unsqueeze(MCMC_DIM),
             output_indices=output_indices,
             observation_noise=observation_noise,
             posterior_transform=posterior_transform,
@@ -542,3 +541,32 @@ class SaasFullyBayesianSingleTaskGP(ExactGP, BatchedMultiOutputGPyTorchModel):
         )
         posterior = GaussianMixturePosterior(distribution=posterior.distribution)
         return posterior
+
+    def condition_on_observations(
+        self, X: Tensor, Y: Tensor, **kwargs: Any
+    ) -> BatchedMultiOutputGPyTorchModel:
+        """Conditions on additional observations for a Fully Bayesian model (either
+        identical across models or unique per-model).
+
+        Args:
+            X: (Tensor): A `(batch_shape) x num_samples x d`-dim Tensor, where `d` is
+                the dimension of the feature space and `batch_shape` is the number of
+                 sampled models.
+            Y (Tensor): A `(batch_shape) x num_samples x 1`-dim Tensor, where `d` is
+                the dimension of the feature space and `batch_shape` is the number of
+                 sampled models.
+
+        Returns:
+            BatchedMultiOutputGPyTorchModel: _description_
+        """
+        if X.ndim < 3 or Y.ndim < 3:
+            # The can either be thrown here or in GPyTorch, when the inference of the
+            # batch dimension fails since the training data by default does not have
+            # a batch shape.
+            raise ValueError(
+                "Conditioning in fully Bayesian models must contain a batch dimension."
+                "Add a batch dimension (the leading dim) with length matching the "
+                "number of hyperparameter sets to the conditioned data."
+            )
+
+        return super().condition_on_observations(X, Y, **kwargs)
--- a/botorch/models/fully_bayesian.py
+++ b/botorch/models/fully_bayesian.py
@@ -549,24 +549,29 @@ class SaasFullyBayesianSingleTaskGP(ExactGP, BatchedMultiOutputGPyTorchModel):
         identical across models or unique per-model).
 
         Args:
-            X: (Tensor): A `(batch_shape) x num_samples x d`-dim Tensor, where `d` is
+            X: A `batch_shape x num_samples x d`-dim Tensor, where `d` is
                 the dimension of the feature space and `batch_shape` is the number of
-                 sampled models.
-            Y (Tensor): A `(batch_shape) x num_samples x 1`-dim Tensor, where `d` is
+                sampled models.
+            Y: A `batch_shape x num_samples x 1`-dim Tensor, where `d` is
                 the dimension of the feature space and `batch_shape` is the number of
-                 sampled models.
+                sampled models.
 
         Returns:
-            BatchedMultiOutputGPyTorchModel: _description_
+            BatchedMultiOutputGPyTorchModel: A fully bayesian model conditioned on
+              given observations. The returned model has `batch_shape` copies of the
+              training data in case of identical observations (and `batch_shape`
+              training datasets otherwise).
         """
-        if X.ndim < 3 or Y.ndim < 3:
-            # The can either be thrown here or in GPyTorch, when the inference of the
-            # batch dimension fails since the training data by default does not have
-            # a batch shape.
-            raise ValueError(
-                "Conditioning in fully Bayesian models must contain a batch dimension."
-                "Add a batch dimension (the leading dim) with length matching the "
-                "number of hyperparameter sets to the conditioned data."
-            )
+        if X.ndim == 2 and Y.ndim == 2:
+            # To avoid an error in GPyTorch when inferring the batch dimension, we add
+            # the explicit batch shape here. The result is that the conditioned model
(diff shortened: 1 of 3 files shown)


hvarfner:
@dme65

sdaulton:
Thanks for putting this up @hvarfner!

dme65:
Thanks for fixing this @hvarfner! These changes look good to me.

botorch/models/fully_bayesian.py
```diff
@@ -534,11 +533,40 @@ def posterior(
...
+            BatchedMultiOutputGPyTorchModel: _description_
+        """
+        if X.ndim < 3 or Y.ndim < 3:
+            # The can either be thrown here or in GPyTorch, when the inference of the
+            # batch dimension fails since the training data by default does not have
+            # a batch shape.
+            raise ValueError(
+                "Conditioning in fully Bayesian models must contain a batch dimension."
+                "Add a batch dimension (the leading dim) with length matching the "
+                "number of hyperparameter sets to the conditioned data."
```
saitcakmak, line 569: Is the number of hyper-parameter sets `self.num_mcmc_samples` or `self.batch_shape` (the two are the same, unless the model is already batched / fantasized)? If I were to attempt fantasizing from an already fantasized model, would this still hold? It'd be good to print out the expected shape as part of the error message.

botorch/models/gpytorch.py
```diff
@@ -224,6 +224,12 @@ def condition_on_observations(self, X: Tensor, Y: Tensor, **kwargs: Any) -> Mode
             >>> model = model.condition_on_observations(X=new_X, Y=new_Y)
         """
         Yvar = kwargs.get("noise", None)
+
+        # for fully bayesian models, the keyword argument "noise": None
+        # throws an error in LinearOperator releted to inferring batch dims
+        if "noise" in kwargs and kwargs["noise"] is None:
+            del kwargs["noise"]
+
```
saitcakmak, line 232:
```suggestion
        Yvar = kwargs.pop("noise", None)
```

botorch/models/gpytorch.py
```diff
@@ -489,7 +496,7 @@ def condition_on_observations(
         fantasy_model._input_batch_shape = fantasy_model.train_targets.shape[
             : (-1 if self._num_outputs == 1 else -2)
         ]
-        fantasy_model._aug_batch_shape = fantasy_model.train_targets.shape[:-1]
+
```
saitcakmak, line 499: I know that this is not really used outside of model initialization, but it'd still be good to keep it consistent with the model. Let's set it conditional on the model type.
```suggestion
        if not self._is_fully_bayesian:
            fantasy_model._aug_batch_shape = fantasy_model.train_targets.shape[:-1]
```

test/models/test_fully_bayesian.py
```diff
@@ -656,6 +666,77 @@ def test_custom_pyro_model(self):
                     atol=5e-4,
                 )
 
+    def test_condition_on_observation(self):
+
+        num_models = 3
+        num_cond = 2
+        for infer_noise, dtype in itertools.product(
+            (True,), (torch.float, torch.double)
+        ):
```
saitcakmak, line 675: Looks like you intending to test `infer_noise=False` as well here? You could also just test one option for each.
```suggestion
        for infer_noise, dtype in ((True, torch.float), (False, torch.double)):
```

test/models/test_fully_bayesian.py
```diff
@@ -656,6 +666,77 @@ def test_custom_pyro_model(self):
                     atol=5e-4,
                 )
 
+    def test_condition_on_observation(self):
```
saitcakmak, line 669: Thanks for adding these tests. For completeness, can you add a test that calls `model.fantasize` and a test that evaluates some acquisition function (e.g., qLogEI) with the fantasized model? I just want to make sure everything works e2e including the sampler related code inside the acquisition functions.

saitcakmak (changes requested):
Thanks for making these changes. I left some, mostly cosmetic, comments in line. 
One thing I want to make sure is that the model batch shape remains compatible with the model itself. To this end, could you also update the `batch_shape` property of `SaasFullyBayesianSingleTaskGP` to reflect the correct shape before & after fantasizing?

botorch/models/fully_bayesian.py
hvarfner: The number of hyperparameter sets is `num_mcmc_samples / thinning` (neither is an attribute, so no attribute indicating the number of models exists for the FBGB). `self.batch_shape` is added when conditioning on observations, since `cond_X`, `cond_Y` have a `batch_shape.` I have added a test for this now, too.

I presume that the logic behind not having a `batch_shape` by default to not carry `batch_shape` sets of training data by default, and so now it is only added when it is necessary.

hvarfner: Had a bit of an incorrect rushed comment before, which I deleted. =)

hvarfner:
@saitcakmak thanks for the feedback! 

I didn't consider fantasizing before, so I had to rework a little bit to add it. Now, the batch shapes are inferred in condition_on_observations for the fantasizing case `X.shape = n x d, Y.shape = fantasy_dim x batch_dim x n x 1` and the no-batch_shape case `X.shape = n x d, Y.shape = n x 1`. In both cases, the output batch shape will be non-empty, as seen in the added tests.

saitcakmak:
Thanks for making the changes. I had not realized that your original changes did not affect the model batch size (as in, the output shape and the shape of the lengthscales does not change in the conditioning tests). But fantasizing (which is the use case of conditioning I am most familiar with) does change the batch shape by adding a another dimension to it.

So, if I tried to use the fantasized model with some acquisition functions, it will now error out because the `batch_shape` property of the model is not updated. E.g., adding 
```
            acqf = qNoisyExpectedImprovement(
                model=fantasy_model, X_baseline=train_X, prune_baseline=False
            )
            acqf(train_X)
``` 
to the end of the fantasize test will raise `AssertionError: Expected the output shape to match either the t-batch shape of X, or the `model.batch_shape` in the case of acquisition functions using batch models; but got output with shape torch.Size([19]) for X with shape torch.Size([1, 10, 4]).` This is because the model batch shape becomes `fantasy_size, num_models` after fantasizing but Line 427-31 defines `batch_shape` as `torch.Size([self.num_mcmc_samples])` (equal to `num_models` in the test). 

One option here would be to update the definition of the `batch_shape` property to correctly return the batch shape of both the basic and fantasized models. Another would be to roll back the fantasize support, which seems to require some more work to get working correctly (probably beyond the scope of this PR). Knowledge gradient is the most common use case for fantasy models that I know of and it seems to trigger additional errors when I add this bit to the fantasize test:
```
        acqf = qKnowledgeGradient(
            model=model, num_fantasies=4
        )
        acqf(train_X)
```

hvarfner:
@saitcakmak Yeah, there seems to be some nuance to this that I failed to consider. I think there are two things that need to be ironed out if we want consistency (and retain only one batch of training data by default):

1. What should the shape of the FBGP posterior be? Currently, the input _must_ be `batch_shape * q * d`, so that `batch_shape * num_models * q * d` is returned by unsqueezing exactly `MCMC_DIM = -3` in the posterior (previously the forward) call. For a the aforementioned `fantasy_model` to work, two dims must be unsqueezed. (or (`len(batch_shape)` dims in general). Moreover, this makes `model.posterior(train_X)` rather awkward, since the shape of the posterior will have the first two dims swapped compared to the conventional call described above. `acqf(train_X)` does not give the intended shape for any FB-acquisition function AFAIK - it has to be three-dimensional.

2. When using `fantasize`, I think that the user would be pretty surprised if they received a 
`batch_shape * num_fantasies * num_models * q * d` (since  the fantasy dim typically goes first, right?). However, this is more in line with the current FBGP structure, and this solution could be accommodated by updating the `batch_shape` and unsqueezing `len(batch_shape)` dimensions in the posterior call. Ultimately, the `models` dim needs to trail the `batch_shape` dim due to [this line](https://github.com/pytorch/botorch/blob/6c546eff212a513f509a96cea9967149d2ea281c/botorch/utils/transforms.py#L263), so I struggle to see a general solution that gives the consistent `num_fantasies * batch_shape  * num_models * q * d` when fantasizing.

I am not quite sure how this can be solved in a consistent manner without reconsidering the FBGP altogether (i.e. making the model dim a leading dim), so I am for rolling back as of now!

hvarfner:
@saitcakmak done! 

I think the long-term fix (which I don't think would be too much work) would be to have `MCMC_DIM` be a leading dim, but the last one, i.e. at index `-4`. I could give this a go going forward if you don't foresee any issues with it.

saitcakmak (approved):
Thanks! I haven't thought too deeply about this, but I think a proper way to support fantasize would start with adding support for batched inputs more generally (in __init__)  and add the MCMC dimension as the outer-most dimension as you suggest. We could also store the train_inputs/targets as batched (along the MCMC) dimension to bring the model internals further in-line with other batched models (e.g., SingleTaskGP). That being said, this is a decent bit of work to make these changes (will also require updating other parts of the code base that have specialized logic for this model), so it might not be worth the effort unless there's a real motivator.

PR #2152, opened 2023-12-18
Add Multi Information Source Augmented GP

andreaponti5:
## Motivation

This pull request introduces the implementation of the **Augmented Gaussian Process** (AGP) and a related acquisition function based on UCB, namely **Augmented UCB** (AUCB), for multi information source problems.

> Candelieri, A., Archetti, F. Sparsifying to optimize over multiple information sources: an augmented Gaussian process based algorithm. Struct Multidisc Optim 64, 239–255 (2021). [https://doi.org/10.1007/s00158-021-02882-7](https://doi.org/10.1007/s00158-021-02882-7)

### AGP and AUCB in a nutshell
The key idea of the AGP is to fit a GP model for each information source and *augment* the observations on the high fidelity source with those from *cheaper* sources which can be considered as *reliable*. The GP model fitted on this *augmented* set of observations is the AGP.

The AUCB is a modification of the standard UCB - computed on the AGP - suitably proposed to also deal with the source-specific query cost.

### Example
This is what the AGP and AUCB look like on the Forrester problem considering 2 sources.
<p align="center">

</p>

### Some implementation details
The Augmented GP implementation is based on the `SingleTaskGP`. 
Each source is implemented as an independent `SingleTaskGP` and all the _reliable_ observations are used to fit the SingleTaskGP representing the AGP, namely `SingleTaskAugmentedGP`. A key difference with the Multi-Fidelity approaches in BoTorch, is that the dimension representing the source is not directly modelled by the AGP. 
In addition, a Fixed-Noise version of the AGP has been implemented based on the `FixedNoiseGP`.

The Augmented UCB implementation is based on the `UpperConfidenceBound`, but it is penalized by the cost of the source and the discrepancy from the AGP.

Yes, I have read the Contributing Guidelines on pull requests.

## Test Plan

A bunch of unit tests for both, the AGP and AUCB, has been implemented. Those are inspired by the `SingleTaskGP` and `UpperConfidenceBound` tests, respectively.

In addition, a notebook tutorial shows how to use the AGP model along with the AUCB acquisition function and a comparison with the Multi-Fidelity MES on the Augmented Branin test function.

Finally, we plan to release an arXiv soon, with an extensive comparison of Multi-Fidelity and Multi Information Source approaches in BoTorch. Here are some preliminary results on the Augmented Hartmann test function, considering three sources (with fidelities $[0.50, 0.75, 1.00]$). The AGP has been compared with the discrete Multi-Fidelity version of the Knowled Gradient (KG), Max-value Entropy Search (MES) and GIBBON. The figure on the left shows the best seen with respect to the query cost, while the figure on the right shows the best seen with respect to the wall-clock time.
<p align="center">

</p>

## Related PRs
The docs has been updated in this PR.


sdaulton:
Thanks for putting up the PR! This works looks quite interesting!

test/models/test_gp_regression_multisource.py
```diff
@@ -0,0 +1,488 @@
...
+def _get_random_data_with_source(batch_shape, n, d, n_source, q=1, **tkwargs):
+    dtype = tkwargs.get("dtype", torch.float32)
+    rep_shape = batch_shape + torch.Size([1, 1])
+    bounds = torch.stack([torch.zeros(d), torch.ones(d)])
+    bounds[-1, -1] = n_source - 1
+    train_x = (
+        get_random_x_for_agp(n=n, bounds=bounds, q=q).repeat(rep_shape).type(dtype)
+    )
+    train_y = torch.sin(train_x[..., :1] * (2 * math.pi)).type(dtype)
+    train_y = train_y + 0.2 * torch.randn(n, 1, **tkwargs).repeat(rep_shape)
```
esantorella, line 44: When run with `self.device` as CUDA, this is failing with the error "Expected all tensors to be on the same device" because `**tkwargs` puts the random noise on the GPU while the other tensors involved are on the CPU. I think we'd want everything to be on the GPU.

test/acquisition/test_multi_source.py
```diff
@@ -0,0 +1,147 @@
...
+            module = AugmentedUpperConfidenceBound(
+                model=mm,
+                beta=1.0,
+                best_f=torch.tensor(0.5, device=self.device, dtype=dtype),
+                cost={0: 1, 1: 0.5},
+            )
+            X = torch.zeros(1, 2, device=self.device, dtype=dtype)
+            ucb = module(X)
+            ucb_expected = torch.tensor([1.8460], device=self.device, dtype=dtype)
+            self.assertAllClose(ucb, ucb_expected, atol=1e-4)
```
esantorella, line 41: I'm seeing this fail on a GPU, when `self.device` is "cuda:0", because `ucb` is on the CPU. Looking at other unit tests, it seems like the convention is that `ucb` should be on the CPU.

test/models/test_gp_regression_multisource.py
andreaponti5: Thanks for the comment, this should be solved now. All the tensors are on the same device.

test/acquisition/test_multi_source.py
andreaponti5: Fixed it. The problem was, as you pointed out, in the computation of `ucb`.

saitcakmak:
Hi @andreaponti5. Thanks for the PR! We're trying out a new support model for community contributions: https://github.com/pytorch/botorch/blob/main/CONTRIBUTING.md#community-contributions-beta. Can you move the new additions to the `botorch_community`, `notebooks_community` & `test_community` directories? Thanks!

botorch_community/models/gp_regression_multisource.py
```diff
@@ -0,0 +1,401 @@
...
+            mean_module=mean_module,
+            outcome_transform=outcome_transform,
+            input_transform=input_transform,
+        )
+        mll = ExactMarginalLogLikelihood(gp.likelihood, gp)
+        fit_gpytorch_mll(mll)
+        return gp
+
+
+class FixedNoiseAugmentedGP(FixedNoiseGP):
```
Balandat, line 231: FWIW we have deprecated `FixedNoiseGP` and incorporated its functionality into `SingleTaskGP` by allowing to pass `train_Yvar` also to `SingleTaskGP` (optionally). Would be great to do the same setup here so that the implementations are analogous.

sphinx/source/models.rst
```diff
@@ -182,4 +182,4 @@ Inducing Point Allocators
 Other Utilties
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 .. automodule:: botorch.models.utils.assorted
-    :members:
+    :members:
```
Balandat, line 185: what happened here?

botorch_community/models/gp_regression_multisource.py
```diff
@@ -0,0 +1,401 @@
...
+
+
+class SingleTaskAugmentedGP(SingleTaskGP):
+    r"""A single-task multi-source GP model.
+
+    The Augmented Gaussian Process is described in [Ca2021ms]_.
+    The basic idea is to use GP sparsification for selecting a subset of
+    the function evaluations, among those performed so far over all the
+    different sources, as inducing locations to generate the AGP approximating f(x).
+    The GP sparsification proposed is an insertion method: the set of inducing
```
Balandat, line 83: The term "inducing locations" is a bit confusing here I feel, since this is typically used for inducing points in approximate GP models. But IIUC here there aren't any approximate GP models, it's just an adaptive data selection scheme to select which inputs to use in the model (in particular, there is no interpolation or the like going on).

Balandat (commented):
Thanks - looks like there are some minor lint and auto-sort code formatting issues. I also left a couple of inline comments that would be great to address (though potentially ok to do this in a follow-up PR).

botorch_community/models/gp_regression_multisource.py
andreaponti5: I incorporated `FixedNoiseAugmentedAGP` in `SingleTaskAugmentedGP` (and consequently the unit tests).

sphinx/source/models.rst
andreaponti5: Not sure what happened. I removed `SingleTaskAugmentedGP` from the doc. Community contributions shouldn't go in the documentation, right?

botorch_community/acquisition/augmented_multisource.py
```diff
@@ -0,0 +1,150 @@
...
+            A `(b1 x ... bk)`-dim tensor of Augmented Upper Confidence Bound values at
+            the given design points `X`.
+        """
+        alpha = torch.zeros(X.shape[0], dtype=X.dtype, device=X.device)
+        agp_mean, agp_sigma = self._mean_and_sigma(X[..., :-1])
+        cb = (self.best_f if self.maximize else -self.best_f) + (
+            (agp_mean if self.maximize else -agp_mean) + self.beta.sqrt() * agp_sigma
+        )
+        source_idxs = {
+            int(s.tolist()): torch.where(torch.round(X[..., -1], decimals=0) == s)[0]
```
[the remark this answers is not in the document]
Balandat: TIL that calling `.tolist()` on a zero-dim tensor produces a scalar. `item()` definitely makes more sense.

sphinx/source/models.rst
Balandat: @saitcakmak what are your thoughts re community contribs? Would be nice if we generated docs as part of the community modules (maybe not part of the core modules) given that people are putting effort into their docstrings (thanks!).

saitcakmak: Yeah, I think it'd be good to have generated docs for community modules on the website. We can build it when building the community tutorials for the website.

botorch_community/models/gp_regression_multisource.py
andreaponti5: Yes, it is based on a GP sparsification technique. We changed the description. I hope it's more clear now.

Balandat: thanks

Balandat:
Merged! Thanks again @andreaponti5 for your contribution.

PR #2165, opened 2024-01-05
Fix for bug that occurs when splitting single-element bins, use default BoTorch kernel for BAxUS.

LeoIV:
This commit does two things:

First, it fixes a bug that occurs when trying to split a bin with a single element.

Also, we now use the default BoTorch Matern kernel instead of using MLE and lengthscale constraints.

## Motivation

I received a bug report via email for a slightly different benchmark setup that affects the code in the BAxUS tutorial. The bug occurs in cases when, after splitting, a bin contains only a single element, but other bins contain more than one element. In that case, the previous code attempted to split that bin which later caused an error. This commit fixes this bug and, at the same time, removes the custom Matern kernel we used in the previous version. The kernel does not improve performance but adds overhead to the tutorial.

Yes

## Test Plan

I tested this version on multiple benchmark setups to ensure this bug is fixed.

## Related PRs

Initial PR for BAxUS tutorial: https://github.com/pytorch/botorch/pull/1559


saitcakmak (approved):
Thanks!

> we now use the default BoTorch Matern kernel instead of using MLE and lengthscale constraints.

Nice to use the existing modules whenever it makes sense. Easier to maintain :)

PR #2167, opened 2024-01-12
New directory for community contributions

saitcakmak:
Summary: Establishes the directory structure for BoTorch community contributions, associated notebooks & unit tests. Updates `CONTRIBUTING.md` with instructions on where each type of contribution should be made.

Differential Revision: D52747159


Balandat (commented):
Thanks for putting this up. Curious why you chose this module structure over a top-level `community` folder (where the import path would be `botorch.community.acquisition...` rather than `botorch_community.acquisition...`). Would this cause any import headaches? 

If we keep the current setup, should we rename `community_notebooks` to `notebooks_community` for consistency?

PR #2169, opened 2024-01-16
`get_acquisition_function` for fully bayesian GPs

jduerholt:
## Motivation

I encountered a bug when using `get_acquisition_function` together with a fully bayesian GP when trying to instantiate `qEI`, which seems also a newer bug as it was  definitly working in the past. The origin is the computation of `Y`, when we have a fully bayesian GP one also has to take also the mean over the individual predictions which is currently missing. This should be in principle also the case for `qLogEI`, but there it is working ...

I added a quick fix, but I am not sure if this is ok in the way I did it, or if I am introducing other bugs by my fix.

Yes.

## Test Plan

In principle unit tests but not yet implemented as I am not sure if the proposed solution is valid.

## Related PRs

### DIFF
--- a/botorch/acquisition/factory.py
+++ b/botorch/acquisition/factory.py
@@ -107,6 +107,11 @@ def get_acquisition_function(
         # Since these are the non-noisy variants, use the posterior mean at the observed
         # inputs directly to compute the best feasible value without sampling.
         Y = model.posterior(X_observed, posterior_transform=posterior_transform).mean
+        # in case of a fully bayesian GP, one has to take also the mean over the
+        # individual sets of hyperparams
+        if len(Y.shape) == 3:
+            Y = Y.mean(dim=0)
+
         obj = objective(samples=Y, X=X_observed)
         best_f = compute_best_feasible_objective(
             samples=Y,


SebastianAment:
Thanks for putting this up @jduerholt! 

Generally, we'd like to average over the different samples of the model as late as possible, so averaging over the different acquisition values in this case, and batching (broadcasting) over the different samples of the model before, e.g. during the computation of the `best_f` values.

This commit might be related: #2058. Do you have more information on the specific bug you ran into?

jduerholt:
> This commit might be related: #2058. Do you have more information on the specific bug you ran into?

For me it worked with the `qlogEI`, I only had problems with qEI. I will put up an MWE to reproduce the bug based on the `saasbo.ipynb` tutorial notebook over the course of the day!

jduerholt:
Here is the MWE:

```python

import os

import torch

from botorch import fit_fully_bayesian_model_nuts
from botorch.acquisition import qExpectedImprovement
from botorch.models.fully_bayesian import SaasFullyBayesianSingleTaskGP
from botorch.models.transforms import Standardize
from botorch.test_functions import Branin
from botorch.acquisition.factory import get_acquisition_function
from botorch.acquisition.objective import IdentityMCObjective, GenericMCObjective

SMOKE_TEST = os.environ.get("SMOKE_TEST")

tkwargs = {
    "device": torch.device("cuda:1" if torch.cuda.is_available() else "cpu"),
    "dtype": torch.double,
}

WARMUP_STEPS = 256 if not SMOKE_TEST else 32
NUM_SAMPLES = 128 if not SMOKE_TEST else 16
THINNING = 16

branin = Branin().to(**tkwargs)


def branin_emb(x):
    """x is assumed to be in [0, 1]^d"""
    lb, ub = branin.bounds
    return branin(lb + (ub - lb) * x[..., :2])


DIM = 30 if not SMOKE_TEST else 10

# Evaluation budget
N_INIT = 10
N_ITERATIONS = 8 if not SMOKE_TEST else 1
BATCH_SIZE = 5 if not SMOKE_TEST else 1
print(f"Using a total of {N_INIT + BATCH_SIZE * N_ITERATIONS} function evaluations")


X = SobolEngine(dimension=DIM, scramble=True, seed=0).draw(N_INIT).to(**tkwargs)
Y = branin_emb(X).unsqueeze(-1)
train_Y = -1 * Y  # Flip the sign since we want to minimize f(x)

gp = SaasFullyBayesianSingleTaskGP(
    train_X=X,
    train_Y=train_Y,
    train_Yvar=torch.full_like(train_Y, 1e-6),
    outcome_transform=Standardize(m=1),
)
fit_fully_bayesian_model_nuts(
    gp,
    warmup_steps=WARMUP_STEPS,
    num_samples=NUM_SAMPLES,
    thinning=THINNING,
    disable_progbar=True,
)

# both options crash for qEI but work for qLogEI
EI = get_acquisition_function(
    "qEI", model=gp, X_observed=X, objective=IdentityMCObjective()
)

EI = get_acquisition_function(
    "qEI",
    model=gp,
    X_observed=X,
    objective=GenericMCObjective(lambda Y, X: Y[..., -1]),
)

candidates, acq_values = optimize_acqf(
    EI,
    bounds=torch.cat((torch.zeros(1, DIM), torch.ones(1, DIM))).to(**tkwargs),
    q=BATCH_SIZE,
    num_restarts=10,
    raw_samples=1024,
)

```

Both options for calling `get_acquisition_function` fail when going into `optimize_acqf`, but it works when asking for `qLogEI` or the noisy variants. 

Maybe it helps you to reproduce the issue.

SebastianAment:
Thanks for adding the example, I was able to reproduce the issue. 

The problem is that qEI assumes `best_f` to be a `batch_shape`-dim Tensor and adds an `unsqueeze` in its forward pass, while `compute_best_feasible_objective` directly returns a `batch_shape x 1`-dim Tensor.

I'm working on a change to standardize the treatment of the shapes and add checks so that this will be easier to diagnose in the future. In the meantime, you can either do `EI.best_f = EI.best_f.squeeze(-1)` directly after the construction of `EI`, or if you are not working on a benchmark of different methods, I'd strongly recommend using `qLogEI` instead, due to its generally improved optimization performance ([paper](https://openreview.net/pdf?id=1vyAG6j9PE)).

jduerholt:
Hi @SebastianAment, thanks for investigating the issue so quickly. I saw that you already put up a PR to harmonize the treatments of `best_f`, so I will close this PR, as my solution was more a quick hack ;)

Regarding the new log ACQFs: I really like the paper and we are already using them as default ACQFs in `BoFire` (`qLogNEI` for SOBO and `qLogNEHVI` for MOBO) as they perform usually better as the original ones.

Issue #2183, opened 2024-01-30
[Bug] qMultiFidelityLowerBoundMaxValueEntropy fails when X_pending is not None


AlexanderMouton:
# 🐛 Bug

When the ```qMultiFidelityLowerBoundMaxValueEntropy``` acquisition function is constructed with a number of points that are pending evaluation (```X_pending```) and with the ```SingleTaskMultiFidelityGP``` model, the ```optimize_acqf``` method fails. This does not happen when the ```qMultiFidelityMaxValueEntropy``` acquisition function is used.

## To reproduce

** Code snippet to reproduce **
```python
import torch
from botorch.acquisition.cost_aware import InverseCostWeightedUtility

from botorch.acquisition.max_value_entropy_search import (
    qMultiFidelityLowerBoundMaxValueEntropy,
    qMultiFidelityMaxValueEntropy
)
from botorch.acquisition.utils import project_to_target_fidelity

from botorch.fit import fit_gpytorch_model
from botorch.models.cost import AffineFidelityCostModel
from botorch.models.gp_regression_fidelity import SingleTaskMultiFidelityGP
from botorch.models.transforms.outcome import Standardize
from botorch.optim.optimize import optimize_acqf
from gpytorch.mlls import ExactMarginalLogLikelihood

# create training data
train_x = torch.rand(20, 4)
train_y = (torch.sin(train_x[:, 0]) + torch.cos(train_x[:, 1])).unsqueeze(-1)

# create model
model = SingleTaskMultiFidelityGP(
    train_X=train_x,
    train_Y=train_y,
    nu=2.5,
    linear_truncated=True,
    outcome_transform=Standardize(m=1),
    data_fidelities=[2, 3],
)
# fit the model
mll = ExactMarginalLogLikelihood(model.likelihood, model)
fit_gpytorch_model(mll)

# define function for projecting to the highest fidelity
def project(X):
    return project_to_target_fidelity(X=X, target_fidelities={2: 1.0, 3: 1.0})

# create cost aware utility
cost_utility = InverseCostWeightedUtility(
    cost_model=AffineFidelityCostModel(
        fidelity_weights={2: 1.0, 3: 1.0},
    )
)

for mve_af in [qMultiFidelityMaxValueEntropy, qMultiFidelityLowerBoundMaxValueEntropy]:
    for pending_points in [None, torch.rand(1, 4)]:
        af = mve_af(
            model=model,
            candidate_set=torch.rand(20, 4),
[...]

PR #2193, opened 2024-02-07
Adds NotImplementedError for bug (#2183) in qMultiFidelityLowerBoundMaxValueEntropy class

AlexanderMouton:
## Motivation

Serves as a quick "fix" to [issue #2183](https://github.com/pytorch/botorch/issues/2183)

Yes.

## Test Plan

No test plan (other than running all unit tests) as only a single `if` statement that raises a `NotImplementedError` when applicable was added.

## Related PRs

None.

### DIFF
--- a/botorch/acquisition/max_value_entropy_search.py
+++ b/botorch/acquisition/max_value_entropy_search.py
@@ -731,6 +731,15 @@ class qMultiFidelityMaxValueEntropy(qMaxValueEntropy):
                 `q_e` additional points in each q-batch correspond to
                 additional ("trace") observations.
         """
+        if X_pending is not None and isinstance(
+            self, qMultiFidelityLowerBoundMaxValueEntropy
+        ):
+            # TODO: Fix bug where passing in pending points causes qMFLBMVE AF to break:
+            # https://github.com/pytorch/botorch/issues/2183
+            raise NotImplementedError(
+                "Pending points (`X_pending`) are not yet supported for the"
+                "`qMultiFidelityLowerBoundMaxValueEntropy` acquisition function."
+            )
         super().__init__(
             model=model,
             candidate_set=candidate_set,
--- a/botorch/acquisition/max_value_entropy_search.py
+++ b/botorch/acquisition/max_value_entropy_search.py
@@ -731,15 +731,6 @@ class qMultiFidelityMaxValueEntropy(qMaxValueEntropy):
                 `q_e` additional points in each q-batch correspond to
                 additional ("trace") observations.
         """
-        if X_pending is not None and isinstance(
-            self, qMultiFidelityLowerBoundMaxValueEntropy
-        ):
-            # TODO: Fix bug where passing in pending points causes qMFLBMVE AF to break:
-            # https://github.com/pytorch/botorch/issues/2183
-            raise NotImplementedError(
-                "Pending points (`X_pending`) are not yet supported for the"
-                "`qMultiFidelityLowerBoundMaxValueEntropy` acquisition function."
-            )
         super().__init__(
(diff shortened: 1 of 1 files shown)


esantorella (changes requested):
Thank you for this! Could you add a unit test for this error? BoTorch requires 100% test coverage. The test could be similar [this line](https://github.com/pytorch/botorch/blob/68fbb68d4a78efd2b803071d883081b2c45a70ae/test/acquisition/test_max_value_entropy_search.py#L249) and could go [here](https://github.com/pytorch/botorch/blob/68fbb68d4a78efd2b803071d883081b2c45a70ae/test/acquisition/test_max_value_entropy_search.py#L296).

I also realized there's a cleaner way of doing this. Rather than raising a `NotImplementedError`, `qMultiFidelityLowerBoundMaxValueEntropy` could just not accept an argument `X_pending`. That's substantially more verbose, but would avoid the unexpectedness of having a superclass reference its subclass. It would require adding an `__init__` method to `qMultiFidelityLowerBoundMaxValueEntropy`, which could look like this:

```
    def __init__(
        self,
        model: Model,
        candidate_set: Tensor,
        num_fantasies: int = 16,
        num_mv_samples: int = 10,
        num_y_samples: int = 128,
        posterior_transform: Optional[PosteriorTransform] = None,
        use_gumbel: bool = True,
        maximize: bool = True,
        cost_aware_utility: Optional[CostAwareUtility] = None,
        project: Callable[[Tensor], Tensor] = lambda X: X,
        expand: Callable[[Tensor], Tensor] = lambda X: X,
    ) -> None:
        r"""Single-outcome max-value entropy search acquisition function.

        Args:
            model: A fitted single-outcome model.
            candidate_set: A `n x d` Tensor including `n` candidate points to
                discretize the design space, which will be used to sample the
                max values from their posteriors.
            cost_aware_utility: A CostAwareUtility computing the cost-transformed
                utility from a candidate set and samples of increases in utility.
            num_fantasies: Number of fantasies to generate. The higher this
                number the more accurate the model (at the expense of model
                complexity and performance) and it's only used when `X_pending`
                is not `None`.
            num_mv_samples: Number of max value samples.
            num_y_samples: Number of posterior samples at specific design point `X`.
            posterior_transform: A PosteriorTransform. If using a multi-output model,
                a PosteriorTransform that transforms the multi-output posterior into a
                single-output posterior is required.
            use_gumbel: If True, use Gumbel approximation to sample the max values.
            maximize: If True, considr the problem a maximization problem.
            cost_aware_utility: A CostAwareUtility computing the cost-transformed
                utility from a candidate set and samples of increases in utility.
            project: A callable mapping a `batch_shape x q x d` tensor of design
                points to a tensor of the same shape projected to the desired
                target set (e.g. the target fidelities in case of multi-fidelity
                optimization).
            expand: A callable mapping a `batch_shape x q x d` input tensor to
                a `batch_shape x (q + q_e)' x d`-dim output tensor, where the
                `q_e` additional points in each q-batch correspond to
                additional ("trace") observations.
        """
        super().__init__(
            model=model,
            candidate_set=candidate_set,
            num_fantasies=num_fantasies,
            num_mv_samples=num_mv_samples,
            posterior_transform=posterior_transform,
            use_gumbel=use_gumbel,
            maximize=maximize,
            cost_aware_utility=cost_aware_utility,
            project=project,
            expand=expand,
        )
```

AlexanderMouton:
Hi @esantorella 

Thanks for the feedback - that's a much cleaner suggestion that also doesn't require adding unit tests!

Regards,
Alex

esantorella (approved):
Thanks!

PR #2196, opened 2024-02-12
Fix EHVI, qEHVI, and qLogEHVI input constructors

esantorella:
## Motivation

WIP, needs some cleaning up.

The EHVI input constructor currently doesn't work because it returns an argument `objective` that EHVI doesn't support. 

[x] In EHVI input constructor inputs, raise an UnsupportedError if an `objective` is provided. No longer use the objective to construct the reference point; if this were done, the reference point would not match the acqf because the objective won’t be passed to the acqf.
[x] Add support for `PosteriorTransform` to EHVI input constructor.
[x] Move logic for using an `objective` to construct the reference point from EHVI input constructor to qEHVI and qLogEHVI input constructor.

## Have you read the [Contributing Guidelines on pull requests](https://github.com/pytorch/botorch/blob/main/CONTRIBUTING.md#pull-requests)?

Yes

## Test Plan

[x] Took these input constructors off the “skip list” for acqf input constructor test factory.
[x] Test for passing `posterior_transform` to EHVI
[ ] move objective tests from test for EHVI input constructor to qEHVI, qLogEHVI input constructor.

### DIFF
--- a/botorch/acquisition/input_constructors.py
+++ b/botorch/acquisition/input_constructors.py
@@ -72,13 +72,10 @@ from botorch.acquisition.multi_objective import (
     qNoisyExpectedHypervolumeImprovement,
 )
 from botorch.acquisition.multi_objective.logei import (
+    qLogExpectedHypervolumeImprovement,
     qLogNoisyExpectedHypervolumeImprovement,
 )
-from botorch.acquisition.multi_objective.objective import (
-    AnalyticMultiOutputObjective,
-    IdentityAnalyticMultiOutputObjective,
-    IdentityMCMultiOutputObjective,
-)
+from botorch.acquisition.multi_objective.objective import IdentityMCMultiOutputObjective
 from botorch.acquisition.multi_objective.utils import get_default_partitioning_alpha
 from botorch.acquisition.objective import (
     ConstrainedMCObjective,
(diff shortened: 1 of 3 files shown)


test/acquisition/test_input_constructors.py
```diff
@@ -772,121 +774,138 @@ def test_construct_inputs_EHVI(self) -> None:
                 constraints=mock.Mock(),
             )
 
-        # test with Y_pmean supplied explicitly
```
esantorella, line 775: TODO move these cases to qEHVI test

botorch/acquisition/input_constructors.py
```diff
@@ -881,42 +884,38 @@ def construct_inputs_EHVI(
...
     # Compute posterior mean (for ref point computation ref pareto frontier)
     # if one is not provided among arguments.
     if Y_pmean is None:
         with torch.no_grad():
             Y_pmean = model.posterior(X).mean
+
     if alpha > 0:
         partitioning = NondominatedPartitioning(
-            ref_point=ref_point,
-            Y=pre_obj(Y_pmean),
```
sdaulton, line 902: It looks like we lost this logic in the diff: namely, applying an objective to the incumbent Y in the case of q(Log)EHVI

esantorella: Good catch! forgot to push a commit :)

PR #2199, opened 2024-02-12
Reap deprecated support for `objective` with 1 arg in `GenericMCObjective`

esantorella:
Summary: This was deprecated as early as BoTorch 0.6.2.

Differential Revision: D53684317


Balandat (commented):
Thanks - tutorial failures are real (we either missed updating those or they were written using the deprecated format).

PR #2200, opened 2024-02-13
Consistent signature for BoTorch `get_objective_weights_transform`

esantorella:
Summary: `get_objective_weights_transform` sometimes returns a function that accepts a second argument and sometimes doesn't. Its docstring and type annotation are incorrect. This will become more problematic when we reap support for passing an `objective` with one argument to `GenericMCObjective` in https://github.com/pytorch/botorch/pull/2199

Differential Revision: D53706138


esantorella:
Closed as https://github.com/pytorch/botorch/commit/5441833907e8bf2281006eb02456e83807579f07 -- not sure why this didn't auto-close.

PR #2219, opened 2024-02-26
Pin torch nightly to known working version (2.3.0.dev20240222)

Balandat:
Recently many of our tests broke on pytorch's nightly newer than 2.3.0.dev20240222. This pins to that version so the CI is green while we can investigate what's going on.


Balandat:
Not the right solution to the problem; closing in favor of #2222

PR #2228, opened 2024-02-27
Remove deprecated args from base MCSampler

saitcakmak:
Summary: These have been deprecated since 0.8.0, time to let them go.

Differential Revision: D54280379


Balandat:
merged as f471f8116a395ea160cf1481847784db98a656c1

PR #2230, opened 2024-02-27
remap task values in MTGP if they are not contiguous integers starting from zero

sdaulton:
Summary: see title. This handles cases where the task values are e.g. {0,2}.

Reviewed By: saitcakmak

Differential Revision: D54121693


Balandat:
merged as bdd370817165a4716a6b3df1a312859823a9e8ab

PR #2231, opened 2024-02-28
unify ModelList and ModelListGP subset_output behavior

sdaulton:
Summary:
`ModelList` and `ModelListGP` have different `subset_output` methods. This is bug prone and is bug in SEBO.

`ModelList.subset_output` considers that `Model`s in the list may themselves be multi-output, and considers that when splitting the input argument `idcs` across models in the list---idcs is used to index into outputs

`ModelListGP.subset_output` does not consider that sub models may be multi-output and instead applies `idcs` is used to index into models.

A significant issue here is that `MultitaskGP` has the number of outputs equal to the number of tasks. This causes issues for typical uses cases where all tasks refer to the same Ax outcome e.g. the real metric and a proxy of it. This is not an issue with the current use of `ModelListGP`, but it is with `Modellist`.

To fix this, this diff unifies the `subset_output` behavior for `ModelListGP` and `Modellist` and prepares the appropriate arguments in Ax.

Ax-side diff will follow.

Reviewed By: saitcakmak, Balandat

Differential Revision: D54203833

### DIFF
--- a/botorch/models/gpytorch.py
+++ b/botorch/models/gpytorch.py
@@ -21,7 +21,11 @@ from typing import Any, List, Optional, Tuple, TYPE_CHECKING, Union
 
 import torch
 from botorch.acquisition.objective import PosteriorTransform
-from botorch.exceptions.errors import BotorchTensorDimensionError, InputDataError
+from botorch.exceptions.errors import (
+    BotorchTensorDimensionError,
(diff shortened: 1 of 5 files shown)


Balandat:
merged as f6773546067086df49a23803f1bee5ba3ffeea39

PR #2237, opened 2024-03-05
Avoid writing a "|" file in `run_tutorials` when runnnig a tutorial

Balandat:
I don't think this is necessary, and it creates a "|" file upon running tutorials that's annoying to deal with in git.

### DIFF
--- a/scripts/run_tutorials.py
+++ b/scripts/run_tutorials.py
@@ -53,7 +53,7 @@ def run_script(
     if env is not None:
         env = {**os.environ, **env}
     run_out = subprocess.run(
-        ["papermill", tutorial, "|"],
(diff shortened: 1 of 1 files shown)


Balandat:
@esantorella I ran into this when fixing some timeouts, is there any reason for this pipe to be here? 

Also, tangentially related, do we still need to write the results to a csv file upon executing tutorials if we don't upload those anymore?

Balandat:
hmm nvm looks like this is needed after all in CI (this ran fine locally when running the module)?

saitcakmak:
@Balandat if this doesn't work, let's abandon the PR to remove it from the queue.

PR #2263, opened 2024-03-25
ConstantKernel

SebastianAment:
Summary:
# `ConstantKernel` vs `ConstantMean`

**Summary**: `ConstantKernel` promises to more stably infer the constant that the default GP converges to as it moves away from the training data, than the current `ConstantMean` approach.

**By default**, GPyTorch, and BoTorch by extension currently uses a `ConstantMean` whose constant value is optimized during the optimization of the marginal likelihood.

**This notebook** compares the default approach with the approach of using a `ConstantKernel` instead, which represents the prior variance of an unknown constant value. In this case, the prior variance of the constant is optimized during the hyper-parameter optimization stage, but the actual value of the constant is computed afterward using the standard linear algebraic approach for the computation of Gaussian process posteriors. Therefore, the inference of the constant value is likely more stable with the kernel approach, in addition to including the uncertainty quantification of the constant. 

**Pros and Cons**: While the number of free parameters is the same for the `ConstantMean` and `ConstantKernel` approaches, the latter infers more information, more stably, That is the `ConstantKernel` approach infers the prior variance in our belief of the constant and the inference of the posterior mean is done in closed form through the standard linear algebraic approach. `ConstantMean` by constrast infers the value of the constant using numerical optimization, which can be fickle, at times giving rise to seemingly non-sensical constants outside of the range of the observed `Y` values.
A current limitation of `ConstantKernel` is that it allocates another intermediate Tensor for the constant kernel matrix, which is wasteful. A more efficient implementation would use a low-rank, or best, a lazily evaluated constant linear operator.

It is also notable that the `LinearKernel` currently does not support a constant offset (though the `PolynomialKernel` of degree 1 does), so the `ConstantKernel` can also be used to introduce the offset to linear GP models.

Differential Revision: D53666027

### DIFF
--- a/botorch/models/kernels/__init__.py
+++ b/botorch/models/kernels/__init__.py
@@ -5,6 +5,7 @@
 # LICENSE file in the root directory of this source tree.
 
 from botorch.models.kernels.categorical import CategoricalKernel
+from botorch.models.kernels.constant import ConstantKernel
 from botorch.models.kernels.downsampling import DownsamplingKernel
 from botorch.models.kernels.exponential_decay import ExponentialDecayKernel
 from botorch.models.kernels.linear_truncated_fidelity import (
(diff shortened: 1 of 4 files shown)


saitcakmak:
merged upstream https://github.com/cornellius-gp/gpytorch/pull/2511

PR #2277, opened 2024-04-03
Fix batch mode cross-validation tutorial

esantorella:
Summary: This notebook somehow got mis-formatted and wouldn't run in GitHub Actions. Trying the same thing again just in case that works.

Differential Revision: D55695922


saitcakmak:
abandoned

Issue #2156, opened 2023-12-20
[Bug] sample_polytope step size is correlated with direction when number of samples is small


sbernasek-albert:
# 🐛 Bug

In the `botorch.utils.sampling.sample_polytope()` function the same seed is used to generate both the random step sizes and the random step directions.  For very small numbers of steps (e.g. n=1), this causes the the step size and step direction to be strongly correlated, yielding strong sampling bias.

This is irrelevant for those using the `optimize_acqf` procedure that leverages large batches of hit-and-run samples generated in parallel. However, for custom optimization routines that require sequential hit-and-run samples, it may be necessary to call `sample_polytope(..., n=1)`. 

Looping through many such sequential calls yields this correlation between step size (`rands`) and step direction (`Rs`):

![image](https://github.com/pytorch/botorch/assets/150554991/59e8eaa7-5ed6-45e8-ad24-6b7f56911f2b)

A super simple fix would be to add 1 to the seed used to generate a step direction, e.g.  `sample_hypersphere(..., seed=seed+1)` . This eliminates any correlation between step-size and step-direction.

![image](https://github.com/pytorch/botorch/assets/150554991/5fd62b9a-0086-4c1c-91b6-b99c0468fe62)

saitcakmak:
Hi @sbernasek-albert. Apologies for the late response. The proposed fix seems quite reasonable. We'd be happy to accept a PR implementing the fix.

Balandat:
Fixed by #2290

PR #2290, opened 2024-04-09
Remove correlation between the step size and the step direction (#2156)

zankin:
## Motivation

This is a quick fix of #2156. 

In the ```botorch.utils.sampling.sample_polytope()```, both the step size ```rands``` and direction ```Rs``` are currently sampled using the same seed value. This can lead to strong correlation and induce sampling bias, particularly with a small number of steps (e.g., ```n=1```), as illustrated in #2156. Adding ```+1``` to the seed value for ```Rs``` helps to avoid such behaviour.

Yes.

## Test Plan

Not sure if any unit tests needed for that change.

### DIFF
--- a/botorch/utils/sampling.py
+++ b/botorch/utils/sampling.py
@@ -249,6 +249,7 @@ def sample_polytope(
     # pre-sample samples from hypersphere
     d = x0.size(0)
     # uniform samples from unit ball in d dims
+    # incrementing the seed by +1 to prevent correlation with the step size
     Rs = sample_hypersphere(
         d=d, n=n_tot, dtype=A.dtype, device=A.device, seed=seed + 1
     ).unsqueeze(-1)
--- a/botorch/utils/sampling.py
+++ b/botorch/utils/sampling.py
@@ -250,7 +250,7 @@ def sample_polytope(
     d = x0.size(0)
     # uniform samples from unit ball in d dims
     Rs = sample_hypersphere(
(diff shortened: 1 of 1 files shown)


botorch/utils/sampling.py
```diff
@@ -250,7 +250,7 @@ def sample_polytope(
     d = x0.size(0)
     # uniform samples from unit ball in d dims
     Rs = sample_hypersphere(
-        d=d, n=n_tot, dtype=A.dtype, device=A.device, seed=seed
+        d=d, n=n_tot, dtype=A.dtype, device=A.device, seed=seed + 1
```
sdaulton, line 253: Would you mind adding a brief comment for why this is necessary, that links to https://github.com/pytorch/botorch/issues/2156 please?

sdaulton:
Thanks for the PR! It looks good, but it would be great to add a comment on why this needed in line, for posterity.

botorch/utils/sampling.py
zankin: Done. Hope it is clearer now 🙂

Balandat:
Sorry for the delay @zankin, sdaulton has been out. Could you add the link to the issue as per my suggestion? Then I'm happy to merge this in.

zankin:
> Sorry for the delay @zankin, sdaulton has been out. Could you add the link to the issue as per my suggestion? Then I'm happy to merge this in.

Thanks! I've included a link to the issue in the comment line.

PR #2293, opened 2024-04-10
bump to minimum python version to 3.10 and test 3.12

sdaulton:
## Motivation
Bump minimum version of python from 3.9 -> 3.10 and start testing 3.12, since scientific python is no longer supporting py3.9: https://scientific-python.org/specs/spec-0000/

(Write your answer here.)

## Test Plan

Circle CI, running manual nightly job: https://github.com/sdaulton/botorch/actions/runs/8638469226

### DIFF
--- a/.github/workflows/deploy_on_release.yml
+++ b/.github/workflows/deploy_on_release.yml
@@ -15,10 +15,7 @@ jobs:
       fail-fast: false
       matrix:
         os: ["ubuntu-latest", "macos-14", "windows-latest"]
-        python-version: ["3.9", "3.11"]
-        exclude:
-          - os: "macos-14"
-            python-version: "3.9"  # not available for macos-14
+        python-version: ["3.10", "3.11", "3.12"]
     steps:
     - uses: actions/checkout@v4
     - name: Set up Python ${{ matrix.python-version }}
@@ -32,7 +29,7 @@ jobs:
       run: |
         pytest -ra --cov=. --cov-report term-missing
     - name: Upload coverage
-      if: ${{ runner.os == 'Linux' && matrix.python-version == 3.9 }}
+      if: ${{ runner.os == 'Linux' && matrix.python-version == 3.10 }}
       run: |
         bash <(curl -s https://codecov.io/bash)
 
@@ -79,7 +76,7 @@ jobs:
       with:
         miniconda-version: "latest"
         activate-environment: test
-        python-version: "3.9"
+        python-version: "3.10"
     - name: Fetch all history for all tags and branches
       run: git fetch --prune --unshallow
     - name: Install dependencies
--- a/.github/workflows/deploy_on_release.yml
+++ b/.github/workflows/deploy_on_release.yml
@@ -15,7 +15,7 @@ jobs:
       fail-fast: false
       matrix:
         os: ["ubuntu-latest", "macos-14", "windows-latest"]
-        python-version: ["3.10", "3.11", "3.12"]
+        python-version: ["3.10", "3.12"]
     steps:
     - uses: actions/checkout@v4
     - name: Set up Python ${{ matrix.python-version }}
@@ -46,7 +46,7 @@ jobs:
     - name: Set up Python
       uses: actions/setup-python@v5
       with:
-        python-version: "3.9"
+        python-version: "3.10"
     - name: Install dependencies
       run: |
         pip install .[test]
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -15,10 +15,7 @@ jobs:
       fail-fast: false
       matrix:
         os: ["ubuntu-latest", "macos-14", "windows-latest"]
-        python-version: ["3.9", "3.11"]
-        exclude:
-        - os: "macos-14"
-          python-version: "3.9"  # not available for macos-14
+        python-version: ["3.10", "3.11", "3.12"]
     steps:
     - uses: actions/checkout@v4
     - name: Set up Python ${{ matrix.python-version }}
@@ -37,7 +34,7 @@ jobs:
       run: |
         pytest -ra --cov=. --cov-report term-missing
     - name: Upload coverage
-      if: ${{ runner.os == 'Linux' && matrix.python-version == 3.9 }}
+      if: ${{ runner.os == 'Linux' && matrix.python-version == 3.10 }}
       run: |
         bash <(curl -s https://codecov.io/bash)
 
@@ -53,7 +50,7 @@ jobs:
     - name: Set up Python
       uses: actions/setup-python@v5
       with:
-        python-version: 3.9
+        python-version: 3.10
     - name: Install dependencies
       env:
         ALLOW_LATEST_GPYTORCH_LINOP: true
@@ -99,7 +96,7 @@ jobs:
       with:
         miniconda-version: "latest"
         activate-environment: test
-        python-version: "3.9"
+        python-version: "3.10"
     - name: Fetch all history for all tags and branches
       run: git fetch --prune --unshallow
     - name: Install dependencies
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -15,7 +15,7 @@ jobs:
       fail-fast: false
       matrix:
         os: ["ubuntu-latest", "macos-14", "windows-latest"]
-        python-version: ["3.10", "3.11", "3.12"]
+        python-version: ["3.10", "3.12"]
     steps:
     - uses: actions/checkout@v4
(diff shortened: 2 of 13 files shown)


.github/workflows/deploy_on_release.yml
```diff
@@ -15,10 +15,7 @@ jobs:
       fail-fast: false
       matrix:
         os: ["ubuntu-latest", "macos-14", "windows-latest"]
-        python-version: ["3.9", "3.11"]
-        exclude:
-          - os: "macos-14"
-            python-version: "3.9"  # not available for macos-14
+        python-version: ["3.10", "3.11", "3.12"]
```
saitcakmak, line 18: We test oldest & newest supported versions. No need to spend extra compute on 3.11.

.github/workflows/deploy_on_release.yml
```diff
@@ -32,7 +29,7 @@ jobs:
       run: |
         pytest -ra --cov=. --cov-report term-missing
     - name: Upload coverage
-      if: ${{ runner.os == 'Linux' && matrix.python-version == 3.9 }}
+      if: ${{ runner.os == 'Linux' && matrix.python-version == 3.10 }}
       run: |
         bash <(curl -s https://codecov.io/bash)
```
saitcakmak, line 35: Github won't let me comment there but there's another python-version 3.9 on line 49

setup.py
```diff
@@ -98,7 +98,7 @@ def read_deps_from_file(filname):
     ],
     long_description=long_description,
     long_description_content_type="text/markdown",
-    python_requires=">=3.9",
+    python_requires=">=3.10",
```
saitcakmak, line 101: Avoids the need for future updates
```suggestion
    python_requires=f">={REQUIRED_MAJOR}.{REQUIRED_MINOR}",
```

saitcakmak (changes requested):

saitcakmak:
Looks like there are 3 tutorial failures. Two of them should be resolved once the Ax PR lands. The last one seems to originate from the `cma` package. It might require an upstream fix before we can land this -- or we can temporarily disable that tutorial.

saitcakmak:
I looked into this a bit more closely. Looks like the failures happen because we're using development version of numpy 2.0 (installed with torch nightly), which changes some things with the copy kwarg.

Balandat:
@saitcakmak we don't need to use the numpy 2.0 dev version though, right? Or do the torch nightlies depend on it? If not we should be able to install numpy <2.0 first and then the torch nightlies in CI

saitcakmak:
I am guessing numpy 2.0 rc1 gets installed because the install command for pytorch nightly is `pip install --pre ...`. If `--pre` is not necessary, removing it could fix the issue without having to pin numpy. I recall this causing an issue with another dependency in the recent past

Balandat:
Yeah that's definitely the reason it gets installed. But if we manually install numpy (without the `--pre` option) prior to installing torch, it shouldn't be upgrading numpy, even when installing torch with `--pre` option.

saitcakmak:
I think removing `--pre` (if it works) is a cleaner solution than pinning `numpy`. Both in that it will not add additional commands in various places, and it won't introduce code that needs to be cleaned up later. By the time numpy 2.0 is released, we should aim to be compatible with it -- which doens't require a lot of work, as I found in https://github.com/CMA-ES/pycma/pull/251. We likely have much less `numpy` code than cma does.

sdaulton:
removing pre did the trick.

.github/workflows/nightly.yml
```diff
@@ -15,10 +15,7 @@ jobs:
       fail-fast: false
       matrix:
         os: ["ubuntu-latest", "macos-14", "windows-latest"]
-        python-version: ["3.9", "3.11"]
-        exclude:
-        - os: "macos-14"
-          python-version: "3.9"  # not available for macos-14
+        python-version: ["3.10", "3.11", "3.12"]
```
Balandat, line 18: let's not test 3.11 here? If things work for 3.10 and 3.12 they should also work for 3.11. And having 3.11 here will cost a bunch of GHA capacity (note that macos hosts are more expensive)

Balandat, line 18:
```suggestion
        python-version: ["3.10", "3.12"]
```

.github/workflows/test.yml
```diff
@@ -16,10 +16,7 @@ jobs:
       fail-fast: false
       matrix:
         os: ["ubuntu-latest", "macos-14", "windows-latest"]
-        python-version: ["3.9", "3.11"]
-        exclude:
-          - os: "macos-14"
-            python-version: "3.9"  # not available for macos-14
+        python-version: ["3.10", "3.11", "3.12"]
```
Balandat, line 19:
```suggestion
        python-version: ["3.10", "3.12"]
```

.github/workflows/test.yml
```diff
@@ -49,10 +46,7 @@ jobs:
       fail-fast: false
       matrix:
         os: ["ubuntu-latest", "macos-14", "windows-latest"]
-        python-version: ["3.9", "3.11"]
-        exclude:
-        - os: "macos-14"
-          python-version: "3.9"  # not available for macos-14
+        python-version: ["3.10", "3.11", "3.12"]
```
Balandat, line 49:
```suggestion
        python-version: ["3.10", "3.12"]
```

.github/workflows/test_stable.yml
```diff
@@ -12,10 +12,7 @@ jobs:
       fail-fast: false
       matrix:
         os: ["ubuntu-latest", "macos-14", "windows-latest"]
-        python-version: ["3.9", "3.11"]
-        exclude:
-          - os: "macos-14"
-            python-version: "3.9"  # not available for macos-14
+        python-version: ["3.10", "3.11", "3.12"]
```
Balandat, line 15:
```suggestion
        python-version: ["3.10", "3.12"]
```

.github/workflows/test_stable.yml
```diff
@@ -37,10 +34,7 @@ jobs:
...
```
Balandat, line 37:
```suggestion
        python-version: ["3.10", "3.12"]
```

.github/workflows/test_stable.yml
```diff
@@ -70,10 +64,7 @@ jobs:
       fail-fast: false
       matrix:
         os: ["ubuntu-latest", "macos-14"]
-        python-version: ["3.9", "3.11"]
-        exclude:
-        - os: "macos-14"
-          python-version: "3.9"  # not available for macos-14
+        python-version: ["3.10", "3.11", "3.12"]
```
Balandat, line 67:
```suggestion
        python-version: ["3.10", "3.12"]
```

.github/workflows/nightly.yml
saitcakmak, line 18:
```suggestion
        python-version: ["3.10", "3.12"]
```

Balandat (changes requested):

.github/workflows/reusable_tutorials.yml
```diff
@@ -43,14 +43,15 @@ jobs:
...
       uses: actions/setup-python@v5
       with:
-        python-version: "3.9"
+        python-version: "3.10"
     - name: Fetch all history for all tags and branches
       # We need to do this so setuptools_scm knows how to set the BoTorch version.
       run: git fetch --prune --unshallow
     - if: ${{ !inputs.use_stable_pytorch_gpytorch }}
       name: Install latest PyTorch & GPyTorch
+      # Ax and cma are not compatible with numpy 2.0 yet
```
saitcakmak, line 52:
```suggestion
```

.github/workflows/test.yml
saitcakmak, line 19:
```suggestion
        python-version: ["3.10", "3.12"]
```

.github/workflows/test.yml
saitcakmak, line 49:
```suggestion
        python-version: ["3.10", "3.12"]
```

.github/workflows/test_stable.yml
saitcakmak, line 15:
```suggestion
        python-version: ["3.10", "3.12"]
```

.github/workflows/test_stable.yml
saitcakmak, line 37:
```suggestion
        python-version: ["3.10", "3.12"]
```

.github/workflows/test_stable.yml
saitcakmak, line 67:
```suggestion
        python-version: ["3.10", "3.12"]
```

saitcakmak (approved):

PR #2313, opened 2024-04-29
Make a reusable workflow for pip tests; split CI testing for botorch & botorch_community 

saitcakmak:
Adds a `reusable_test_pip` workflow that runs tests with stable or latest PyTorch & GPyTorch and uploads to codecov (optionally). This reduces repetition within the workflow definitions a decent bit. One remaining pip test that is not covered by this workflow is "test against min supported", which didn't seem worth merging in due to its complicated dependency installation.

Also split up tests for botorch & botorch_community into two separate steps, with each step measuring coverage for the corresponding code base.


.github/workflows/reusable_test_pip.yml
```diff
@@ -0,0 +1,54 @@
...
+        pytest -ra test/ --cov botorch/ --cov-report term-missing --cov-report xml:botorch_cov.xml
+    - name: Unit tests and coverage -- BoTorch Community
+      run: |
+        pytest -ra test_community/ --cov botorch_community/ --cov-report term-missing --cov-report xml:botorch_community_cov.xml
+    - name: Upload coverage
+      if: ${{ inputs.upload_coverage && runner.os == 'Linux' && matrix.python-version == 3.10 }}
+      uses: codecov/codecov-action@v4
+      with:
+        fail_ci_if_error: true
+        token: ${{ secrets.CODECOV_TOKEN }}
```
Balandat, line 54: Does this by default upload all xml files with coverage info? There seems to be a `files` argument and the docs aren't fully clear what happens if that is not provided: https://github.com/codecov/codecov-action

Balandat (approved):

.github/workflows/reusable_test_pip.yml
saitcakmak: According to https://docs.codecov.com/docs/github-4a-merging-reports, it should merge them by default. Without writing to xml, I think the last run was overwriting the previous reports, so we got a weird report. Waiting to see what happens with the xmls

saitcakmak: Looks like it didn't merge them. Trying again with the `files` argument

PR #2314, opened 2024-04-29
Add test for GaussianMixturePosterior.mixture_covariance_matrix

saitcakmak:
This was only being tested as part of `test_community`, which does not count for coverage of `botorch` after https://github.com/pytorch/botorch/pull/2313. Adding a simple test to bump coverage.


Balandat (approved):
thanks

PR #2345, opened 2024-05-14
Swap out deprecated runner

ZainRizvi:
Github is deprecating the `N-core-ubuntu` runner label, replacing it with `N-core`.  We need to migrate PyTorch runners accordingly

This only affects how github queues jobs, not where those jobs actually run

More details in https://github.com/pytorch/pytorch/issues/125721


saitcakmak (approved):
Lgtm!

PR #2349, opened 2024-05-22
Swap out deprecated runner

ZainRizvi:
Github is deprecating the N-core-ubuntu runner label, replacing it with N-core. We need to migrate PyTorch runners accordingly. Turns out github had originally asked for the change to be in the wrong direction

This only affects how github queues jobs, not where those jobs actually run

More details in pytorch/pytorch#125721

### DIFF
--- a/.github/workflows/deploy_on_release.yml
+++ b/.github/workflows/deploy_on_release.yml
@@ -46,7 +46,7 @@ jobs:
 
   package-deploy-conda:
     name: Package conda and deploy to anaconda.org
-    runs-on: 8-core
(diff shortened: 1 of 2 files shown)


Balandat:
@ZainRizvi so this just undoes #2345? Was this intended to be a temporary thing? what is going on?

ZainRizvi:
@Balandat yeah, unfortunately. Turns out Github had accidentally asked for the change to be made in the opposite direction to what they actually wanted.  They've now clarified their actual request

Balandat (approved):
..... and back again

PR #2354, opened 2024-05-29
WIP Test codecov upload

saitcakmak:
Testing if passing in token this way makes a difference. It is annoying that I can't test this on main repo, since the behavior only reproduces if the PR is from a forked repo.

### DIFF
--- a/botorch/__init__.py
+++ b/botorch/__init__.py
@@ -51,6 +51,8 @@ gp_settings.max_eager_kernel_size._global_value = 4096
 
 
 
+
+
 __all__ = [
     "acquisition",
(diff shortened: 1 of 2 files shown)


saitcakmak:
Interesting that workflows are not starting. Maybe it is some sort of attack protection?

botorch/__init__.py
```diff
@@ -49,6 +49,8 @@
 gp_settings.max_eager_kernel_size._global_value = 4096
 
 
+
```
saitcakmak, line 52:
```suggestion



```

saitcakmak:
This doesn't make a difference

PR #2358, opened 2024-06-02
Update polytope sampling code and add thinning capability

Balandat:
This set of changes does the following:
* adds an `n_thinning` argument to `sample_polytope` and `HitAndRunPolytopeSampler`; changes the defaults for `HitAndRunPolytopeSampler` args to `n_burnin=200` and `n_thinning=20`
* Changes `HitAndRunPolytopeSampler` to take the `seed` arg in its constructor, and removes the arg from the `draw()` method (the method on the base class is adjusted accordingly). The resulting behavior is that if a `HitAndRunPolytopeSampler` is instantiated with the same args and seed, then the sequence of `draw()`s will be deterministic. `DelaunayPolytopeSampler` is stateless, and so retains its existing behavior.
* normalizes the (inequality and equality) constraints in `HitAndRunPolytopeSampler` to avoid the same issue as https://github.com/pytorch/botorch/issues/1225. If `bounds` are note provided, emits a warning that this cannot be performed (doing this would require vertex enumeration of the constraint polytope, which is NP-hard and too costly).
* introduces `normalize_dense_linear_constraints` to normalize constraint given in dense format to the unit cube
* removes `normalize_linear_constraint`; `normalize_sparse_linear_constraints` is to be used instead
* simplifies some of the testing code

Note: This change is in preparation for fixing https://github.com/facebook/Ax/issues/2373

### DIFF
--- a/botorch/optim/initializers.py
+++ b/botorch/optim/initializers.py
@@ -180,7 +180,7 @@ def sample_q_batches_from_polytope(
     q: int,
     bounds: Tensor,
     n_burnin: int,
-    thinning: int,
+    n_thinning: int,
     seed: int,
     inequality_constraints: Optional[List[Tuple[Tensor, Tensor, float]]] = None,
     equality_constraints: Optional[List[Tuple[Tensor, Tensor, float]]] = None,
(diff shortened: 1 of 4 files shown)


botorch/optim/initializers.py
```diff
@@ -367,7 +368,7 @@ def gen_batch_initial_conditions(
                     q=q,
                     bounds=bounds,
                     n_burnin=options.get("n_burnin", 10000),
-                    thinning=options.get("thinning", 32),
+                    n_thinning=options.get("n_thinning", 32),
```
saitcakmak, line 371: If someone passes in `thinning`, I am guessing it'll get silently ignored. Would something like this offer better backwards compatibility? Ideally we'd just error out if we got some unsupported option cc @esantorella 
```suggestion
                    n_thinning=options.get("n_thinning", options.get("thinning", 32)),
```

botorch/utils/sampling.py
```diff
@@ -16,14 +16,17 @@
 
 from __future__ import annotations
 
+import warnings
+
 from abc import ABC, abstractmethod
```
saitcakmak, line 21:
```suggestion
import warnings
from abc import ABC, abstractmethod
```

saitcakmak (approved):

PR #2406, opened 2024-07-01
qLogNEI **kwargs removal

hvarfner:
Summary: Removed the ability for qLogNEI to swallow unused kwargs.

Differential Revision: D59238948


saitcakmak:
The docs failure is unrelated and should be fixed by a diff that just landed.

Issue #2412, opened 2024-07-07
[Bug] _get_noiseless_fantasy_model for LogNoisyExpectedImprovement and NoisyExpectedImprovement only works for default GP with default Matern & no input or outcome transforms


71c:
# 🐛 Bug

In `botorch/acquisition/analytic.py`, the `LogNoisyExpectedImprovement` and `NoisyExpectedImprovement` use the function `_get_noiseless_fantasy_model` in order to repeatedly sample from fantasy model. But `_get_noiseless_fantasy_model` only works for default GP (i.e. with default Matern kernel) & also with no input or outcome transforms.
I think that it would make sense if this code were written to work with any kind of `SingleTaskGP`, not just the default one with no input and outcome transforms.

## To reproduce

Setup code:
```python
import torch
from botorch.models.gp_regression import SingleTaskGP
from botorch.models.transforms.outcome import Standardize
from botorch.models.transforms.input import Normalize
from botorch.acquisition.analytic import LogNoisyExpectedImprovement
from gpytorch.kernels import ScaleKernel, RBFKernel
train_X = torch.rand(10, 2, dtype=torch.float64) * 2.3 - 4.9
train_Y = torch.sin(train_X).sum(dim=1, keepdim=True) * 3. + 2.
train_Yvar = torch.full_like(train_Y, 1e-3)
```

Now, say we do,
```python
model = SingleTaskGP(
     train_X, train_Y,
     outcome_transform=Standardize(m=1),
     input_transform=Normalize(d=2),
     train_Yvar=train_Yvar)
LogNEI = LogNoisyExpectedImprovement(model, train_X)
```
then this is the output:
```
Traceback (most recent call last):
  File "/Users/aloja/Library/Mobile Documents/com~apple~CloudDocs/cornell/RESEARCH/BayesOpt/CODE/botorch/alon_test.py", line 139, in <module>
    LogNEI = LogNoisyExpectedImprovement(model, train_X)
  File "/Users/aloja/Library/Mobile Documents/com~apple~CloudDocs/cornell/RESEARCH/BayesOpt/CODE/botorch/botorch/acquisition/analytic.py", line 634, in __init__
    fantasy_model = _get_noiseless_fantasy_model(
  File "/Users/aloja/Library/Mobile Documents/com~apple~CloudDocs/cornell/RESEARCH/BayesOpt/CODE/botorch/botorch/acquisition/analytic.py", line 1088, in _get_noiseless_fantasy_model
[...]

PR #2414, opened 2024-07-08
Make (Log)NoisyExpectedImprovement create a correct fantasy model with non-default SingleTaskGP

71c:
## Motivation

In `botorch/acquisition/analytic.py`, the `LogNoisyExpectedImprovement` and `NoisyExpectedImprovement` use the function `_get_noiseless_fantasy_model` in order to repeatedly sample from fantasy model. But `_get_noiseless_fantasy_model` only works for default GP (i.e. with default Matern kernel) & also with no input or outcome transforms.
I think that it would make sense if this code were written to work with any kind of `SingleTaskGP`, not just the default one with no input and outcome transforms.

Yes

## Test Plan

Since the code is now meant to work even if there are input or outcome transforms or different covar_module or mean_module, I updated the test code to try all these things, as well as try different input bounds to make sure the input transform is working correctly. However, the tests now fail, specifically when either the input data range is not [0,1], or when the kernel is the RBF kernel (not Matern). I believe that the tests failing when RBF is used is simply because certain constants were used in the code that are only valid for particular GP settings. However, I think that when the code fails due to input range not being [0,1], this might be a slight problem with the code -- or might not -- I'm not completely sure.

I also added a line that makes sure that the state_dict() is the same between the original model and fantasy model.

## Related PRs

I put this in an issue #2412 and was told that it is OK if not all the tests pass.

### DIFF
--- a/botorch/acquisition/analytic.py
+++ b/botorch/acquisition/analytic.py
@@ -25,7 +25,6 @@ from botorch.acquisition.acquisition import AcquisitionFunction
 from botorch.acquisition.objective import PosteriorTransform
 from botorch.exceptions import UnsupportedError
 from botorch.models.gp_regression import SingleTaskGP
-from botorch.models.gpytorch import GPyTorchModel
 from botorch.models.model import Model
 from botorch.utils.constants import get_constants_like
 from botorch.utils.probability import MVNXPB
@@ -651,7 +650,6 @@ class LogNoisyExpectedImprovement(AnalyticAcquisitionFunction):
         """
         # add batch dimension for broadcasting to fantasy models
         mean, sigma = self._mean_and_sigma(X.unsqueeze(-3))
-        print("Mean and sigma:", mean, sigma)
         u = _scaled_improvement(mean, sigma, self.best_f, self.maximize)
         log_ei = _log_ei_helper(u) + sigma.log()
         # this is mathematically - though not numerically - equivalent to log(mean(ei))
@@ -1069,55 +1067,12 @@ def _get_noiseless_fantasy_model(
             "Only SingleTaskGP models with known observation noise "
             "are currently supported for fantasy-based NEI & LogNEI."
         )
-    
-    # # initialize a copy of SingleTaskGP on the original training inputs
-    # # this makes SingleTaskGP a non-batch GP, so that the same hyperparameters
-    # # are used across all batches (by default, a GP with batched training data
-    # # uses independent hyperparameters for each batch).
-    # fantasy_model = SingleTaskGP(
-    #     train_X=model.train_inputs[0],
-    #     train_Y=model.train_targets.unsqueeze(-1),
-    #     train_Yvar=model.likelihood.noise_covar.noise.unsqueeze(-1),
-    # )
-    # # update training inputs/targets to be batch mode fantasies
-    # fantasy_model.set_train_data(
-    #     inputs=batch_X_observed, targets=Y_fantasized, strict=False
-    # )
-    # # use noiseless fantasies
-    # fantasy_model.likelihood.noise_covar.noise = torch.full_like(Y_fantasized, 1e-7)
-    # # load hyperparameters from original model
-    # state_dict = deepcopy(model.state_dict())
-    # fantasy_model.load_state_dict(state_dict)
-    # return fantasy_model
 
-
-    # # initialize a copy of SingleTaskGP on the original training inputs
-    # # this makes SingleTaskGP a non-batch GP, so that the same hyperparameters
-    # # are used across all batches (by default, a GP with batched training data
-    # # uses independent hyperparameters for each batch).
-    # fantasy_model = deepcopy(model)
-    # # update training inputs/targets to be batch mode fantasies
-    # fantasy_model.set_train_data(
-    #     inputs=batch_X_observed, targets=Y_fantasized, strict=False
-    # )
-    # # use noiseless fantasies
-    # fantasy_model.likelihood.noise_covar.noise = torch.full_like(Y_fantasized, 1e-7)
-    # # load hyperparameters from original model
-    # state_dict = deepcopy(model.state_dict())
-    # fantasy_model.load_state_dict(state_dict)
-    # return fantasy_model
-
-
-
-
-    outcome_transform = deepcopy(getattr(model, "outcome_transform", None))
     # initialize a copy of SingleTaskGP on the original training inputs
     # this makes SingleTaskGP a non-batch GP, so that the same hyperparameters
     # are used across all batches (by default, a GP with batched training data
     # uses independent hyperparameters for each batch).
-    # print(model.train_inputs[0])
-    # exit()
-    # from botorch.models.transforms.outcome import Standardize
+    outcome_transform = deepcopy(getattr(model, "outcome_transform", None))
     fantasy_model = SingleTaskGP(
         train_X=model.train_inputs[0],
         train_Y=model.train_targets.unsqueeze(-1),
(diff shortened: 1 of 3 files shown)


esantorella (changes requested):
Thanks so much for reporting this issue and for doing so much of the work towards fixing it! I'm not sure the handling of transforms is right -- see comments. And if you don't have time to get this PR past the finish line yourself, just let us know.

71c:
> Thanks so much for reporting this issue and for doing so much of the work towards fixing it! I'm not sure the handling of transforms is right -- see comments. And if you don't have time to get this PR past the finish line yourself, just let us know.

Hi thanks for reviewing my pull request. I don't know what you mean by "see comments" - I don't see any comments.
And regarding whether I can finish it myself, I see that the tests were hard-coded, but I'm not sure how the person who wrote those tests came up with those constants...so I'm not sure how to adapt these test to test this where there could be input and outcome transforms and different kernel.

I think that I won't have the time or ability to finish this any time in the near future.
But one thing is for sure, which is that what I wrote is better than what there was previously...it's just that, of course, ideally we will make sure that the code is completely correct before letting it go through.

botorch/acquisition/analytic.py
```diff
@@ -1068,24 +1067,40 @@ def _get_noiseless_fantasy_model(
...
     # uses independent hyperparameters for each batch).
+    outcome_transform = deepcopy(getattr(model, "outcome_transform", None))
     fantasy_model = SingleTaskGP(
         train_X=model.train_inputs[0],
         train_Y=model.train_targets.unsqueeze(-1),
         train_Yvar=model.likelihood.noise_covar.noise.unsqueeze(-1),
+        covar_module=deepcopy(model.covar_module),
+        mean_module=deepcopy(model.mean_module),
+        outcome_transform=outcome_transform,
+        input_transform=deepcopy(getattr(model, "input_transform", None)),
```
esantorella, line 1083: `model.train_inputs[0]` and `model.train_targets` are already in the transformed space, so I don't think we want to  (re)apply the transforms here.

esantorella:
Ah, I'm sorry, I failed to submit my comments yesterday. No worries if you don't have time to work on this; I just wanted to check on what you intentions are.

I think we'd want to test combining transforms with (Log)NEI by comparing against the behavior without transforms. For example, if we evaluate LogNEI on a `train_X` that has already been normalized to [0, 1] and don't use transforms, we should get the same acquisition values if we add a constant to `train_X` and apply the `Normalize` input transform.

botorch/acquisition/analytic.py
71c: It is true that in the SingleTaskGP constructor, it will transform the already transformed train_targets (but not the train_inputs[0]). But that doesn't matter, because the data is reset afterward.
But okay, sure, I guess it's unnecessary to apply them yet again.

Also by the way, I found the problem - it was a subtle problem with the testing code. I fixed it and now all the tests pass! However I needed to increase one of the error tolerances for comparing the gradients in float32 case when RBFKernel was used instead of MaternKernel - probably because, for some reason, RBFKernel gave numerical ill-conditioning warnings but MaternKernel didn't, I'm not sure why.

71c:
I fixed the problems now and all the tests pass. So I think it is ready to be committed to main.
However, I see that Sebastian just made a significant change to test_analytic.py... I'll update my code for that.

71c:
@esantorella  There's still conflicts. I'm not sure how to merge them correctly though (either through GitHub or locally on command line). Any help would be appreciated!

test/acquisition/test_analytic.py
```diff
@@ -819,96 +833,172 @@ def _get_model(self, dtype=torch.float):
...
-            LogNEI.best_f = nEI.best_f
-
-            X_test = torch.tensor(
-                [[[0.25]], [[0.75]]],
-                device=X_observed.device,
-                dtype=dtype,
+        # Same as the default Matern kernel
+        # botorch.models.utils.gpytorch_modules.get_matern_kernel_with_gamma_prior,
+        # except RBFKernel is used instead of MaternKernel.
+        # For some reason, RBF gives numerical problems but Matern does not.
```
SebastianAment, line 864: Which numerical problems did you run into? The RBF kernel is generally smoother as the Matern kernel, meaning its eigen-spectrum decays faster, which implies that associated covariance matrices are more likely to be numerically low rank than for the Matern. Evaluating the kernel on fewer points, or alternatively decreasing its lengthscale should make this type of numerical issue go away.

test/acquisition/test_analytic.py
```diff
@@ -916,8 +1001,18 @@ def _test_noisy_expected_imrpovement(self, dtype: torch.dtype) -> None:
         # testing gradient through exp of log computation
         exp_log_val.sum().backward()
         # testing that first gradient element coincides. The second is in the
-        # regime where the naive implementation looses accuracy.
-        atol = 2e-5 if dtype == torch.float32 else 1e-12
+        # regime where the naive implementation loses accuracy.
+
+        # Note - for some reason, when the RBFKernel is used instead of
```
SebastianAment, line 1006: See my comment above, RBF generally gives rise to lower rank covariance matrices. Can you try lowering its lengthscale or spacing out the observations more? That should make it more likely that the RBF kernel gives rise to a well conditioned kernel matrix.

SebastianAment (changes requested):

test/acquisition/test_analytic.py
```diff
@@ -902,11 +910,17 @@ def test_noisy_expected_improvement(self):
...
         ):
-            self._test_noisy_expected_imrpovement(
-                dtype, use_octf, use_intf, bounds, covar_module
-            )
+            with catch_warnings():
+                simplefilter("ignore", category=NumericsWarning)
+                self._test_noisy_expected_improvement(
+                    dtype=dtype, 
+                    use_octf=use_octf,
+                    use_inf=use_intf,
```
SebastianAment, line 918: This should be a `use_inft`, couldn't test this before adding the merge commit through the web UI.

test/acquisition/test_analytic.py
71c: You mean increasing the lengthscale right? But in any case, I think that would be too much of a pain because firstly, `NEI_NOISE` is fixed at 10 values. And secondly, the lengthscale really matters for making the tests work; whoever made this test I assume knew what they were doing and I don't want to arbitrarily fudge numbers (namely, the lengthscale) too much because I saw that that can make the tests fail. Why not just keep it how it is? 

How about this, if we really don't want those numerical warnings: test Matern on both float32 and float64, but only test RBF on float64?

SebastianAment: > You mean increasing the lengthscale right?

Increasing the lengthscale will tend to exacerbate the numerical ill conditioning of the RBF kernel matrix, decreasing the lengthscale will help it. You can think of the lengthscale as controlling the potential complexity of the function, and functions with larger lengthscales are smoother, i.e. less complex, than functions with shorter lengthscales, which can exhibit many more variations. This in turn shows up as an increase in the numerical rank for more complex functions, which helps the conditioning. 

In the most extreme case, a lengthscale of zero would imply that all points are independent, i.e. the covariance would be diagonal, and in the case of the unscaled RBF kernel, it'd be the identity, which has perfect conditioning.

> but only test RBF on float64

Sure!

SebastianAment (approved):

SebastianAment:
This looks good to me, thanks! I just imported the PR, which will run some additional integration tests. Two remaining items to get this in:
- @esantorella could you approve the changes if they were addressed?
- @71c we can't merge the PR into the main brach while it's still in draft. Could you put it out of draft mode?

71c:
> * @71c we can't merge the PR into the main branch while it's still in draft. Could you put it out of draft mode?

@SebastianAment This PR used to be in draft mode but I don't see anything on this page about it being in draft mode; it looks to me like it's not in draft mode anymore.

SebastianAment:
Yes same here, maybe it was changed automatically by the import.

botorch/acquisition/analytic.py
```diff
@@ -1080,24 +1079,50 @@ def _get_noiseless_fantasy_model(
...
+
     # initialize a copy of SingleTaskGP on the original training inputs
     # this makes SingleTaskGP a non-batch GP, so that the same hyperparameters
     # are used across all batches (by default, a GP with batched training data
     # uses independent hyperparameters for each batch).
+
+    # Could pass in the outcome_transform and input_transform here,
+    # however that would make them be applied in SingleTaskGP.__init__ which is
+    # unnecessary. So we will instead set them afterwards.
     fantasy_model = SingleTaskGP(
```
esantorella, line 1091: This change would allow for supporting more model types. However, @saitcakmak  had a good question: Why do we need `_get_noiseless_fantasy_model` at all? Can we use the `fantasize` method on the model instead? I'm a bit afraid of the change I'm suggesting, since this instantiation logic won't be right for every model.

```suggestion
    fantasy_model = cls(model)(
```

esantorella:
Thanks! This looks good to me. I checked the acquisition values with and without having transforms on the model [here](https://gist.github.com/esantorella/595ac3ceb060d3f307bcf64b91cc3de3) and see that the transforms are being handled correctly -- you can't see all the (Log)NEI lines here because they're all on top of each other. And they're all reasonably close to qLogNEI, which hasn't changed.

One thing I don't quite understand is why fantasization needs to happen this way instead of using the model's `fantasize` method. It's strange that there's this ad-hoc logic for just NEI and LogNEI. It would be great if the fantasization could be handled by the model, because then we'd accommodate a broader set of model types.

image

botorch/acquisition/analytic.py
```diff
@@ -608,7 +607,7 @@ class LogNoisyExpectedImprovement(AnalyticAcquisitionFunction):
 
     def __init__(
         self,
-        model: GPyTorchModel,
+        model: SingleTaskGP,
```
esantorella, line 610: Although it's a very ugly hack, I'd suggest leaving this as `SingleTaskGP` and raising an `UnsupportedError` when the model is something other than a `SingleTaskGP`, because this will create a lot of downstream typecheck errors.

botorch/acquisition/analytic.py
saitcakmak: Yeah, it'd be a great simplification (& removal of duplicate logic) if we can simply use `model.fantasize(...)` rather than defining a custom `_get_noiseless_fantasy_model`.

71c: I think you mean to write `model.__class__(...)`; `cls(model)` is not a thing in Python.
But anyway, how about just not making that change because the model is currently assumed to be a `SingleTaskGP` anyway, and if we can find a way to use `fantasize` for a wider variety of models later, then just do that.

botorch/acquisition/analytic.py
71c: OK yeah, that's what's currently done in `_get_noiseless_fantasy_model`. And I think that it would be a good idea to make this explicit in the documentation as well. Which is, like you say, hacky since it says that the type is `GPyTorchModel`.

This makes me think of another thing, which is, `SingleTaskGP` can be multiple-outcome, but in `LogNoisyExpectedImprovement` and `NoisyExpectedImprovement` it is assumed to be single-outcome, but there is no explicit check for this.

71c: But then, there's the fact that all the other classes in `analytic.py`, and probably most of the BoTorch code, doesn't explicitly check that the models are single-outcome...

saitcakmak (approved):

PR #2426, opened 2024-07-11
An improved elliptical slice sampling implementation

kayween:
## Motivation

Elliptical slice sampling for truncated normal distributions (e.g., [Gessner et al., 2020](https://arxiv.org/abs/1910.09328)) requires constructing the active intervals corresponding to the intersection of the ellipse and the domain. One method constructing the active intervals is based on likelihood testing, which tests whether each intersection angle is active or not. The current BoTorch implementation follows this idea and takes $\mathcal{O}(m^2 d)$ time per iteration, where $d$ is the dimensionality and $m$ is the number of linear inequality constraints.

However, there exists a faster algorithm computing the active intervals in $\mathcal{O}(m \log m)$ time ([Wu and Gardner, 2024](https://arxiv.org/abs/2407.10449)). This PR implements Algorithm 3 in this paper and dramatically accelerates truncated normal sampling in high dimensions.

In addition, this PR also implements batch MCMC to launch multiple Markov chains, which further speeds up sampling by exploiting GPU parallelism better. Users can pass an additional argument `chains` to specify how many chains to run in parallel:
```
batch_sampler = LinearEllipticalSliceSampler(
    inequality_constraints=(A, b),
    check_feasibility=True,
    chains=100,  # launch 100 Markov chains in parallel
)

samples = batch_samplers.draw(n)  # returns 100 * n samples
```

Yes.

## Test Plan

1. This implementation passes all existing test cases in `test/utils/probability/test_lin_ess.py`
2. Though, two test cases [here](https://github.com/pytorch/botorch/blob/0455dc3945c5e89eefa2dfeb7abab8c4f4079b36/test/utils/probability/test_lin_ess.py#L363-L396) have to be modified. Because the function `self._find_active_intersections` is replaced with `self._find_active_intersection_angles`, which has a different output format.
3. I have added additional test cases for batch MCMC.

### Some Plots
Attached is a plot demonstrating that this new algorithm accelerates truncated normal sampling in high dimensions. This experiments is ran with `torch.float32`. I have experienced a similar speed-up with double precision.

<p>

 
</p>

The following is a plot of samples by running 500 Markov chains in parallel for a univariate truncated normal distribution.

```
torch.manual_seed(0)

lb, ub = -1, 3

# domain is lb <= x <= ub
A = torch.tensor([[-1.], [1.]], device=device)
b = torch.tensor([-lb, ub], device=device)
x = torch.zeros(1, device=device) + 0.5 * (lb + ub)

sampler = LinearEllipticalSliceSampler(
    inequality_constraints=(A, b.unsqueeze(-1)),
    interior_point=x,
    check_feasibility=False,
    burnin=500,
    thinning=0,
    chains=500,
)
samples = sampler.draw(n=500)
```

## Related PRs

NA. But I am happy to create a new PR (or within this PR) updating the documentation to describe the argument `chains`.


botorch/utils/probability/lin_ess.py
```diff
@@ -316,130 +331,84 @@ def _get_cart_coords(self, nu: Tensor, theta: Tensor) -> Tensor:
         """
         return self._z * torch.cos(theta) + nu * torch.sin(theta)
 
-    def _find_rotated_intersections(self, nu: Tensor) -> Tuple[Tensor, Tensor]:
-        r"""Finds rotated intersections.
+    def _trim_intervals(self, left, right):
```
esantorella, line 334: Could you add type annotations here and throughout?

botorch/utils/probability/lin_ess.py
```diff
@@ -316,130 +331,84 @@ def _get_cart_coords(self, nu: Tensor, theta: Tensor) -> Tensor:
...
+        """Trim the intervals by a small positive constant."""
+        gap = torch.clamp(right - left, min=0.0)
+        eps = gap.mul(0.25).clamp(max=1e-6 if gap.dtype == torch.float32 else 1e-12)
+
+        return left + eps, right - eps
 
-        Rotates the intersections by the rotation angle and makes sure that all
-        angles lie in [0, 2*pi].
+    def _find_active_intersection_angles(self, nu) -> Tuple(Tensor, Tensor):
```
esantorella, line 341:
```suggestion
    def _find_active_intersection_angles(self, nu) -> Tuple[Tensor, Tensor]:
```

botorch/utils/probability/lin_ess.py
```diff
@@ -316,130 +331,84 @@ def _get_cart_coords(self, nu: Tensor, theta: Tensor) -> Tensor:
...
+
+        srted, indices = torch.sort(alpha, descending=False)
+        cummax = beta[self.indices_batch.unsqueeze(-1), indices].cummax(dim=-1).values
+
+        srted = torch.cat([srted, self.ones * 2 * math.pi], dim=-1)
+        cummax = torch.cat([self.zeros, cummax], dim=-1)
+
+        return cummax, srted
+
+    def _find_intersection_angles(self, nu: Tensor) -> Tuple(Tensor, Tensor):
```
esantorella, line 368:
```suggestion
    def _find_intersection_angles(self, nu: Tensor) -> Tuple[Tensor, Tensor]:
```

botorch/utils/probability/lin_ess.py
```diff
@@ -316,130 +331,84 @@ def _get_cart_coords(self, nu: Tensor, theta: Tensor) -> Tensor:
...
-            # TODO: What about `self.ellipse_in_domain = False` in the original code?
-        elif delta_active[0] == -1:  # ensuring that the first interval is feasible
+        p = self._Az @ self._z
+        q = self._Az @ nu
 
-            theta_active = torch.cat((theta_active[1:], theta_active[:1]))
+        radius = torch.sqrt(p**2 + q**2)
+        # if radius.abs().lt(1e-6).any():
+        #     warnings.warn("The ellipse has an extremely small volume.
+        # This may cause numerical issues.")
```
esantorella, line 388:
```suggestion
        radius = torch.sqrt(p**2 + q**2)
```

botorch/utils/probability/lin_ess.py
```diff
@@ -316,130 +331,84 @@ def _get_cart_coords(self, nu: Tensor, theta: Tensor) -> Tensor:
...
 
-        return theta_active.view(-1)
+        ratio = self._bz / radius
+        # if ratio.min().le(-1. + 1e-6):
+        #     warnings.warn("The ellipse is almost outside the domain.
+        # This may cause numerical issues.")
```
esantorella, line 393:
```suggestion
        ratio = self._bz / radius
```

test/utils/probability/test_lin_ess.py
```diff
@@ -373,15 +373,12 @@ def test_multivariate(self):
...
-            rot_angle, slices = sampler._find_rotated_intersections(nu)
-            self.assertEqual(rot_angle, 0.0)
-            self.assertAllClose(
-                slices, torch.tensor([[0.0, 2 * torch.pi]], **tkwargs), atol=atol
-            )
+
+            # Get the left and right endpoints of the active intervals.
+            # As the entire ellipse is feasible, the cumulative length should be 2 * pi.
+            left, right = sampler._find_active_intersection_angles(nu)
+            csum = right.sub(left).clamp(min=0.0).cumsum(dim=-1)
```
esantorella, line 380: I'm not sure if this works, but I'd find it clearer because I had to look up `torch.sub` . Also, why clamp?

```suggestion
            csum = (right - left).clamp(min=0.0).cumsum(dim=-1)
```

test/utils/probability/test_lin_ess.py
```diff
@@ -462,3 +462,44 @@ def test_multivariate(self):
...
+
+                # Run 100 Markov chains.
+                batch_sampler = LinearEllipticalSliceSampler(
+                    bounds=bounds,
+                    check_feasibility=True,
+                    burnin=100,
+                    chains=100,
+                )
+                batch_samples = batch_sampler.draw(n=100)
+                self.assertEqual(batch_samples.shape, torch.Size([100 * 100, d]))
```
esantorella, line 497: About how long does this take to run on a CPU? We'll want tests to run within a second or two (not counting any slow imports) on any reasonable infra

test/utils/probability/test_lin_ess.py
```diff
@@ -462,3 +462,44 @@ def test_multivariate(self):
             self.assertEqual(X_high_d.shape, torch.Size([16, d]))
             self.assertTrue(sampler._is_feasible(X_high_d.T).all())
             self.assertEqual(sampler.lifetime_samples, num_samples)
+
+    def test_batch_mcmc(self):
+        d = 5
+        for i in range(2):
+            torch.manual_seed(i)
```
esantorella, line 469: This doesn't need to be checked for both seeds every time. One option would be to confirm it works for a range of seeds (say 0-10) and then randomly choose one of those seeds each time the test is run.

esantorella (commented):
Thanks so much for this, especially for the clear summary, runtime comparisons, and plot against the ground truth! I'm not sure if we'll want to accept this before there's a paper on the faster algorithm, as future developers will need to understand and debug this. We could probably get the usage of multiple chains in soon though. Did you use multiple chains for the performance comparisons, and how much of the speedup does that account for?

Is there any loss in accuracy with this new algorithm? Or change in autocorrelation or the amount of burn-in needed?

Balandat:
Great stuff, thanks so much for the PR, looks awesome. Do you have a sense of when the tech report will be ready? As @esantorella says it would be great if we could reference that in the code.

kayween:
> Great stuff, thanks so much for the PR, looks awesome. Do you have a sense of when the tech report will be ready? As @esantorella says it would be great if we could reference that in the code.

@Balandat @esantorella I am currently in the final stage of editing. The tech report will be on arXiv early next week. But I can post a pdf of this paper here on github in the next two days.

kayween:
> Thanks so much for this, especially for the clear summary, runtime comparisons, and plot against the ground truth! I'm not sure if we'll want to accept this before there's a paper on the faster algorithm, as future developers will need to understand and debug this. We could probably get the usage of multiple chains in soon though. Did you use multiple chains for the performance comparisons, and how much of the speedup does that account for?
> 
> Is there any loss in accuracy with this new algorithm? Or change in autocorrelation or the amount of burn-in needed?

@esantorella Both implementation use a single Markov chain in the above CPU/GPU running time comparison. Therefore, the speed-up solely comes from the improved time complexity $\mathcal{O}(m \log m)$.

There should be no loss of accuracy. This new implementation has the same burnin and autocorrelation, because it computes the exactly same active intervals.

Those technical details will be more clear once I post the tech report soon. Stay tuned for that!

Balandat:
Awesome. My recommendation would be to clean this PR up to make it pretty much ready to go in, and then add the reference to the docstring before merging the PR, assuming there is no rush to get this merged in right away.

test/utils/probability/test_lin_ess.py
kayween: Will remove the for loop.

test/utils/probability/test_lin_ess.py
kayween: On a CPU, this code block takes about 1.5 second.

After removing the for loop as suggested in this [comment](https://github.com/pytorch/botorch/pull/2426#discussion_r1673846160
), the entire test file `test_lin_ess.py` runs in 2 second.

Is that a reasonable running time? We can always reduce the number of MCMC steps or the number of chains if needed.

botorch/utils/probability/lin_ess.py
kayween: Sure, will do.

test/utils/probability/test_lin_ess.py
kayween: Yes, this also works.

The clamping is needed because not all entries in `left` and `right` are active intersection angles. A pair of angles `left[i]` and `right[j]` are active if `left[i] <= right[i]`. In this case, the interval `[left[i], right[i]]` is an active interval. Otherwise, `left[i]` and `right[j]` are dummy angles that do not contribute to the cumulative length. Therefore, we have to clamp them.

test/utils/probability/test_lin_ess.py
esantorella: 2 seconds is okay, but if we can make it faster without making the tests less informative, I'd vote to do that.

kayween: I am reducing the problem size to `burnin=50`, `num_chains=50`, and `n=10`. Running a few hundred MCMC steps should be fast on any reasonable machines.

Now the entire test takes about 1.6 seconds. Note that the rest of tests already takes 1.5 seconds. Will push the change shortly.

kayween:
The tech report is on arXiv now. It's been added to the document string. Let me know if there are any code changes that I need to make.

https://arxiv.org/abs/2407.10449

botorch/utils/probability/lin_ess.py
```diff
@@ -47,7 +49,6 @@ class LinearEllipticalSliceSampler(PolytopeSampler):
     r"""Linear Elliptical Slice Sampler.
 
     Ideas:
-    - Add batch support, broadcasting over parallel chains.
```
esantorella, line 50: Nice :)

botorch/utils/probability/lin_ess.py
```diff
@@ -158,10 +161,17 @@ def __init__(
         self._x = self.x0.clone()
         self._z = self._transform(self._x)
 
+        # Expand the shape to (d, num_chains) for running parallel Markov chains.
```
esantorella, line 164: The helpful comment is appreciated!

test/utils/probability/test_lin_ess.py
```diff
@@ -462,3 +462,45 @@ def test_multivariate(self):
...
+            # Run a single Markov chain.
+            sampler = LinearEllipticalSliceSampler(
+                bounds=bounds,
+                check_feasibility=True,
+                burnin=50,
+            )
+            samples = sampler.draw(n=10)
+            self.assertEqual(samples.shape, torch.Size([10, d]))
+
+            # Run 100 Markov chains.
```
esantorella, line 490:
```suggestion
            # Run 50 Markov chains.
```

test/utils/probability/test_lin_ess.py
```diff
@@ -462,3 +462,45 @@ def test_multivariate(self):
...
             self.assertTrue(sampler._is_feasible(X_high_d.T).all())
             self.assertEqual(sampler.lifetime_samples, num_samples)
+
+    def test_batch_mcmc(self):
+        # all random seeds in [0, 99] should pass
+        torch.manual_seed(torch.randint(100, (1,)))
+
+        d = 5
+
+        for dtype, atol in zip((torch.float, torch.double), (2e-5, 1e-12)):
```
esantorella, line 472: `atol` isn't used here. Is this suggestion right? If so, this looks good to go :)

```suggestion
        for dtype in (torch.float, torch.double):
```

esantorella (changes requested):

test/utils/probability/test_lin_ess.py
esantorella: By the way, I merged in main and made a minor change to a comment, so you'll need to either pull down those changes or make further changes directly from the UI.

kayween: > `atol` isn't used here. Is this suggestion right? If so, this looks good to go :)

Yep, `atol` is not used. It should be removed.

esantorella (approved):

dme65:
Thanks for your contribution, @kayween!

kayween:
> Thanks for your contribution, @kayween!

Happy to do that. Hope you find it helpful in the internal applications!

Issue #2419, opened 2024-07-10
[Bug] Documentation not rendering for classes that have no class docstring (but have an __init__ docstring)


esantorella:
This seems to happen for classes that lack a class docstring. See https://github.com/pytorch/botorch/pull/1313 for a slightly different issue.

## To reproduce

* Go to https://botorch.org/v/latest/api/acquisition.html
* Notice `qLogNoisyExpectedHypervolumeImprovement` is not there

## System information

**Please complete the following information:**
Computer OS: OS X

esantorella:
This appears to have affected classes that don't have a class docstring of their own and inherit from a super class that had :meta private: in its docstring. The subclass "inherits" the "private" attribute and does not show up.  This affected `qLogNoisyExpectedHypervolumeImprovement` because it inherits from `NoisyExpectedHypervolumeMixin` and `AffineInputTransform` because it inherits from `ReversibleInputTransform`. There are probably many other cases.

PR #2437, opened 2024-07-21
Remove :meta private: from documentation so Sphinx doesn't hide subclasses

esantorella:
## Motivation
* Addresses #2419 , classes missing from Sphinx-rendered API documentation
* Makes ABCs that were previously manually hidden (by me) show up in the documentation. I now think it was a mistake to hide these, as users frequently need to use these.

# Changes in this PR
* Remove `:meta private:` from docstrings. `:meta_private` was present in many ABCs; it prevents them from appearing in the Sphinx-rendered API documentation. It also seems to prevent their subclasses from appearing if the subclasses don't "override" the docstring.
* Fixed up a docstring in `batch_cross_validation`
* Removed ABC `HomotopySchedule`, which only had one subclass.

We might want to consider adding a lint to check that `:meta private:` is not added again, but I don't expect it to come up too often.

## Test Plan

Built the website locally and confirmed that the documentation looks okay and that formerly missing classes such as `qLogNoisyExpectedHypervolumeImprovement` and `AffineInputTransform` now show up.

### DIFF
--- a/botorch/acquisition/acquisition.py
+++ b/botorch/acquisition/acquisition.py
@@ -28,8 +28,6 @@ class AcquisitionFunction(Module, ABC):
     Please note that if your acquisition requires a backwards call,
     you will need to wrap the backwards call inside of an enable_grad
     context to be able to optimize the acquisition. See #1164.
-
-    :meta private:
     """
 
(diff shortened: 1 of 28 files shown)


saitcakmak (approved):
Thanks!

PR #2382, opened 2024-06-17
Pin numpy to <2.0

Balandat:
The update caused some issues on windows tests, pinning this until this is root-caused & fixed. Failures: https://github.com/pytorch/botorch/actions/runs/9546224666/job/26308628524, https://github.com/pytorch/botorch/actions/runs/9551785679/job/26326926744?pr=2381

PR #2459, opened 2024-08-04
Unpin numpy from <2.0

Balandat:
This unpins numpy from <2.0, which was done as a temporary workaround in #2382 to avoid issues with windows builds. 

The underlying issue was an incompatibility between pytorch 2.4.0 windows builds and numpy 2.0. Pointed pip to the updated pytorch pip index to pick up newer 2.5.0 dev versions to resolve the issue: https://github.com/pytorch/botorch/pull/2459#issuecomment-2267577174

There is still the issue of existing torch release windows binaries potentially not being compatible with numpy 2.0, but that shouldn't be something that we need to handle.

### DIFF
--- a/.github/workflows/deploy_on_release.yml
+++ b/.github/workflows/deploy_on_release.yml
@@ -62,8 +62,6 @@ jobs:
       run: |
         # avoid conda bug in >=23.10.0: https://github.com/conda/conda/issues/13412
         conda config --set solver classic
-        # avoid issues with numpy 2.0 on windows builds and in tensorboard
-        conda install -y "numpy<2.0"
         conda install -y setuptools_scm conda-build conda-verify anaconda-client
         conda install -y scipy sphinx pytest flake8 multipledispatch
         conda install -y -c pytorch pytorch cpuonly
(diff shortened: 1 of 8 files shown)


Balandat:
I guess this still causes issues, will need to dig deeper.

Balandat:
The issue appears to be that the windows binaries for pytorch 2.4.0 are not compatible with `numpy>=2.0`:  https://github.com/pytorch/pytorch/issues/131668.

https://github.com/pytorch/builder/pull/1945 should fix this going forward for the nightly builds in our CI, but fundamentally this isn't actually a botorch issue but a torch/numpy compatibility issue on windows. 

I also realized that in our CI we were using `pip install torch torchvision -f https://download.pytorch.org/whl/nightly/cpu/torch_nightly.html` whereas the pytorch website suggests `pip install --pre torch torchvision --index-url https://download.pytorch.org/whl/nightly/cpu`. Making this change; hopefully this will pick up newer nightly builds and resolve the windows numpy compatibility issue.

Balandat:
> Making this change; hopefully this will pick up newer nightly builds and resolve the windows numpy compatibility issue.

This appears to be the case. There is still the issue of existing torch release windows binaries potentially not being compatible with numpy 2.0, but that shouldn't be something that we need to handle.

PR #2462, opened 2024-08-11
Speed up Warped GP tutorial by only running one replication and not comparing against other methods

esantorella:
Summary:
Context: This tutorial has been taking too long to run. Also, a tutorial doesn't need to serve both as a demonstration tha the method works better than other methods (in a statistically significant way) and as a demonstration of how to use it.

This PR:
* Only ones one replication, rather than 3. (Putting a CI on 3 data points is a little silly anyway.)
* Removes the comparision methods, Sobol and qNEI with a non-warped GP.
* Uses qLogNEI instead of qNEI
* Use SingleTaskGP instead of deprecated FixedNoiseGP
* No longer manually specifies outcome transform (building on #2458)
* Makes copy edits

Differential Revision: D61054473


Balandat (approved):
Thanks for these updates (including improving the writing), this is great.

Issue #2306, opened 2024-04-23
Document how to implement custom models


HaoLiu56:
## Issue description

Provide a short description.

I wonder if we could implement a custom model instead of GP class model. 
Specifically, suppose we build a random forest model, and it also have mean and variance for data points. Is it possible to do Bayesian Optimization with this model under BoTorch framework?

## Code example

For example:
```model = CustomModel(train_X=init_x, train_Y=init_y )```

## System Info
Please provide information about your setup, including
- BoTorch Version (0.10.0)
- GPyTorch Version (1.11)
- PyTorch Version (2.2.0.post100)
- Computer OS: mac OS Sonoma

**Comment by Balandat** [comment]
Yes, this is generally possible. BoTorch's [`Model`](https://github.com/pytorch/botorch/blob/main/botorch/models/model.py#L57) class only requires you to define a `posterior()` method that returns a [`Posterior`](https://github.com/pytorch/botorch/blob/main/botorch/posteriors/posterior.py#L22) object, the only requirement of which is to implement an `rsample()` function for drawing posterior samples (the [`EnsemblePosterior`](https://github.com/pytorch/botorch/blob/main/botorch/posteriors/ensemble.py#L20) class provides an implementation for that). Then the MC acquisition function machinery will work with such a setup (`EnsemblePosterior` also implements `mean` and `variance` properties, so some other analytic acquisition functions will also work). 

That said, note that in order to use gradient-based optimization of the acquisition function (via the standard `optimize_acqf()` method) you will need to have the samples drawn from the posterior be differentiable w.r.t. to the input to the `posterior()` method (this is not the case for Random Forest models). In that case you'll have to perform the acquisition function optimization with gradient-free methods (e.g. CMA-ES or similar methods).

**Comment by HaoLiu56** [comment]
Great! Could you include some tutorials notebook for this? I'm fairly new to BoTorch.

**Comment by jakee417** [comment]
@esantorella @Balandat here is an attempt at documenting a custom model that I'd like to submit a PR for: https://github.com/jakee417/botorch/blob/main/tutorials/custom_model.ipynb

One question I have is trying to get `MCSampler` to work with `TorchPosterior`. I am getting:
```
NotImplementedError: A registered `MCSampler` for posterior <botorch.posteriors.torch.TorchPosterior object at 0x1334cf410> is not found. You can implement and register one using `@GetSampler.register`.
```

But when I try:
```python
from botorch.sampling import get_sampler
[...]

PR #2474, opened 2024-08-20
Document how to implement custom models

jakee417:
## Motivation

Issue https://github.com/pytorch/botorch/issues/2306

Yes. Added a tutorial which can be used for smoke tests.

## Test Plan

Probabilistic linear regression, bayesian linear regression, and ensemble linear regression all yield optimization results close to (0, 0) which is groundtruth answer.

Random Forest doesn't seem to achieve groundtruth answer, likely due to its inability to incorporate gradient information into the optimization of the acquisition function.

## Related PRs

N/A


tutorials/custom_model.ipynb
```diff
@@ -0,0 +1,1014 @@
...
+   "execution_count": 8,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "from botorch.sampling.base import MCSampler\n",
+    "from botorch.sampling.get_sampler import GetSampler\n",
+    "from botorch.sampling.stochastic_samplers import StochasticSampler\n",
+    "\n",
+    "\n",
+    "@GetSampler.register(distributions.Distribution)\n",
```
saitcakmak, line 276: The `Distribution` base class seems like a generic fallback here. I think it'd be better to register this for the specific distribution instead. The sampler we want will often depend on the particular distribution.
```suggestion
    "@GetSampler.register(distributions.StudentT)\n",
```

saitcakmak (approved):
This is a great tutorial, thank you very much for the contribution! I just have minor point about the sampler registration. Once it is ready, you can import the PR to fbcode. Landing it in fbcode will sync the changes to GitHub and close the PR.

Balandat (commented):
This is an awesome tutorial, thanks a lot for the contribution!

Some additional linear algebraic comments:
```
        # Inverse of the gram matrix.
        self.V = torch.linalg.inv(x.T @ x)
```
It's generally inadvisable to compute the inverse of a matrix. Rather than doing that, you'd typically compute a matrix decomposition of `x.T @ x` and use that for solves down the line using forward-backward substitutions. Could you modify the code to that end? Happy to provide more specifics / code changes if that would be useful.

Also, some of this can get hairy in 32bit precision if the gram matrix is ill-conditioned. In general we recommend folks use botorch with 64bit precision (except when running on larger data sets on a GPU where perf really counts). Could you modify the tutorial to use `torch.float64`? Either by choosing the dtype explicitly or by setting the default torch dtype at the beginning of the tutorial.

tutorials/custom_model.ipynb
```diff
@@ -0,0 +1,1014 @@
...
+    "def _get_sampler_torch(\n",
+    "    posterior: TorchPosterior,\n",
+    "    sample_shape: torch.Size,\n",
+    "    *,\n",
+    "    seed: Optional[int] = None,\n",
+    ") -> MCSampler:\n",
+    "    return StochasticSampler(sample_shape=sample_shape, seed=seed)"
```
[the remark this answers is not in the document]
jakee417: I am curious to know what the use case is for `StochasticSampler` then? And yes, happy to switch in `ForkedRNGSampler`, it also gives the correct result

jakee417:
> This is an awesome tutorial, thanks a lot for the contribution!
> 
> Some additional linear algebraic comments:
> 
> ```
>         # Inverse of the gram matrix.
>         self.V = torch.linalg.inv(x.T @ x)
> ```
> 
> It's generally inadvisable to compute the inverse of a matrix. Rather than doing that, you'd typically compute a matrix decomposition of `x.T @ x` and use that for solves down the line using forward-backward substitutions. Could you modify the code to that end? Happy to provide more specifics / code changes if that would be useful.
> 
> Also, some of this can get hairy in 32bit precision if the gram matrix is ill-conditioned. In general we recommend folks use botorch with 64bit precision (except when running on larger data sets on a GPU where perf really counts). Could you modify the tutorial to use `torch.float64`? Either by choosing the dtype explicitly or by setting the default torch dtype at the beginning of the tutorial.

torch.Double seems to improve the results, I see why this is recommended now. Switched the `torch.linalg.inv` out for `torch.linalg.cholesky` followed by `torch.cholesky_inverse`. I am not sure if `torch.cholesky_inverse` does forwards-backwards under the hood, but it gives the same result as `torch.linalg.inv`.

Balandat:
> torch.Double seems to improve the results, I see why this is recommended now.

Great

> Switched the torch.linalg.inv out for torch.linalg.cholesky followed by torch.cholesky_inverse. I am not sure if torch.cholesky_inverse does forwards-backwards under the hood, but it gives the same result as torch.linalg.inv.

`torch.cholesky_inverse` still computes the inverse explicitly. It's better than `torch.linalg.inv` since you're using the knowledge that the matrix is pd, but instead of computing the inverse explicitly and then doing matmuls you should use `torch.cholesky_solve` (which does forward-backward substitution).  So if you have `A x = b` you want to do `x = torch.cholesky_solve(b, torch.linalg.cholesky(A))` instead of `` = torch.cholesky_inverse(torch.linalg.cholesky(A)) @ x`. You can assign `self.L = torch.linalg.cholesky(A)` instead of `self.V = torch.cholesky_inverse(L)` and use that both to compute the Least squares estimate as well as in the posterior call (swapping out `self.V @ X.squeeze().T` for the respective `cholesky_solve()` call.

jakee417:
okay, now I got it. Replaced the `V @ b` with `torch.cholesky_solve(b, L)` in the appropriate places.

tutorials/custom_model.ipynb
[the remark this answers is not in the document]
saitcakmak: You can use `StochasticSampler` with a stochastic optimizer. In BoTorch, we tend to use deterministic optimizers, which requires making the acquisition function evaluations deterministic (an approach called sample-average approximation). But you can also use a stochastic optimizer, like Adam, to optimize the acquisition function. This could be done by passing `gen_candidates=gen_candidates_torch` to `optimize_acqf`. In that case, it'd be perfectly fine to use `StochasticSampler`.

tutorials/custom_model.ipynb
```diff
@@ -0,0 +1,1037 @@
...
+    }
+   ],
+   "source": [
+    "import torch\n",
+    "\n",
+    "# Set the seed for reproducibility\n",
+    "torch.manual_seed(1)\n",
+    "# Double precision is highly recommended for BoTorch.\n",
+    "# See https://github.com/pytorch/botorch/discussions/1444\n",
+    "torch.set_default_tensor_type(torch.DoubleTensor)\n",
```
esantorella, line 37: To resolve the deprecation warning
```suggestion
    "torch.set_default_dtype(torch.DoubleTensor)\n",
```

tutorials/custom_model.ipynb
```diff
@@ -0,0 +1,1037 @@
...
+    }
+   ],
+   "source": [
+    "from matplotlib import pyplot as plt\n",
+    "from matplotlib.axes import Axes\n",
+    "from torch import Tensor\n",
+    "\n",
+    "\n",
+    "def plot_toy_data(x: Tensor, y: Tensor) -> Axes:\n",
+    "    ax = plt.figure().add_subplot(projection=\"3d\")\n",
```
esantorella, line 75: The 3d projection is cool!

tutorials/custom_model.ipynb
```diff
@@ -0,0 +1,1037 @@
...
+    "        # Linear layer that will compute the regression output.\n",
+    "        self.linear = nn.Linear(train_X.shape[-1], self.num_outputs)\n",
+    "\n",
+    "    @property\n",
+    "    def num_outputs(self) -> int:\n",
+    "        return self._num_outputs\n",
+    "\n",
+    "    def forward(self, x: Tensor) -> distributions.Distribution:\n",
+    "        n, p = x.squeeze().shape\n",
+    "        # For now, lets suppose we have known variance 1.\n",
```
esantorella, line 143:
```suggestion
    "        # For now, let's suppose we have known variance 1.\n",
```

tutorials/custom_model.ipynb
```diff
@@ -0,0 +1,1037 @@
...
+    "        X: Tensor,\n",
+    "        output_indices: Optional[list[int]] = None,\n",
+    "        observation_noise: Union[bool, Tensor] = False,\n",
+    "        posterior_transform: Optional[PosteriorTransform] = None,\n",
+    "    ) -> Posterior:\n",
+    "        if output_indices:\n",
+    "            X = X[..., output_indices]\n",
+    "        # TorchPosterior directly wraps our torch.distributions.Distribution output.\n",
+    "        posterior = TorchPosterior(distribution=self(X))\n",
+    "        if posterior_transform:\n",
```
esantorella, line 157: Nit: Can't hurt to be explicit, although it's not likely to be problematic in this case

```suggestion
    "        if posterior_transform is not None:\n",
```

tutorials/custom_model.ipynb
```diff
@@ -0,0 +1,1037 @@
...
+    "from botorch.acquisition.analytic import LogExpectedImprovement\n",
+    "from botorch.optim.optimize import optimize_acqf\n",
+    "\n",
+    "optimize_acqf(\n",
+    "    LogExpectedImprovement(model=prob_regression_model, best_f=Y.max()),\n",
+    "    bounds=bounds,\n",
+    "    q=1,\n",
+    "    num_restarts=5,\n",
+    "    raw_samples=10,\n",
+    ")"
```
esantorella, line 269:
```suggestion
    "candidate, acq_val = optimize_acqf(\n",
    "    LogExpectedImprovement(model=prob_regression_model, best_f=Y.max()),\n",
    "    bounds=bounds,\n",
    "    q=1,\n",
    "    num_restarts=5,\n",
    "    raw_samples=10,\n",
    ")",
    "candidate, acq_val"
```

esantorella (approved):
Awesome, this is a great tutorial and I may refer to it myself in the future. I left a couple minor suggestions, but this looks pretty good to me as-is.

tutorials/custom_model.ipynb
jakee417: Yes, unfortunately I don't think the 3d projection is compatible with internal building. Will refactor this into a Y vs. Y_hat style plot and collapse the dimensions to 2 to make it compatible.

PR #2481, opened 2024-08-22
[Website] Use mathjax svg renderer

CristianLara:
We identified an issue where mathjax would fail to render certain math symbols. This happens on MacOS versions Ventura and later (13+) because of a change in the OS-supplied STIX fonts. We fix this here by changing the renderer from HTML+CSS (which tries to use local STIX fonts) to SVG.

More details on the issue can be found here: https://meta.stackexchange.com/a/389308

| **Before**    | image |
| -------- | ------- |
| **After**  | image    |

### Considerations

The [official docs](https://docs.mathjax.org/en/stable/output.html#mathjax-output-formats) outline a couple considerations for using the SVG renderer, the biggest being lack of IE8 support. If IE8 support is crucial then we should consider alternate approaches.

An alternative fix could be to upgrade our mathjax version as later versions may not be as reliant on local STIX fonts but the 3.0 update was a complete rewrite and appears to [not be a straightforward upgrade](https://docs.mathjax.org/en/latest/upgrading/v2.html).

### DIFF
--- a/website/siteConfig.js
+++ b/website/siteConfig.js
@@ -74,7 +74,7 @@ const siteConfig = {
     'https://cdnjs.cloudflare.com/ajax/libs/clipboard.js/2.0.0/clipboard.min.js',
     // Mathjax for rendering math content
     `${baseUrl}js/mathjax.js`,
-    'https://cdnjs.cloudflare.com/ajax/libs/mathjax/2.7.5/latest.js?config=TeX-AMS_SVG',
+    'https://cdn.jsdelivr.net/npm/mathjax@2/MathJax.js?config=TeX-AMS_SVG',
   ],
 
   // CSS sources to load
--- a/website/siteConfig.js
+++ b/website/siteConfig.js
@@ -74,7 +74,7 @@ const siteConfig = {
     'https://cdnjs.cloudflare.com/ajax/libs/clipboard.js/2.0.0/clipboard.min.js',
     // Mathjax for rendering math content
     `${baseUrl}js/mathjax.js`,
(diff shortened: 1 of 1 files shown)


website/siteConfig.js
```diff
@@ -74,7 +74,7 @@ const siteConfig = {
     'https://cdnjs.cloudflare.com/ajax/libs/clipboard.js/2.0.0/clipboard.min.js',
     // Mathjax for rendering math content
     `${baseUrl}js/mathjax.js`,
-    'https://cdnjs.cloudflare.com/ajax/libs/mathjax/2.7.5/latest.js?config=TeX-AMS_HTML',
+    'https://cdnjs.cloudflare.com/ajax/libs/mathjax/2.7.5/latest.js?config=TeX-AMS_SVG',
```
Balandat, line 77:
```suggestion
    'https://cdn.jsdelivr.net/npm/mathjax@2/MathJax.js?config=TeX-AMS_SVG',
```

Balandat: > An alternative fix could be to upgrade our mathjax version as later versions may not be as reliant on local STIX fonts but the 3.0 update was a complete rewrite and appears to not be a straightforward upgrade.

Let's not bother spending time on the 3.0 upgrade at this point. But how about doing this which [should give us the latest 2.x MathJax version](https://www.mathjax.org/#gettingstarted)?

Balandat (approved):
Awesome, thanks for figuring this out and the fix!

> the biggest being lack of IE8 support

Oh no, Lord, have mercy! Pretty sure you're trolling me here...

website/siteConfig.js
CristianLara: That makes sense to me! Tested this change to verify everything continues to work as expected

PR #2511, opened 2024-09-06
Require PyTorch>=2.0.1, GPyTorch==1.13 and linear_operator==0.5.3

saitcakmak:
As titled.

Also includes some updates to test against stable workflow to make it pass.

Test plan: https://github.com/pytorch/botorch/actions/runs/10751171192

### DIFF
--- a/.github/workflows/test_stable.yml
+++ b/.github/workflows/test_stable.yml
@@ -38,10 +38,12 @@ jobs:
         conda install -y -c gpytorch gpytorch
         conda install -y -c conda-forge pyro-ppl>=1.8.4
         pip install .[test]
-    - name: Unit tests
-      shell: bash -l {0}
+    - name: Unit tests and coverage -- BoTorch
       run: |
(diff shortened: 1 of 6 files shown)


.github/workflows/test_stable.yml
```diff
@@ -62,8 +62,6 @@ jobs:
         python setup.py egg_info
         req_txt="botorch.egg-info/requires.txt"
         min_torch_version=$(grep '\btorch[>=]=' ${req_txt} | sed 's/[^0-9.]//g')
-        # HACK around the fact that pytorch does not offer a mac binary for 1.13.1 for py3.11 - TODO: Remove when bumping torch to 2.0.1
```
esantorella, line 65: Thanks for catching this! Do you know if there's a reason it's specifying 2.0.1 rather than 2.0? I guess it's fine if this flow passes.

esantorella (approved):

Issue #2526, opened 2024-09-11
[Bug] Memory leak when using a DataLoader for SingleTaskVariationalGP


mikkelbue:
# 🐛 Bug
When training the `SingleTaskVariationalGP`  using a `DataLoader`, there is  memory leak that causes the device to run out of memory as training progresses. This is not happening immediately after starting the optimization, but after couple of training iterations.  I believe this is because [`fval` in `torch_minimize` is not detached before passing it to `stopping_criterion`](https://github.com/pytorch/botorch/blob/4d49bf7c97b971c4d2149577a7c8f252531f4b7f/botorch/optim/core.py#L205).

## To reproduce

```python
import math
import torch
from torch.utils.data import TensorDataset, DataLoader
from botorch.models import SingleTaskVariationalGP
from gpytorch.mlls import VariationalELBO
from botorch import fit_gpytorch_mll

n_train = 100000

train_x = torch.linspace(0, 1, n_train).unsqueeze(-1)
train_y = torch.sin(train_x * (4 * math.pi) + torch.randn(train_x.size()) * 0.2)

n_inducing = 4096
idx_inducing = torch.randint(0, train_x.shape[0], size=(n_inducing,))

if torch.cuda.is_available():
    train_x, train_y = train_x.cuda(), train_y.cuda()

train_dataset = TensorDataset(train_x, train_y)
train_loader = DataLoader(train_dataset, batch_size=2048, shuffle=True)

model = SingleTaskVariationalGP(train_x[:10,:],
                                train_y[:10,:],
                                inducing_points=train_x[idx_inducing,:])
mll = VariationalELBO(model.likelihood, model.model, train_x.shape[0])

if torch.cuda.is_available():
    mll = mll.cuda()

_ = fit_gpytorch_mll(mll, data_loader=train_loader)

```

```
---------------------------------------------------------------------------
OutOfMemoryError                          Traceback (most recent call last)
Cell In[8], line 1
----> 1 _ = fit_gpytorch_mll(mll, data_loader=train_loader)

File ~/.cache/pypoetry/virtualenvs/new-project-mAQ5cvJS-py3.10/lib/python3.10/site-packages/botorch/fit.py:104, in fit_gpytorch_mll(mll, closure, optimizer, closure_kwargs, optimizer_kwargs, **kwargs)
    101 if optimizer is not None:  # defer to per-method defaults
    102     kwargs["optimizer"] = optimizer
--> 104 return FitGPyTorchMLL(
    105     mll,
[...]

PR #2529, opened 2024-09-11
Detach fval in torch_minimize to avoid memory leak

mikkelbue:
## Motivation

This fixes https://github.com/pytorch/botorch/issues/2526

Yes.

## Test Plan
The tests are passing.

## Related PRs
N/A

### DIFF
--- a/botorch/optim/core.py
+++ b/botorch/optim/core.py
@@ -202,7 +202,7 @@ def torch_minimize(
         )
 
         # TODO: Update stopping_criterion API to return a message.
-        if stopping_criterion and stopping_criterion(fval):
+        if stopping_criterion and stopping_criterion(fval.detach()):
             result.status = OptimizationStatus.STOPPED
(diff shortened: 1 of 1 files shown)


botorch/optim/core.py
```diff
@@ -202,7 +202,7 @@ def torch_minimize(
         )
 
         # TODO: Update stopping_criterion API to return a message.
-        if stopping_criterion and stopping_criterion(fval):
+        if stopping_criterion and stopping_criterion(fval.detach()):
```
esantorella, line 205: It looks like `fval` is detached everywhere it's used. Perhaps line 195 should be `fval = closure()[0].detach()`?

Balandat: yeah that would be better

mikkelbue: Good idea.

esantorella (approved):
Thanks!

PR #2537, opened 2024-09-14
OptimizationGradientError

jduerholt:
## Motivation

I am currently experimenting with some custom kernels, which exhibit some numeric instabilities, which lead in some situations during optimization to `NaN`s in the gradient. To better examine these situations, I thought it could be nice if the error which is thrown in this situation also contains the current x which leads to the `NaN`s in the gradient. For this purpose, I implemented an `OptimizationGradientError` like the `OptimizationTimeoutError` which holds the current x.

Im also thinking about catching the error in `timeout.py` as the `OptimizationTimeoutError`, throwing a warning and returning a result of the optimization. What do you think about this? I think this could be a useful and elegant way of dealing with it.

Yes.

## Test Plan

Unit tests.

### DIFF
--- a/botorch/exceptions/errors.py
+++ b/botorch/exceptions/errors.py
@@ -74,7 +74,7 @@ class OptimizationTimeoutError(BotorchError):
         self.runtime = runtime
 
 
-class OptimizationGradientError(BotorchError):
+class OptimizationGradientError(BotorchError, RuntimeError):
     r"""Exception raised when gradient array `gradf` containts NaNs."""
 
     def __init__(self, /, *args: Any, current_x: np.ndarray, **kwargs: Any) -> None:
--- a/botorch/exceptions/errors.py
+++ b/botorch/exceptions/errors.py
@@ -72,3 +72,17 @@ class OptimizationTimeoutError(BotorchError):
         super().__init__(*args, **kwargs)
         self.current_x = current_x
         self.runtime = runtime
+
+
+class OptimizationGradientError(BotorchError):
+    r"""Exception raised when gradient array `gradf` containts NaNs."""
+
+    def __init__(self, /, *args: Any, current_x: np.ndarray, **kwargs: Any) -> None:
+        r"""
+        Args:
+            *args: Standard args to `BoTorchError`.
+            current_x: A numpy array representing the current iterate.
+            **kwargs: Standard kwargs to `BoTorchError`.
+        """
+        super().__init__(*args, **kwargs)
+        self.current_x = current_x
--- a/test/exceptions/test_errors.py
+++ b/test/exceptions/test_errors.py
@@ -12,6 +12,7 @@ from botorch.exceptions.errors import (
     CandidateGenerationError,
     DeprecationError,
     InputDataError,
+    OptimizationGradientError,
     OptimizationTimeoutError,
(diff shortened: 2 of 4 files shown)


test/exceptions/test_errors.py
```diff
@@ -49,3 +50,9 @@ def test_OptimizationTimeoutError(self):
         self.assertTrue(np.array_equal(error.current_x, np.array([1.0])))
         with self.assertRaises(OptimizationTimeoutError):
             raise error
+
+    def test_OptimizationGradientError(self):
+        error = OptimizationGradientError("message", current_x=np.array([1.0]))
+        self.assertTrue(np.array_equal(error.current_x, np.array([1.0])))
+        with self.assertRaises(OptimizationGradientError):
+            raise error
```
Balandat, line 58:
```suggestion
        with self.assertRaisesRegex(OptimizationGradientError, "message"):
            raise error
```

test/generation/test_gen.py
```diff
@@ -318,7 +319,7 @@ def test_gen_candidates_scipy_nan_handling(self):
             test_grad = torch.tensor([0.5, 0.2, float("nan")], **ckwargs)
             # test NaN in grad
             with mock.patch("torch.autograd.grad", return_value=[test_grad]):
-                with self.assertRaisesRegex(RuntimeError, expected_regex):
+                with self.assertRaisesRegex(OptimizationGradientError, expected_regex):
```
Balandat, line 322: One thing to consider here is that this could break people's code if they rely on this being a `RuntimeError` - how about subclassing `OptimizationGradientError` from both `BoTorchError` and `RuntimeError`? (this is somehting we could probably consider doing more generally)

Balandat (commented):
Thanks for putting this up - this makes sense to me. 

> Im also thinking about catching the error in timeout.py as the OptimizationTimeoutError, throwing a warning and returning a result of the optimization. What do you think about this? I think this could be a useful and elegant way of dealing with it.

Yeah I think this generally make sense. One issue with warnings is that they're often overlooked or ignored - so I wouldn't want to change the default behavior that something that would previously throw an error would now just return a result. But maybe it could be gated by some `raise_on_nan` flag?

test/generation/test_gen.py
jduerholt: Good point, I will change it accordingly.

botorch/exceptions/errors.py
```diff
@@ -72,3 +72,17 @@ def __init__(
         super().__init__(*args, **kwargs)
         self.current_x = current_x
         self.runtime = runtime
+
+
+class OptimizationGradientError(BotorchError):
```
Balandat, line 77:
```suggestion
class OptimizationGradientError(BotorchError, RuntimeError):
```

esantorella (approved):
Thanks! I'm all for making it easier to debug optimization errors :)

jduerholt:
I think it is also fine as it is now, advanced users can now catch the error and see which is the `x` that creates the `NaN`s and further inspect them. From my side, no urgent need to further catch in `minimize_with_timeout`, based on an argument to the function. Maybe, I add this in a seperate PR, but from my side, this is good to go now. Thanks for the quick review!

PR #2539, opened 2024-09-16
Use log1p(x) instead of log(1+x)

kit1980:
This function is more accurate than torch.log() for small values of input - https://pytorch.org/docs/stable/generated/torch.log1p.html

Found with https://github.com/pytorch-labs/torchfix/


saitcakmak (approved):
Thanks @kit1980! Happy to stamp the diff if you want to import the PR into fbcode.

PR #2561, opened 2024-10-01
Apply PEP 604 union type syntax codemod

Balandat:
This codemods all `Optional[X]` type definitions to use the PEP 604 syntax `X | None` instead.


esantorella (approved):
Thanks!

Issue #2506, opened 2024-09-03
[Question] How can I see the debugging logs from `fit.py` ?


rashisht1:
**Question:** How can I see the `logging.DEBUG` logs in `fit.py` for example [here](https://github.com/pytorch/botorch/blob/main/botorch/fit.py#L232) ? It seems that file uses `logging` module directly instead of the `logger` object from `botorch.logging` --- is this a bug ?

esantorella:
I believe that the log messages should show up in your terminal if you add `logger = logging.getLogger(); logger.setLevel(logging.DEBUG)` at the beginning of a Python file you're running -- at least it works for me. If you're using a notebook, that may be trickier, as it appears that different IPython versions handle logging differently. If you still can't get it working, can you share a reproducible example of code that you believe should generate these logs, as well as the BoTorch version, Python version, and (if applicable) IPython version you're using?

As an aside, it took me way too long to figure out how to get the logs to show up myself, and I presume many of our users are using IPython notebooks, making this additionally difficult. We might consider getting rid of the logging.

PR #2574, opened 2024-10-15
Use botorch.logging.logger instead of logging to log in fit.py

Balandat:
This was raised in #2506 - replacing this for consistency with the rest of the codebase.


saitcakmak (approved):
Thanks!

Issue #2579, opened 2024-10-16
[Feature Request] optimize_acqf_homotopy doesn't handle constraints


CompRhys:
# 🚀 Feature Request

Would you accept a PR to add constraints to `optimize_acqf_homotopy`, e.g. `inequality_constraints`, `equality_constraints` and `nonlinear_inequality_constraints`. Looking at the code this seems more like an omission than a technical challenge to add?

## Motivation

I want to compare gradient free MOO vs SEBO for a mixtures problem requiring sparsity. I believe other people are working on similar problems and would benefit from this functionality (https://github.com/facebook/Ax/issues/2790)

## Pitch

**Describe the solution you'd like**
The change appears as though it should be as simple as making all the arguements of `optimize_acqf` arguements of `optimize_acqf_homotopy`. This can either be done explicitly or via kwargs (whatever the maintainers prefer).

**Are you willing to open a pull request?** (See [CONTRIBUTING](../CONTRIBUTING.md))

I am happy to make the PR.

saitcakmak:
Hi @CompRhys. Thanks for flagging this. A PR adding constraint support to `optimize_acqf_homotopy` would be very welcome!

saitcakmak:
I'd like to avoid adding catch-all dict valued arguments to the code. `options` kwarg goes against this but it is a common arg across `optimize_acqf` variant that is used to modify low-level settings that are not worth exposing as top level kwargs. 

For constraints, would we expect them to be different between the homotopy steps and final optimization? They only constrain the search space, so I would expect them to be the same for both. If that's the case, we can just mirror the signature of `optimize_acqf` to add the three args and pass them down to both steps.
```
    inequality_constraints: list[tuple[Tensor, Tensor, float]] | None = None,
    equality_constraints: list[tuple[Tensor, Tensor, float]] | None = None,
    nonlinear_inequality_constraints: list[tuple[Callable, bool]] | None = None,
```

CompRhys:
I think one problem with not doing it via catch alls it that you end up with what's currently in the code which is very confusing about whether arguments to the loop and final optimizations are intentionally missing or accidentally missing. For example currently in the code if you passed `fixed_features` it would only be used in the loop `optimize_acqf` and not the final `optimize_acqf`. I think that the catch all here avoids the chances that this code needs repeated (and not clearly telegraphed) maintenance over time.

[...]

PR #2580, opened 2024-10-16
Expose all kwargs to optimize_acqf in optimize_acqf_homotopy

CompRhys:
## Motivation

This PR seeks to address https://github.com/pytorch/botorch/issues/2579

Yes

## Test Plan

Added test incorporating a linear inequality constraint

## Related PRs

If this is merged as is with breaking changes then a follow-on PR will be needed in Ax

### DIFF
--- a/botorch/optim/optimize_homotopy.py
+++ b/botorch/optim/optimize_homotopy.py
@@ -3,7 +3,8 @@
 # This source code is licensed under the MIT license found in the
 # LICENSE file in the root directory of this source tree.
 
-from collections.abc import Callable
+from typing import Any
+import warnings
 
 import torch
 from botorch.acquisition import AcquisitionFunction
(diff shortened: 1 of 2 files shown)


saitcakmak (changes requested):
As discussed on the issue, this is quite a big divergence from the signature of the other `optimize_acqf` implementations. Rather than diverging, we want to unify the signatures as much as possible, so that the code that uses one optimizer could easily switch over to another.

In addition, dict valued inputs are generally not desirable. They do not contain any meaningful type hints, which provides very limited information to the user to what the inputs should be and requires them to dig through the codebase to figure out what is acceptable. We want to reserve these only for rarely modified options.

CompRhys:
I really am loath to contribute a PR that creates increased maintenance burden going forward for sake of documentation/type hinting, this method has these missing constraints options and has the apparent `fixed_features` bug because it leaves the surface area for such issues to exist by hard coding the args to the function it wraps. There will 100% be divergences again when more functionality is added to the base method.

However, pragmatically as I would like this functionality, if you get back to me with decisions about the need for `final_options` (I think it's most consistent with what you propose in terms of arg unification to remove this as not used here or in Ax) and whether the apparent issues with `fixed_features` are indeed bugs or whether it was intended (I can't see how it would be intended as the test would be flaky if the test optimization was harder), I can implement it the other way.

CompRhys:
close in favour of https://github.com/pytorch/botorch/pull/2588

PR #2584, opened 2024-10-18
Upgrade fbcode/pytorch to Python Scientific Stack 2

igorsugak:
Differential Revision: D64008689


saitcakmak:
merged as https://github.com/pytorch/botorch/pull/2586

PR #2589, opened 2024-10-21
fix wrong variable name in docs

fabiankannenberg:
Fixed a wrong variable name in the getting started doc.


Balandat (approved):
Thanks for this fix!

Issue #2596, opened 2024-10-28
[Bug] Bounds in `test_functions/base.py` are hard-coded to double precision


AVHopp:
# 🐛 Bug

Within `test_functions/base.py`, the bounds are hard-coded to double precision. This makes it impossible to use single-precision which is necessary for e.g. MPS support on Mac.

## To reproduce

```python
import torch
from botorch.test_functions import Rastrigin

torch.set_default_device("mps")
torch.set_default_dtype(torch.float32)

test = Rastrigin()
```

This yields the following error message (paths shortened for better readability):

```
Traceback (most recent call last):
  File "[...]/test.py", line 7, in <module>
    test = Rastrigin()
  File "[...]/lib/python3.10/site-packages/botorch/test_functions/synthetic.py", line 664, in __init__
    super().__init__(noise_std=noise_std, negate=negate, bounds=bounds)
  File "[...]/lib/python3.10/site-packages/botorch/test_functions/synthetic.py", line 83, in __init__
    super().__init__(noise_std=noise_std, negate=negate)
  File "[...]/lib/python3.10/site-packages/botorch/test_functions/base.py", line 51, in __init__
    "bounds", torch.tensor(self._bounds, dtype=torch.double).transpose(-1, -2)
  File "[...]/lib/python3.10/site-packages/torch/utils/_device.py", line 79, in __torch_function__
    return func(*args, **kwargs)
TypeError: Cannot convert a MPS Tensor to float64 dtype as the MPS framework doesn't support float64. Please use float32 instead.
```

## Expected Behavior

The initialization of the `BaseTestProblem` class should not enforce double precision in the buffer for the bounds but probably use `torch.get_default_dtype()`

## System information

**Please complete the following information:**
- BoTorch Version 0.12.0
- PyTorch Version 2.4.1
- MacOS

## Additional context

These seem to be the problematic lines:
https://github.com/pytorch/botorch/blob/9d37e905639e0d4983e52ce425306b8161760ee4/botorch/test_functions/base.py#L49-L51

If the error is in fact just this one line of code, I'd be more than happy to create a mini Pull Request if it makes sense :)

AVHopp:
NOTE: I have verified locally that changing the line of code that I mentioned under "Additional context" solves the problem.

Balandat:
> The initialization of the BaseTestProblem class should not enforce double precision in the buffer for the bounds but probably use torch.get_default_dtype()

[...]

PR #2597, opened 2024-10-28
Fix hard-coded `double` precision in test_functions to default dtype

AVHopp:
## Motivation

This PR replaces the hard-coded double precision that was used in the initialization of `test_functions/base.py` to use `torch.get_default_dtype()` instead.

See #2596 for more details.

Yes, I have read it.

## Test Plan

I ran code formatting via `ufmt` and checked the code via `pytest -ra`. All tests related to the changes here were adjusted in the second commit of the branch.
Locally, two tests failed for me, but it seems to me that these are not related to the fix implemented here. If it turns out they are, I'd be more than happy to further adjust.

## Related PRs

None, but #2596 is related.

### DIFF
--- a/botorch/test_functions/base.py
+++ b/botorch/test_functions/base.py
@@ -170,6 +170,7 @@ class MultiObjectiveTestProblem(BaseTestProblem, ABC):
         self,
         noise_std: None | float | list[float] = None,
         negate: bool = False,
+        dtype: torch.dtype = torch.double,
     ) -> None:
         r"""Base constructor for multi-objective test functions.
 
@@ -184,8 +185,8 @@ class MultiObjectiveTestProblem(BaseTestProblem, ABC):
                 f"If specified as a list, length of noise_std ({len(noise_std)}) "
                 f"must match the number of objectives ({len(self._ref_point)})"
             )
-        super().__init__(noise_std=noise_std, negate=negate)
-        ref_point = torch.tensor(self._ref_point, dtype=torch.get_default_dtype())
+        super().__init__(noise_std=noise_std, negate=negate, dtype=dtype)
+        ref_point = torch.tensor(self._ref_point, dtype=self.dtype)
         if negate:
             ref_point *= -1
         self.register_buffer("ref_point", ref_point)
--- a/botorch/test_functions/base.py
+++ b/botorch/test_functions/base.py
@@ -47,7 +47,10 @@ class BaseTestProblem(Module, ABC):
                 f"Got {self.dim=} and {len(self._bounds)=}."
             )
         self.register_buffer(
-            "bounds", torch.tensor(self._bounds, dtype=torch.double).transpose(-1, -2)
+            "bounds",
+            torch.tensor(self._bounds, dtype=torch.get_default_dtype()).transpose(
+                -1, -2
+            ),
         )
 
     def forward(self, X: Tensor, noise: bool = True) -> Tensor:
--- a/botorch/test_functions/base.py
+++ b/botorch/test_functions/base.py
@@ -43,7 +43,6 @@ class BaseTestProblem(Module, ABC):
         super().__init__()
         self.noise_std = noise_std
         self.negate = negate
-        self.dtype = dtype
         if len(self._bounds) != self.dim:
             raise InputDataError(
                 "Expected the bounds to match the dimensionality of the domain. "
@@ -51,7 +50,7 @@ class BaseTestProblem(Module, ABC):
             )
         self.register_buffer(
             "bounds",
-            torch.tensor(self._bounds, dtype=self.dtype).transpose(-1, -2),
+            torch.tensor(self._bounds, dtype=dtype).transpose(-1, -2),
         )
 
     def forward(self, X: Tensor, noise: bool = True) -> Tensor:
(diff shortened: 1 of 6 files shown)


AVHopp:
I am aware that I still need to sign the CLA and will do so soon.

AVHopp:
I have signed the CLA, but it seems like the test was not retriggered, does this just take some more time or did something go wrong with me signing it?

Balandat:
> I have signed the CLA, but it seems like the test was not retriggered, does this just take some more time or did something go wrong with me signing it?

No need to re-test. I can now import the PR and land it even if GitHub says that check failed.

AVHopp:
@Balandat @esantorella Pinging both of you since we already started discussions on #2596 .

**TL;DR:** Changed to use `dtype` in `__init__`. Looking forward to feedback, how to deal with the branch history?

As proposed there, I have re-implemented this now such that there is an explicit `dtype` argument in all of the test functions. I just added this as an argument to every explicit `__init__` function, always setting it to `torch.double`. Not sure if there is a more elegant way, if so feel free to point me to it :)

For better reviewing, each commit just touches a single file - however, I'd be more than happy to rebase/squash the commits at a later point, but since I could not find any guidelines regarding whether or not this is done here, I thought I'd better just have a clean but "true" history for now 😄 

I tested the code again locally using `pytest` and also in my own code where I observed this problem and everything seems to work.

Looking forward to your feedback/comments :)

botorch/test_functions/base.py
```diff
@@ -180,8 +185,8 @@ def __init__(
                 f"If specified as a list, length of noise_std ({len(noise_std)}) "
                 f"must match the number of objectives ({len(self._ref_point)})"
             )
-        super().__init__(noise_std=noise_std, negate=negate)
-        ref_point = torch.tensor(self._ref_point, dtype=torch.get_default_dtype())
+        super().__init__(noise_std=noise_std, negate=negate, dtype=dtype)
+        ref_point = torch.tensor(self._ref_point, dtype=self.dtype)
```
AVHopp, line 189: Note: I spotted another use of `torch.get_default_dtype` while implementing my changes - I guess that this should be changed, only question is to what. For now, I used `self.dtype`, but I guess it should rather be `X.dtype`?

botorch/test_functions/base.py
```diff
@@ -37,17 +38,20 @@ def __init__(
                 provided, specifies separate noise standard deviations for each
                 objective in a multiobjective problem.
             negate: If True, negate the function.
+            dtype: The dtype that is used for the bounds of the function.
         """
         super().__init__()
         self.noise_std = noise_std
         self.negate = negate
+        self.dtype = dtype
```
Balandat, line 46: why assign this? Seems like the only place we'd need this is to register the `bounds` buffer?

botorch/test_functions/base.py
Balandat: At this point during init we don't know what `X` will be - so I think the only reasonable thing to do here would be to be consistent with how we handle the `bounds`.

This raises the question what to do when getting an `X` with a different `dtype` during `forward` or `evaluate_true`. But I think we should be able to just rely on type promotion there. 


```suggestion
        ref_point = torch.tensor(self._ref_point, dtype=dtype)
```

botorch/test_functions/multi_fidelity.py
```diff
@@ -74,13 +74,18 @@ class AugmentedHartmann(SyntheticTestFunction):
...
+    def __init__(
+        self,
+        noise_std: float | None = None,
+        negate: bool = False,
+        dtype: torch.dtype = torch.double,
+    ) -> None:
         r"""
         Args:
             noise_std: Standard deviation of the observation noise.
             negate: If True, negate the function.
```
Balandat, line 86: missing dtype arg in docstring

botorch/test_functions/multi_fidelity.py
```diff
@@ -126,7 +131,11 @@ class AugmentedRosenbrock(SyntheticTestFunction):
     _optimal_value = 0.0
 
     def __init__(
-        self, dim=3, noise_std: float | None = None, negate: bool = False
+        self,
+        dim=3,
+        noise_std: float | None = None,
+        negate: bool = False,
+        dtype: torch.dtype = torch.double,
```
Balandat, line 138: missing dtype arg in docstring below

botorch/test_functions/synthetic.py
```diff
@@ -261,6 +264,7 @@ def __init__(
         noise_std: float | None = None,
         negate: bool = False,
         bounds: list[tuple[float, float]] | None = None,
+        dtype: torch.dtype = torch.double,
```
Balandat, line 267: missing dtype arg in docstring below

Balandat:
> For better reviewing, each commit just touches a single file - however, I'd be more than happy to rebase/squash the commits at a later point, but since I could not find any guidelines regarding whether or not this is done here, I thought I'd better just have a clean but "true" history for now 😄

I think it's fine to have these as one commit. But it doesn't really matter for the branch history here; the way we merge things into the main branch in botorch is that our github bot will auto-squash every change from this PR into a single commit.

botorch/test_functions/base.py
AVHopp: You are right, there is no reason for this. Will remove, commit and resolve this comment then.

AVHopp: Removed the assignment, hence resolving this comment.

botorch/test_functions/multi_fidelity.py
AVHopp: Added the argument to the docstring, hence resolving this conversation.

botorch/test_functions/multi_fidelity.py
AVHopp: Adding argument to docstring, hence resolving this conversation.

botorch/test_functions/synthetic.py
AVHopp: Added the docstring, hence resolving this conversation.

Balandat (approved):

Balandat:
Thanks!

Issue #2602, opened 2024-10-31
[Bug] `fixed_features` does not support negative indices


slishak-PX:
# 🐛 Bug

If `fixed_features` has a negative index, the initial conditions will not be constructed with the correct reduced dimensionality.

## To reproduce

** Code snippet to reproduce **
```python
import torch
from botorch.acquisition import qLogExpectedImprovement
from botorch.fit import fit_gpytorch_mll
from botorch.models import SingleTaskGP
from botorch.models.transforms import Normalize
from botorch.optim import optimize_acqf
from gpytorch.mlls import ExactMarginalLogLikelihood

n_inputs = 4
n_outputs = 1
n_train = 256
n_test = 16
device = torch.device("cpu")

train_x = torch.rand(n_train, n_inputs, dtype=torch.float64, device=device)
train_y = torch.randn(n_train, n_outputs, dtype=torch.float64, device=device)

model = SingleTaskGP(train_x, train_y, input_transform=Normalize(n_inputs))

mll = ExactMarginalLogLikelihood(model.likelihood, model)
fit_gpytorch_mll(mll)

acqf = qLogExpectedImprovement(model, best_f=train_y.max())

bounds = torch.vstack([torch.zeros(1, n_inputs), torch.ones(1, n_inputs)])

candidates, value = optimize_acqf(
    acqf,
    bounds,
    q=2,
    num_restarts=4,
    raw_samples=8,
    fixed_features={-1: 0.5},
)
```

** Stack trace/error message **
```
Traceback (most recent call last):
  File "repro.py", line 27, in <module>
    candidates, value = optimize_acqf(
  File ".../lib/python3.10/site-packages/botorch/optim/optimize.py", line 547, in optimize_acqf
    return _optimize_acqf(opt_acqf_inputs)
  File ".../lib/python3.10/site-packages/botorch/optim/optimize.py", line 568, in _optimize_acqf
    return _optimize_acqf_batch(opt_inputs=opt_inputs)
  File ".../lib/python3.10/site-packages/botorch/optim/optimize.py", line 332, in _optimize_acqf_batch
    batch_candidates, batch_acq_values, ws = _optimize_batch_candidates()
  File ".../lib/python3.10/site-packages/botorch/optim/optimize.py", line 316, in _optimize_batch_candidates
    ) = opt_inputs.gen_candidates(
  File ".../lib/python3.10/site-packages/botorch/generation/gen.py", line 159, in gen_candidates_scipy
[...]

PR #2603, opened 2024-10-31
Make optimizers raise an error when provided negative fixed features

esantorella:
Summary:
Context: See https://github.com/pytorch/botorch/issues/2602

This PR:
* Adds a check for negative fixed_features keys to input validation for optimizers. This applies to all of the optimizers that take fixed_features.
* Updates docstrings

Differential Revision: D65272024

### DIFF
--- a/botorch/optim/optimize.py
+++ b/botorch/optim/optimize.py
@@ -125,6 +125,10 @@ class OptimizeAcqfInputs:
                     "Must specify `raw_samples` when "
                     "`batch_initial_conditions` is None`."
                 )
+        if self.fixed_features is not None and any(
+            (k < 0 for k in self.fixed_features)
+        ):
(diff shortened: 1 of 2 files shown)


Balandat (approved):
I think it's good to get this in for now, but I feel like it shouldn't be to hard to allow this by canonicalizing the indices to `index % num_features`? Being able to use backward indexing can be quite convenient. Maybe we can make a backlog task for this?

esantorella:
> I think it's good to get this in for now, but I feel like it shouldn't be to hard to allow this by canonicalizing the indices to `index % num_features`? Being able to use backward indexing can be quite convenient. Maybe we can make a backlog task for this?

That makes sense to me. I didn't want to do it right now since I'd want to check a lot more carefully that that wouldn't break anything (e.g. if there were a place in the code that added a feature as the last column).

PR #2608, opened 2024-11-01
Robust Gaussian Processes via Relevance Pursuit

SebastianAment:
Summary: This commit adds the implementation of the Robust Gaussian Processes via Relevance Pursuit models and algorithms of the NeurIPS 2024 article.

Differential Revision: D65343571


jduerholt:
I just read the paper, and really like it! We will definitely integrate it into our workflows, as this problem is very common for our experimental data. Currently we are sometimes using iterative trimming. 

Any plan when this will land in main?

SebastianAment:
Great to hear @jduerholt! I'll just need to get test coverage to 100%, will try to get the time for this within the next two weeks. Would be curious to learn about your experience if / when you start using the model.

jduerholt:
I will update you, as soon as it will land here in main, I will integrate it into our workflows ;)

Issue #2609, opened 2024-11-01
[Feature Request] allow both get_ic_generator and batch_initial_conditions to be used in the same call.


CompRhys:
# 🚀 Feature Request

When targeting a campaign I might want to include some points explicitly and then fill up the rest of the initial candidates using the strategies in botorch. This behaviour would be triggered by having `n_repeats > len(batch_initial_conditions)` 

## Motivation

Makes it much easier to do https://github.com/facebook/Ax/pull/2938 as then all I have to worry about to get the desired behaviour for that to get merged is dropping the points from the given X_pareto front that do not satisfy the constraints

## Pitch

this should be fairly minimal change in the logic to check that `raw_samples is not None` and then just using `raw_samples-len(batch_initial_conditions)` as the number of requested samples to the ic generator.

**Are you willing to open a pull request?** (See [CONTRIBUTING](../CONTRIBUTING.md))

yes

## Additional context

saitcakmak:
Hi @CompRhys! That sounds reasonable to me. Happy to review a PR for this.

PR #2610, opened 2024-11-04
Add ability to mix batch initial conditions and internal IC generation

CompRhys:
## Motivation

See https://github.com/pytorch/botorch/issues/2609

Yes

## Test Plan

Basic testing of the code is easy the challenge is working out what the run on implications might be, will this break people's code?

## Related PRs

https://github.com/facebook/Ax/pull/2938


botorch/optim/optimize.py
```diff
@@ -114,6 +114,19 @@ def __post_init__(self) -> None:
...
+                and len(batch_initial_conditions_shape) == 3
+                and self.num_restarts is not None
+                and self.num_restarts != batch_initial_conditions_shape[0]
+            ):
+                raise ValueError(
+                    "If using `batch_initial_conditions` together with `raw_samples`, "
+                    "the first repeat dimension of `batch_initial_conditions` must "
+                    "match `num_restarts`."
+                )
+
```
CompRhys, line 129: @saitcakmak most of the tests that break appear to do so based on this setting in which we see that if we give a 3 dimension initial condition tensor if the first dimension isn't equal to num_restarts we can't concat. Not sure the cleanest way to handle this edge case, in intentional/correct use of the code it shouldn't occur but  it may create issues depending on how people have set these arguements.

botorch/acquisition/input_constructors.py
```diff
@@ -1780,7 +1780,7 @@ def optimize_objective(
         bounds=free_feature_bounds,
         q=q,
         num_restarts=optimizer_options.get("num_restarts", 60),
-        raw_samples=optimizer_options.get("raw_samples", 1024),
+        raw_samples=optimizer_options.get("raw_samples", 1024),  # NOTE potential behaviour change
```
Balandat, line 1783: Not sure I understand the NOTE here. Is this in the right place? Did you mean to change the code to return a `None` default here?

botorch/optim/optimize.py
```diff
@@ -257,22 +270,49 @@ def _optimize_acqf_batch(opt_inputs: OptimizeAcqfInputs) -> tuple[Tensor, Tensor
 
     initial_conditions_provided = opt_inputs.batch_initial_conditions is not None
 
+    required_raw_samples = opt_inputs.raw_samples
     if initial_conditions_provided:
-        batch_initial_conditions = opt_inputs.batch_initial_conditions
+        provided_initial_conditions = opt_inputs.batch_initial_conditions
+        if opt_inputs.raw_samples is not None:
+            required_raw_samples -= provided_initial_conditions.shape[-2]
```
Balandat, line 277: seems that this would error out if `opt_inputs.raw_samples` were `None` (which seems possible)?

test/optim/test_optimize.py
```diff
@@ -1919,3 +1954,7 @@ def my_gen():
         )
         ic_generator = opt_inputs.get_ic_generator()
         self.assertIs(ic_generator, my_gen)
+
+if __name__ == "__main__":
+    import pytest
+    pytest.main([__file__])
```
Balandat, line 1960:
```suggestion

```

botorch/optim/optimize.py
```diff
@@ -257,22 +270,49 @@ def _optimize_acqf_batch(opt_inputs: OptimizeAcqfInputs) -> tuple[Tensor, Tensor
...
+            )
+        batch_initial_conditions = torch.cat(
+            [provided_initial_conditions, generated_initial_conditions], dim=-2
+        )  # should this be shuffled?
+    elif provided_initial_conditions is not None:
+        batch_initial_conditions = provided_initial_conditions
+    elif generated_initial_conditions is not None:
+        batch_initial_conditions = generated_initial_conditions
+    else:
+        raise ValueError("Either `batch_initial_conditions` or `raw_samples` must be set.")
```
Balandat, line 315: This has some  nontrivial complexity and makes it more challenging to read - could you factor this out into some helper function?

botorch/optim/optimize.py
```diff
@@ -1239,11 +1306,54 @@ def optimize_acqf_discrete_local_search(
...
     for i in range(q):
         # generate some starting points
-        if i == 0 and batch_initial_conditions is not None:
-            X0 = _filter_invalid(X=batch_initial_conditions.squeeze(1), X_avoid=X_avoid)
-            X0 = _filter_infeasible(
-                X=X0, inequality_constraints=inequality_constraints
-            ).unsqueeze(1)
+        if i == 0:
+
+            if batch_initial_conditions is not None:
```
Balandat, line 1311: same here - a lot of stuff happening here, can you factor this out into a helper function so this is easier to read?

botorch/optim/optimize.py
Balandat: So you mean in existing use cases in which previously folks were providing both `batch_initial_conditions` and `raw_samples` but only the `batch_initial_conditions` were used? I think it's probably fine to error out like this. An alternative would be to fall back to the previous behavior of ignoring `raw_samples` in this case but raising some `DeprecationWarning` that this behavior will change going forward. Then you can both achieve backward compatibility and allow for your use case (for some slight increase in complexity, at least temporarily).

botorch/acquisition/input_constructors.py
CompRhys: The note here is just a reminder to ask about this edge case in the review :)

Elsewhere in the code the default is indeed None whereas here it defaults to a value. If defaulted to None then there would be no behaviour change for someone giving initial batch conditions but since it defaults to a value if the user gave less than 1024 initial conditions then they would see a behaviour change.

test/optim/test_optimize.py
CompRhys: yep was planning to clean this out before changing the status from draft. It makes it more convenient to run with debugger whilst trying to work out the test failures induced by the change. Will rebase the commits at the end into logical parts.

botorch/optim/optimize.py
CompRhys: previous the `n_restarts` kwarg was just unused if giving initial batch conditions. As far as I understand the 3D are `n_restarts, n_samples, n_dims`. If we are giving initial conditions we can within the code choose to give those conditions either as `n_samples, 1, n_dims`, `1, n_samples, n_dims`, I even think it wouldn't matter if they were given as `2, n_samples/2, n_dims`, or something like that. 

If n_samples-n_ic_samples is bigger than 1 we would expect to see different solutions from the optimization and so it's useful to have repeats. However, the points from the IC generator will always be returned as `n_restarts, n_samples-n_ic_samples, n_dims` so if the batch initial conditions are not size 1 (so they can be easily repeated) or already equal to `n_restarts` in the 0th dimension then it makes it fairly messy, in principal you could do some sort of repeat and interleaving to get the combinations right but then the number of repeats considered wouldn't be `n_repeats`.

CompRhys: In terms of the solution I am happy to either add the deprecation warning or leave as is, whatever makes more sense to you.

botorch/optim/optimize.py
CompRhys: I don't think so as the logic is based on `required_raw_samples` below which is assigned in 269. I think that the check in 272 should be `if required_raw_samples is not None:` and then it's easier to follow

botorch/optim/optimize.py
CompRhys: Actually I think that I have the wrong logic here as the size of the initial_batch_constraints is compared to `n_repeats` not `raw_samples`.

CompRhys:
In the homotopy the previously unnoticed test errors are due to `prune_candidates` getting rid of the `num_restarts` channel and this was interacting badly with the logic changes. I am not sure why you wouldn't want the ability to have multiple restarts go through the homotopy optimization.

test/optim/test_optimize.py
```diff
@@ -562,7 +605,10 @@ def test_optimize_acqf_runs_given_batch_initial_conditions(self):
         ]
         q = 1
 
-        ic_shapes = [(1, 2, dim), (2, 1, dim), (1, dim)]
+        ic_shapes = [
+            (1, 2, dim),
+            (1, dim),
+        ]  # NOTE removed test on (2, 1, dim) as error tested elsewhere
```
Balandat, line 611:
```suggestion
        ]
```

botorch/optim/optimize.py
```diff
@@ -252,27 +271,64 @@ def _optimize_acqf_sequential_q(
     return candidates, torch.stack(acq_value_list)
 
 
+def _combine_initial_conditions(
+    provided_initial_conditions, generated_initial_conditions, num_restarts
+):
```
Balandat, line 276: add type annotations here?

botorch/optim/optimize.py
```diff
@@ -252,27 +271,64 @@ def _optimize_acqf_sequential_q(
...
+            provided_initial_conditions.shape[0] == 1
+            and num_restarts is not None
+            and num_restarts > 1
+        ):
+            provided_initial_conditions = provided_initial_conditions.repeat(
+                num_restarts, *([1] * (provided_initial_conditions.dim() - 1))
+            )
+        return torch.cat(
+            [provided_initial_conditions, generated_initial_conditions], dim=-2
+        )  # should this be shuffled?
```
Balandat, line 291: probably not a bad idea since we're solving some of the optimization in batches by default and it would be good to distribute the provided and generated ones across the different batches.

botorch/optim/optimize.py
```diff
@@ -1177,6 +1240,86 @@ def _gen_batch_initial_conditions_local_search(
...
+        acqvals_init = _split_batch_eval_acqf(
+            acq_function=acq_function,
+            X=X_init.unsqueeze(1),
+            max_batch_size=max_batch_size,
+        ).unsqueeze(-1)
+
+    return X_init[acqvals_init.topk(k=min_points, largest=True, dim=0).indices]
+
+
+def _gen_first_candidate_starting_points_local_search(
```
Balandat, line 1273: It seems like you could simplify this by just having a single funciton `_gen_starting_points_local_search` and just check whether `batch_initial_conditions` is provided and branch on that? Then you can call that recursively with  `batch_initial_conditions=None` in that funciton.

Balandat (changes requested):
Thanks. Seems like you have some `dtype` test issues to fix.

Also, if you rebase on the most recent main branch you don't have to run the conda test in the CI :)

botorch/optim/optimize.py
```diff
@@ -1227,6 +1373,13 @@ def optimize_acqf_discrete_local_search(
         - a `q x d`-dim tensor of generated candidates.
         - an associated acquisition value.
     """
+    if (
+        batch_initial_conditions is not None
+        and len(batch_initial_conditions.shape) != 3
+        and batch_initial_conditions.shape[-1] != 1
+    ):
+        raise ValueError("`batch_initial_conditions` must be of shape `n x 1 x d`.")
```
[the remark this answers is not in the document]
CompRhys: This logic was wrong

CompRhys:
locally tests still failing for `FAILED test/optim/test_optimize.py::TestOptimizeAcqf::test_optimize_acqf_runs_given_batch_initial_conditions - RuntimeError: expected scalar type Long but found Float` although not really sure how i've caused that

test/optim/test_optimize.py
```diff
@@ -546,9 +655,32 @@ def test_optimize_acqf_batch_limit(self) -> None:
...
+                    options=options,
+                    return_best_only=False,
+                    gen_candidates=gen_candidates,
+                    batch_initial_conditions=ics,
+                )
+
+                self.assertEqual(acq_value_list.shape, (ics.shape[0],))
+
     def test_optimize_acqf_runs_given_batch_initial_conditions(self):
         num_restarts, raw_samples, dim = 1, 2, 3
```
saitcakmak, line 685: GH won't let me comment there, but updating the second tensor definitions below from `torch.tensor([1])` to `torch.tensor([1.])` fixes that issue. I believe it expects the coefficients in the constraints to be floats. The test still fails with a shape error when `shape=(1, dim)`.

CompRhys: thanks, I'll look at the other test case issue tomorrow. I was trying to understand what I had done that was changing dtypes but if this is okay then will just update the test accordingly

saitcakmak: I don't think you changed any dtypes. Previously, it would've just used the given ICs and not called `ic_generator` with those constraints (with wrong dtypes), since it didn't need to generate more. The scipy optimizer uses a different format for the constraints, so they're modified before passing in. It just wasn't executing the code path (polytope sampler) that triggered the error.

CompRhys:
```
            if (
                self.raw_samples is not None
                and (self.raw_samples - batch_initial_conditions_shape[-2]) > 0
                and len(batch_initial_conditions_shape) == 3
                and self.num_restarts is not None
                and batch_initial_conditions_shape[0] not in [1, self.num_restarts]
            ):
```

this logic should have been catching the case but wasn't, will fix

botorch/optim/optimize.py
CompRhys: This is now shuffled but there isn't a PRNG seed being set, it looks like the best way to do that would be via `ic_gen_kwargs` but elsewhere in the code `randperm` has been used without seeding and so will leave it to a future refactor/PR if desired

saitcakmak (approved):

saitcakmak:
Looks like the PR is out of sync with the main branch. Can you do another rebase please?

CompRhys:
In this PR as stands I was confused about the roles of q and raw_samples in the `initial_batch_generator` argument

CompRhys:
@saitcakmak I redid this whole thing after realizing that I had got confused and failed to implement what I originally intended by following `raw_samples` rather than `num_repeats` and then got distracted focussing on the trees rather than the forest.

CompRhys:
Apologies for linting.. 

Two solutions to the homotopy test failure: 
1) either change the following linked lines to the snippet testing that we prune and then generate new restarts.  https://github.com/pytorch/botorch/blob/5d3760633cae6d76b10df57e9f8478557e4e1946/test/optim/test_homotopy.py#L237-L246

```python

        for i in range(5):
            # assert that the number of restarts are repopulated at each step
            self.assertEqual(
                prune_candidates_mock.call_args_list[i][1]["candidates"].shape,
                torch.Size([4, 1]),
            )
            # assert that the candidates are pruned to just one candidate
            self.assertEqual(
                prune_candidates(**prune_candidates_mock.call_args_list[i][1]).shape,
                torch.Size([1, 1]),
            )

```

2) set `shared_optimize_acqf_kwargs["raw_samples"]=None` at https://github.com/pytorch/botorch/blob/5d3760633cae6d76b10df57e9f8478557e4e1946/botorch/optim/optimize_homotopy.py#L194 to prevent the dimensions from the pruned restarts being filled with random designs in subsequent loops.

To get current behaviour the second option is better, but in principal despite the fact that random IC designs introduced later in the homotopy might not be able to be as readily optimized to sparsity the first option doesn't seem harmful (beyond there being no computational benefit to pruning) and may have upsides.

saitcakmak:
There is one remaining test failure and a rebase seems to be necessary. Lgtm otherwise

Balandat:
> reset: squash into single commit

Btw @CompRhys for future reference this is not really necessary - the actual commit on the main branch will be made by our github bot (while maintaining attribution to the PR author), which automatically squashes all the changes together.

botorch/optim/optimize.py
```diff
@@ -109,12 +109,41 @@ def __post_init__(self) -> None:
...
                     f"shape is {batch_initial_conditions_shape}."
                 )
 
+            if len(batch_initial_conditions_shape) == 2:
+                warnings.warn(
+                    "If using a 2-dim `batch_initial_conditions` botorch will "
+                    "default to old behavior of ignoring `num_restarts` and just "
+                    "use the given `batch_initial_conditions` by setting "
+                    "`raw_samples` to None.",
+                    RuntimeWarning,
```
Balandat, line 125: Otherwise our linter complains here
```suggestion
                    RuntimeWarning,
                    stacklevel=3,
```

botorch/optim/optimize.py
```diff
@@ -109,12 +109,41 @@ def __post_init__(self) -> None:
...
+                and batch_initial_conditions_shape[-2] != self.q
+            ):
+                warnings.warn(
+                    "If using a 3-dim `batch_initial_conditions` where the "
+                    "first dimension is less than `num_restarts` and the second "
+                    "dimension is not equal to `q`, botorch will default to "
+                    "old behavior of ignoring `num_restarts` and just use the "
+                    "given `batch_initial_conditions` by setting `raw_samples` "
+                    "to None.",
+                    RuntimeWarning,
```
Balandat, line 142:
```suggestion
                    RuntimeWarning,
                    stacklevel=3,
```

botorch/optim/optimize.py
```diff
@@ -1247,6 +1372,16 @@ def optimize_acqf_discrete_local_search(
...
         - an associated acquisition value.
     """
+    if batch_initial_conditions is not None:
+        if not (
+            len(batch_initial_conditions.shape) == 3
+            and batch_initial_conditions.shape[-2] == 1
+        ):
+            raise ValueError(
+                "batch_initial_conditions must have shape `n x 1 x d` if "
+                f"given (recieved {batch_initial_conditions})."
```
Balandat, line 1382: This may not be very useful / possible to parse if `batch_initial_conditions` is large
```suggestion
                f"given (received shape {batch_initial_conditions.shape})."
```

Balandat (commented):
couple of minor nits, o/w this is good to go. If you merge in the recent changes on master that will also fix the tutorial failure.

botorch/optim/optimize.py
CompRhys: this was an oversight I meant it to be the shape

PR #2615, opened 2024-11-06
Remove `as_tensor argument` of `set_tensors_from_ndarray_1d`

AVHopp:
## Motivation

This PR removes the `as_tensor` argument of the `set_tensors_from_ndarray_1d` function in `numpy_utils.py`. 

The reason for this change is that the previous implementation
- messed up the float precision: That is, even when trying to use `float32` precision, this always transformed floats into `float64` precision
- did only use the default function anyway.

This change was discussed previously with @Balandat and @esantorella in #2596 .

Yes, I have read it.

## Test Plan

I checked the format using `ufmt format .` and verified that all tests are still running using `pytest -ra`.

## Related PRs

This PR is related to #2597 and #2596, fixing a similar issue (`float32` being broken by some internal calculations).

### DIFF
--- a/botorch/optim/utils/numpy_utils.py
+++ b/botorch/optim/utils/numpy_utils.py
@@ -125,9 +125,7 @@ def set_tensors_from_ndarray_1d(
                 size = tnsr.numel()
                 vals = array[index : index + size] if tnsr.ndim else array[index]
                 tnsr.copy_(
-                    torch.as_tensor(vals, device=tnsr.device, dtype=tnsr.dtype).view(
-                        tnsr.shape
-                    )
+                    torch.from_numpy(vals.reshape(tnsr.shape)).to(tnsr)
                 )
                 index += size
             except Exception as e:
--- a/botorch/optim/utils/numpy_utils.py
+++ b/botorch/optim/utils/numpy_utils.py
@@ -126,9 +126,7 @@ def set_tensors_from_ndarray_1d(
                 vals = array[index : index + size] if tnsr.ndim else array[index]
                 tnsr.copy_(
                     torch.as_tensor(vals, device=tnsr.device, dtype=tnsr.dtype)
-                    .to(tnsr)
                     .view(tnsr.shape)
-                    .to(tnsr)
                 )
                 index += size
             except Exception as e:
--- a/botorch/optim/utils/numpy_utils.py
+++ b/botorch/optim/utils/numpy_utils.py
@@ -125,8 +125,9 @@ def set_tensors_from_ndarray_1d(
                 size = tnsr.numel()
                 vals = array[index : index + size] if tnsr.ndim else array[index]
                 tnsr.copy_(
-                    torch.as_tensor(vals, device=tnsr.device, dtype=tnsr.dtype)
-                    .view(tnsr.shape)
+                    torch.as_tensor(vals, device=tnsr.device, dtype=tnsr.dtype).view(
+                        tnsr.shape
+                    )
                 )
                 index += size
             except Exception as e:
--- a/botorch/optim/utils/numpy_utils.py
+++ b/botorch/optim/utils/numpy_utils.py
@@ -124,9 +124,7 @@ def set_tensors_from_ndarray_1d(
             try:
                 size = tnsr.numel()
                 vals = array[index : index + size] if tnsr.ndim else array[index]
-                tnsr.copy_(
-                    torch.from_numpy(vals.reshape(tnsr.shape)).to(tnsr)
-                )
+                tnsr.copy_(torch.from_numpy(vals.reshape(tnsr.shape)).to(tnsr))
                 index += size
             except Exception as e:
                 raise RuntimeError(
--- a/botorch/optim/utils/numpy_utils.py
+++ b/botorch/optim/utils/numpy_utils.py
@@ -9,7 +9,6 @@ r"""Utilities for interfacing Numpy and Torch."""
 from __future__ import annotations
 
 from collections.abc import Callable, Iterator
-
 from itertools import tee
 
 import numpy as np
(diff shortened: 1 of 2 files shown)


botorch/optim/utils/numpy_utils.py
```diff
@@ -125,7 +124,12 @@ def set_tensors_from_ndarray_1d(
             try:
                 size = tnsr.numel()
                 vals = array[index : index + size] if tnsr.ndim else array[index]
-                tnsr.copy_(as_tensor(vals).to(tnsr).view(tnsr.shape).to(tnsr))
+                tnsr.copy_(
+                    torch.as_tensor(vals, device=tnsr.device, dtype=tnsr.dtype)
+                    .to(tnsr)
+                    .view(tnsr.shape)
+                    .to(tnsr)
```
Balandat: It seems that these `.to()` are noops in this setup given that the tensor is initially constructed on the right device with the right dtype? 
```suggestion
                    torch.as_tensor(vals, device=tnsr.device, dtype=tnsr.dtype)
                    .view(tnsr.shape)
```

saitcakmak, line 131: These `.to` calls are redundant, since you already specified device and dtype in `as_tensor` call.
```suggestion
                    .view(tnsr.shape)
```

saitcakmak (approved):

botorch/optim/utils/numpy_utils.py
Balandat: Actually, let's do the reshaping first and then use `from_numpy`: `torch.from_numpy(vals.reshape(tnsr.shape)).to(tnsr)`. I did some lightweight benchmarking and this generally is a tad faster.

Balandat (commented):
Thanks for making this change!

botorch/optim/utils/numpy_utils.py
```diff
@@ -125,7 +124,11 @@ def set_tensors_from_ndarray_1d(
             try:
                 size = tnsr.numel()
                 vals = array[index : index + size] if tnsr.ndim else array[index]
-                tnsr.copy_(as_tensor(vals).to(tnsr).view(tnsr.shape).to(tnsr))
+                tnsr.copy_(
+                    torch.as_tensor(vals, device=tnsr.device, dtype=tnsr.dtype).view(
+                        tnsr.shape
+                    )
```
saitcakmak, line 130:
```suggestion
                    torch.from_numpy(vals.reshape(tnsr.shape)).to(tnsr)
```

botorch/optim/utils/numpy_utils.py
saitcakmak: The benchmarking explains the 22 minute gap between the two comments :D. I updated the diff directly rather than waiting for back & forth.

botorch/optim/utils/numpy_utils.py
```diff
@@ -125,7 +124,9 @@ def set_tensors_from_ndarray_1d(
             try:
                 size = tnsr.numel()
                 vals = array[index : index + size] if tnsr.ndim else array[index]
-                tnsr.copy_(as_tensor(vals).to(tnsr).view(tnsr.shape).to(tnsr))
+                tnsr.copy_(
+                    torch.from_numpy(vals.reshape(tnsr.shape)).to(tnsr)
+                )
```
saitcakmak, line 129: Ufmt
```suggestion
                tnsr.copy_(torch.from_numpy(vals.reshape(tnsr.shape)).to(tnsr))
```

botorch/optim/utils/numpy_utils.py
saitcakmak: Looks like this doesn't work if vals is a scalar. Reverting to the previous solution

botorch/optim/utils/numpy_utils.py
```diff
@@ -125,7 +124,7 @@ def set_tensors_from_ndarray_1d(
             try:
                 size = tnsr.numel()
                 vals = array[index : index + size] if tnsr.ndim else array[index]
-                tnsr.copy_(as_tensor(vals).to(tnsr).view(tnsr.shape).to(tnsr))
+                tnsr.copy_(torch.from_numpy(vals.reshape(tnsr.shape)).to(tnsr))
```
saitcakmak, line 127:
```suggestion
                torch.as_tensor(vals, device=tnsr.device, dtype=tnsr.dtype).view(
                    tnsr.shape
                )
```

botorch/optim/utils/numpy_utils.py
AVHopp: Thanks for pointing out :)

botorch/optim/utils/numpy_utils.py
```diff
@@ -125,7 +124,10 @@ def set_tensors_from_ndarray_1d(
             try:
                 size = tnsr.numel()
                 vals = array[index : index + size] if tnsr.ndim else array[index]
-                tnsr.copy_(as_tensor(vals).to(tnsr).view(tnsr.shape).to(tnsr))
+                tnsr.copy_(
+                    torch.as_tensor(vals, device=tnsr.device, dtype=tnsr.dtype)
+                    .view(tnsr.shape)
```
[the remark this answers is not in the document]
AVHopp: Huh, weird I did run `ufmt format`, maybe my IDE did some autoformatting. Thanks for fixing :)

botorch/optim/utils/numpy_utils.py
AVHopp: Since there have been change after this comment, is there anything I can still do to assist? In any case, thanks for already implementing the necessary changes :)

botorch/optim/utils/numpy_utils.py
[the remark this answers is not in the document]
saitcakmak: This was caused by my previous update :). Updating PRs directly on GH can get messy

botorch/optim/utils/numpy_utils.py
saitcakmak: I think the PR is ready as is. I'll land it after the internal checks pass. Thanks for the contribution!

AVHopp: Glad that I could help :)

PR #2632, opened 2024-11-18
Add pre-commit hook to stop linting errors being pushed.

CompRhys:
## Motivation

I am so used to relying on `pre-commit` that I have forgotten to run manual linting when making PRs to `botorch`. This adds a simple `pre-commit` config to automate linting.

Yes

## Test Plan

This PR should not affect anything. There are tools to run pre-commit in ci that maintainers could add to the package: https://pre-commit.ci/

## Related Issues:
https://github.com/omnilib/ufmt/issues/251 The pre-commit hook only seems to work for v2.3.0 of ufmt. This issue tracks the error following from ruff_api.

### DIFF
--- a/scripts/check_pre_commit_reqs.py
+++ b/scripts/check_pre_commit_reqs.py
@@ -1,8 +1,8 @@
 #!/usr/bin/env python3
-+# Copyright (c) Meta Platforms, Inc. and affiliates.
-+#
-+# This source code is licensed under the MIT license found in the
-+# LICENSE file in the root directory of this source tree.
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+#
+# This source code is licensed under the MIT license found in the
+# LICENSE file in the root directory of this source tree.
 
 import sys
 from pathlib import Path
--- a/scripts/check_pre_commit_reqs.py
+++ b/scripts/check_pre_commit_reqs.py
@@ -1,0 +1,98 @@
+#!/usr/bin/env python3
+
+import sys
+from pathlib import Path
+
+import yaml
+
+
+def parse_requirements(filepath):
+    """Parse requirements file and return a dict of package versions."""
+    versions = {}
+    with open(filepath) as f:
+        for line in f:
+            line = line.strip()
+            if line and not line.startswith("#"):
+                # Handle different requirement formats
+                if "==" in line:
+                    pkg, version = line.split("==")
+                    versions[pkg.strip().lower()] = version.strip()
+    return versions
+
+
+def parse_precommit_config(filepath):
+    """Parse pre-commit config and extract ufmt repo rev and hook dependencies."""
+    with open(filepath) as f:
+        config = yaml.safe_load(f)
+
+    versions = {}
+    for repo in config["repos"]:
+        if "https://github.com/omnilib/ufmt" in repo.get("repo", ""):
+            # Get ufmt version from rev - assumes fixed format: vX.Y.Z
+            versions["ufmt"] = repo.get("rev", "").replace("v", "")
+
+            # Get dependency versions
+            for hook in repo["hooks"]:
+                if hook["id"] == "ufmt":
+                    for dep in hook.get("additional_dependencies", []):
+                        if "==" in dep:
+                            pkg, version = dep.split("==")
+                            versions[pkg.strip().lower()] = version.strip()
+            break
+    return versions
+
+
+def main():
+    # Find the pre-commit config and requirements files
+    config_file = Path(".pre-commit-config.yaml")
+    requirements_file = Path("requirements-fmt.txt")
+
+    if not config_file.exists():
+        print(f"Error: Could not find {config_file}")
+        sys.exit(1)
+
+    if not requirements_file.exists():
+        print(f"Error: Could not find {requirements_file}")
+        sys.exit(1)
+
+    # Parse both files
+    req_versions = parse_requirements(requirements_file)
+    config_versions = parse_precommit_config(config_file)
+
+    # Packages to check
+    packages = ["ufmt", "black", "usort", "ruff-api"]
+
+    # Check versions
+    mismatches = []
+    for pkg in packages:
+        req_ver = req_versions.get(pkg, None)
+        config_ver = config_versions.get(pkg, None)
+
+        if req_ver != config_ver:
+            found_version_str = f"{pkg}: {requirements_file} has {req_ver},"
+            if pkg == "ufmt":
+                mismatches.append(
+                    f"{found_version_str} pre-commit config rev has v{config_ver}"
+                )
+            else:
+                mismatches.append(
+                    f"{found_version_str} pre-commit config has {config_ver}"
+                )
+
+    # Report results
+    if mismatches:
+        msg_str = "".join("\n\t" + msg for msg in mismatches)
+        print(
+            f"Version mismatches found:{msg_str}"
+            "\nPlease update the versions in `.pre-commit-config.yaml` to be "
+            "consistent with those in `requirements-fmt.txt` (source of truth)."
+            "\nNote: all versions must be pinned exactly ('==X.Y.Z') in both files."
+        )
+        sys.exit(1)
+    else:
+        print("All versions match!")
+        sys.exit(0)
+
+
+if __name__ == "__main__":
+    main()
--- a/scripts/check_pre_commit_reqs.py
+++ b/scripts/check_pre_commit_reqs.py
@@ -59,12 +59,9 @@ def main():
     req_versions = parse_requirements(requirements_file)
     config_versions = parse_precommit_config(config_file)
 
-    # Packages to check
-    packages = ["ufmt", "black", "usort", "ruff-api"]
-
     # Check versions
     mismatches = []
-    for pkg in packages:
+    for pkg, req_ver in req_versions.items():
         req_ver = req_versions.get(pkg, None)
         config_ver = config_versions.get(pkg, None)
 
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -10,39 +10,18 @@ on:
 
 jobs:
 
-  ufmt:
-    name: Code formatting and sorting with ufmt
+  lint:
     runs-on: ubuntu-latest
-    strategy:
-      fail-fast: false
     steps:
-    - uses: actions/checkout@v4
-    - name: Set up Python
-      uses: actions/setup-python@v5
-      with:
-        python-version: "3.10"
-    - name: Install dependencies
-      run: |
-        # pin dependencies to match Meta-internal versions
-        pip install -r requirements-fmt.txt
-    - name: ufmt
-      run: |
-        ufmt diff .
+      - uses: actions/checkout@v4
 
-  flake8:
-    name: Lint with flake8
-    runs-on: ubuntu-latest
-    strategy:
-      fail-fast: false
-    steps:
-    - uses: actions/checkout@v4
-    - name: Set up Python
-      uses: actions/setup-python@v5
-      with:
-        python-version: "3.10"
-    - name: Install dependencies
-      run: |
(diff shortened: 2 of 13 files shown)


Balandat:
In general I am supportive of this.

The main question I have is how we can ensure that this stays in sync with the versions as defined in https://github.com/pytorch/botorch/blob/main/requirements-fmt.txt. Note that the versions in that file are synced with our internal linting tools, and so these may be updated without us actively doing those updates. If we hard-code versions of the tools in `.pre-commit-config.yaml` then we're going to end up with version inconsistencies that may be annoying to deal with (i.e. the pre-commit hook will format things in a way that is inconsistent with our internal linters). Is there a way to programmatically source the versions for the pre-commit hooks from the`requirements-fmt.txt` file?

CompRhys:
So your concern is immediately apparent given this v2.3.0 vs v2.8.0 issue there were 3 lines of code which were changed by `ufmt` here that differ to running `ufmt format .` on v2.8.0 manually with the most recent version.

Is the SSOT for the internal tools `requirements-fmt.txt` or is that file produced based on some other process?

Balandat:
> Is the SSOT for the internal tools requirements-fmt.txt or is that file produced based on some other process?

So the SSOT for this lives in our internal repos; the way it works is that if changes are made there this automatically gets exported to the different github repos that we have for OSS projects. So for all intents and purposes, from the PoV of OSS versioning and pre-commit hooks, the versions defined in `requirements-fmt.txt` are the SOT.

CompRhys:
Having had a think about this I think the best I can do is define a script that parses the file and the action file and then fails if the two do not match. This isn't automatic but it would mean that anyone using the pre-commit would be forced to fix the divergence in order to satisfy the hook with would give eventual consistency to the SSOT -- it would only significantly diverge if no-one used the pre-commit.

That said given the linked issue that would put this PR on ice until https://github.com/omnilib/ufmt/issues/251 is resolved.

Has the internal team considered `ruff` as a replacement to `ufmt` and `flake8`, single tool, highly configurable and I've had great experience with its CI/pre-commit ease of use.

CompRhys:
This now implements the proposed solution and it's easy to do the downstream fix and set ruff-api in the action I defined and so not actually blocked by https://github.com/omnilib/ufmt/issues/251.

scripts/check_requirements.py
```diff
@@ -0,0 +1,110 @@
...
+            # Get ufmt version from rev
+            rev = repo.get("rev", "")
+            if rev.startswith("v"):
+                versions["ufmt"] = rev[1:]  # Remove 'v' prefix
+
+            # Get dependency versions
+            for hook in repo["hooks"]:
+                if hook["id"] == "ufmt":
+                    for dep in hook.get("additional_dependencies", []):
+                        if "==" in dep:
```
Balandat, line 45: When would `==` not be part of this string? Let's error out if that occurs since we don't know how to properly handle it?

scripts/check_requirements.py
```diff
@@ -0,0 +1,110 @@
...
+    # Packages to check
+    packages = ["ufmt", "black", "usort", "ruff-api"]
+
+    # Check versions
+    mismatches = []
+    for pkg in packages:
+        req_ver = req_versions.get(pkg)
+        config_ver = config_versions.get(pkg)
+
+        if not req_ver:
```
Balandat, line 78:
```suggestion
        if req_ver is None:
```

scripts/check_requirements.py
```diff
@@ -0,0 +1,110 @@
...
+            print(f"Warning: {pkg} not found in {requirements_file}")
+            continue
+
+        if not config_ver:
```
Balandat, line 82:
```suggestion
        if config_ver is None:
```

scripts/check_requirements.py
```diff
@@ -0,0 +1,110 @@
...
+    with open(filepath) as f:
+        config = yaml.safe_load(f)
+
+    versions = {}
+    for repo in config["repos"]:
+        if "https://github.com/omnilib/ufmt" in repo.get("repo", ""):
+            # Get ufmt version from rev
+            rev = repo.get("rev", "")
+            if rev.startswith("v"):
+                versions["ufmt"] = rev[1:]  # Remove 'v' prefix
```
Balandat, line 39: could do this
```suggestion
            rev = repo.get("rev", "").replace("v", "")
```

scripts/check_requirements.py
```diff
@@ -0,0 +1,110 @@
...
+            if pkg == "ufmt":
+                mismatches.append(
+                    f"{pkg}: {requirements_file} has {req_ver}, "
+                    f"pre-commit config rev has v{config_ver}"
+                )
+            else:
+                mismatches.append(
+                    f"{pkg}: {requirements_file} has {req_ver}, "
+                    f"pre-commit config has {config_ver}"
+                )
```
Balandat, line 96:
```suggestion
            found_version_str = f"{pkg}: {requirements_file} has {req_ver}, "
            if pkg == "ufmt":
                mismatches.append(
                    found_version_str + 
                    f"pre-commit config rev has v{config_ver}"
                )
            else:
                mismatches.append(
                    found_version_str + 
                    f"pre-commit config has {config_ver}"
                )
```

amyreese:
>  Is there a way to programmatically source the versions for the pre-commit hooks from the requirements-fmt.txt file? 

No, and this is a known/intentional limitation of pre-commit — they don't offer any option beyond explicit package lists in `additional_dependencies`, and have no desire to support requirements files. See https://github.com/pre-commit/pre-commit/issues/730

scripts/check_requirements.py
```diff
@@ -0,0 +1,110 @@
...
+
+    # Report results
+    if mismatches:
+        print("Version mismatches found:")
+        for msg in mismatches:
+            print(f"  {msg}")
+        sys.exit(1)
```
Balandat, line 103:
```suggestion
        sys.exit(1)
        print(
            "Please update the versions in `.pre-commit-config.yaml` " +
            " to be consistent with those in `requirements-fmt.txt` (source of truth)"
        )
```

Balandat (commented):
Thanks this overall lgtm, just some nits. 

The formatting changes here, are those things that snuck through previously on our end? It doesn't seem like there are any changes in terms of the versions?

CompRhys:
> Thanks this overall lgtm, just some nits.
> 
> The formatting changes here, are those things that snuck through previously on our end? It doesn't seem like there are any changes in terms of the versions?

Will get the nits and then double check it still works.

They were flake8 errors I got running flake globally with `pre-commit run --all-files`, the versions of flake aren't pinned so potentially a divergence issue but to me at least they all seemed to make sense, f"{thing=}" rather than f"{thing = }" was most common.

scripts/check_requirements.py
CompRhys:
```suggestion
            versions["ufmt"] = repo.get("rev", "").replace("v", "")
```

scripts/check_requirements.py
CompRhys: as setup now and assuming `requirements-fmt.txt` will always be a lock file of sorts then it would always be there. This code would already error as written if `==` wasn't in the string as we'd get a comparison against none.

scripts/check_requirements.py
CompRhys:
```suggestion
            found_version_str = f"{pkg}: {requirements_file} has {req_ver},"
	            if pkg == "ufmt":
	                mismatches.append(
	                    f"{found_version_str} pre-commit config rev has v{config_ver}"
	                )
	            else:
	                mismatches.append(
	                    f"{found_version_str} pre-commit config has {config_ver}"
	                )
```

Balandat:
Makes sense, let's just import it and see if the internal linter complains... Main thing would be that flake8 honors the settings in `settings.cfg`: https://github.com/pytorch/botorch/blob/main/setup.cfg#L1

scripts/check_requirements.py
```diff
@@ -0,0 +1,111 @@
...
+        if parse_version(req_ver) != parse_version(config_ver):
+            found_version_str = f"{pkg}: {requirements_file} has {req_ver},"
+	            if pkg == "ufmt":
+	                mismatches.append(
+	                    f"{found_version_str} pre-commit config rev has v{config_ver}"
+	                )
+	            else:
+	                mismatches.append(
+	                    f"{found_version_str} pre-commit config has {config_ver}"
+	                )
```
Balandat, line 93:
```suggestion
	        if pkg == "ufmt":
	            mismatches.append(
	                f"{found_version_str} pre-commit config rev has v{config_ver}"
	            )
	        else:
	            mismatches.append(
	                f"{found_version_str} pre-commit config has {config_ver}"
	            )
```

CompRhys:
Checked the edge cases manually and tested that it takes the flake8 settings from `setup.cfg` by setting line length to 60 and getting huge numbers of faults. Added a clarifying note that everything should be pinned with `==` for the logic to work.

I think this is good to go.

scripts/check_pre_commit_reqs.py
```diff
@@ -0,0 +1,97 @@
...
+                    f"{found_version_str} pre-commit config has {config_ver}"
+                )
+
+    # Report results
+    if mismatches:
+        print(
+            f"Version mismatches found:{''.join(f'\n\t{msg}' for msg in mismatches)}"
+            "\nPlease update the versions in `.pre-commit-config.yaml` to be "
+            "consistent with those in `requirements-fmt.txt` (source of truth)."
+            "\nNote: all versions must be pinned exactly ('==X.Y.Z') in both files."
```
Balandat, line 88: flake8 is complaining about this

CompRhys: My solution would be to recommend turning off the rule based on the fact that backslashes are now allowed in python. I guess the issues is that in the monorepo you have internally there may be python versions at EOL where this is important. 

If locally changing the rule in the repo isn't a viable option because the config is global on the mirror then can look at ignoring the rule for this line?

Balandat:
Hmm @CompRhys this seems like this is need of another rebase :(

CompRhys:
> Hmm @CompRhys this seems like this is need of another rebase :(

I think this is actually a real error based on the behaviour of f-strings f"{xyz=}" being a special when f"{xyz = }" may not be

```
(base) ➜  botorch git:(pre-commit) python
Python 3.12.7 | packaged by Anaconda, Inc. | (main, Oct  4 2024, 08:22:19) [Clang 14.0.6 ] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> a = "abc"
>>> f"{a=}"
"a='abc'"
>>> f"{a = }"
"a = 'abc'"
>>> 
```

Balandat:
I wasn't referring to the error as much as that I wasn't able to re-import the PR. 

Also, did you mean to  close this?

CompRhys:
force-pushing on reset/rebase closes PRs

CompRhys:
> So one thought I had to ensure that the hard-coded yaml versions stay in sync with the SOT in the `requirements-fmt.txt`: Can we add a simple unit test that validates that? That way we'd avoid that divergence upfront rather than breaking things for contributors. We have that validation logic already; ideally we could just factor that out and reuse the same code for both the validation and the unit tests - (though we'd probably have to do some hacking to import it in the script).

could I change the lint CI to be this:
```
      - name: Run pre-commit
        run: pre-commit run --all-files --show-diff-on-failure
```
as then that would check in CI and also creates a SSOT connection between the pre-commit and the lint action, or is the lint action created by some internal tool?

I think doing it with the unit tests is more messy but could do if needed

Balandat:
> could I change the lint CI to be this:

Yeah I guess that could work? Though that would require setting up `pre-commit` and installing the deps into a virtualenv on every lint run, right? Would we have to edit the lint workflow to do that or does that happen automatically on the first run in a virgin setup?

Balandat:
I realized we shold probably have updated the `CONTRIBUTING` readme - will do that in a follow up

PR #2636, opened 2024-11-20
Add TopK downselection for initial batch generation.

CompRhys:
## Motivation

In order to get https://github.com/facebook/Ax/pull/2938 over the line with initial candidate generation that obey the constraints we want to use the existing tooling within `botorch`. The hard coded logic currently in Ax uses topk to downselect the sobol samples. To make a change there that will not impact existing users we then need to implement topk downselection in `botorch`.

## Test Plan

TODO:
- [x] add tests for initialize_q_batch_topk

## Related PRs

https://github.com/facebook/Ax/pull/2938. (https://github.com/pytorch/botorch/pull/2610 was initially intended to play part of this solution but then I realized that the pattern I wanted to use was conflating repeats and the batch dimension.)


botorch/optim/initializers.py
```diff
@@ -1057,6 +1079,53 @@ def initialize_q_batch_nonneg(
     return X[idcs], acq_vals[idcs]
 
 
+def initialize_q_batch_topk(
+    X: Tensor, acq_vals: Tensor, n: int, largest: bool = True, sorted: bool = True
+) -> tuple[Tensor, Tensor]:
+    r"""Take the top `n` initial conditions for candidate generation.
```
Balandat, line 1085: Seems a bit odd to call this `topk` but then call the arg `n` instead of `k`.
```suggestion
def initialize_q_batch_topk(
    X: Tensor, acq_vals: Tensor, k: int, largest: bool = True, sorted: bool = True
) -> tuple[Tensor, Tensor]:
    r"""Take the top `k` initial conditions for candidate generation.
```

botorch/optim/initializers.py
```diff
@@ -1057,6 +1079,53 @@ def initialize_q_batch_nonneg(
...
+    if n > n_samples:
+        raise RuntimeError(
+            f"n ({n}) cannot be larger than the number of "
+            f"provided samples ({n_samples})"
+        )
+    elif n == n_samples:
+        return X, acq_vals
+
+    Ystd = acq_vals.std(dim=0)
+    if torch.any(Ystd == 0):
```
Balandat, line 1115: I guess by the above logic we'd never have to worry about there only being one element, right?

botorch/optim/initializers.py
```diff
@@ -1057,6 +1079,53 @@ def initialize_q_batch_nonneg(
...
+            "All acquisition values for raw samples points are the same for "
+            "at least one batch. Choosing initial conditions at random.",
+            BadInitialCandidatesWarning,
+            stacklevel=3,
+        )
+        idcs = torch.randperm(n=n_samples, device=X.device)[:n]
+        return X[idcs], acq_vals[idcs]
+
+    idcs = acq_vals.topk(n, largest=largest, sorted=sorted).indices
+    return X[idcs], acq_vals[idcs]
```
Balandat, line 1126: µ-optimization here
```suggestion
    topk_out = acq_vals.topk(n, largest=largest, sorted=sorted).indices
    return X[topk_out.indices], topk_out.values
```

botorch/utils/sampling.py
```diff
@@ -98,14 +98,12 @@ def draw_sobol_samples(
...
-    rng = bounds[1] - bounds[0]
     sobol_engine = SobolEngine(q * d, scramble=True, seed=seed)
-    samples_raw = sobol_engine.draw(batch_size * n, dtype=lower.dtype)
-    samples_raw = samples_raw.view(*batch_shape, n, q, d).to(device=lower.device)
+    samples_raw = sobol_engine.draw(batch_size * n, dtype=bounds.dtype)
+    samples_raw = samples_raw.view(*batch_shape, n, q, d).to(device=bounds.device)
     if batch_shape != torch.Size():
         samples_raw = samples_raw.permute(-3, *range(len(batch_shape)), -2, -1)
-    return lower + rng * samples_raw
+    return unnormalize(samples_raw, bounds)
```
Balandat, line 106: nice

Balandat (commented):
couple of small nits, o/w this looks great.

saitcakmak:
Other initializers use Boltzmann sampling through `initialize_q_batch` helper to achieve a more diverse set of ICs. The motivation is that this will hopefully lead to different local-optima, making it more likely to discover the global optimum. If you have multiple raw samples that are close by, TopK can get you ICs that all lead to the same local optima. Should we follow the same logic in `initialize_q_batch` instead?

Balandat:
That would make sense, I'd be interested in seeing a benchmark how that affects performance of SEBO

botorch/optim/initializers.py
CompRhys: I think to match the sister functions it should stay as `n`? happy to change the method to be called topn but used topk just to match the name of the `Tensor` method

botorch/optim/initializers.py
CompRhys: Honestly this was just verbatim copied from the boltzman sampled equivalent but I believe that if `raw_samples=1` and `n=1` we would just return the single sample and if it was bigger it would raise the error.

CompRhys:
I would be entirely happy to not merge this if it will end up as dead code if theres a decision to swap the ic default for SEBO to use Boltzmann sampling. The motivation here was that I want to use the polytope sampling from botorch such that I can add inequality constraints to SEBO, adding a bunch of switching logic is bad in terms of maintenance there so the minimal change to keep the same behavior for how it's currently setup is to add this here.

The longer term idea for the cleanest way to do it I had was based on this https://github.com/pytorch/botorch/pull/2610 where SEBO default could just pass the `X_pareto` points as initial batch conditions and then have botorch just fill up the rest of the requested `n_restarts` without needing to call anything in Ax from botorch to do the initial sampling that currently uses topk to get 1/2 of the initial conditions.

botorch/optim/initializers.py
Balandat: Yeah good point needs to have the same signature. I think we can call this `topn` instead.

botorch/optim/initializers.py
```diff
@@ -363,7 +373,7 @@ def gen_batch_initial_conditions(
                         X_rnd_nlzd = torch.rand(
                             n, q, bounds_cpu.shape[-1], dtype=bounds.dtype
                         )
-                    X_rnd = bounds_cpu[0] + (bounds_cpu[1] - bounds_cpu[0]) * X_rnd_nlzd
+                    X_rnd = X_rnd_nlzd * (bounds_cpu[1] - bounds_cpu[0]) + bounds_cpu[0]
```
Balandat, line 376: you could use `unnormalize` from https://github.com/pytorch/botorch/blob/2e143c9c7876f4d3534cc46e56a80bd8120f8d62/botorch/utils/transforms.py#L96 here

botorch/utils/sampling.py
```diff
@@ -98,14 +98,12 @@ def draw_sobol_samples(
...
-    rng = bounds[1] - bounds[0]
     sobol_engine = SobolEngine(q * d, scramble=True, seed=seed)
-    samples_raw = sobol_engine.draw(batch_size * n, dtype=lower.dtype)
-    samples_raw = samples_raw.view(*batch_shape, n, q, d).to(device=lower.device)
+    samples_raw = sobol_engine.draw(batch_size * n, dtype=bounds.dtype)
+    samples_raw = samples_raw.view(*batch_shape, n, q, d).to(device=bounds.device)
     if batch_shape != torch.Size():
         samples_raw = samples_raw.permute(-3, *range(len(batch_shape)), -2, -1)
-    return lower + rng * samples_raw
+    return bounds[0] + (bounds[1] - bounds[0]) * samples_raw
```
Balandat, line 106: same here, let's use `unnormalize` since we have it...

test/optim/test_initializers.py
```diff
@@ -246,6 +280,86 @@ def test_gen_batch_initial_conditions(self):
...
 
+    def test_gen_batch_initial_conditions_topn(self):
+        bounds = torch.stack([torch.zeros(2), torch.ones(2)])
+        mock_acqf = MockAcquisitionFunction()
+        mock_acqf.objective = lambda y: y.squeeze(-1)
+        mock_acqf.maximize = True  # Add maximize attribute
+        for dtype in (torch.float, torch.double):
+            bounds = bounds.to(device=self.device, dtype=dtype)
+            mock_acqf.X_baseline = bounds  # for testing sample_around_best
+            mock_acqf.model = MockModel(MockPosterior(mean=bounds[:, :1]))
```
Balandat, line 291: if you pull this into the `product()` below you can save one indent ...

botorch/utils/sampling.py
CompRhys: The reason I reverted this having originally change to unnormalize is that unnormalize is intended as the reverse of the normalize function which has special behaviour for bounds where they're both equal. When I swapped these two cases to normalize initially I ended up breaking some multi-fidelity contribution's tests that relied on the behaviour than unnormalize with equal bounds would return the lower bound rather than 1. I could add an option to toggle that behaviour but that seems a little out of scope. https://github.com/pytorch/botorch/pull/2636#discussion_r1856802825

test/optim/test_initializers.py
CompRhys: happy to change this but it matches the other testcases perhaps to avoid reassigning lines 289-291 in every iteration of the product. If I change here will change in all the equivalent tests for consistency.

botorch/optim/initializers.py
```diff
@@ -395,6 +408,8 @@ def gen_batch_initial_conditions(
                     )
             # Keep X on CPU for consistency & to limit GPU memory usage.
             X_rnd = fix_features(X_rnd, fixed_features=fixed_features).cpu()
+
+            # Append the fixed fantasies to the randomly generated points
```
esantorella, line 412: Thanks for the comments!

test/optim/test_initializers.py
```diff
@@ -155,6 +157,38 @@ def test_initialize_q_batch(self):
...
+            ics, _ = initialize_q_batch_topn(X=X, acq_vals=acq_vals, n=2)
+            self.assertEqual(ics.shape, torch.Size([2, 3, 4]))
+            self.assertEqual(ics.device, X.device)
+            self.assertEqual(ics.dtype, X.dtype)
+            # ensure raises correct warning
+            acq_vals = torch.zeros(5, device=self.device, dtype=dtype)
+            with warnings.catch_warnings(record=True) as w:
+                ics, _ = initialize_q_batch_topn(X=X, acq_vals=acq_vals, n=2)
+            self.assertEqual(len(w), 1)
+            self.assertTrue(issubclass(w[-1].category, BadInitialCandidatesWarning))
```
esantorella, line 187: Nit:

```suggestion
            with self.assertWarns(
                 BadInitialCandidatesWarning,
            ):
                ics, _ = initialize_q_batch_topn(X=X, acq_vals=acq_vals, n=2)
```
You could also check that the warning message is right using `assertWarnsRegex`

test/optim/test_initializers.py
```diff
@@ -187,64 +221,152 @@ def test_gen_batch_initial_conditions(self):
...
+                    else torch.Size([init_batch_limit])
+                )
+                raw_samps = mock_acqf_call.call_args[0][0]
+                batch_shape = (
+                    torch.Size([20 if sample_around_best else 10])
+                    if init_batch_limit is None
+                    else torch.Size([init_batch_limit])
+                )
+                expected_raw_samps_shape = batch_shape + torch.Size([1, 2])
+                self.assertEqual(raw_samps.shape, expected_raw_samps_shape)
```
esantorella, line 284: Nit: This can all be un-indented

esantorella:
Thank you!

esantorella:
Could you update the documentation for the `options` argument to `gen_batch_initial_conditions`? It might be too much work to explain all of the possible options, but it would be good to at least document the options newly added here such as `topn`. Thanks again for getting this in!

test/optim/test_initializers.py
CompRhys: sure, updated a bunch of places in this file when the context managers scope could be reduced

test/optim/test_initializers.py
```diff
@@ -187,64 +221,152 @@ def test_gen_batch_initial_conditions(self):
...
+            ffs,
+            sample_around_best,
+        ) in product(
+            (torch.float, torch.double),
+            [True, False],
+            [None, 1234],
+            [None, 1],
+            [None, {0: 0.5}],
+            [True, False],
+        ):
```
saitcakmak, line 238: This runs 64 combinations of tests, which is a bit too much. It'd be a lot more efficient to just run a handful of combinations that cover all options.
```suggestion
        ) in (
            (torch.float, True, None, None, None, True),
            (torch.double, False, 1234, 1, {0: 0.5}, False),
            (torch.double, True, 1234, None, {0: 0.5}, True),  # could add a few more that are interesting
        ):
```

test/optim/test_initializers.py
```diff
@@ -187,64 +221,152 @@ def test_gen_batch_initial_conditions(self):
...
+        ) in product(
+            [torch.float, torch.double],
+            [True, False],
+            [True, False, None],
+            [True, False],
+            [None, 1234],
+            [None, 1],
+            [None, {0: 0.5}],
+            [True, False],
+        ):
```
saitcakmak, line 315: Same comment here. This one has 384 combinations.

test/optim/test_initializers.py
CompRhys: will do, here the change was just combining the two loops based on a comment from @Balandat to remove an indent.

test/optim/test_initializers.py
CompRhys: have reduced the number of testcases, `pytest --cov` says it's still 100% coverage so hopefully that matches you're codecov setup

test/optim/test_initializers.py
Balandat: Fair enough. I think we can go either way here. FWIW the overhead of reassigning lines 289-291 in every iteration should be negligible.

CompRhys:
@saitcakmak is there anything blocking this currently or can it be merged such that https://github.com/facebook/Ax/pull/2938 can hopefully be merged in short term?

saitcakmak:
Shouldn't be any blockers. I'll work with @Balandat to get this landed

saitcakmak:
We have a test failure when running on GPU. Looks like cuda & cpu tensors are being mixed together somewhere.
```
test: test_gen_batch_initial_conditions_highdim (optim.test_initializers.TestGenBatchInitialCandidates)
error: Traceback (most recent call last):
  File "/re_cwd/buck-out/v2/gen/fbcode/03967c653c571e2c/pytorch/botorch/__test_optim_cuda__/test_optim_cuda#link-tree/pytorch/botorch/test/optim/test_initializers.py", line 383, in test_gen_batch_initial_conditions_highdim
    batch_initial_conditions = gen_batch_initial_conditions(
  File "/re_cwd/buck-out/v2/gen/fbcode/03967c653c571e2c/pytorch/botorch/__test_optim_cuda__/test_optim_cuda#link-tree/botorch/optim/initializers.py", line 378, in gen_batch_initial_conditions
    X_rnd = unnormalize(
  File "/re_cwd/buck-out/v2/gen/fbcode/03967c653c571e2c/pytorch/botorch/__test_optim_cuda__/test_optim_cuda#link-tree/botorch/utils/transforms.py", line 129, in unnormalize
    return X * (bounds[1] - bounds[0]) + bounds[0]
RuntimeError: Expected all tensors to be on the same device, but found at least two devices, cuda:0 and cpu!
```
X & bounds must be on different devices here. Since it is a single test failure, it could just be on the specific test

test/optim/test_initializers.py
```diff
@@ -187,113 +220,196 @@ def test_gen_batch_initial_conditions(self):
...
-                            "nonnegative": nonnegative,
-                            "eta": 0.01,
-                            "alpha": 0.1,
-                            "seed": seed,
-                            "sample_around_best": sample_around_best,
-                        },
-                    )
+            with warnings.catch_warnings(record=True) as ws:
+                warnings.simplefilter("ignore", category=BadInitialCandidatesWarning)
+                batch_initial_conditions = gen_batch_initial_conditions(
```
saitcakmak, line 383: The failure is on this line. `X` is generated internally and `bounds` is on the correct device above. So, the code generating `X` must be keeping it on CPU

botorch/optim/initializers.py
```diff
@@ -363,7 +375,9 @@ def gen_batch_initial_conditions(
                         X_rnd_nlzd = torch.rand(
                             n, q, bounds_cpu.shape[-1], dtype=bounds.dtype
                         )
-                    X_rnd = bounds_cpu[0] + (bounds_cpu[1] - bounds_cpu[0]) * X_rnd_nlzd
+                    X_rnd = unnormalize(
+                        X_rnd_nlzd, bounds, update_constant_bounds=False
```
saitcakmak, line 379: I suspect this is the fix
```suggestion
                        X_rnd_nlzd, bounds_cpu, update_constant_bounds=False
```

PR #2639, opened 2024-11-27
Optimization over mixed spaces in `optimize_acqf_homotopy`

jduerholt:
## Motivation

So far, `optimize_acqf_homotopy` can only handle `fixed_features` but not `fixed_features_list` which is useful for optimization over mixed spaces. In the spirit of `optimize_acqf_list`, this PR adds the option for `fixed_features_list` by using `optimize_acqf_mixed` instead of `optimize_acqf` in `optimize_acqf_homotopy`, when `fixed_features_list` is provided.

Currently, it is not working as `optimize_acqf_mixed` has no option `return_best_only=False` like `optimize_acqf`. The easiest way to solve it (maybe not the best), would be to provide an option to return the candidates from all restarts for the finally picked fixed features combination in `optimize_acqf_mixed`. For the start, it would be sufficient to only implement it for `q=1`, and error out for the case of `q>1`.

What do you think?

Yes.

## Test Plan

Unit tests.


botorch/optim/optimize_homotopy.py
```diff
@@ -155,13 +160,17 @@ def optimize_acqf_homotopy(
         ic_gen_kwargs: Additional keyword arguments passed to function specified by
             `ic_generator`
     """
+    if fixed_features and fixed_features_list:
+        raise ValueError(
+            "Èither `fixed_feature` or `fixed_features_list` can be provided, not both."
+        )
```
Balandat, line 166: `fixed_features` as an arg seems redundant if we have `fixed_features_list`. We should probably deprecate the  `fixed_features` arg and raise a warning when passed. And if `fixed_features_list` contains only one element we can just use that with `optimize_acqf`

test/optim/test_homotopy.py
```diff
@@ -143,6 +143,38 @@ def test_optimize_acqf_homotopy(self):
...
+        # test fixed feature list
+        fixed_features_list = [{0: 1.0}]
+        model = GenericDeterministicModel(
+            f=lambda x: 5 - (x - p).sum(dim=-1, keepdims=True) ** 2
+        )
+        acqf = PosteriorMean(model=model)
+        # test raise error when fixed_features and fixed_features_list are both provided
+        with self.assertRaisesRegex(
+            ValueError,
+            "Èither `fixed_feature` or `fixed_features_list` can be provided, not both.",
```
Balandat, line 155:
```suggestion
            "Either `fixed_feature` or `fixed_features_list` can be provided, not both.",
```

Balandat:
> Currently, it is not working as optimize_acqf_mixed has no option return_best_only=False like optimize_acqf. The easiest way to solve it (maybe not the best), would be to provide an option to return the candidates from all restarts for the finally picked fixed features combination in optimize_acqf_mixed. For the start, it would be sufficient to only implement it for q=1, and error out for the case of q>1. What do you think?

I think that makes sense. Starting with only supporting the `q=1` case seems very reasonable (let's add the idea above in the comments though where we raise that `UnsupportedError`).

jduerholt:
I added the possibility `return_best_only=False` for `q=1` in `optimize_acqf_mixed` and adapted the tests. Also `optimize_acqf_homotopy` works now with a fixed_features_list. For this, I had to remove the following options from `optimize_acqf_homotopy` as they are not available for `optimize_acqf_mixed`:
- `gen_candidates`: would it make sense for you to add this option to `optimize_acqf_mixed`?
- `sequential`: this keyword make for me no sense in the kontext of `optimize_homotopy`, as we are always using `q=1` for the calls to `optimize_acqf`
 - `return_full_tree`: I am not sure about this as I do not exactly know what it does. Does it make sense in kontext of `optimize_homotopy`?

What do you think?

botorch/optim/optimize_homotopy.py
```diff
@@ -155,19 +160,23 @@ def optimize_acqf_homotopy(
...
         "raw_samples": raw_samples,
         "inequality_constraints": inequality_constraints,
         "equality_constraints": equality_constraints,
         "nonlinear_inequality_constraints": nonlinear_inequality_constraints,
-        "fixed_features": fixed_features,
         "return_best_only": False,  # False to make n_restarts persist through homotopy.
-        "gen_candidates": gen_candidates,
-        "sequential": sequential,
+        # "gen_candidates": gen_candidates,
+        # "sequential": sequential, this is not needed as we are always using q=1 here.
```
Balandat, line 176: yup, makes sense

botorch/optim/optimize_homotopy.py
Balandat: yes, that would be great

botorch/optim/optimize_homotopy.py
```diff
@@ -155,19 +160,23 @@ def optimize_acqf_homotopy(
...
         "ic_generator": ic_generator,
         "timeout_sec": timeout_sec,
-        "return_full_tree": return_full_tree,
+        # "return_full_tree": return_full_tree, is this really needed here
```
Balandat, line 179: No we can probably just remove this from `optimize_acqf_homotopy`

botorch/optim/optimize_homotopy.py
```diff
@@ -199,14 +220,26 @@ def optimize_acqf_homotopy(
...
+        else:
+            candidates, acq_values = optimize_acqf(
+                acq_function=acq_function,
+                bounds=bounds,
+                q=1,
+                options=final_options,
+                batch_initial_conditions=candidates,
+                fixed_features=fixed_features,
+                **shared_optimize_acqf_kwargs,
+            )
```
Balandat, line 242: You can make this more concise  - you should define `optimization_fn` and `fixed_feature_kwargs` above before the `for _ in range(q):`  block so you can reuse this below as well (but GH doesn't let me make that code suggestion). 
```suggestion
        if fixed_features_list:
            optimization_fn = optimize_acqf_mixed
            fixed_feature_kwargs = {"fixed_features_list": fixed_features_list}
        else:
            optimization_fn = optimize_acqf
            fixed_feature_kwargs = {"fixed_features": fixed_features}
        
        candidates, acq_values = optimization_fn(
                acq_function=acq_function,
                bounds=bounds,
                q=1,
                options=final_options,
                batch_initial_conditions=candidates,
                **fixed_feature_kwargs,
                **shared_optimize_acqf_kwargs,
            )
```

Balandat (commented):
> gen_candidates: would it make sense for you to add this option to optimize_acqf_mixed?

Yeah I think that would make sense. Inside of `optimize_acqf_mixed` we'd just pass this to `optimize_acqf`.

> sequential: this keyword make for me no sense in the kontext of optimize_homotopy, as we are always using q=1 for the calls to optimize_acqf

True - I think we can remove this from `optimize_acqf_homotopy`.

> return_full_tree: I am not sure about this as I do not exactly know what it does. Does it make sense in kontext of optimize_homotopy?

This is used by lookahead acquisition functions. I don't think we need to support this for `optimize_acqf_homotopy`.

botorch/optim/optimize.py
```diff
@@ -956,15 +971,17 @@ def optimize_acqf_mixed(
                     post_processing_func=post_processing_func,
                     batch_initial_conditions=batch_initial_conditions,
                     ic_generator=ic_generator,
-                    return_best_only=True,
+                    return_best_only=False,
+                    timeout_sec=timeout_sec,
```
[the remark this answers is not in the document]
Balandat: Did you? I don't see this change.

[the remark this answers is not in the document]
jduerholt: Yeah, I only did it locally. It is kind of a todolist for me here, with this comment it was for me "abgehakt" :D Will push all at once later ;)

botorch/optim/optimize.py
```diff
@@ -956,15 +978,20 @@ def optimize_acqf_mixed(
                     post_processing_func=post_processing_func,
                     batch_initial_conditions=batch_initial_conditions,
                     ic_generator=ic_generator,
-                    return_best_only=True,
+                    return_best_only=False,
+                    gen_candidates=gen_candidates,
+                    timeout_sec=timeout_sec / len(fixed_features_list)
+                    if timeout_sec
+                    else None,
```
Balandat, line 985: Can you move this before the `for` loop? Don't need to redo this each iteration and it will also make things more readable.

botorch/optim/optimize_homotopy.py
```diff
@@ -199,14 +209,16 @@ def optimize_acqf_homotopy(
...
         # Optimize one more time with the final options
-        candidates, acq_values = optimize_acqf(
-            acq_function=acq_function,
-            bounds=bounds,
-            q=1,
-            options=final_options,
-            batch_initial_conditions=candidates,
-            **shared_optimize_acqf_kwargs,
-        )
+        if fixed_features_list:
```
Balandat, line 212: I don't think you want to make things conditional on this - we did this even before supporting `fixed_features_list`

test/optim/test_homotopy.py
```diff
@@ -143,6 +143,38 @@ def test_optimize_acqf_homotopy(self):
...
+                homotopy=Homotopy(homotopy_parameters=[hp]),
+                num_restarts=2,
+                raw_samples=16,
+                fixed_features_list=fixed_features_list,
+                fixed_features=fixed_features,
+            )
+        candidate, acqf_val = optimize_acqf_homotopy(
+            q=1,
+            acq_function=acqf,
+            bounds=torch.tensor([[-10, -10], [5, 5]]).to(**tkwargs),
```
Balandat, line 170:
```suggestion
            bounds=torch.tensor([[-10, -10], [5, 5]], **tkwargs),
```

test/optim/test_homotopy.py
```diff
@@ -143,6 +143,38 @@ def test_optimize_acqf_homotopy(self):
...
+        acqf = PosteriorMean(model=model)
+        # test raise error when fixed_features and fixed_features_list are both provided
+        with self.assertRaisesRegex(
+            ValueError,
+            "Either `fixed_feature` or `fixed_features_list` can be provided, not both.",
+        ):
+            optimize_acqf_homotopy(
+                q=1,
+                acq_function=acqf,
+                bounds=torch.tensor([[-10, -10], [5, 5]]).to(**tkwargs),
```
Balandat, line 160:
```suggestion
                bounds=torch.tensor([[-10, -10], [5, 5]], **tkwargs),
```

test/optim/test_optimize.py
```diff
@@ -1459,13 +1461,28 @@ def test_optimize_acqf_mixed_q1(self, mock_optimize_acqf):
...
+            ).to(acq_val_rvs[0])
+            best_batch_idx = torch.argmax(best_acq_values)
+
+            if return_best_only:
+                best_batch_candidates = candidate_rvs[best_batch_idx]
+                best_batch_acq_values = acq_val_rvs[best_batch_idx]
+                best_idx = torch.argmax(best_batch_acq_values)
+                expected_candidates = best_batch_candidates[best_idx]
+                expected_acq_value = best_batch_acq_values[best_idx]
+                assert expected_candidates.dim() == 2
```
Balandat, line 1479:
```suggestion
                self.assertEqual(expected_candidates.dim(), 2)
```

test/optim/test_optimize.py
```diff
@@ -1459,13 +1461,28 @@ def test_optimize_acqf_mixed_q1(self, mock_optimize_acqf):
...
+            else:
+                expected_candidates = candidate_rvs[best_batch_idx]
+                expected_acq_value = acq_val_rvs[best_batch_idx]
+                assert expected_candidates.dim() == 3
+                assert expected_acq_value.dim() == 1
```
Balandat, line 1484:
```suggestion
                self.assertEqual(expected_candidates.dim(), 3)
                self.assertEqual(expected_acq_value.dim(), 1)
```

test/optim/test_optimize.py
```diff
@@ -1562,6 +1597,22 @@ def test_optimize_acqf_mixed_empty_ff(self):
                 raw_samples=10,
             )
 
+    def test_optimize_acqf_mixed_return_best_only_q2(self):
+        with self.assertRaises(
+            NotImplementedError,
+            msg="`return_best_only=False` is only supported for q=1.",
+        ):
+            mock_acq_function = MockAcquisitionFunction()
```
Balandat, line 1605: nit: isolate the expected Exception in the context manager
```suggestion
        mock_acq_function = MockAcquisitionFunction()
        with self.assertRaises(
            NotImplementedError,
            msg="`return_best_only=False` is only supported for q=1.",
        ):
```

test/optim/test_optimize.py
```diff
@@ -1520,10 +1539,24 @@ def test_optimize_acqf_mixed_q2(self, mock_optimize_acqf):
...
+                candidate_rvs_q = [
+                    torch.rand(num_restarts, 1, 3, **tkwargs) for _ in range(num_ff)
+                ]
+                acq_val_rvs_q = [
+                    torch.rand(num_restarts, **tkwargs) for _ in range(num_ff)
+                ]
+
+                best_acq_values = torch.tensor(
+                    [torch.max(acq_values) for acq_values in acq_val_rvs_q]
+                ).to(acq_val_rvs_q[0])
```
Balandat, line 1551: Don't think this `to` is needed here
```suggestion
                )
```

test/optim/test_optimize.py
```diff
@@ -1459,13 +1461,28 @@ def test_optimize_acqf_mixed_q1(self, mock_optimize_acqf):
...
                 post_processing_func=rounding_func,
             )
             # compute expected output
-            ff_acq_values = torch.stack(acq_val_rvs)
-            best = torch.argmax(ff_acq_values)
-            expected_candidates = candidate_rvs[best]
-            expected_acq_value = ff_acq_values[best]
+            best_acq_values = torch.tensor(
+                [torch.max(acq_values) for acq_values in acq_val_rvs]
+            ).to(acq_val_rvs[0])
```
Balandat, line 1470: This shouldn't be needed here?
```suggestion
            )
```

Balandat (commented):
Thanks - largely nits except for the conditional when "Optimize one more time with the final options".

Also looks like you have a `KeyError` test failure as well as line length lint issue

botorch/optim/optimize_homotopy.py
jduerholt: Good catch. Thanks! I overlooked the if when I was changing it to `optimization_fn`.

botorch/optim/optimize_homotopy.py
jduerholt: I implemeted it now for `optimize_acqf_homotopy`, if it is fine for you how it looks there, I will port it to `optimize_acqf_list`.

jduerholt:
Hi @Balandat,

thanks for your review(s). I think, I have now implemented all the suggestions.

Could you have a look on the issue that I always have too wait for approval before the pipelines start? In the past, they were starting always automatically which helped me to faster finish the PRs ;) And maybe it is also less annoying for you ...

Best,

Johannes

Balandat:
> Could you have a look on the issue that I always have to wait for approval before the pipelines start? In the past, they were starting always automatically which helped me to faster finish the PRs ;) And maybe it is also less annoying for you ...

Yeah I pinged some folks to understand if we can avoid this, let's see what they say.

Balandat:
Looks like you're missing a unit test for more than 1 elements in the `fixed_features_list` which seems important :)

jduerholt:
> Looks like you're missing a unit test for more than 1 elements in the `fixed_features_list` which seems important :)

Ah, this is now a consequence of the new structure where we call `optimize_acqf` when `fixed_features_list` has only one element. Before it was also in this case going to `òptimize_acqf_mixed`. I will add a new test case.

jduerholt:
Now, the test is there ;)

Balandat (approved):
Tutorial failure is unrelated (and will be fixed in #2641). Thanks!

botorch/optim/optimize.py
```diff
@@ -956,15 +979,18 @@ def optimize_acqf_mixed(
                     post_processing_func=post_processing_func,
                     batch_initial_conditions=batch_initial_conditions,
                     ic_generator=ic_generator,
-                    return_best_only=True,
+                    return_best_only=False,
```
esantorella, line 982: Why not `return_best_only=return_best_only`?

botorch/optim/optimize.py
```diff
@@ -1003,8 +1038,12 @@ def optimize_acqf_mixed(
             nonlinear_inequality_constraints=nonlinear_inequality_constraints,
             post_processing_func=post_processing_func,
             batch_initial_conditions=batch_initial_conditions,
+            gen_candidates=gen_candidates,
             ic_generator=ic_generator,
             ic_gen_kwargs=ic_gen_kwargs,
+            timeout_sec=timeout_sec,
+            retry_on_optimization_warning=retry_on_optimization_warning,
+            return_best_only=True,
```
esantorella, line 1046: Why True here and False earlier?

test/optim/test_homotopy.py
```diff
@@ -132,14 +132,62 @@ def test_optimize_acqf_homotopy(self):
             f=lambda x: 5 - (x - p).sum(dim=-1, keepdims=True) ** 2
         )
         acqf = PosteriorMean(model=model)
+        # test raise warning on using `fixed_features` argument
+        message = (
+            "The `fixed_features` argument is deprecated, "
+            "use `fixed_features_list` instead."
+        )
+        with self.assertWarnsRegex(DeprecationWarning, message):
```
esantorella, line 140: Thanks for the tests!

test/optim/test_optimize.py
```diff
@@ -1551,7 +1586,9 @@ def test_optimize_acqf_mixed_q2(self, mock_optimize_acqf):
             self.assertTrue(torch.equal(acq_value, expected_acq_value))
 
     def test_optimize_acqf_mixed_empty_ff(self):
-        with self.assertRaises(ValueError):
+        with self.assertRaises(
```
esantorella, line 1589:
```suggestion
        with self.assertRaisesRegex(
```
`assertRaises` with `msg` doesn't actually check message equality -- the msg keyword is for the error message for a failure. See here: https://docs.python.org/3/library/unittest.html#unittest.TestCase.debug

test/optim/test_optimize.py
```diff
@@ -1551,7 +1586,9 @@ def test_optimize_acqf_mixed_q2(self, mock_optimize_acqf):
...
+            ValueError, msg="fixed_features_list must be non-empty."
```
esantorella, line 1590:
```suggestion
            ValueError, "fixed_features_list must be non-empty."
```

test/optim/test_optimize.py
```diff
@@ -1562,6 +1599,22 @@ def test_optimize_acqf_mixed_empty_ff(self):
                 raw_samples=10,
             )
 
+    def test_optimize_acqf_mixed_return_best_only_q2(self):
+        mock_acq_function = MockAcquisitionFunction()
+        with self.assertRaises(
+            NotImplementedError,
+            msg="`return_best_only=False` is only supported for q=1.",
```
esantorella, line 1606:
```suggestion
        with self.assertRaisesRegex(
            NotImplementedError,
            regex="`return_best_only=False` is only supported for q=1.",
```

botorch/optim/optimize.py
jduerholt: We get here all candidates and then filter later based on the `return_best_only` argument. I will add a comment.

botorch/optim/optimize.py
jduerholt: Because, currently `return_best_only` is only implemented for `q=1` and here we are in the loop for `q>1`.

jduerholt:
Thanks for the review @esantorella. I adressed your comments and resolved the merge conflicts with main. Best, Johannes

PR #2641, opened 2024-12-02
Fix custom_botorch_model_in_ax tutorial

saitcakmak:
Updates the tutorial to pass in `input_transform_classes=None` (to disable default transforms), since the custom model does not support transforms.


Balandat (approved):
thanks

PR #2651, opened 2024-12-12
Stop redirecting for the archived docs

zpao:
We won't need it. This will prevent a separate archive site from working.

## Test Plan

Generated static version and served locally. Redirected previously. Doesn't anymore.


Balandat:
@CristianLara I assume you have the necessary context on this?

PR #2653, opened 2024-12-16
Upgrade Website

CristianLara:
## Motivation

We are already using most of the correct tooling but it has become very out of date and is tied together with scripts that are difficult to maintain. Upgrade our tooling by updating to latest versions and dramatically simplifying our setup/deploy scripts.

Includes:
- Upgrade Docusaurus from V1 to V3
- Utilize new MDX support in Docusaurus by converting our tutorials to MDX (previously converted to html)
- Tutorials link to Colab for easy execution
- Website versioning is now handled natively through Docusaurus
- New website versions are only created for major and minor releases (not patches)
- Build and Deploy workflows have been dramatically simplified
- Change Sphinx theme to ReadTheDocs
- API Reference is still built using Sphinx but is now hosted by ReadTheDocs. ReadTheDocs also handles versioning.
- Legacy website will be frozen and available at [archive.botorch.org](archive.botorch.org)
    - Served by Vercel from the `archive` branch
- Website now supports dark mode

| Before    | After |
| -------- | ------- |
|        image         |      image      |
| image | image  |
| image | image |
| image | image |

The commits for most of the major changes are grouped into themed PRs on my fork if that makes the review easier: https://github.com/CristianLara/botorch/pulls?q=is%3Apr+is%3Aclosed

## Considerations
- Our tutorials are natively embedded and versioned in the website using Docusaurus, but our sphinx-generated API reference is not. ReadTheDocs will create new version of the sphinx docs for every new release tag in our github repo.
- The main branch will now contain copies of the website documentation for every version. This includes the converted tutorial MDX but not the original ipynb files, those are accessible linked to from github using release tags.
- Continue using same integrations for Algolia Search and Google Analytics
- The number and type of github workflows we have remain the same, but the complexity of the workflows and the supporting scripts has been simplified.

## Test Plan

- I've enabled all the workflows in my PR and everything is passing [https://github.com/CristianLara/botorch/actions](https://github.com/CristianLara/botorch/actions)
    - Pypy deploy and codecov are failing as expected due to missing credentials
- Website for my fork is being actively built and served at [cristianlara.me/botorch](cristianlara.me/botorch)
- API Reference is being built and served by ReadTheDocs at [https://botorch-dev.readthedocs.io/](https://botorch-dev.readthedocs.io/)


.github/workflows/publish_website.yml
```diff
@@ -0,0 +1,62 @@
...
+      new_version:
+        required: false
+        type: string
+      run_tutorials:
+        required: false
+        type: boolean
+        default: false
+  workflow_dispatch:
+  push:
+    branches: [ main ]
```
saitcakmak, line 15: Do we want to publish the website on each push to the main branch? Leftover from testing?

.github/workflows/publish_website.yml
```diff
@@ -0,0 +1,62 @@
...
+        git config --global user.email "41898282+github-actions[bot]@users.noreply.github.com"
+
+        python3 scripts/convert_ipynb_to_mdx.py --clean
+        cd website
+        yarn
+        yarn docusaurus docs:version ${{ inputs.new_version }}
+
+        git add --all
+        git commit -m "Create version ${{ inputs.new_version }} of site in Docusaurus"
+        git push --force origin HEAD:main
```
saitcakmak, line 54: Where is this bot pushing to? main branch? If so, that'll get blocked

.github/workflows/publish_website.yml
```diff
@@ -0,0 +1,62 @@
...
+        ALLOW_BOTORCH_LATEST: true  # Allow Ax to install w/ new BoTorch release.
+      run: |
+        pip install ."[dev, tutorials]"
+        # There may not be a compatible Ax pip version, so we use the development version.
+        pip install git+https://github.com/facebook/Ax.git
+    - if: ${{ inputs.new_version }}
+      name: Create new docusaurus version
+      run: |
+        git config --global user.name "github-actions[bot]"
+        git config --global user.email "41898282+github-actions[bot]@users.noreply.github.com"
```
saitcakmak, line 45: Does this bot require access to some secrets? How does it get permissions?

.github/workflows/publish_website.yml
```diff
@@ -0,0 +1,62 @@
...
+        git commit -m "Create version ${{ inputs.new_version }} of site in Docusaurus"
+        git push --force origin HEAD:main
+    - name: Build website
+      run: |
+        bash scripts/build_docs.sh -b
+    - name: Deploy
+      uses: JamesIves/github-pages-deploy-action@releases/v4
+      with:
+          branch: gh-pages # The branch the action should deploy to.
+          folder: website/build # The folder the action should deploy.
```
saitcakmak, line 62: Is a third party action necessary for deployment? Is this to avoid directly pushing to the gh-pages branch (though we seem to be pushing somewhere above)?

.readthedocs.yaml
```diff
@@ -0,0 +1,25 @@
...
+          pip install git+https://github.com/cornellius-gp/linear_operator.git
+          pip install git+https://github.com/cornellius-gp/gpytorch.git
+        fi
+
+python:
+  install:
+    - method: pip
+      path: .
+      extra_requirements:
+        - dev
```
saitcakmak, line 22: Does this need the allow lastest gpytorch env var?

scripts/convert_ipynb_to_mdx.py
```diff
@@ -0,0 +1,1080 @@
...
+import json
+import os
+import re
+import shutil
+import subprocess
+import uuid
+from pathlib import Path
+from typing import Any, Dict, List, Optional, Tuple, Union
+
+import mdformat  # @manual=fbsource//third-party/pypi/mdformat:mdformat
```
saitcakmak, line 17: This seems misplaced in OSS code. You can probably add it directly to the targets file
```suggestion
import mdformat
```

scripts/convert_ipynb_to_mdx.py
```diff
@@ -0,0 +1,1080 @@
...
+    print("--------------------------------------------")
+    if args.clean:
+        clean_up_directories()
+    for metadata in tutorials_metadata:
+        path = (
+            LIB_DIR / "tutorials" / metadata["id"] / (metadata["id"] + ".ipynb")
+        ).resolve()
+        print(f"{path.stem}")
+        mdx = transform_notebook(path, metadata)
+    print("")
```
saitcakmak, line 1080: I guess we still need 1000 lines of custom conversion code for the notebooks :( We can't use a third party tool like nbconvert to avoid the need for maintaining our own scripts for this?

tutorials/GIBBON_for_efficient_batch_entropy_search/GIBBON_for_efficient_batch_entropy_search.ipynb
saitcakmak, line 1: Does each tutorial need its own directory now? I don't like needlessly deep directory structures.

saitcakmak (changes requested):

.github/workflows/publish_website.yml
CristianLara: You're right this is leftover from testing. Removing

.github/workflows/publish_website.yml
CristianLara: Correct it pushes to the main branch... and you're also right that the GH bot respects branch protection rules. Thanks for pointing this out, I didn't have branch protections in my fork.

There's a discussion about this with some workarounds here: https://github.com/orgs/community/discussions/25305#discussioncomment-10728028

.github/workflows/publish_website.yml
CristianLara: This uses the `GITHUB_TOKEN` provided directly by the github actions workflow: https://docs.github.com/en/actions/security-for-github-actions/security-guides/automatic-token-authentication#about-the-github_token-secret

The granted permissions are configurable with the defaults listed here: https://docs.github.com/en/actions/security-for-github-actions/security-guides/automatic-token-authentication#permissions-for-the-github_token

CristianLara: Depending on the solution I choose for https://github.com/pytorch/botorch/pull/2653?new_mergebox=true#discussion_r1887693194 I may use a different approach for auth though

CristianLara: Using the GH actions bot in this way for creating a commit is an example provided by the official `actions/checkout` documentation: https://github.com/actions/checkout?tab=readme-ov-file#push-a-commit-to-a-pr-using-the-built-in-token

.github/workflows/publish_website.yml
CristianLara: No it's not necessary. There's actually a more official one that I can switch to: https://github.com/actions/deploy-pages

.readthedocs.yaml
CristianLara: I don't think so. The order of operations here is different from how we usually install deps because ReadTheDocs doesn't have a way for us to conditionally set an environment variable. There's a `pre_install` step available to us but environment variables set there don't actually persist to the installation step.

My solution here was to install pinned deps first then conditionally install the latest versions as a `post_install` step.

scripts/convert_ipynb_to_mdx.py
CristianLara: Docusaurus only natively accepts MD and MDX for docs. There's a hacky way to force html into the page and that's what we were previously doing but it's much less flexible and doesn't get picked up by docusaurus's sidebar or versioning system.

There _is_ an MD output option in nbconvert but we would lose the interactive plots and html/react features of MDX. It might be possible to shave down some of this script by pre-processing using nbconvert but I think the loss of ipynb metadata might actually make some parts more difficult to parse/convert.

This file is mostly a copy of a script written by the BeanMachine team so there may be pieces of it that are not relevant to our tutorials and can trim. Here's the original: https://github.com/facebookresearch/beanmachine/blob/main/website/scripts/convert_ipynb_to_mdx.py

And here's the PR where I originally added support for it in the Ax repo first with the most context: https://github.com/CristianLara/Ax/pull/13

CristianLara: This is definitely the most complex part of the new website and I also would prefer to avoid it if there is a better alternative.

Balandat: > This file is mostly a copy of a script written by the BeanMachine team so there may be pieces of it that are not relevant to our tutorials and can trim

Yeah at the very least we'd want to cut this down to only the parts that are essential to our setup.

.github/workflows/publish_website.yml
saitcakmak: Do we need to be pushing to the main branch to publish the website? Previously, we would've simply pushed the updated website to gh-pages branch, which would automatically trigger the deploy action there. Would changing this to push to gh-pages work here?

.github/workflows/publish_website.yml
saitcakmak: This is the same action that's triggered when we push to gh-pages branch, right? If so, that'd be my preferred solution.

tutorials/GIBBON_for_efficient_batch_entropy_search/GIBBON_for_efficient_batch_entropy_search.ipynb
saitcakmak: Makes sense. I guess it's a fair price to pay for consistency between the two setups.

.github/workflows/publish_website.yml
CristianLara: The blessed Docusaurus way of versioning is to make versioned copies of the docs directly in the main branch. Running `yarn build` then builds all the versions together such that all versions are properly aware of each other.

Previously we were handling our own versioning by always building a single version of the website (two actually if latest _and_ vX.Y.Z for the different path handling), merging with the previously built versions in gh-pages, and performing quite a bit of surgery so that previous versions can point to the new one.

I was initially concerned that storing all versions in the main branch would result in a lot of bloat and disk space, but I don't think this will be an issue because:
1. We aren't storing duplicate ipynb files. We're essentially converting them to markdown files which are very small. We still provide download links to the ipynb files by linking directly to the tagged/release version stored in github history.
2. Going forward we will no longer create new versions of the website for patch releases, only major and minor releases.

CristianLara: For some concrete numbers: in commit f13759a0b I have two test versions of the botorch website and each copy of the versioned docs is around 5MB

```shell
(venv) ➜  website git:(f13759a0b) du -hs versioned_docs/*                          git:(f13759a0b|⚑3) 
4.9M    versioned_docs/version-v0.13.0
4.9M    versioned_docs/version-v0.14.0

(venv) ➜  website git:(f13759a0b) du -hs versioned_sidebars/*                      git:(f13759a0b|⚑3) 
8.0K    versioned_sidebars/version-v0.13.0-sidebars.json
8.0K    versioned_sidebars/version-v0.14.0-sidebars.json
```

.github/workflows/publish_website.yml
CristianLara: Yes exactly, when we push to the gh-pages branch github will build an artifact from that repo and deploy it using the https://github.com/actions/deploy-pages action

Here's a recent gh-pages deploy in this repo that shows this action being used: https://github.com/pytorch/botorch/actions/runs/12371594841/job/34528079206

.github/workflows/publish_website.yml
saitcakmak: I see several issues with this.

- We can't have a bot pushing to the main branch. It'd have to create PRs that are imported and landed in fbcode. 
- Even if we're not storing duplicate ipynb files, the commit history will store them each time they're modified. This is not as much of an issue for BoTorch but it's an issue for Ax since the plotly plots are a lot larger than the static plots used in BoTorch tutorials.
- Are we going to stop publishing latest / nightly website versions? Those will add up fast in the commit history.

CristianLara: > Are we going to stop publishing latest / nightly website versions? Those will add up fast in the commit history.

The gh workflows will only create commit on `main` when we release a new version / create a release tag. 

latest / nightly deploys will still run which build the existing versions + latest version of the docs. This build is currently deployed by making a new commit in the `gh-pages` branch but [I'm changing that logic to use the upload-artifact + deploy-pages actions directly](https://github.com/pytorch/botorch/pull/2653?new_mergebox=true#discussion_r1888855698) which will result in no gh-pages commit.

> Even if we're not storing duplicate ipynb files, the commit history will store them each time they're modified. This is not as much of an issue for BoTorch but it's an issue for Ax since the plotly plots are a lot larger than the static plots used in BoTorch tutorials.

I'm not following. The new website links to the pre-existing tagged versions of the ipynb files in the `main` branch. I'm not adding any additional commits anywhere to track the ipynb files and I don't plan on erasing/squashing git history for the `main` branch. Here's an example of what the link I'm generating looks like: https://github.com/pytorch/botorch/blob/v0.12.0/tutorials/Multi_objective_multi_fidelity_BO.ipynb

> We can't have a bot pushing to the main branch. It'd have to create PRs that are imported and landed in fbcode.

Hmm I could use more context on this. Are you saying that merging a PR to `main` in OSS without importing to fbcode is either not possible or will cause problems?

.github/workflows/publish_website.yml
CristianLara: Removed in [3e6d0b6](https://github.com/pytorch/botorch/pull/2653/commits/3e6d0b6ca6cd31e737bd25d0e0a1771f4f583474)

.github/workflows/publish_website.yml
CristianLara: Implemented in [99196eb](https://github.com/pytorch/botorch/pull/2653/commits/99196eb7629d52b886073f207a4f5489e0dd37b8). See the PR for test plan details: https://github.com/CristianLara/botorch/pull/10.

tl;dr We will now deploy to github pages directly via the official github actions by directly deploying the build output. The `gh-pages` branch is no longer used at all and can be deleted.

CristianLara: I had to modify my fork's GH Pages settings to build from github actions instead of build from branch, will need to make the same settings change in this main repo after merging.

.github/workflows/publish_website.yml
CristianLara: > latest / nightly deploys will still run which build the existing versions + latest version of the docs. This build is currently deployed by making a new commit in the gh-pages branch but https://github.com/pytorch/botorch/pull/2653?new_mergebox=true#discussion_r1888855698 which will result in no gh-pages commit.

This has been implemented. See https://github.com/pytorch/botorch/pull/2653#discussion_r1889058756

Balandat: > Are you saying that merging a PR to main in OSS without importing to fbcode is either not possible or will cause problems?

Yes, I believe this will cause issues with ShipIt since the repos won't be in sync. Unless there is a way to exclude some of the files/directories in the GitHub repo, but I'm not sure that's possible.

CristianLara: Okay I've read through the ShipIt documentation and understand the issue now.

Something we can do is have a branch `docusaurus-versions` that is only for keeping track of docusaurus versions. It would be a fork of `main` but include autogenerated docusaurus versioned files. The behavior would be:
1. We always build the website from the `docusaurus-versions` branch
2. Before building the website we always sync `docusaurus-versions` with `main`
3. For new releases our github workflow will create the new docusaurus version in the `docusaurus-versions` branch

The `main` branch would then only be responsible for tracking the latest documentation, and our github workflows can continue to automate the release process from the `docusaurus-versions` branch without touching `main`.

I'll make a PR in my fork to test this out.

CristianLara: I've implemented my proposal above. There is now a release branch `docusaurus-versions` in my fork which is used by our workflows to deploy the website and keep track of docusaurus versions. If this approach is accepted I'll create this branch for this repo before merging.

47a42f2f915b531a2d78be378be923dc2ccd33a1

CristianLara: Here's an example of the updated workflow running to deploy a test version in my Ax fork using the release branch: https://github.com/CristianLara/Ax/actions/runs/12397040263/job/34606475835

saitcakmak: Using the bot to commit to a separate `docusaurus-versions` branch seems reasonable to me.

Re the file size issues: I think I mixed together parts of the old setup with the new setup in my head. If we're only adding a commit for the new website versions but not for the nightly builds, this shouldn't be an issue.

.github/workflows/publish_website.yml
```diff
@@ -0,0 +1,80 @@
...
+    - name: Sync release branch with main
+      run: |
+        git config --global user.name "github-actions[bot]"
+        git config --global user.email "41898282+github-actions[bot]@users.noreply.github.com"
+        git merge origin/main
+        # To avoid a large number of commits we don't push this sync commit to github until a new release.
+    - name: Set up Python
+      uses: actions/setup-python@v5
+      with:
+        python-version: "3.10"
```
Balandat, line 34: why only 3.10?

.readthedocs.yaml
```diff
@@ -0,0 +1,25 @@
+version: "2"
+
+build:
+  os: "ubuntu-22.04"
+  tools:
+    python: "3.10"
+  jobs:
+    post_install:
+      # Install latest botorch if not on a released version
```
Balandat, line 9: should we set `ALLOW_LATEST_GPYTORCH_LINOP = true` here as well?

.github/workflows/deploy_on_release.yml
```diff
@@ -44,11 +44,40 @@ jobs:
...
+    runs-on: ubuntu-latest
+    outputs:
+      major_minor_changed: ${{ steps.compare.outputs.major_minor_changed }}
+    steps:
+    - uses: actions/checkout@v4
+      with:
+        fetch-depth: 0
+        fetch-tags: true
+        ref: ${{ github.sha }}
+    - name: Check if major or minor version changed
```
Balandat, line 59: When would that not be the case? This workflow is executed upon creating a GitHub release, so this should involve updating at least the minor version? Or is the idea that there may be some post-release update bugfix or the like?

.github/workflows/publish_website.yml
saitcakmak: We need to pick one version to build the website with. Do you prefer a different version, like 3.12?

.github/workflows/deploy_on_release.yml
saitcakmak: There is also the patch versions. v<major>.<minor>.<patch> The idea is to not release website for the patch versions (primarily for Ax, since we'll follow semver there after 1.0), since these are not intended to include any new features, but only minimal bug / documentation fixes.

.github/workflows/publish_website.yml
Balandat: yeah, might as well use a newer version, o/w we'd have to update it soon anyway.

CristianLara: This came from the existing actions which all use 3.10. I've updated this action and all the others to use 3.12 instead: 931505207eb993eda5a684d277f721225e5e0d60

.readthedocs.yaml
CristianLara: Sait had the same question 🙂  answered here: https://github.com/pytorch/botorch/pull/2653#discussion_r1887703528

The gist is that readthedocs does not have a way for us to programmatically set an env variable to be consumed by the install step in their CI. My solution here was to install pinned deps first then conditionally install the latest versions as a `post_install` step.

scripts/convert_ipynb_to_mdx.py
```diff
@@ -0,0 +1,928 @@
...
+from nbformat.notebooknode import NotebookNode
+
+SCRIPTS_DIR = Path(__file__).parent.resolve()
+LIB_DIR = SCRIPTS_DIR.parent.resolve()
+
+WEBSITE_DIR = LIB_DIR.joinpath("website")
+DOCS_DIR = LIB_DIR.joinpath("docs")
+TUTORIALS_DIR = DOCS_DIR.joinpath("tutorials")
+
+ORGANIZATION = "cristianlara"
```
CristianLara, line 29: Reminder to change this  to `ORGANIZATION = "pytorch"` before merging

CristianLara: [ad1a4f1](https://github.com/pytorch/botorch/pull/2653/commits/ad1a4f12c7bfd7781e95735b75621f3e6f592670)

saitcakmak (approved):
Looks like we're ready to test in prod!

saitcakmak (approved):

PR #2660, opened 2024-12-24
Use expm1 instead of exp - 1

kit1980:
Similarly to https://github.com/pytorch/botorch/pull/2540 and https://github.com/pytorch/botorch/pull/2541: use more numerically stable `expm1`.

Found with TorchFix https://github.com/pytorch-labs/torchfix/.  This should be the last instance to replace in the current codebase.


saitcakmak (approved):
Thanks! Happy to stamp the diff after you import it to fbsource

Issue #2684, opened 2025-01-18
[FEATURE REQUEST]: Implementing Region Averaged Acquisition Functions


SaiAakash:
### Motivation

A very recent paper published in AAAI conference on AI (2025) talks about region-averaged acqusition functions that can perform efficient trust region selection for high-dimensional and multi-modal Bayesian Optimization problems.

Paper: [Regional Expected Improvement for Efficient Trust Region Selection in High-Dimensional Bayesian Optimization](https://arxiv.org/abs/2412.11456)

Code: [Regional-Expected-Improvement](https://github.com/nobuo-namura/regional-expected-improvement) 

This improves upon the existing `TuRBO` method by picking trust regions efficiently during initializations and during local optimization restarts using region-averaged acquisition functions instead of randomly sampling trust regions. The code already uses BoTorch infrastructure to construct the `REI`, `qREI`, `LogREI` and `qLogREI` acqusition functions. This might be a nice feature to add to the optimization suite present in BoTorch.

Any comments on the feasibility of integration, suggestions are welcome ! 

### Describe the solution you'd like to see implemented in BoTorch.

Probably a new script in the `acquisition` folder that implements all 4 of the region-averaged acquisition functions directly. The usefulness can be demonstrated through another tutorial for TuRBO with efficient trust region selection.

### Describe any alternatives you've considered to the above solution.

NA

### Is this related to an existing issue in BoTorch or another repository? If so please include links to those Issues here.

NA

### Pull Request

I am willing to open a PR if everyone thinks this is something worth adding to BoTorch !

### Code of Conduct

- [x] I agree to follow BoTorch's Code of Conduct

Balandat:
I've only skimmed the paper, but this seems simple enough that including it would be reasonable assuming this actually provides meaningful performance improvements. @dme65, @saitcakmak what do you think about the approach?

We could also first include this in botorch_community if we're not sure about the performance value-add.

saitcakmak:
I'd be open to adding them to botorch_community. If we see meaningful improvements, we can then promote one of the methods (I'd guess this would be qLogREI, due to log related improvements and generality that comes from q-batch support) to core botorch.

Balandat:
@SaiAakash yes that'd be great, thanks!

SaiAakash:
Closing this in favor of #2699.

PR #2699, opened 2025-01-25
Add Region Averaged Acquisition Functions

SaiAakash:
## Motivation

This PR addresses #2684. `LogREI` and `qLogREI` have been implemented in the `botorch_community` folder. 

Yes

## Test Plan

Test for both the acquisition functions have been added in `test_community/acquisition/test_rei.py`.

## Related PRs

NA


esantorella (changes requested):
Thanks for getting this in! A couple things:

* Could you add a module-level docstring to rei.py giving a broad overview of what acquisition functions are in this file and what they do, with pointers to the relevant academic literature? And could you also add docstrings to the acquisition functions?
* Some usage examples in the docstrings would be particularly helpful, especially regarding  `X_dev`. How should one choose those points? Can this be used in a similar way to other acquisition functions, or do the trust regions make things very different?
* If possible, it might be nice to have an input constructor in `botorch_community/acquisition/input_constructors.py`  (following the pattern of the constructors in botorch/acquisition/input_constructors.py`), but I don't think we are requiring that.

botorch_community/acquisition/rei.py
```diff
@@ -0,0 +1,171 @@
...
+
+    def __init__(
+        self,
+        model: Model,
+        best_f: Union[float, Tensor],
+        X_dev: Union[float, Tensor],
+        posterior_transform: Optional[PosteriorTransform] = None,
+        maximize: bool = True,
+        length: float = 0.8,
+        bounds: Optional[Union[float, Tensor]] = None,
```
esantorella, line 34: Nit: PEP 604

```suggestion
        best_f: float | Tensor,
        X_dev: float | Tensor,
        posterior_transform: PosteriorTransform | None = None,
        maximize: bool = True,
        length: float = 0.8,
        bounds: float | Tensor | None = None,
```

botorch_community/acquisition/rei.py
```diff
@@ -0,0 +1,171 @@
...
+    ):
+        r"""Log-Regional Expected Improvement (analytic).
+
+        Args:
+            model: A fitted single-outcome model.
```
esantorella, line 39: Thanks for documenting all these arguments!

botorch_community/acquisition/rei.py
```diff
@@ -0,0 +1,171 @@
...
+        """
+
+        super().__init__(model=model, posterior_transform=posterior_transform)
+        self.register_buffer("best_f", torch.as_tensor(best_f))
+        self.maximize = maximize
+
+        dim = X_dev.shape[1]
+        self.n_region = X_dev.shape[0]
+        self.X_dev = X_dev.reshape(self.n_region, 1, 1, -1)
+        self.length = length
```
esantorella, line 60:
```suggestion
        self.maximize: bool = maximize

        dim: int = X_dev.shape[1]
        self.n_region: int = X_dev.shape[0]
        self.X_dev: Tensor = X_dev.reshape(self.n_region, 1, 1, -1)
        self.length: float = length
```

botorch_community/acquisition/rei.py
```diff
@@ -0,0 +1,171 @@
...
+    def __init__(
+        self,
+        model: Model,
+        best_f: Union[float, Tensor],
+        X_dev: Union[float, Tensor],
+        posterior_transform: Optional[PosteriorTransform] = None,
+        maximize: bool = True,
+        length: float = 0.8,
+        bounds: Optional[Union[float, Tensor]] = None,
+    ):
```
esantorella, line 35:
```suggestion
    ) -> None:
```

botorch_community/acquisition/rei.py
```diff
@@ -0,0 +1,171 @@
...
+        X_min = (X - 0.5 * self.length).clamp_min(self.bounds[0]).unsqueeze(0)
+        X_max = (X + 0.5 * self.length).clamp_max(self.bounds[1]).unsqueeze(0)
+        Xs = (self.X_dev * (X_max - X_min) + X_min).reshape(-1, q, d)
+
+        mean, sigma = self._mean_and_sigma(Xs)
+        u = _scaled_improvement(mean, sigma, self.best_f, self.maximize)
+        logei = _log_ei_helper(u) + sigma.log()
+        logrei = torch.log(
+            torch.exp(logei.reshape(self.n_region, batch_shape)).mean(axis=0)
+        )
```
esantorella, line 84: It would probably be more numerically stable to use `botorch.utils.safe_math.logsumexp` (and then subtract a constant to change `log(sum(exp(...)))` to `log(mean(exp(...)))`)

botorch_community/acquisition/rei.py
```diff
@@ -0,0 +1,171 @@
...
+        model: Model,
+        best_f: Union[float, Tensor],
+        X_dev: Union[float, Tensor],
+        sampler: Optional[MCSampler] = None,
+        objective: Optional[MCAcquisitionObjective] = None,
+        posterior_transform: Optional[PosteriorTransform] = None,
+        X_pending: Optional[Tensor] = None,
+        length: float = 0.8,
+        bounds: Optional[Union[float, Tensor]] = None,
+        **kwargs: Any,
```
esantorella, line 101: It looks like these `kwargs` will be ignored; can they be removed?

botorch_community/acquisition/rei.py
```diff
@@ -0,0 +1,171 @@
...
+        Xs = (self.X_dev * (X_max - X_min) + X_min).reshape(-1, q, d)
+
+        posterior = self.model.posterior(
+            X=Xs, posterior_transform=self.posterior_transform
+        )
+        samples = self.get_posterior_samples(posterior)
+        obj = self.objective(samples, X=Xs)
+        obj = (
+            (obj - self.best_f.unsqueeze(-1).to(obj))
+            .clamp_min(0)
```
esantorella, line 167: @SebastianAment , would it make sense to use `botorch.utils.safe_math.fatminimum` or something for differentiability here?

test_community/acquisition/test_rei.py
```diff
@@ -0,0 +1,246 @@
...
+        X_dev = torch.rand(1, 1, device=self.device, dtype=dtype)
+        module = LogRegionalExpectedImprovement(
+            model=mm, best_f=0.0, X_dev=X_dev, maximize=False
+        )
+        X = torch.empty(1, 1, device=self.device, dtype=dtype)
+        log_rei = module(X)
+        rei_expected = torch.tensor([0.6978], device=self.device, dtype=dtype)
+        self.assertAllClose(log_rei, rei_expected.log(), atol=1e-4)
+
+        with self.assertRaises(UnsupportedError):
```
esantorella, line 58: Nit: It's nice to use `assertRaisesRegex` to check the error message

test_community/acquisition/test_rei.py
```diff
@@ -0,0 +1,246 @@
...
+from botorch.sampling.normal import IIDNormalSampler, SobolQMCNormalSampler
+from botorch.utils.testing import BotorchTestCase, MockModel, MockPosterior
+from gpytorch.distributions import MultitaskMultivariateNormal, MultivariateNormal
+
+
+class TestLogRegionalExpectedImprovement(BotorchTestCase):
+    def test_log_regional_expected_improvement(self):
+        for dtype in (torch.float, torch.double):
+            with catch_warnings():
+                simplefilter("ignore", category=NumericsWarning)
```
esantorella, line 32: Thanks for silencing the warnings; it's annoying when unit tests are polluted by tons of numerical warnings. What kind of numerics warnings were coming up? Could you leave a comment? This is nitpicking, but if there's one warning in particular that was an issue, you could use `warnings.filterwarnings` with the `message` argument to isolate that warning.

botorch_community/acquisition/rei.py
SebastianAment: Yes, let's use `logmeanexp` here. Calling `log`, `mean`, and `exp` separately very commonly leads to numerical underflow.

botorch_community/acquisition/rei.py
SebastianAment: Depends on what the goal is: 
- If the goal is to add REI as it was proposed in the [REI paper](https://arxiv.org/pdf/2412.11456v1) ([code](https://github.com/Nobuo-Namura/regional-expected-improvement/blob/main/src/rei.py)), we should keep the hard non-linearities.
- If we seek to make this acquisition function as performant as it could be, you could add a `qLogRegionalExpectedImprovement` that computes the log acquisition value based on smoothed fat-tailed non-linearities similar to `qLogExpectedImprovement` ([code](https://github.com/pytorch/botorch/blob/851df1fef632c596b9afea484a540d331d020c55/botorch/acquisition/logei.py#L140), [NeurIPS '23 spotlight](https://arxiv.org/pdf/2310.20708)). It seems like you should be able to take the existing qLogEI code and adapt it to this use case with few modification. If you decide to do this, let me know if you have questions about this.

Also, @Nobuo-Namura great work on REI! cc'ing you here since this conversation on leveraging qLogEI techniques for qLogREI might be of interest to you.

botorch_community/acquisition/rei.py
```diff
@@ -0,0 +1,171 @@
...
+from botorch.acquisition.monte_carlo import MCAcquisitionFunction
+from botorch.acquisition.objective import MCAcquisitionObjective, PosteriorTransform
+from botorch.models.model import Model
+from botorch.sampling.base import MCSampler
+from botorch.utils.transforms import concatenate_pending_points, t_batch_mode_transform
+from botorch.acquisition.analytic import (
+    AnalyticAcquisitionFunction,
+    _scaled_improvement,
+    _log_ei_helper,
+)
```
SebastianAment, line 16: @SaiAakash Could you add a citation to the REI paper, similar to we do this in other places, e.g. [here](https://github.com/pytorch/botorch/blob/851df1fef632c596b9afea484a540d331d020c55/botorch/acquisition/logei.py#L12)? Thanks!

test_community/acquisition/test_rei.py
SaiAakash: Actually, they don't come up with any numerical warnings which is surprising ( I had a similar thought about the numerical stability of the `mean`, `exp` and `log` operation in `LogREI`). But, I have confirmed that we don't get any warnings in this case. I have removed these lines now.

botorch_community/acquisition/rei.py
SaiAakash: @SebastianAment this has been added now. Thanks for bringing this up !

botorch_community/acquisition/rei.py
SaiAakash: Hi @SebastianAment ! Theoretically, any improvements that `qLogEI` can provide over `qEI` should translate to `qLogREI` as well. So, I think it's worth trying modifying `qREI` to `qLogREI`. I'll try to do this now. Thanks !

SaiAakash: I have modified qREI to qLogREI and the corresponding tests. Let me know if that looks alright.

SaiAakash:
Hi @esantorella @SebastianAment ! I have made the requested changes. I have also modified `qREI` to `qLogREI` to benefit from robustness of `Log` based acquisition functions against numerical instabilities. I have not written an example usage in the docstring for the module because the usage of these acquisition functions cannot be readily seen in a short code snippet (requires a bit of boilerplate code). We can demonstrate this in a notebook in the `notebooks_community` folder that implements efficient TuRBO with the new acquisition functions in a separate PR if that's okay. Thanks for all your inputs ! Let me know if you have any further comments.

test_community/acquisition/test_rei.py
SaiAakash: This is done now !

botorch_community/acquisition/rei.py
SaiAakash: Yes. They have been removed.

botorch_community/acquisition/rei.py
SaiAakash: This is done. We use `logmeanexp` to perform the same operation now !

botorch_community/acquisition/rei.py
```diff
@@ -0,0 +1,215 @@
...
+Region Averaged Acquisition Functions for Efficient Trust Region Selection
+for High-Dimensional Trust Region Bayesian Optimization.
+See [eriksson2019TuRBO]_, [namura2024rei]_
+
+Two versions of the Regional Expected Improvement (REI) acquisition
+function are implemented here:
+1. Analytic version: LogRegionalExpectedImprovement
+2. Monte Carlo version: qLogRegionalExpectedImprovement
+
+LogREI has been implemented from the original paper [namura2024rei]_
```
esantorella, line 17:
```suggestion
LogREI has been implemented from the original paper [namura2024rei]_
and code
(https://github.com/Nobuo-Namura/regional-expected-improvement/)
```

botorch_community/acquisition/rei.py
```diff
@@ -0,0 +1,215 @@
+#!/usr/bin/env python3
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+#
+# This source code is licensed under the MIT license found in the
+# LICENSE file in the root directory of this source tree.
+
+r"""
+Region Averaged Acquisition Functions for Efficient Trust Region Selection
```
esantorella, line 8: Thanks very much for this docstring, very clarifying!

esantorella:
Thanks for these changes; this is looking good! I'm reluctant to give @SaiAakash even more to-dos, but it would be ideal to check that performance is not worse after some of the more substantive changes such as using `logmeanexp`. Anyone have  any thoughts on how to most efficiently confirm that this is still working? It looks like some benchmarking functionality is provided at https://github.com/Nobuo-Namura/regional-expected-improvement/

botorch_community/acquisition/rei.py
Nobuo-Namura: Thank you for the mention and for implementing REI. I’ll definitely refer to your qLogREI implementation.

botorch_community/acquisition/rei.py
SaiAakash: I have added the link to the code now. Thanks !

SaiAakash:
Hi @esantorella ! After these changes, I have tested both of these acquisition functions on a problem I am working on at the moment about which I cannot disclose any details. But I can confirm that there are no significant performance degradations since these changes have been made. I can put some convergence plots here if you're interested though.

SebastianAment:
Hi @SaiAakash, just making sure you’ve seen the  pending suggestions above. Once we have these in and tests pass, we should be ready to land this!

SaiAakash:
Hi @SebastianAment ! Those changes are already in. I can see the Outdated flag on those conversations. So, those can be resolved ?

SebastianAment:
Hi @SaiAakash, there are a few more that are still `pending`, e.g. replacing `MCAcquisitionFunction ` with `LogImprovementMCAcquisitionFunction` changing `forward` to `_sample_forward` etc.

Attaching a screenshot of the first comment that is still pending above:
Screenshot 2025-01-29 at 10 19 47 PM

SaiAakash:
Aaha ! I see, it's weird that I am unable to see the comments shown in your screenshot ! I'll make the changes, thanks !

SebastianAment (changes requested):

SebastianAment:
> Aaha I see ! This is weird, because for some reason I am unable to see these comments shown in the screenshot on this PR ! Anyways, I will make these changes and get back to you. Thanks !

I forgot to submit the review, sorry about that! Hopefully it's visible now.

SaiAakash:
@SebastianAment I have incorporated your suggested changes. `qLogREI` now inherits from `LogImprovementMCAcquisitionFunction` and implements a `_sample_forward` method. Let me know if you have any further comments.

SaiAakash:
@SebastianAment I have a question regarding this change though. The `_sample_forward` method in general, expects the MC objective function values instead of `X` in other places. Here, we want to provide `X` to the evaluation function and compute the objective function value within the evaluation function. Do you feel this inconsistency is okay ? I feel sometimes this can lead to some dimension mismatches of the tensors maybe. Let me know if you have any thoughts about this ?

SebastianAment:
> The _sample_forward method in general, expects the MC objective function values instead of X in other places. Here, we want to provide X to the evaluation function and compute the objective function value within the evaluation function.

@SaiAakash That's right, so this is more subtle, thanks for raising this! There are two adjustments that would solve this issue:

1) We should move the code that generates the samples and objectives to a custom [`_get_samples_and_objectives`](https://github.com/pytorch/botorch/blob/472de87c7132f216bb38c18973fcc522c35dc5a9/botorch/acquisition/monte_carlo.py#L108), and can just reuse the existing `_sample_forward` implemented by `qLogExpectedImprovement` (you could even derive from this class). Something like adding
```python
def _get_samples_and_objectives(self, X: Tensor) -> tuple[Tensor, Tensor]:
       """ doc-string ... """
        # region-averaged EI specific code 
        q = X.shape[1]
        d = X.shape[2]

        # make N_x samples in design space
        X_min = (X - 0.5 * self.length).clamp_min(self.bounds[0]).unsqueeze(0)
        X_max = (X + 0.5 * self.length).clamp_max(self.bounds[1]).unsqueeze(0)
        Xs = (self.X_dev * (X_max - X_min) + X_min).reshape(-1, q, d)

        # calling the original method with the modified inputs
        return super()._get_samples_and_objectives(Xs)
```

2) I overlooked that the `mean` call goes over both `(0, 1)` before, i.e. the function-sample and the input-sample dimension. This can be taken care of with an adjustment in the constructor: 
```python
# difference to `SampleReducingMCAcquisitionFunction` constructor being the "+1"
sample_dim = tuple(range(len(self.sample_shape) + 1))   
self._sample_reduction = partial(logmeanexp, dim=sample_dim)
```
It is important to do this assignment *after* the `super().__init__` call, which itself assigns this variable [here](https://github.com/pytorch/botorch/blob/472de87c7132f216bb38c18973fcc522c35dc5a9/botorch/acquisition/monte_carlo.py#L247).

Let me know if you have thoughts or questions on this!

botorch_community/acquisition/rei.py
```diff
@@ -0,0 +1,241 @@
...
+
+        """
+        super().__init__(
+            model=model,
+            sampler=sampler,
+            objective=objective,
+            posterior_transform=posterior_transform,
+            X_pending=X_pending,
+        )
+
```
SebastianAment, line 183:
```suggestion
        # adding + 1 to account for the additional MC sampling dimension for points inside the trust region surrounding `X`
        sample_dim = tuple(range(len(self.sample_shape) + 1))   
        self._sample_reduction = partial(logmeanexp, dim=sample_dim)
```

SebastianAment (changes requested):
Hi @SaiAakash, I tested this locally and there are a few more changes needed. The tests currently pass because the sample shape is mocked out and not representative of the shapes when using the modified `_get_samples_and_objectives`. I added the changes I needed to make this work as suggestions below. Let me know if you have any thoughts or questions about this.

SaiAakash:
Hi @SebastianAment, I have implemented your changes. However, it doesn't seem to work with non-trivial `batch_shape`.

For example, changing the `batch_shape` value in this test results in a failed test
```
from botorch.models.transforms import Normalize

n = 3
d = 2
X = torch.randn(n, d, dtype=torch.float64)
Y = X.sin().sum(dim=-1, keepdim=True)

model = SingleTaskGP(X, Y, input_transform=Normalize(d=X.shape[-1]))

n_tr = 17
X_dev = torch.rand(n_tr, d)
acqf = qLogRegionalExpectedImprovement(model, best_f=0.0, X_dev=X_dev)
batch_shape = (2, )  # can modify this here to use it as a test for the batch case (`_test_q_log_regional_expected_improvement_batch`)
X_test = torch.randn(*batch_shape, n, d)
acqf(X_test)  # check that this executes and returns a tensor compatible with `batch_shape`.
```
and gives the following error:
`AssertionError: Expected the output shape to match either the t-batch shape of X, or the `model.batch_shape` in the case of acquisition functions using batch models; but got output with shape torch.Size([1, 2]) for X with shape torch.Size([2, 3, 2]).`

Any thoughts ?

botorch_community/acquisition/rei.py
```diff
@@ -0,0 +1,252 @@
...
+        # adding + 1 to account for the additional MC sampling dimension for points inside the trust region surrounding `X`
+        sample_dim = tuple(range(len(self.sample_shape) + 1))
+        self._sample_reduction = partial(logmeanexp, dim=sample_dim)
+
+        self.register_buffer("best_f", torch.as_tensor(best_f, dtype=float))
+        self.fat: bool = fat
+        self.tau_relu: float = check_tau(tau_relu, "tau_relu")
+        dim: int = X_dev.shape[1]
+        self.n_region: int = X_dev.shape[0]
+        self.X_dev: Tensor = X_dev.reshape(self.n_region, 1, 1, -1)
```
SebastianAment, line 195:
```suggestion
        self.X_dev: Tensor = X_dev
```

test_community/acquisition/test_rei.py
```diff
@@ -0,0 +1,298 @@
...
+
+        model = SingleTaskGP(X, Y, input_transform=Normalize(d=X.shape[-1]))
+
+        n_tr = 17
+        X_dev = torch.rand(n_tr, d)
+        acqf = qLogRegionalExpectedImprovement(model, best_f=0.0, X_dev=X_dev)
+        # can modify this here to use it as a test for the batch case
+        batch_shape = ()
+        X_test = torch.randn(*batch_shape, n, d)
+        acqf(X_test)
```
SebastianAment, line 201: Can you add a check at the end here that tests some property of `acqf(X_test)`, e.g. that it's a tensor with the expected shape?

test_community/acquisition/test_rei.py
```diff
@@ -0,0 +1,298 @@
...
+            X,
+            Y,
+            input_transform=Normalize(d=X.shape[-1]),
+        )
+
+        n_tr = 17
+        X_dev = torch.rand(n_tr, d)
+        acqf = qLogRegionalExpectedImprovement(model, best_f=0.0, X_dev=X_dev)
+        # can modify this here to use it as a test for the batch case
+        batch_shape = ()
```
SebastianAment, line 296: Adding a non-trivial batch shape here.
```suggestion
        batch_shape = (5,)
```

test_community/acquisition/test_rei.py
```diff
@@ -0,0 +1,298 @@
...
+        X_test = torch.randn(*batch_shape, n, d)
+        acqf(X_test)
```
SebastianAment, line 298: Similar to the test above, could you test the output of `acqf(X_test)`?

SebastianAment (commented):
>  changing the batch_shape value in this test results in a failed test

Fortunately this already already passed in my local tests, I forgot to send you a suggestion in the previous comments, particularly removing the reshaping of `X_dev` in the constructor. I hope that does it, we should (really) be close :)

botorch_community/acquisition/rei.py
SaiAakash: So, I made this change, but the test still fails for non-trivial `batch shape`.

botorch_community/acquisition/rei.py
```diff
@@ -0,0 +1,254 @@
...
+            X: A `batch_shape x q x d`-dim Tensor of model inputs.
+
+        Returns:
+            A two-tuple `(samples, obj)`, where `samples` is a tensor of posterior
+            samples with shape `sample_shape x n_region x batch_shape x q x m`,
+            and `obj` is a tensor of MC objective values with shape
+            `sample_shape x n_region x batch_shape x q`.
+        """
+        # region-averaged EI specific code
+        batch_shape = X.shape[0]
```
SebastianAment, line 222:
```suggestion
        batch_shape = X.shape[:-2]
```

botorch_community/acquisition/rei.py
```diff
@@ -0,0 +1,254 @@
...
+        d = X.shape[2]
+
+        # make N_x samples in design space
+        X_min = (X - 0.5 * self.length).clamp_min(self.bounds[0]).unsqueeze(0)
+        X_max = (X + 0.5 * self.length).clamp_max(self.bounds[1]).unsqueeze(0)
+
+        # n_region x (1, ..., 1) x 1 x d
+        X_dev = self.X_dev.reshape(
+            (self.n_region,) + (tuple(1 for _ in range(batch_shape)) + (1, d))
```
SebastianAment, line 231:
```suggestion
            (self.n_region,) + (tuple(1 for _ in batch_shape) + (1, d))
```

botorch_community/acquisition/rei.py
SebastianAment: Found two more differences below, we usually refer to `X.shape[:-2]` as the `batch_shape`, rather than the batch of candidates `q`.

botorch_community/acquisition/rei.py
SaiAakash: Aah, stuck with the old code. Thanks for helping out !

botorch_community/acquisition/rei.py
SaiAakash: This is fixed now ! Thanks !

SaiAakash:
@SebastianAment I have added the checks on tensor dimensions for both the tests (batch and non-batch). Thanks for helping out with this PR !

Issue #725, opened 2021-03-01
Classification Model as Output Constraint


jduerholt:
Dear botorch developers,

I have a question regarding output constraints. So far they are used and implemented in the following way:

- There is a property which should be larger than a user provided threshold.
- A GP regression model is build to model the property. The output of the regression model is used as input for a sigmoid function that is chosen in a way that it is 1 above the thresshold. 
- The actual optimization target is then multiplied by the output of the sigmoid function and this product forms the actual objective function
- The combination of regression model and sigmoid can be interpreted as a classification model --> related to logistic regression

Now comes my question: 

How to set up a optimization in which a "real" classification model is needed. Imagine achemical experiment in which only certain input combinations lead to a "successful" experiments. Successful in this context means that it is possible to measure the properties that should be optimized. For the failed experiments, no outputs are available, and within the BO loop one wants to  propose only experiments that lead to succesful experiments.

I found one [paper](https://www.cell.com/cell-reports-physical-science/fulltext/S2666-3864(20)30321-0) in this context. They are using an in the loop classifier to judge if a experiment will be succesful or not. I think there should be better solutions using the technique on the output constraints described above. One could for example train directly a classification model using gpytorch and multiply its output with the actual objective.

What are your thoughts on this and have you ever tried something related within botorch?

Best,

Johannes

Balandat:
Hi @jduerholt, great question. 

Using a Probit classification model with a Bernoulli likelihood for actual classification should  work in this context. You're right that since such model maps the latent function directly to a probability of success, one should be able to use the same approach as with the existing constrained formulation, but using this success probability directly (here it is: https://github.com/cornellius-gp/gpytorch/blob/master/gpytorch/likelihoods/bernoulli_likelihood.py#L26).

We could probably just use the vanilla GP Regression model from https://github.com/cornellius-gp/gpytorch/blob/master/examples/04_Variational_and_Approximate_GPs/Non_Gaussian_Likelihoods.ipynb to prototype something (note that this uses variational inference). The main reason we haven't tried this out is that we so far didn't run into any concrete use cases in our work, but it would be interesting to try this out. 

I recall reading a paper that did this within the last year or two, but I can't find it right now. 

An interesting question is whether one would want to use a multi-task model for modeling the latent constraint function and the objective(s), the rationale being that performance and feasibility may be quite correlated. But that's a refinement and not necessary as a first step.

jduerholt:
Hi @Balandat,

[...]

PR #2700, opened 2025-01-26
Adding a tutorial on BO constrained by probability of classification model 

FrankWanger:
## Motivation

There is no current tutorial on using the probability pulled from classification results as constraints in acquisition functions. And such an application holds strong interest from BO guided laboratory experimentation. A prior discussion (#725) was formed.

Yes

## Test Plan

In the present tutorial we show how to deal with feasibility constraints that are observed alongside the optimization process (referred to as 'outcome constriants' in BoTorch document, or sometimes as 'black-box constraints'). More specifically, the feasibility is modelled by a classification model, followed by feeding this learned probability to the acquisition funtion through the `constraint` argument in `SampleReducingMCAcquisitionFunction`. Namely, this is achieved through re-weighting the acquisition function by $\alpha_{\text{acqf-con}}=\mathbb{P}(\text{Constraint satisfied})*\alpha_{\text{acqf}}$. To achieve this, the pulled probability of classification model underwent un-sigmoid function and was inversed to fit into the API (as negative values treated as feasibility).

A 2D syntheic problem of [Townsend function](https://www.chebfun.org/examples/opt/ConstrainedOptimization.html) was used. For the classification model, we implemented [approximate GP](https://docs.gpytorch.ai/en/v1.13/examples/04_Variational_and_Approximate_GPs/Non_Gaussian_Likelihoods.html) with a Bernoulli likelihood. `qLogExpectedImprovement` was selected as the aquisition function.

Below are the plots of the problem landscape, acquisition function value, constraint probability, and the EI value (before weighting) at different iterations:

At iter=1:
image

At iter=10:
image

At iter=50:
image

The log regret after 50 iterations are plotted against random (sobel).
image

All images can be reproduced by the notebook.

## Related PRs

not related to any change of functionality


notebooks_community/clf_constrained_bo.ipynb
```diff
@@ -0,0 +1,840 @@
...
+    "        self._optimal_value = 1.660\n",
+    "        self.name = \"Townsend\"\n",
+    "\n",
+    "    def __call__(self, x):\n",
+    "        return self.objective(x)\n",
+    "\n",
+    "    def constraint(self, x):\n",
+    "        x1, x2 = x[..., 0], x[..., 1]\n",
+    "        t = torch.atan2(x1, x2) \n",
+    "        constraint_value = (((2 * torch.cos(t) - 0.5 * torch.cos(2 * t) - 0.25 * torch.cos(3 * t) - 0.125 * torch.cos(4 * t)) ** 2 + (2 * torch.sin(t)) ** 2 - x1 ** 2 - x2 ** 2) > 0).float()\n",
```
Balandat, line 97: This is kind of hard to read; maybe split this up in to a couple of lines?

notebooks_community/clf_constrained_bo.ipynb
```diff
@@ -0,0 +1,840 @@
...
+    "        self.lb = torch.tensor([-2.25, 2.25], **tkwargs)\n",
+    "        self.ub = torch.tensor([-2.5, 1.75], **tkwargs)\n",
+    "        self.bounds = torch.stack([self.lb, self.ub]).T\n",
+    "        self._optimal_value = 1.660\n",
+    "        self.name = \"Townsend\"\n",
+    "\n",
+    "    def __call__(self, x):\n",
+    "        return self.objective(x)\n",
+    "\n",
+    "    def constraint(self, x):\n",
```
Balandat, line 94: Maybe calling this `is_feasible()` is more intuitive in that it it makes it clear this is a boolean observation?

notebooks_community/clf_constrained_bo.ipynb
```diff
@@ -0,0 +1,840 @@
...
+      "      import os\n",
+      "      os.environ['CXX'] = 'g++-8'\n",
+      "    \n",
+      "[KeOps] Warning : Cuda libraries were not detected on the system or could not be loaded ; using cpu only mode\n"
+     ]
+    }
+   ],
+   "source": [
+    "from botorch.utils.sampling import draw_sobol_samples\n",
+    "def generate_initial_data(n):\n",
```
Balandat, line 191: formatting nit
```suggestion
    "from botorch.utils.sampling import draw_sobol_samples\n",
    "\n",
    "def generate_initial_data(n):\n",
```

notebooks_community/clf_constrained_bo.ipynb
```diff
@@ -0,0 +1,840 @@
...
+     "text": [
+      "[KeOps] Warning : \n",
+      "    The default C++ compiler could not be found on your system.\n",
+      "    You need to either define the CXX environment variable or a symlink to the g++ command.\n",
+      "    For example if g++-8 is the command you can do\n",
+      "      import os\n",
+      "      os.environ['CXX'] = 'g++-8'\n",
+      "    \n",
+      "[KeOps] Warning : Cuda libraries were not detected on the system or could not be loaded ; using cpu only mode\n"
+     ]
```
Balandat, line 186: Is there way to get rid of this warning as this is completely unrelated to the tutorial?

notebooks_community/clf_constrained_bo.ipynb
```diff
@@ -0,0 +1,840 @@
...
+    "from botorch.acquisition.objective import GenericMCObjective\n",
+    "from botorch.sampling.normal import SobolQMCNormalSampler\n",
+    "from botorch.optim import optimize_acqf\n",
+    "\n",
+    "BATCH_SIZE = 1\n",
+    "NUM_RESTARTS = 10 \n",
+    "RAW_SAMPLES = 512\n",
+    "\n",
+    "def optimize_acqf_and_get_observation(model, train_obj, train_con, \n",
+    "):\n",
```
Balandat, line 313:
```suggestion
    "def optimize_acqf_and_get_observation(model, train_obj, train_con):\n",
```

notebooks_community/clf_constrained_bo.ipynb
```diff
@@ -0,0 +1,840 @@
...
+    "    therefore we need to unsigmoid our probability (0-1) to (-inf,inf)\n",
+    "    also we need to invert the probability, where -inf means the constraint is satisfied. Finally,we add 1e-8 to avoid log(0).\n",
+    "    '''\n",
+    "    y_con = Z[...,1] #get the constraint\n",
+    "\n",
+    "    prob = model_con.likelihood(y_con).probs #obtain the probability of y_con(when constraint satisfied)\n",
+    "    prob_unsigmoid_neg = torch.log(1-prob+1e-8)-torch.log(prob+1e-8) #unsigmoid the probability and invert it to adapt to BoTorch's constraint API\n",
+    "    \n",
+    "    return prob_unsigmoid_neg\n"
+   ]
```
Balandat, line 366: If the classification model already produces the probabilities of feasibility, it would be great if we could directly use that in the acquisition function, rather than converting it back first. @SebastianAment do you see any major challenges to just accept an additional "probability_of_feasibility" argument to `SampleReducingMCAcquisitionFunction` (and possibly in other places) and then just use that in the probability weighting? 

Even if there are no issues, getting such a change into botorch would require some eng work so I wouldn't want to block this PR on that. That said, the probability of feasibility conversion is not a standard sigmoid though internally, see https://github.com/pytorch/botorch/blob/main/botorch/utils/objective.py#L178 - ideally for the time being (until we can accept the probability directly) we could apply the actual inverse of what is being applied in botorch.

notebooks_community/clf_constrained_bo.ipynb
```diff
@@ -0,0 +1,840 @@
...
...1ZSs9tGgDffyJeVOGw+FIyfPLKy2ORWvebTH09fVRc3Nz0dsAZOdClNl7771FETk5F58x3l1eiXEsemCxXDt+yS1iWOAsV+mqU78tJh7HdswI0wLaZkaXC0RkwSH+r3/9K81BjvvIKVdjv/32EzEs2E7m008/FUI69lfIPhlGUENCajGCt1HO+LAzIGJY4CxXadKpPhEWj2M7hjEaFswLRBYzlaK5LJxDNFdzi+cSypXbG+UYznQ3G4YG0VyPGF1M9EopwXty25PktEkUk7Y31zHF5xWX7CTFY+SkZNbCq3CZZwKXeb4CoIXAIjvD1BAV7AbXKoobLZ67bUmyEzLL1SfcOZwu6h+O0lPvb6D/vLPOUoO+WhH9k9CwpYt9Ihs+3yLn3SIKRV6uv/56Q48D4sptt91G3//+97Nu8/jjj9M777xDZ5xxRs59XX311WLbl19+mU444QSRnYt9M9bH6gU/mdy4JYfILEcMixoxSojHsR1TOVx44YV077330kMPPURLly6ls88+W8RnyW0xorIuvfTS1PZ4vLu7W8wKglD+3HPPiaKfGMjUuk+GqdU4Fisdd8zmFpnliGFRwyXFxOPYLtfr1mzxT7jli8QVbBSRNrmX6owd5EiWIpBzppVxGdkiWnKhJYqlWOHUcGEcjzUUJlhoKf5pVULhYYo32shlkyiqEM1lnNsKgMZto8eitGaZ5yJXNAvDMIxh0SwlEt+LEb7x3GId0r2hCCXJQYk4rrGj2+JEPEZxslHPcFJcu0ohOg9G4uTzGNM9W7Shm255aSUtWbhy1GOV5i7Xk3d7ySWX0A033JBzXxBHZs+enbqPbFvEs5x44okix1yNV199VQgpEFh22WWXnPu/7LLLUv+fP3++EGGQnXveeeflfB5TmLu8kPzycqInjqUW3OXhTWHD9mW3xWjY00pSMk5RGm1UcpODhu1R6ukLUVhKN+j4x/kNOw7GWE466STq6Oigyy+/XMSqzJs3j1544YVU0c61a9eS3b799xcGWF988UW64IILRB2KCRMmCPEc9Si07pMxkP5uIp+Fom5K0AdC1AciPyodvVnmRov8LikqCnwisxwxLJnEbC7xOLYr5HgtTwX216sJFswLRM6a1iKag2zCebXklpuB1bLMQV/UQVvCNpoQiG/LMLelXOYuj5Pq7THqjHtp66BE9dv63MVmmWc
```
Balandat, line 766: Maybe split this across a few lines to make it more readable.

notebooks_community/clf_constrained_bo.ipynb
```diff
@@ -0,0 +1,840 @@
...
+    "    plt.colorbar(c_ei, ax=axes[3])\n",
+    "    plt.tight_layout()"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "### Perform BO loop\n",
+    "We initialize the BO loop with 10 random points from SOBEL sequence. We then perform 50 iterations of BO. The acquisition function value, the underlying constraint probability, and the expected improvement value are plotted at every 10 iterations."
```
Balandat, line 433:
```suggestion
    "We initialize the BO loop with 10 random points from SOBOL sequence. We then perform 50 iterations of BO. The acquisition function value, the underlying constraint probability, and the expected improvement value are plotted at every 10 iterations."
```

notebooks_community/clf_constrained_bo.ipynb
```diff
@@ -0,0 +1,840 @@
...
+  {
+   "cell_type": "code",
+   "execution_count": 9,
+   "metadata": {},
+   "outputs": [
+    {
+     "name": "stdout",
+     "output_type": "stream",
+     "text": [
+      "Iteration 0: log Regret (CEI, Sobel) = (-1.27, -1.27)\n",
```
Balandat, line 445:
```suggestion
      "Iteration 0: log Regret (CEI, Sobol) = (-1.27, -1.27)\n",
```

Balandat (commented):
Thanks a lot for putting this up, this is great. 

My main comment (see inline) is on how to leverage the probability of feasibility produced by the classification model directly, rather than converting it twice, but that would require some changes to botorch itself.

Other than that mostly cosmetic comments.

I see that the method finds what appears to be the optimum very quickly - is this consistent across runs? If so it may make sense to reduce the number of iterations somewhat to cut down the runtime of the tutorial.

notebooks_community/clf_constrained_bo.ipynb
```diff
@@ -0,0 +1,840 @@
...
+    "\\end{align*}\n",
+    "$$\n",
+    "\n",
+    "Subjected to the constraint:\n",
+    "$$\n",
+    "y_{\\text{con}} = \\left(\\left(2 \\cos(t) - 0.5 \\cos(2t) - 0.25 \\cos(3t) - 0.125 \\cos(4t)\\right)^2 + \\left(2 \\sin(t)\\right)^2 - x_1^2 - x_2^2\\right)>0\n",
+    "$$ \n",
+    "where $t = \\arctan\\left(\\frac{x_1}{x_2}\\right)$\n",
+    "\n",
+    "Here, we follow a natural representation where $y_{\\text{con}}=1$ indicates a feasible condition. We will train a classification model to predict the feasibility of the point. Note that in BoTorch's implementation, **negative values** indicate feasibility, thus we need to do conversion later when feeding feasibility into the pipeline.\n",
```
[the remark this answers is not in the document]
FrankWanger: Indeed I have realised the problem of notation. I wanted to add that in many situations in experiments the numerical value of the constraint is not directly observable. So what we have as the data is only binary outcomes of success or failure - and yes here we applied this binary mask to our synthetic problem to throw away information so that we can simulate what we obtain in lab.

notebooks_community/clf_constrained_bo.ipynb
FrankWanger: I wasn't sure where this is about but I've modified the output to make it clearer

FrankWanger:
> Thanks a lot for putting this up, this is great.
> 
> My main comment (see inline) is on how to leverage the probability of feasibility produced by the classification model directly, rather than converting it twice, but that would require some changes to botorch itself.
> 
> Other than that mostly cosmetic comments.
> 
> I see that the method finds what appears to be the optimum very quickly - is this consistent across runs? If so it may make sense to reduce the number of iterations somewhat to cut down the runtime of the tutorial.

Thank you so much! I've addressed most of the formatting issues, there is only one that I am not sure how to remove - the KeOps warnings. I've switched to macOS and it did not help. In terms of the results, yes I can see that it is quite consistent so I have halved the iterations to 25 and slighted added the freq of plots.

Balandat:
Great. I may just manually strip the output from the notebook source to keep it clean. 

I'll get this merged in since it's in great shape already, but still curious to hear @SebastianAment's thoughts on supporting this better in the acquisition functions themselves (which would be a separate PR anyway).

notebooks_community/clf_constrained_bo.ipynb
```diff
@@ -0,0 +1,729 @@
...
+     "name": "stdout",
+     "output_type": "stream",
+     "text": [
+      "[KeOps] Warning : omp.h header is not in the path, disabling OpenMP. To fix this, you can set the environment\n",
+      "                  variable OMP_PATH to the location of the header before importing keopscore or pykeops,\n",
+      "                  e.g. using os.environ: import os; os.environ['OMP_PATH'] = '/path/to/omp/header'\n",
+      "[KeOps] Warning : Cuda libraries were not detected on the system or could not be loaded ; using cpu only mode\n"
+     ]
+    }
+   ],
```
Balandat, line 194:
```suggestion
```

Balandat, line 194:
```suggestion
   "outputs": [],
```

notebooks_community/clf_constrained_bo.ipynb
SebastianAment: > do you see any major challenges to just accept an additional "probability_of_feasibility" argument to SampleReducingMCAcquisitionFunction (and possibly in other places) and then just use that in the probability weighting?

That should be pretty straightforward, mainly taking care of appropriate reshaping, since we are usually applying the feasibility weighting on a per-sample basis, and `probability_of_feasibility` won't share the MC dimension. 

Regarding the inversion of the sigmoid, we are currently using a [sigmoid with inverse quadratic asymptotic behavior](https://github.com/pytorch/botorch/blob/aeda83a4e22c053dd57dab3d3632d5b77a6445f1/botorch/utils/safe_math.py#L435), which could likely be inverted analytically as well, but that will not be necessary once we support this in the acquisition function directly.

PR #2711, opened 2025-01-29
Update cost function, use qLogEI in discrete MFKG tutorial

Balandat:
The cost function has pretty high fixed cost, whcih isn't the best to illustrate the algorithm (see https://github.com/pytorch/botorch/discussions/2523#discussioncomment-11105963 for discussion).

Also updates to use `qLogEI` instead of `qEI` to get rid of the deprecation warning.

### DIFF
--- a/tutorials/discrete_multi_fidelity_bo.ipynb
+++ b/tutorials/discrete_multi_fidelity_bo.ipynb
@@ -1,571 +1,540 @@
 {
-  "cells": [
-    {
-      "cell_type": "markdown",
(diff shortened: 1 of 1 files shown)


saitcakmak (approved):

saitcakmak:
Do we need the comparison to LogEI in the notebook? We're not even plotting the comparison

Balandat:
I think the comparison is still useful, you can see that you get a competitive suggestion for less budget even with one eval

PR #2713, opened 2025-01-30
[RFC] Use `uv` package manager in GHA

saitcakmak:
I came across [`uv`](https://docs.astral.sh/uv/guides/integration/github/) while debugging the scipy slowdowns and noticed how much faster it is than vanilla `pip`. Using `uv` saves a couple minutes in installing dependencies, which can add up to meaningful compute savings over many CI runs we do every day. Should we update all workflows to use `uv`?

This PR updates `reusable_tutorials.yml` as an example.
Before, with `pip`: https://github.com/pytorch/botorch/actions/runs/13058152281/job/36434279975
- Total setup time: 2 + 2 + 12 + 44 + 50 + 9 = 119s
Screenshot 2025-01-30 at 12 59 40 PM

After, with `uv pip`: https://github.com/pytorch/botorch/actions/runs/13058212963/job/36434483617
- Total setup time: 2 + 1 + 10 + 10 + 7 + 6 = 36s
Screenshot 2025-01-30 at 1 00 19 PM

### DIFF
--- a/.github/workflows/deploy_on_release.yml
+++ b/.github/workflows/deploy_on_release.yml
@@ -20,9 +20,14 @@ jobs:
     runs-on: ubuntu-latest
     permissions:
       id-token: write # This is required for PyPI OIDC authentication.
+    env:
+      # `uv pip ...` requires venv by default. This skips that requirement.
+      UV_SYSTEM_PYTHON: 1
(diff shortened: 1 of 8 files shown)


Balandat:
sure that sounds like a good change, let's update this across the board.

PR #2721, opened 2025-02-03
[GHA] Specify permissions when deviating from defaults

CristianLara:
## Motivation

The `publish_website.yml` workflow requires write permissions to
1. create new docusaurus versions by pushing a commit to `docusaurus-versions` branch
2. push new website to gh-pages

This was not an issue in the fork that introduced these changes because Meta's organization / the official repo has more restrictive permissions than the defaults. More restrictive default permissions are definitely the way to go, here we elevate permissions only when necessary.

## Test Plan

I made the default permissions in my fork more restrictive such that the same workflows would fail then verified that this change results in successful workflow runs. https://github.com/CristianLara/botorch/actions/runs/13107635487/job/36565023833

### DIFF
--- a/.github/workflows/deploy_on_release.yml
+++ b/.github/workflows/deploy_on_release.yml
@@ -85,4 +85,7 @@ jobs:
     uses: ./.github/workflows/publish_website.yml
     with:
       new_version: ${{ github.event.release.tag_name }}
-    secrets: inherit
+    permissions:
+      pages: write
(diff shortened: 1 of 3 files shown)


saitcakmak (approved):
Lgtm. AFAIK, the publish_website doesn't rely on secrets anymore (previously, docusaurus bot needed it, I think), so removing those also seems safe there.

Issue #2724, opened 2025-02-04
Make SAASBO available on MPS


Hrovatin:
### Motivation

The current SAASBO implementation does not support MPS as one of the operations uses float64 instead of float32, whit the former not being supported on MPS.
```
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Cell In[10], line 34
     20 gp = SaasFullyBayesianSingleTaskGP(
     21     train_X=X,
     22     train_Y=train_Y,
     23     train_Yvar=torch.full_like(train_Y, 1e-6),
     24     outcome_transform=Standardize(m=1),
     25 )
     26 fit_fully_bayesian_model_nuts(
     27     gp,
     28     warmup_steps=WARMUP_STEPS,
   (...)
     31     disable_progbar=True,
     32 )
---> 34 EI = qLogExpectedImprovement(model=gp, best_f=train_Y.max())
     35 candidates, acq_values = optimize_acqf(
     36     EI,
     37     bounds=torch.cat((torch.zeros(1, DIM), torch.ones(1, DIM))).to(**tkwargs),
   (...)
     40     raw_samples=1024,
     41 )
     43 Y_next = torch.cat([branin_emb(x).unsqueeze(-1) for x in candidates]).unsqueeze(-1)

File [~/miniforge3/envs/baybe/lib/python3.10/site-packages/botorch/acquisition/logei.py:218](http://localhost:8888/lab/tree/baybe-analysis/method_improvements/methods/feature_selection/~/miniforge3/envs/baybe/lib/python3.10/site-packages/botorch/acquisition/logei.py#line=217), in qLogExpectedImprovement.__init__(self, model, best_f, sampler, objective, posterior_transform, X_pending, constraints, eta, fat, tau_max, tau_relu)
    177 r"""q-Log Expected Improvement.
    178 
    179 Args:
   (...)
    205         approximations to ReLU.
    206 """
    207 super().__init__(
    208     model=model,
    209     sampler=sampler,
   (...)
    216     fat=fat,
    217 )
--> 218 self.register_buffer("best_f", torch.as_tensor(best_f, dtype=float))
    219 self.tau_relu = check_tau(tau_relu, name="tau_relu")

TypeError: Cannot convert a MPS Tensor to float64 dtype as the MPS framework doesn't support float64. Please use float32 instead.
```

Besides, there are some operations which still require fallback to CPU*. Although PyTorch aims to implement them on MPS in the future, which makes the issue less important.

[...]

PR #2725, opened 2025-02-04
Remove hard-coded dtype from best_f buffers

saitcakmak:
Summary:
Specifiying `dtype=float` (equivalent to `torch.float64`) causes issues if the user wants to use single precision. See https://github.com/pytorch/botorch/issues/2724

Removing the `dtype` argument from `torch.as_tensor` will lead to using the existing dtype if the input is a tensor, and using `torch.get_default_dtype()` if the input is a python float or a list of floats. If the input is a numpy array, the corresponding torch dtype is used (e.g. `torch.float64` for `np.float64`).

Differential Revision: D69121327

### DIFF
--- a/botorch/acquisition/logei.py
+++ b/botorch/acquisition/logei.py
@@ -18,11 +18,8 @@ References
 from __future__ import annotations
 
 from collections.abc import Callable
-
 from copy import deepcopy
-
 from functools import partial
-
 from typing import TypeVar
 
 import torch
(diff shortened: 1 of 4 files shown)


Balandat:
One thing to note here is that if the input is a python list with ints then this will register an int tensor which will cause breakages downstream. This shouldn't happen a lot but I think we should auto-convert the dtype in that case.

saitcakmak:
> One thing to note here is that if the input is a python list with ints then this will register an int tensor which will cause breakages downstream. This shouldn't happen a lot but I think we should auto-convert the dtype in that case.

Yeah, that's a fair point. I can't find a clean way to convert to the default float type without explicitly checking for integers and casting after, which I don't like. The type hints specify `float | Tensor`. I think it's ok to expect the user to respect that and specify an appropriate inputs. I'm sure various other parts of the code will also break if you pass in integer tensors.

PR #2726, opened 2025-02-04
[GHA] Creating a new version only tracks relevant files to git

CristianLara:
Fixes an issue where unrelated files could get added to the `docusaurus-versions` branch. This broke GHA in the Ax repo since this was adding the results of running tutorials. Botorch doesn't run tutorials in CI but we still fix this to avoid other possible issues down the road.

Same fix as the Ax counterpart: https://github.com/facebook/Ax/pull/3306


CristianLara:
Propagate this change to `docusaurus-versions` branch by catching it up with `main` once this lands.

PR #2743, opened 2025-02-12
Fix output tensor dimension in degenerate `optimize_acqf` call

### LINKED ISSUE #2740: [Bug]: Inconsistent output dimensions
### What happened?

Not sure if this is intentional, but it seems like a bug to me. When fixing all features in `optimize_acqf`, the returned acquisition value changes from a float to a 1-D tensor. Haven't checked what happens for the other `optimize_*` functions.

Of course, this is a degenerate case since calling the optimization routing with zero degrees of freedom is sort of meaningless, but it can still be useful when invoking BoTorch procedurally from wrappers as in my case.

Here a minimal example adjusted from the landing page code:

```python
import torch
from botorch.acquisition import LogExpectedImprovement
from botorch.models.transforms import Normalize, Standardize

def run(fixed_features):

    train_X = torch.rand(10, 2, dtype=torch.double) * 2
    Y = 1 - torch.linalg.norm(train_X - 0.5, dim=-1, keepdim=True)
    Y = Y + 0.1 * torch.randn_like(Y)

    gp = SingleTaskGP(
        train_X=train_X,
        train_Y=Y,
        input_transform=Normalize(d=2),
        outcome_transform=Standardize(m=1),
    )
    fit_gpytorch_mll(mll)
    logEI = LogExpectedImprovement(model=gp, best_f=Y.max())
    bounds = torch.stack([torch.zeros(2), torch.ones(2)]).to(torch.double)
        logEI,
        bounds=bounds,
        q=1,
        num_restarts=5,
        raw_samples=20,
        fixed_features=fixed_features,
    )
    print(acq_value)

run({0: 0})  
run({0: 0, 1: 0})
```

```shell
tensor(-4.7173, dtype=torch.float64)
tensor([-8.1727], dtype=torch.float64)
```

### BoTorch Version

0.13.0

### Python Version

3.10

### Operating System

macOS

### Code of Conduct

AdrianSosic:
## Motivation

Fixes #2740 

@sdaulton: This is a draft for fixing the dimensionality issue. While the PR technically solves the problem, think it only addresses the symptom and not the underlying root cause. For the latter, I'd need some guidance from your end. 

Specifically, something seems fishy about the tensor dimensions promised by the [acquisition function base class](https://github.com/pytorch/botorch/blob/d74a11331c2efb89c9b5ea688d2a61c1d9fb1420/botorch/acquisition/acquisition.py#L64-L73). The docstring states that the input is a `(b) x q x d`-dim tensor and the output is a `(b)`-dim. The way I understand this is that `b` is an optional batching dimension. But if you read the contract, it would imply:
* If you pass a `b x q x d`-dim tensor, you get a `(b)`-dim tensor back.
* If you pass a `q x d`-dim tensor, you get a dimensionless tensor (i.e. scalar value) back.

Now the problem is that the second part seems to be actually violated, which causes the the problem in #2740. Interestingly, when writing the test, I wanted to use a `MockAcquisitionFunction()` object for it like done in other tests. However, that object *does seem* to fulfill the contract, so I couldn't even reproduce the problem with the mock but had to reuse the code from my original minimal example.

Any thoughts how we should go about this, i.e. is this actually a more fundamental problem in the acquisition function code?

Yes

## Test Plan

A corresponding test has been included.

### DIFF
--- a/botorch/optim/optimize.py
+++ b/botorch/optim/optimize.py
@@ -192,7 +192,7 @@ def _optimize_acqf_all_features_fixed(
     X = X.expand(q, *X.shape)
     with torch.no_grad():
         acq_value = acq_function(X)
-    return X, acq_value
+    return X, acq_value[0]
 
 
 def _validate_sequential_inputs(opt_inputs: OptimizeAcqfInputs) -> None:
--- a/test/optim/test_optimize.py
+++ b/test/optim/test_optimize.py
@@ -14,10 +14,12 @@ from unittest import mock
 
 import numpy as np
 import torch
+
 from botorch.acquisition.acquisition import (
     AcquisitionFunction,
     OneShotAcquisitionFunction,
 )
+from botorch.acquisition.analytic import LogExpectedImprovement
 from botorch.acquisition.knowledge_gradient import qKnowledgeGradient
 from botorch.acquisition.monte_carlo import qExpectedImprovement
 from botorch.acquisition.multi_objective.hypervolume_knowledge_gradient import (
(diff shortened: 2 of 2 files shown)


sdaulton:
Thanks for the PR! It looks like a test is failing. Would you mind fixing that please?

AdrianSosic:
> Thanks for the PR! It looks like a test is failing. Would you mind fixing that please?

Sorry, my bad, forgot to run the entire test suite. Should now be fixed – though there are 4 other tests failing also on the main branch?

Still, like I mentioned earlier: while the change fixes the particular problem reported, I don't think it properly addresses the root cause. So input would be welcome 🙃

PR #2748, opened 2025-02-18
Performance & runtime improvements to info-theoretic acquisition functions (1/N)

hvarfner:
A series of improvements directed towards improving the performance of PES & JES, as well as their MultiObj counterparts.

## Motivation
As pointed out by @SebastianAment in [this paper](https://arxiv.org/pdf/2310.20708), the BoTorch variant of JES, and to an extent PES, is brutally slow an suspiciously ill-performing. To bring them up to their potential, I've added a series of performance improvements:

**1. Improvement to get_optimal_samples and optimal_posterior_samples**: As this is an integral part of their efficiency, I've added `suggestions` (similar to `sample_around_best`) to `optimize_posterior_samples`. 
Marginal runtime improvement in acquisition optimization (sampling time practically unchanged):
runtime_pr1
Substantial performance improvement: pr1_regret.

**2. Added initializer to acquisition funcction optimization**: Similar to KG, ES methods have sensible suggestions for acquisition function optimization in the form of the sampled optima. This drastically reduces the time of acquisition function optimization, which could on occasion take 30+ seconds when `num_restarts` was large `>4`.

Benchmarking INC

**2b. Multi-objective support for initializer**: By re-naming arguments of the multi-objective variants, we get consistency and support for MO variants.

**3. Enabled gradient-based optimization for PES**: The current implementation contains a while-loop which forces the gradients to be recursively computed. This commonly causes NaN gradients, which is why the recommended option is `"with_grad": False` in the tutorial. One `detach()` alleviates this issue, enabling gradient-based optimization. 

NOTE: this has NOT been ablated, since the non-grad optimization is extremely computationally demanding.

## Test Plan

Unit tests and benchmarking.

## Related PRs

First of a couple!

Bonus: while benchmarking, I had issues repro'ing the LogEI performance initially. I found that `sample_around_best` made LogEI worse on Mich5. All experiments are otherwise a repro of the settings used in the LogEI paper.
LogEI_sample_around_best

### DIFF
--- a/botorch/utils/sampling.py
+++ b/botorch/utils/sampling.py
@@ -27,12 +27,14 @@ import numpy as np
 import numpy.typing as npt
 import scipy
 import torch
+
 from botorch.exceptions.errors import BotorchError, InfeasibilityError
 from botorch.exceptions.warnings import UserInputWarning
 from botorch.sampling.qmc import NormalQMCEngine
 from botorch.utils.transforms import unnormalize
 from scipy.spatial import Delaunay, HalfspaceIntersection
 from torch import LongTensor, Tensor
+from torch.distributions import Multinomial
 from torch.quasirandom import SobolEngine
 
 
@@ -990,10 +992,12 @@ def sparse_to_dense_constraints(
 def optimize_posterior_samples(
     paths: GenericDeterministicModel,
     bounds: Tensor,
-    raw_samples: int = 1024,
-    num_restarts: int = 20,
+    raw_samples: int = 2048,
+    num_restarts: int = 4,
     sample_transform: Callable[[Tensor], Tensor] | None = None,
     return_transformed: bool = False,
+    suggested_points: Tensor | None = None,
+    options: dict | None = None,
 ) -> tuple[Tensor, Tensor]:
     r"""Cheaply maximizes posterior samples by random querying followed by
     gradient-based optimization using SciPy's L-BFGS-B routine.
@@ -1008,6 +1012,9 @@ def optimize_posterior_samples(
             negate the objective or otherwise transform the output.
         return_transformed: A boolean indicating whether to return the transformed
             or non-transformed samples.
+        suggested_points
+        options: Options for generation of initial candidates, passed to
+            gen_batch_initial_conditions.
 
     Returns:
         A two-element tuple containing:
@@ -1015,6 +1022,7 @@ def optimize_posterior_samples(
             - f_opt: A `num_optima x [batch_size] x m`-dim, optionally
                 `num_optima x [batch_size] x 1`-dim,  tensor of optimal outputs f*.
     """
+    options = {} if options is None else options
 
     def path_func(x) -> Tensor:
         res = paths(x)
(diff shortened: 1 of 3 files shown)


sdaulton:
Thanks! It seems like `sample_around_best` could definitely lead to the AF optimization getting stuck in a local optima

hvarfner:
@sdaulton for sure! I currently observe similar things for JES, but I'm not sure whether the found points are actually higher in acquisition function value or not (for either LogEI or JES)

sdaulton:
That would be interesting to see

saitcakmak (changes requested):
Hi Carl! This seems like a decent improvement. Just a few comments in-line

botorch/utils/sampling.py
```diff
@@ -1023,21 +1031,46 @@ def path_func(x) -> Tensor:
...
+    weights = torch.exp(eta * weights)
+
+    # weights can be more than 2D, which is not supported by torch.multinomial
+    # the argsort picks out the indices that are nonzero, i.e. those that are drawn
+    # (without replacement, so we will always have num_restarts nonzero ones)
+    idx = (
+        Multinomial(num_restarts, probs=weights)
+        .sample()
+        .argsort(descending=True)[..., :num_restarts]
+    )
```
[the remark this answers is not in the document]
hvarfner: It is extremely similar, but torch.multinomial does not support 3D multinomial sampling which is why I re-used the logic (see the comment). I can try and combine the two and see if it meshes.

botorch/utils/sampling.py
```diff
@@ -1023,21 +1031,46 @@ def path_func(x) -> Tensor:
...
-        SobolEngine(dimension=bounds.shape[1], scramble=True).draw(n=raw_samples),
-        bounds=bounds,
-    )
     # queries all samples on all candidates - output shape
     # raw_samples * num_optima * num_models
+    frac_random = options.get("frac_random", 0.9)
+    candidate_set = draw_sobol_samples(
+        bounds=bounds, n=round(raw_samples * frac_random), q=1
+    ).squeeze(-2)
+    if suggested_points is not None:
```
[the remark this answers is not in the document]
hvarfner: Ah, good catch. Absolutely

botorch/utils/sampling.py
```diff
@@ -1023,21 +1031,46 @@ def path_func(x) -> Tensor:
...
+        from botorch.optim.initializers import sample_truncated_normal_perturbations
```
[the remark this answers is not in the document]
hvarfner: I think so! let me check

botorch/optim/initializers.py
```diff
@@ -979,18 +980,12 @@ def initialize_q_batch(
...
-    while torch.isinf(weights).any():
-        etaZ *= 0.5
-        weights = torch.exp(etaZ)
-    if batch_shape == torch.Size():
-        idcs = torch.multinomial(weights, n)
-    else:
-        idcs = batched_multinomial(
-            weights=weights.permute(*range(1, len(batch_shape) + 1), 0), num_samples=n
-        ).permute(-1, *range(len(batch_shape)))
+    idcs = boltzmann_sample(
```
saitcakmak, line 983: Cool, this deduplicated a bunch of repeated logic here.

botorch/utils/sampling.py
```diff
@@ -1002,19 +1013,27 @@ def optimize_posterior_samples(
         paths: Random Fourier Feature-based sample paths from the GP
         bounds: The bounds on the search space.
         raw_samples: The number of samples with which to query the samples initially.
+            Raw samples are cheap to evaluate, so this should ideally be set much higher
+            than num_restarts.
         num_restarts: The number of points selected for gradient-based optimization.
+            Should be set low relative to the number of raw
```
saitcakmak, line 1019:
```suggestion
            Should be set low relative to the number of raw samples.
```

saitcakmak (approved):

test/utils/test_sampling.py
```diff
@@ -616,15 +623,16 @@ def test_optimize_posterior_samples(self):
...
             Y_queries = paths(X)
             # this is when we negate, so the values should be smaller
             if posterior_transform:
                 self.assertTrue(torch.all(f_opt < Y_queries.min(dim=-2).values))
 
             # otherwise, larger
             else:
+                print(torch.all(f_opt > Y_queries.max(dim=-2).values))
+                print(f_opt - Y_queries.max(dim=-2).values)
+                breakpoint()
```
sdaulton, line 635: This needs cleaning up

sdaulton (changes requested):

test/utils/test_sampling.py
hvarfner: Oh shoot, that shouldn't have made it in there. Thanks.

PR #2751, opened 2025-02-20
Performance & runtime improvements to info-theoretic acquisition functions (1/N) - AcqOpt initializer

hvarfner:
## Motivation

(Write your motivation here.)

(Write your answer here.)

## Test Plan

## Related PRs


botorch/optim/initializers.py
```diff
@@ -468,6 +468,91 @@ def gen_batch_initial_conditions(
...
+            [
+                acq_function(x_.to(device=device)).cpu()
+                for x_ in X.split(split_size=batch_limit, dim=0)
+            ],
+            dim=0,
+        )
+    idx = boltzmann_sample(
+        function_values=acq_vals,
+        num_samples=num_restarts,
+        eta=options.get("eta", 2.0),
```
esantorella, line 550: By passing these individually rather than as a dict, we help static analysis tools (and people) see that the code isn't obviously wrong, and prevent unused options from being passed and silently dropped. That can be especially helpful in guarding against typos or when refactoring.

You could then update the call sites to pass `**options` instead of `options` -- personally I'd pass them individually everywhere, but it may be a matter of taste.

```suggestion
    frac_random: float = 0.0,
    batch_limit: int | None = None,
    n_burnin: int = 10000,
    n_thinning: int = 32,
    sample_around_best_sigma: float = 1e-2,
    eta: float = 2.0,
    inequality_constraints: list[tuple[Tensor, Tensor, float]] | None = None,
    equality_constraints: list[tuple[Tensor, Tensor, float]] | None = None,
):
    options = options or {}
    device = bounds.device
    if not hasattr(acq_function, "optimal_inputs"):
        raise AttributeError(
            "gen_optimal_input_initial_conditions can only be used with "
            "an AcquisitionFunction that has an optimal_inputs attribute."
        )
    frac_random: float = options.get("frac_random", 0.0)
    if not 0 <= frac_random <= 1:
        raise ValueError(
            f"frac_random must take on values in (0,1). Value: {frac_random}"
        )

    batch_limit = options.get("batch_limit")
    num_optima = acq_function.optimal_inputs.shape[:-1].numel()
    suggestions = acq_function.optimal_inputs.reshape(num_optima, -1)
    X = torch.empty(0, q, bounds.shape[1], dtype=bounds.dtype)
    num_random = round(raw_samples * frac_random)
    if num_random > 0:
        X_rnd = sample_q_batches_from_polytope(
            n=num_random,
            q=q,
            bounds=bounds,
            n_burnin=options.get("n_burnin", 10000),
            n_thinning=options.get("n_thinning", 32),
            equality_constraints=equality_constraints,
            inequality_constraints=inequality_constraints,
        )
        X = torch.cat((X, X_rnd))

    if num_random < raw_samples:
        X_perturbed = sample_points_around_best(
            acq_function=acq_function,
            n_discrete_points=q * (raw_samples - num_random),
            sigma=options.get("sample_around_best_sigma", 1e-2),
            bounds=bounds,
            best_X=suggestions,
        )
        X_perturbed = X_perturbed.view(
            raw_samples - num_random, q, bounds.shape[-1]
        ).cpu()
        X = torch.cat((X, X_perturbed))

    if options.get("sample_around_best", False):
        X_best = sample_points_around_best(
            acq_function=acq_function,
            n_discrete_points=q * raw_samples,
            sigma=options.get("sample_around_best_sigma", 1e-2),
            bounds=bounds,
        )
        X_best = X_best.view(raw_samples, q, bounds.shape[-1]).cpu()
        X = torch.cat((X, X_best))

    with torch.no_grad():
        if batch_limit is None:
            batch_limit = X.shape[0]
        # Evaluate the acquisition function on `X_rnd` using `batch_limit`
        # sized chunks.
        acq_vals = torch.cat(
            [
                acq_function(x_.to(device=device)).cpu()
                for x_ in X.split(split_size=batch_limit, dim=0)
            ],
            dim=0,
        )
    idx = boltzmann_sample(
        function_values=acq_vals,
        num_samples=num_restarts,
        eta=options.get("eta", 2.0),
```

botorch/optim/initializers.py
```diff
@@ -468,6 +468,91 @@ def gen_batch_initial_conditions(
     return batch_initial_conditions
 
 
+def gen_optimal_input_initial_conditions(
```
esantorella, line 471: Could you add a docstring explaining the behavior of this function?

botorch/optim/initializers.py
```diff
@@ -468,6 +468,91 @@ def gen_batch_initial_conditions(
...
+            bounds=bounds,
+            n_burnin=options.get("n_burnin", 10000),
+            n_thinning=options.get("n_thinning", 32),
+            equality_constraints=equality_constraints,
+            inequality_constraints=inequality_constraints,
+        )
+        X = torch.cat((X, X_rnd))
+
+    if num_random < raw_samples:
+        X_perturbed = sample_points_around_best(
```
esantorella, line 513: It's a bit nonintuitive that we do this even when `sample_around_best` is `False`, no?

esantorella:
Thanks for this! I'm looking forward to seeing the plots.

botorch/optim/initializers.py
hvarfner: Added!

botorch/optim/initializers.py
hvarfner: Possibly! My though was, since it is not actually sampling around the incumbent but around the sampled optima, I could keep it and re-use its logic. I tried to mimic the KG logic for it, and that uses `frac_random` for a similar reason.

botorch/optim/initializers.py
```diff
@@ -468,6 +468,128 @@ def gen_batch_initial_conditions(
...
+    bounds: Tensor,
+    q: int,
+    num_restarts: int,
+    raw_samples: int,
+    fixed_features: dict[int, float] | None = None,
+    options: dict[str, bool | float | int] | None = None,
+    inequality_constraints: list[tuple[Tensor, Tensor, float]] | None = None,
+    equality_constraints: list[tuple[Tensor, Tensor, float]] | None = None,
+):
+    r"""Generate a batch of initial conditions for random-restart optimziation of
```
esantorella, line 482:
```suggestion
    r"""Generate a batch of initial conditions for random-restart optimization of
```

botorch/optim/initializers.py
```diff
@@ -468,6 +468,128 @@ def gen_batch_initial_conditions(
...
+    information-theoretic acquisition functions (PES & JES), where sampled optimizers
+    of the posterior constitute good initial guesses for further optimization. A
+    fraction of initial samples (by default: 100%) are drawn as perturbations around
+    `acq.optimal_inputs`. On average, this drastically decreases the runtime of
+    acquisition function optimization and yields higher-valued candidates by acquisition
+    function value. See https://github.com/pytorch/botorch/pull/2751 for more info.
```
esantorella, line 488:
```suggestion
    function value. See https://github.com/pytorch/botorch/pull/2751 for more info, relative to [...].
```

botorch/optim/initializers.py
esantorella: Thanks!

botorch/optim/initializers.py
```diff
@@ -1154,60 +1277,62 @@ def sample_points_around_best(
...
+                # find pareto set
+                is_pareto = is_non_dominated(f_pred)
+                best_X = X[is_pareto]
+            else:
+                if f_pred.shape[-1] == 1:
+                    f_pred = f_pred.squeeze(-1)
+                n_best = max(1, round(X.shape[0] * best_pct / 100))
+                # the view() is to ensure that best_idcs is not a scalar tensor
+                best_idcs = torch.topk(f_pred, n_best).indices.view(-1)
+                best_X = X[best_idcs]
```
esantorella, line 1334: `best_X` needs to be on the same device as `bounds`. I'm not sure which branch it's coming from, but I'm seeing a failure when `test_gen_optimal_input_initial_conditions` is run with tensors on CUDA.

botorch/acquisition/utils.py
```diff
@@ -576,12 +578,20 @@ def get_optimal_samples(
         sample_transform = None
 
     paths = get_matheron_path_model(model=model, sample_shape=torch.Size([num_optima]))
+    suggested_points = prune_inferior_points(
+        model=model,
+        X=model.train_inputs[0],
```
esantorella, line 583: This doesn't work if the model is a `ModelListGP`, where `train_inputs is a list of tuples of tensors rather than a tuple of tensors: https://github.com/cornellius-gp/gpytorch/blob/b017b9c3fe4de526f7a2243ce12ce2305862c90b/gpytorch/models/model_list.py#L83-L86

Any thoughts on what we should do there?

esantorella (changes requested):

PR #2752, opened 2025-02-20
[Tutorials] Consistently use LaTeX delimiter for Docusaurus

CristianLara:
Docusaurus 3 requires math to be wrapped in $ delimiters. We were trying to be backwards compatible with our existing tutorials by wrapping math blocks in $ delimiters while converting to MDX but properly handling every type of math block (which this was failing to do) is beyond the scope of this script.

Since  these tutorials are intended to be consumed by Docusaurus lets just enforce the $ delimiters when writing tutorials.


saitcakmak (approved):
Thanks! Enforcing correct formatting from start makes sense

PR #2754, opened 2025-02-21
[Feature] Variational Bayesian last layer models as surrogate models

brunzema:
## Motivation

This PR adds variational Bayesian last layers (VBLLs) [1], which demonstrated very promising results in the context of BO in our last paper [2], to BoTorch. The goal is to provide a BoTorch-compatible implementation of VBLL surrogates for standard use cases (single-output models), making them accessible to the community as quickly as possible. This PR does not yet contain all the features discussed in [2] such as the continual learning.  If there is the interest to also add the continual learning, I am happy to add them down the line! 

The VBLLs can be used in standard acquisition functions such as (log)EI but are especially nice for Thompson sampling as the Thompson sample of a Bayesian last layer model is a differentiable standard feed forward neural network which is useful for (almost) global optimization of the sample for the next query location. 

## Implementation details

This PR adds the implementation to the community folders--also here, if there is a large interest in the model, I am happy to help merge them into the main part of the repo. The added files of this PR are the following

```
botorch_community
|-- acquisition
|   |-- bll_thompson_sampling.py # TS for Bayesian last layer models
|-- models
|   |-- vblls.py # BoTorch wrapper for VBLLs
|-- posteriors
|   |-- bll_posterior.py # Posterior class for Bayesian last layer models
notebooks_community
|-- vbll_thompson_sampling.ipynb # Tutorial on how use the VBLL model
test_community
|-- models
|   |-- test_vblls.py # test for the VBLL models functionality (backbone freezing for feature reuse, etc)
```

The current implementation build directly on the [VBLL repo](https://github.com/VectorInstitute/vbll), which is actively maintained and depends only on PyTorch. Using this repo allows improvements—e.g., better variational posterior initialization—to be directly beneficial for BO.

Yes.

## Test Plan

The PR does not change any functionality of the current code base. The core functionality of the VBLLs should be covered by `test_vblls.py`. Let me know if further tests are required.

## Related PRs

This PR does not change functionality and I did not see any PRs regarding last layer models in BoTorch. Maybe this implementation can useful also for other BLLs.

## References

[1] P. Brunzema, M. Jordahn, J. Willes, S. Trimpe, J. Snoek, J. Harrison. [Bayesian Optimization via Continual Variational Last Layer Training](https://arxiv.org/abs/2412.09477). International Conference on Learning Representations (ICLR), 2025.

[2] J. Harrison, J. Willes, J. Snoek. [Variational Bayesian Last Layers](https://arxiv.org/abs/2404.11599). International Conference on Learning Representations (ICLR), 2024.


eytan:
Thanks for this contribution, Paul! This was a very intriguing paper and
it’s great to see it in botch.
We typically don’t take on new external dependencies but if it’s pure
PyTorch that should be OK,, as long as it’s possible for your package to
have some sort of unit or integration test that ensures any future changes
to your package do not cause problems for the botch implementation/
functionality demonstrated in the nb.

Balandat:
Thanks for putting this up, @brunzema. The notebook looks great, and I plan to review this PR in more detail over the next day or two. 

Regarding the dependency on vbll: Right now it looks like the code only uses ~120 lines of pure torch code from the vbll repo (namely this 
https://github.com/VectorInstitute/vbll/blob/main/vbll/layers/regression.py#L34-L149 plus the minimal function https://github.com/VectorInstitute/vbll/blob/main/vbll/utils/distributions.py#L94-L98). It seems questionable to me to take that dependency, especially since the vllb repo doesn't include unit tests and/or CI. We spend a nontrivial amount of time fixing issues with downstream dependencies, so we're really careful about adding them only if really necessary.

My preference would be to move the relevant pieces of the vbll code mentioned above into a helper module (and clearly attribute the source there, of course) so we can avoid the dependency for now. If we do end up expanding the functionality and use additional features from vbll, then I'd be happy to reconsider (provided the vbll repo adds proper unit tests and a CI setup).

botorch_community/acquisition/bll_thompson_sampling.py
```diff
@@ -0,0 +1,133 @@
+from typing import Optional
```
Balandat, line 1: We'll need to add this license header to each of the files in this PR.

```suggestion
#!/usr/bin/env python3
# Copyright (c) Meta Platforms, Inc. and affiliates.
#
# This source code is licensed under the MIT license found in the
# LICENSE file in the root directory of this source tree.

from typing import Optional
```

botorch_community/acquisition/bll_thompson_sampling.py
```diff
@@ -0,0 +1,133 @@
...
+import numpy as np
+
+import scipy
+
+import torch
+from torch.func import grad
+
+from botorch_community.models.vblls import AbstractBLLModel
+
+torch.set_default_dtype(torch.float64)
```
Balandat, line 12: This will have side effects on other code as well. Can you instead explicitly set the dtype at the places in the code (shouldn't be too many occurences).

botorch_community/acquisition/bll_thompson_sampling.py
```diff
@@ -0,0 +1,133 @@
...
+class BLLMaxPosteriorSampling:
+    """
+    Implements Maximum Posterior Sampling for Bayesian Linear Last (VBLL) models.
+
+    This class provides functionality to sample from the posterior distribution of a BLL model,
+    with optional optimization to refine the sampling process.
+
+    Args:
+        model (AbstractBLLModel):
+            The VBLL model from which posterior samples are drawn. Must be an instance of `AbstractBLLModel`.
```
Balandat, line 24: Formatting consistency nit (applies throughout this PR). Also, we generally don't include the type in the docstring since this is evident from the type signature (and will be included automatically in the sphinx API docs).
```suggestion
        model: The VBLL model from which posterior samples are drawn. Must be an instance of `AbstractBLLModel`.
```

botorch_community/acquisition/bll_thompson_sampling.py
```diff
@@ -0,0 +1,133 @@
...
+
+    Notes:
+        - If `bounds` is not provided, the default range [0,1] is assumed for each input dimension.
+        - The lower (`lb`) and upper (`ub`) bounds are stored as CPU tensors for compatibility
+          with initial condition generation and `scipy.optimize.minimize`.
+    """
+
+    def __init__(
+        self,
+        model,
```
Balandat, line 45:
```suggestion
        model: AbstractBLLModel,
```

botorch_community/acquisition/bll_thompson_sampling.py
```diff
@@ -0,0 +1,133 @@
...
+
+
+class BLLMaxPosteriorSampling:
+    """
+    Implements Maximum Posterior Sampling for Bayesian Linear Last (VBLL) models.
+
+    This class provides functionality to sample from the posterior distribution of a BLL model,
+    with optional optimization to refine the sampling process.
+
+    Args:
```
Balandat, line 22: Consistency nit: We usually add the docstring for the `Args` to the docstring of the `__init__()` method.

botorch_community/acquisition/bll_thompson_sampling.py
```diff
@@ -0,0 +1,133 @@
...
+            self.bounds = [(0, 1)] * self.model.num_inputs
+            self.lb = torch.zeros(self.model.num_inputs).cpu()
+            self.ub = torch.ones(self.model.num_inputs).cpu()
+        else:
+            # Ensure bounds are on CPU for compatibility with scipy.optimize.minimize
+            self.lb = bounds[0, :].cpu()
+            self.ub = bounds[1, :].cpu()
+            self.bounds = [tuple(bound) for bound in bounds.T.cpu().tolist()]
+
+    def __call__(self, X_cand: torch.Tensor = None, num_samples: int = 1):
```
Balandat, line 69:
```suggestion
    def __call__(self, X_cand: torch.Tensor = None, num_samples: int = 1) -> torch.Tensor:
```

botorch_community/acquisition/bll_thompson_sampling.py
```diff
@@ -0,0 +1,133 @@
...
+        if self.discrete and X_cand is None:
+            raise ValueError("X_cand must be provided if `discrete` is True.")
+
+        sampled_functions = [self.model.sample() for _ in range(num_samples)]
```
Balandat, line 73: Is there any significance to sampling these first and then looping over them? As compared to the alternative of sampling these one-by-one in the loop below?

botorch_community/acquisition/bll_thompson_sampling.py
```diff
@@ -0,0 +1,133 @@
...
+        X_next = torch.empty(num_samples, self.model.num_inputs, device=self.device)
+
+        # get max of sampled functions at candidate points for each function
+        for i, f in enumerate(sampled_functions):
+            # Note that optimization overwrites sampling-based approach
```
Balandat, line 78: I'm not sure I understand this comment, can you make this more clear?

botorch_community/acquisition/bll_thompson_sampling.py
```diff
@@ -0,0 +1,133 @@
...
+            if not self.discrete:
+                X_cand = torch.empty(
```
Balandat, line 80: Seems like this whole block inside this if statement could be moved into a helper function to make the code more readable.

botorch_community/acquisition/bll_thompson_sampling.py
```diff
@@ -0,0 +1,133 @@
...
+                    res = scipy.optimize.minimize(
+                        func,
+                        x0,
+                        jac=grad_func,
+                        bounds=self.bounds,
+                        method="L-BFGS-B",
+                    )
+
+                    if not res.success:
+                        print(f"Optimization failed with message: {res.message}")
```
Balandat, line 118: should we make this a `logger.warn`?

botorch_community/acquisition/bll_thompson_sampling.py
```diff
@@ -0,0 +1,133 @@
...
+                    # store the candidate
+                    X_cand[j, :] = torch.from_numpy(res.x)
+                    Y_cand[j] = torch.tensor([-res.fun])
+
+                # select the best candidate
+                X_next[i, :] = X_cand[Y_cand.argmax()]
+            else:
+                # sampling based approach on candidate points
+                Y_cand = f(X_cand)
+                X_next[i, :] = X_cand[Y_cand.argmax()]
```
Balandat, line 129:
```suggestion
            else:
                # sampling based approach on candidate points
                Y_cand = f(X_cand)
                
            # select the best candidate
            X_next[i, :] = X_cand[Y_cand.argmax()]
```

botorch_community/models/vblls.py
```diff
@@ -0,0 +1,418 @@
...
+
+from gpytorch.distributions import MultivariateNormal
+
+from botorch.posteriors import Posterior
+from botorch.models.model import Model
+from botorch.posteriors.gpytorch import GPyTorchPosterior
+
+from botorch_community.posteriors.bll_posterior import BLLPosterior
+
+import vbll
```
Balandat, line 36: As per [my earlier comment](https://github.com/pytorch/botorch/pull/2754#issuecomment-2682140981), I would prefer that at this point we copy the small amount of code required from vbll to a helper module.

Balandat (commented):
Unit tests are failing b/c vbll is not installed in the CI. This will not be necessary if we move the minimal code required in a helper module that we include here.

notebooks_community/vbll_thompson_sampling.ipynb
```diff
@@ -0,0 +1,803 @@
...
+    "\n",
+    "Variational Bayesian Last Layer (VBLL) [2] models provide a practical and scalable way to approximate Bayesian inference in neural networks. Instead of placing a prior over all network weights, VBLL only model the uncertainty on the parameters on the last layer while keeping the feature extractor deterministic. This setup allows the model to retain the expressive power of deep learning while maintaining well-calibrated uncertainty estimates.\n",
+    "\n",
+    "In VBLL, we model the output as a generalized linear model with learned features $\\phi_{\\theta}$ as:\n",
+    "$$\n",
+    "    y = \\mathbf{w}^\\top \\phi_{\\theta} (\\mathbf{x}) + \\varepsilon\n",
+    "$$\n",
+    "where $\\varepsilon \\sim \\mathcal{N}(0, \\sigma^2)$.\n",
+    "VBLL models use a variational posterior distribution on the weights as $\\mathbf{w} \\sim q(\\mathbf{w})$ with $q(\\mathbf{w}) = \\mathcal{N}(\\bar{\\mathbf{w}}, S)$ where $\\bar{\\mathbf{w}}$ is the mean and $S$ the full covariance of a multivariate normal distribution (for more information, see [1,2]).\n",
+    "Through conjugany, we yield the posterior predictive\n",
```
Balandat, line 38:
```suggestion
    "Through conjugacy, we yield the posterior predictive\n",
```

notebooks_community/vbll_thompson_sampling.ipynb
```diff
@@ -0,0 +1,803 @@
...
+    "$$\n",
+    "    p(y \\mid \\mathbf{x}, \\mathcal{D}_T, \\theta) = \\mathcal{N} \\big( \\bar{\\mathbf{w}}_T^\\top \\phi_{\\theta}(\\mathbf{x}), \\, \\phi_{\\theta}(\\mathbf{x})^\\top S_T \\phi_{\\theta}(\\mathbf{x}) + \\sigma^{2} \\big).\n",
+    "$$\n",
+    "This predictive can then be used in downsteam tasks such as Bayesian optimization (BO).\n",
+    "\n",
+    "## The VBLLs on Toy Data\n",
+    "\n",
+    "Before using the VBLL model in the context of Bayesian Optimization, let's see how the interface of the model looks. The current implementation allows for passing a full backbone to the model (which could be anything). If no backbone is provided, a standard MLP is created. For training, [1] discusses continual learning approaches which will be added to the implementation, but in the following, we will not go into detail. See the docstring of `model.fit` for the different options for configuring the optimization and the training.\n",
```
Balandat, line 46: anything? Can you be a bit more specific what API it needs to satisfy?

notebooks_community/vbll_thompson_sampling.ipynb
```diff
@@ -0,0 +1,803 @@
...
+    "    in_features=1,\n",
+    "    hidden_features=64,\n",
+    "    num_layers=3,\n",
+    "    out_features=1,\n",
+    ")\n",
+    "\n",
+    "# lets print the model, we can see the MLP backbone and the VBLL regression head\n",
+    "print(model)\n",
+    "\n",
+    "# fit the model on the data -- it is also possible to specify the optimizer throught `optimizer_settings` (see docstring)\n",
```
Balandat, line 139: split this line for readability

notebooks_community/vbll_thompson_sampling.ipynb
```diff
@@ -0,0 +1,803 @@
...
+   "source": [
+    "## Thompson sampling with VBLLs\n",
+    "\n",
+    "VBLLs yield a Gaussian predictive distribution, and thus most acquisition functions that are straightforward to compute for Gaussian Processes (GPs) are also straightforward for VBLLs. Moreover, parametric VBLLs are especially well-suited for Thompson sampling.\n",
+    "\n",
+    "For a Thompson sample, we simply sample from the variational posterior of $\\mathbf{w}$ at iteration $t$ and then construct a sample from the predictive $\\hat{f}$ as a generalized linear model:\n",
+    "\n",
+    "\\begin{aligned}\n",
+    "&\\textcircled{1} \\quad \\hat{\\mathbf{w}} \\sim q(\\mathbf{w}) \\\\\n",
+    "&\\textcircled{2} \\quad \\hat{f} (\\mathbf{x}) \\coloneqq \\hat{\\mathbf{w}}^\\top \\phi_{\\theta} (\\mathbf{x}).\n",
```
Balandat, line 196: This doesn't render properly (at least on the GitHub viewer)

notebooks_community/vbll_thompson_sampling.ipynb
```diff
@@ -0,0 +1,803 @@
...
+    "    x,\n",
+    "    y,\n",
+    "    new_data=None,\n",
+    "    show_opjective=False,\n",
+    "    title=\"VBLL Predictive on Toy Data with Thompson Samples\",\n",
+    "):\n",
+    "    x_test = torch.linspace(0, 1, 100)\n",
+    "    with torch.no_grad():\n",
+    "        posterior = model.posterior(x_test.view(-1, 1))\n",
+    "    mean, std = posterior.mean.squeeze(), posterior.variance.sqrt().squeeze()\n",
```
Balandat, line 231: Nit: Make this two lines for readability

notebooks_community/vbll_thompson_sampling.ipynb
```diff
@@ -0,0 +1,803 @@
...
+     },
+     "metadata": {},
+     "output_type": "display_data"
+    }
+   ],
+   "source": [
+    "test_X = torch.linspace(0, 1, 100)\n",
+    "with torch.no_grad():\n",
+    "    posterior = model.posterior(test_X.view(-1, 1))\n",
+    "mean, std = posterior.mean.squeeze(), posterior.variance.sqrt().squeeze()\n",
```
Balandat, line 172: Nit: Make this two lines for readability

notebooks_community/vbll_thompson_sampling.ipynb
```diff
@@ -0,0 +1,803 @@
...
+     "output_type": "display_data"
+    }
+   ],
+   "source": [
+    "def plot_model(\n",
+    "    model,\n",
+    "    x,\n",
+    "    y,\n",
+    "    new_data=None,\n",
+    "    show_opjective=False,\n",
```
Balandat, line 225:
```suggestion
    "    show_objective=False,\n",
```

notebooks_community/vbll_thompson_sampling.ipynb
```diff
@@ -0,0 +1,803 @@
...
+    "    if new_data is not None:\n",
+    "        x_new, y_new = new_data\n",
+    "        ax.scatter(x_new, y_new, c=\"r\", label=\"New observations\")\n",
+    "\n",
+    "    if show_opjective:\n",
+    "        ax.plot(x_test, objective(x_test, noise=False), label=\"Objective\", color=\"k\")\n",
+    "\n",
+    "    ax.plot(x_test, mean, label=\"Posterior predictive\", color=\"tab:blue\")\n",
+    "\n",
+    "    # Posterior samples\n",
```
Balandat, line 246: The TS samples shown are not the samples that were optimized to obtain the next candidate point. This initially caused some confusion; can you make the plots in a way so that this is consistent (i.e. that we show the set of TSs that were optimized to obtain the next candidate)?

notebooks_community/vbll_thompson_sampling.ipynb
```diff
@@ -0,0 +1,803 @@
...
...J2f+kF0NBhBhxACCiGj8kyoGCSY6Bydien2wmM0DiZgR253w9+lcDt2d6KiCua/nkz8zmTVXwZ0hwcRU+ByZl+2gKUmW9s7aQePCl2sfi8uRwWavHa5DzYf7UO47CrupDyk2E2bmJGFqThqsVlv/iPAvDSR/MoAIIwYQRESxRV6mpIGVBBOyO9GqiZhe+P3QQCLZboVd+j5H5Iv7tdLCLrsTHVWwuoa2MvA5MvWYw50+FX3JhUOPOgxD+1OknN87EIRI6+2uoluuWEra4/Fi1wcfo+3INhT0nZNPhEyLGyXZSZhUkA97YrJWfEhZKgOIMGIAQUQU22SIlQQT2hGzy62JmB6vAZsl8mWilt522J2BYOLSvAnpCeFJnaTBhAzSCu4SyPFISv1+rdiQoEKORKTfgzSEGtw7YrhdmIrKQ+j6cCvQ2w4H3Mg2dSMtOxcl+dmo7U3F3p7JmF9agFWfuwUWy+iPeBhAMIAgIooLfknE9PSXiXYFZnX0uL1ayRHpjpgmn1vzJaSttt1ZPfSoo39Gh+RMuNOn6O6EtNxOrZXeEU2f9I4o/hy8yXlX/Do+rwe1h3ei9fhetHe7kI/ALkhtex+amy/ihT2nYSSkacPFr33ta6P6nhhAMIAgIorbMlGp6pBJoi3BMlHpiGk1R3aSqGHA2tOkzaCk5bWtp3FIiWhgd6IYnpRiLftMajky0DtCdiJkR+KKyZlSNeJqwdHX1qOu6gQy05NRaLqI80Y21u5tx7H9O/Q5mzZtGlUQwQCCAQQRUdyT/hLBJlYylrzTFUjEtEoiZkJkjzpMXldgd0ICCue5gSmaQdIoSgZiWTydOlnTZ0/V3IjerOsu23/C5/fh9tu/jIl99fjSdAcmFBZhQkYy1v3zIA6eatRdl6KiIlRXV4/4OGNcTuMkIiIKJzm6yEiy61WUGZgkKrkTsjMhFQ8dHR6YzSYk2awaUISzqsOwJsKdNVMv3Z1wNWsgoVd3PcxeFwyzDX6TBY7WY5Jdoa23ZZJo+7TVwx5rHK6sRHNTM5phxZn3+3Br/UnMLkzGsYbA0YnsB9TW1mrn52DDxkhiAEFERHE1STQ/PXHgqKO9O9DESkpF/Yah/SYkmAhre22TSZtAydWTt1BzJ2yddUjolICiRpMtpcW2rasOqV11WrXRmz1Hx5YHumJO1M9x8eLFgU/p9pnwem0C3nSm
```
Balandat, line 479: It would be really nice to run both TS and logEI and then show these in a two-panel figure to have the comparison. This would also ensure that both code paths work properly

Balandat, line 552: Also here, comparing TS with logEI would be interesting (though not really meaningful if we only do one sample, but we can point out that caveat prominently.

Balandat, line 588: This seems to be producing a numpy `DeprecationWarning`, can  you update this to avoid this?

Balandat, line 607: I don't know how much extra value this adds over the previous demonstration, esp. if we don't use it to compare performance. I suggest removing the Ackley part to focus on the key messages / examples.

Balandat (commented):
oops, submitted prematurely. Here is the rest of the review.

botorch_community/acquisition/bll_thompson_sampling.py
brunzema: added this to all files 👍

botorch_community/acquisition/bll_thompson_sampling.py
brunzema: changed this here as well as the other occurrence 👍

botorch_community/acquisition/bll_thompson_sampling.py
brunzema: tried to make the docstring as consistent as possible and removed the dtypes

botorch_community/acquisition/bll_thompson_sampling.py
brunzema: no, there is no difference. I now changed it to the one-by-one for readability :)

botorch_community/acquisition/bll_thompson_sampling.py
brunzema: hope the new version clarifies this. I also added a `logger.warning` that if `self.discrete=True` but no `X_cand` are provided, then it will default to numerically optimizing the sample path.

botorch_community/acquisition/bll_thompson_sampling.py
brunzema: agree, with a helper function its much nicer; changed it accordingly 👍

botorch_community/models/vblls.py
brunzema: As discussed, I now included a file with the core vbll functionality that is relevant for BO. 👍

botorch_community/models/vblls.py
```diff
@@ -0,0 +1,418 @@
...
+torch.set_default_dtype(torch.float64)
+
+
+class SampleModel(nn.Module):
+    def __init__(self, backbone: nn.Module, sampled_params: Tensor):
+        super().__init__()
+        self.backbone = backbone
+        self.sampled_params = sampled_params
+
+    def forward(self, x: Tensor) -> Tensor:
```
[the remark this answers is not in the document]
brunzema: Added the dimensions for the forward method here and also for the vbll networks 👍

botorch_community/acquisition/bll_thompson_sampling.py
```diff
@@ -0,0 +1,133 @@
...
+        - If `bounds` is not provided, the default range [0,1] is assumed for each input dimension.
+        - The lower (`lb`) and upper (`ub`) bounds are stored as CPU tensors for compatibility
+          with initial condition generation and `scipy.optimize.minimize`.
+    """
+
+    def __init__(
+        self,
+        model,
+        num_restarts: int = 10,
+        bounds: Optional[torch.Tensor] = None,
```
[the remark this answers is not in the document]
brunzema: Adapted the style in the updated PR

botorch_community/models/vblls.py
```diff
@@ -0,0 +1,418 @@
...
+
+        # Merge defaults with provided settings
+        optimization_settings = (
+            default_opt_settings
+            if optimization_settings is None
+            else {**default_opt_settings, **optimization_settings}
+        )
+
+        # Make dataloader based on train_X, train_y
+        device = torch.device("cuda:0" if torch.cuda.is_available() else "cpu")
```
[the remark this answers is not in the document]
brunzema: Changed it to pulling the device from `self.model` which can explicitly be initialized with a device

botorch_community/models/vblls.py
```diff
@@ -0,0 +1,418 @@
...
+            else:
+                epochs_no_improve += 1
+
+            if epochs_no_improve >= optimization_settings["patience"]:
+                early_stop = True
+
+        # load best model
+        if best_model_state is not None:
+            self.model.load_state_dict(best_model_state)
+            print("Early stopping at epoch ", epoch, " with loss ", best_loss)
```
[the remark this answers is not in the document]
brunzema: nice, yes moved all print statements to the logger

botorch_community/models/vblls.py
```diff
@@ -0,0 +1,418 @@
...
+
+        # Extract mean and variance
+        mean = posterior.mean.squeeze()
+        variance = posterior.variance.squeeze()
+        cov = torch.diag_embed(variance)
+
+        K = self.num_outputs
+        mean = mean.reshape(B, N * K)
+
+        # Cov must be `(B, Q*K, Q*K)`
```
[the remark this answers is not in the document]
brunzema: Ups, yes!

botorch_community/posteriors/bll_posterior.py
```diff
@@ -0,0 +1,54 @@
+from typing import Optional
+
+import torch
+from torch import Tensor
+
+from botorch.posteriors import Posterior
+
+
+class BLLPosterior(Posterior):
+    def __init__(self, posterior, model, X, output_dim):
```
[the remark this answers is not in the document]
brunzema: Included types here, but I had to do a construction with `TYPE_CHECKING` from  the `typing` package to avoid circular imports

botorch_community/models/vblls.py
```diff
@@ -0,0 +1,418 @@
...
+            mean = mean.squeeze(0)
+            cov = cov.squeeze(0)
+
+        # pass as MultivariateNormal to GPyTorchPosterior
+        mvn_dist = MultivariateNormal(mean, cov)
+        post_pred = GPyTorchPosterior(mvn_dist)
+        return BLLPosterior(post_pred, self, X, self.num_outputs)
+
+    @abstractmethod
+    def sample(self, sample_shape: Optional[torch.Size] = None):
```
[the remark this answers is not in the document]
brunzema: added `nn.Module` as return type

test_community/models/test_vblls.py
```diff
@@ -0,0 +1,186 @@
...
+            )
+
+            model = VBLLModel(
+                backbone=copy.deepcopy(test_backbone),  # copy.deepcopy(test_backbone)
+                hidden_features=num_hidden,  # match the output of the backbone
+            )
+
+            X, y = _reg_data_singletask(d)
+            optim_settings = {
+                "num_epochs": 10,
```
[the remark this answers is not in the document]
brunzema: no, not necessary--changed it to 1 and increased the learning rate to make sure we get significant changes

notebooks_community/vbll_thompson_sampling.ipynb
brunzema: ...again, ups and thank you!

notebooks_community/vbll_thompson_sampling.ipynb
brunzema: sounds good, removed Ackley

brunzema: I changed it to loop over both logEI and TS and then show the models for all iterations. I think the model "evolution" is very similar. I wanted to keep the main BO loop as simple as possible rather than optimizing in parallel with both acquisition function for easier reuse of the loop. If you think two-panel figures would be more interesting, I'm happy to make that change!

brunzema:
@Balandat Thank you again for the detailed review--I really appreciate it! I’ve updated the PR to address all your points, but please let me know if any further changes are needed. I’m happy to make any additional updates!

Biggest change is of course the added code from the vbll package, let me know if the way I know included it is ok (additional file + credit at the top).

Balandat:
@brunzema thanks a lot for the updates - will review this within the next couple of days!

botorch_community/acquisition/bll_thompson_sampling.py
```diff
@@ -0,0 +1,158 @@
...
+    # get gradient and create wrapper
+    grad_f = grad(lambda x: f(x).mean())
+
+    def grad_func(x):
+        return -grad_f(torch.from_numpy(x).to(device)).detach().cpu().numpy()
+
+    for j in range(num_restarts):
+        # generate initial condition from within the bounds
+        x0 = np.random.rand(f.num_inputs)
+        x0 = lb + (ub - lb) * x0
```
Balandat, line 144: Nit: Could do this once in a vectorized fashion outside the loop and then just index in the loop (more a µ-optimization though)

botorch_community/acquisition/bll_thompson_sampling.py
```diff
@@ -0,0 +1,158 @@
...
+
+        # optimize sample path
+        res = scipy.optimize.minimize(
+            func, x0, jac=grad_func, bounds=bounds, method="L-BFGS-B"
+        )
+
+        if not res.success:
```
Balandat, line 151: Does this work reliably? I've run into cases where the line search fails in odd ways sometimes, if there is a chance of this happening in any of the samples and we use a lot of samples then this could cause frequent failures. If this doesn't happen then great, o/w we may want to handle this, possibly by retrying the optimization

botorch_community/models/vbll_helper.py
```diff
@@ -0,0 +1,487 @@
...
+        new_cov = self.covariance_weighted_inner_prod(
+            inp.unsqueeze(-3), reduce_dim=False
+        )
+        return Normal(self.loc @ inp, torch.sqrt(torch.clip(new_cov, min=1e-12)))
+
+    def squeeze(self, idx):
+        return DenseNormalPrec(self.loc.squeeze(idx), self.tril.squeeze(idx))
+
+
+cov_param_dict = {
```
Balandat, line 319: Can you move this up to the top of the module and capitalize this to indicate this is a constant?

botorch_community/models/vbll_helper.py
```diff
@@ -0,0 +1,487 @@
...
+def tp(M):
+    return M.transpose(-1, -2)
+
+
+def sym(M):
+    return (M + tp(M)) / 2.0
+
+
+# Credit to https://github.com/brentyi/fannypack/blob/2888aa5d969824ac1e1a528264674ece3f4703f9/fannypack/utils/_math.py
+def cholesky_inverse(u: torch.Tensor, upper: bool = False) -> torch.Tensor:
```
Balandat, line 39: `torch.cholesky_inverse()` supports batch dimensions since early 2021 (https://github.com/pytorch/pytorch/pull/50269), so I don't think function is needed here

botorch_community/models/vbll_helper.py
```diff
@@ -0,0 +1,487 @@
...
+        new_cov = self.covariance_weighted_inner_prod(
+            inp.unsqueeze(-3), reduce_dim=False
+        )
+        return Normal(self.loc @ inp, torch.sqrt(torch.clip(new_cov, min=1e-12)))
+
+    def squeeze(self, idx):
+        return Normal(self.loc.squeeze(idx), self.scale.squeeze(idx))
+
+
+class DenseNormal(torch.distributions.MultivariateNormal):
```
Balandat, line 138: Side note: the term "dense" normal for MVN is a bit confusing here, but I guess that's fine

botorch_community/models/vbll_helper.py
```diff
@@ -0,0 +1,487 @@
...
+    @property
+    def covariance(self):
+        return self.scale_tril @ tp(self.scale_tril)
+
+    @property
+    def inverse_covariance(self):
+        warnings.warn(
+            "Direct matrix inverse for dense covariances is O(N^3), consider using eg inverse weighted inner product"
+        )
+        return tp(torch.linalg.inv(self.scale_tril)) @ torch.linalg.inv(self.scale_tril)
```
Balandat, line 159: Ugh, this is pretty bad. You're calling `inv` on the same tensor twice. Also, you already have the Cholesky decomposition here, why are you calling `inv` in the first place? You can do this with a simple triangular solve.
```suggestion
        Eye = torch.eye(self.scale_tril.shape[-1], device=self.scale_tril.device, dtype=self.scale_tril.dtype)
        W = torch.linalg.solve_triangular(self.scale_tril, Eye, upper=False)
        return tp(W) @ W
```

botorch_community/models/vbll_helper.py
```diff
@@ -0,0 +1,487 @@
...
+        return (self.scale_tril**2).sum(-1).sum(-1)  # compute as frob norm squared
+
+    def covariance_weighted_inner_prod(self, b, reduce_dim=True):
+        assert b.shape[-1] == 1
+        prod = ((tp(self.scale_tril) @ b) ** 2).sum(-2)
+        return prod.squeeze(-1) if reduce_dim else prod
+
+    def precision_weighted_inner_prod(self, b, reduce_dim=True):
+        assert b.shape[-1] == 1
+        prod = (torch.linalg.solve(self.scale_tril, b) ** 2).sum(-2)
```
Balandat, line 176: should use `solve_triangular` here as well

botorch_community/models/vbll_helper.py
```diff
@@ -0,0 +1,487 @@
...
+    def mean(self):
+        return self.loc
+
+    @property
+    def chol_covariance(self):
+        raise NotImplementedError()
+
+    @property
+    def covariance(self):
+        return self.cov_factor @ tp(self.cov_factor) + torch.diag_embed(self.cov_diag)
```
Balandat, line 205: seems like the `torch.distributions.LowRankMultivariateNormal` does this just fine, why overwrite?

botorch_community/models/vbll_helper.py
```diff
@@ -0,0 +1,487 @@
...
+                * noise.trace_precision
+            )
+
+            kl_term = gaussian_kl(W, self.prior_scale)
+            wishart_term = (
+                self.dof * noise.logdet_precision
+                - 0.5 * self.wishart_scale * noise.trace_precision
+            )
+            total_elbo = torch.mean(pred_likelihood - trace_term)
+            total_elbo += self.regularization_weight * (wishart_term - kl_term)
```
Balandat, line 476: in-pace updates can be problematic for backprop, suggest avoiding that
```suggestion
            total_elbo = total_elbo + self.regularization_weight * (wishart_term - kl_term)
```

botorch_community/posteriors/bll_posterior.py
```diff
@@ -0,0 +1,69 @@
...
+
+import torch
+from torch import Tensor
+
+from botorch.posteriors import Posterior, GPyTorchPosterior
+
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from botorch_community.models.vblls import AbstractBLLModel
```
Balandat, line 16: is this still needed if you use `from __future__ import annotations` in all the modules to do postponed type evaluations? We should generally use that.

Balandat (changes requested):
Thanks for the updates and in particular for avoiding another external dependency! I left a number of inline comments, here are some higher level ones: 
1. Please rebase this on a recent version of `master` (the base of commit of this PR is pretty old)
2. Please address flake8 (incl. line length) and import sorting errors - you can install the `pre-commit` hooks to make sure that your code confirms to the standard (see https://github.com/pytorch/botorch/blob/main/CONTRIBUTING.md#pre-commit-hooks)
3. I spotted some some rather problematic numerical code in the vbll helpers, let's update that (highlighted inline), shall we?

brunzema:
@Balandat no, the delay is fully on me, sorry! Everything is pretty much done, flake8 is adressed and also with the import sorting +  `from __future__ import annotations` I no longer require the not so nice type_checking condition 👍 I will make a fresh rebase tomorrow and a final pass and hopefully in the morning your time, I have submitted the updated PR :)

Balandat:
Excellent!

botorch_community/acquisition/bll_thompson_sampling.py
```diff
@@ -0,0 +1,158 @@
...
+            self.bounds = [tuple(bound) for bound in bounds.T.cpu().tolist()]
+
+    def __call__(
+        self, X_cand: torch.Tensor = None, num_samples: int = 1
+    ) -> torch.Tensor:
+        if self.discrete and X_cand is None:
+            raise ValueError("X_cand must be provided if `discrete` is True.")
+
+        if X_cand is not None and not self.discrete:
+            logger.warning(
```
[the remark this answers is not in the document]
brunzema: yeah makes sense, changed it to raise an error

botorch_community/models/vbll_helper.py
brunzema: true, thank you!

botorch_community/models/vbll_helper.py
brunzema: Yeah, I wanted to keep consistency with the vbll package to avoid confusion. But I see how avoiding confusion can lead to more confusion… lol.

botorch_community/models/vbll_helper.py
brunzema: uff, thank you! Will also push this to the vbll repo

botorch_community/models/vbll_helper.py
brunzema: Actually not super sure if there was a deeper reason here, I will double check!

botorch_community/acquisition/bll_thompson_sampling.py
```diff
@@ -0,0 +1,158 @@
...
+        X_next = torch.empty(
+            num_samples, self.model.num_inputs, dtype=torch.float64, device=self.device
+        )
+
+        # get max of sampled functions at candidate points for each function
+        for i in range(num_samples):
+            f = self.model.sample()
+
+            # NOTE: If self.discrete is False, we will always numerically optimize the sample
+            # path for X_next. X_cand is ignored in this case.
```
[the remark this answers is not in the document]
brunzema: Happy to include this down the line! Seems very relevant for incorporating some one-hot encoding-like inputs + some parameters in R^d. I am mainly thinking about some autoML applications. Would it be ok to include this in a later PR?

brunzema:
@Balandat I updated the PR. Not sure why I thought the importing was fixed yesterday—I still had the type checking in place. I noticed that some files in the main repo also use it, so it should be fine?

That said, I think a cleaner approach (which I’ve implemented in the updated PR) is to extract the abstract Bayesian last-layer (BLL) class and use it as a parent class/interface. This should also make it more extensible for future BLL models.

```
botorch_community
|-- acquisition
|   |-- bll_thompson_sampling.py # AbstractBLL to specify "interface"
|-- models
|   |-- blls.py # Define AbstractBLL model
|   |-- vblls.py # VBLL inherit from AbstractBLL
|-- posteriors
|   |-- bll_posterior.py # AbstractBLL to specify "interface"
...
```

botorch_community/acquisition/bll_thompson_sampling.py
```diff
@@ -0,0 +1,173 @@
...
+
+import torch
+
+from botorch.logging import logger
+
+from botorch_community.models.blls import AbstractBLLModel
+from torch.func import grad
+
+
+class BLLMaxPosteriorSampling:
```
Balandat, line 19: There is a [`SamplingStrategy`](https://github.com/pytorch/botorch/blob/main/botorch/generation/sampling.py#L39) interface here that that defines the API for this (and subclasses `torch.nn.Module`) but doesn't support not providing an `X_cand`. I assume this is why you didn't subclass from that? 

I don't think it's necessary to use this if it would require changes to `SamplingStrategy` but maybe leave a comment here that we may want to consider doing that in the future.

botorch_community/acquisition/bll_thompson_sampling.py
```diff
@@ -0,0 +1,173 @@
...
+            )
+        else:
+            # Ensure bounds are on CPU for compatibility with scipy.optimize.minimize
+            self.lb = bounds[0, :].cpu()
+            self.ub = bounds[1, :].cpu()
+            self.bounds = [tuple(bound) for bound in bounds.T.cpu().tolist()]
+
+    def __call__(
+        self, X_cand: torch.Tensor = None, num_samples: int = 1
+    ) -> torch.Tensor:
```
Balandat, line 79: Can you add a docstring describing this method, including the expected input and return shapes of the tensors?

botorch_community/acquisition/bll_thompson_sampling.py
Balandat: By this pointer you mean my comment above? I guess right now you're not erroring out but just warning, so that should be fine. Though if the majority of of the `num_restarts` have an optimization failure that's probably not good.

botorch_community/models/vbll_helper.py
Balandat: Maybe add a comment to the docstring clarifying this?

botorch_community/models/vbll_helper.py
Balandat: any intel on this?

Balandat (commented):
Thanks! Overall this looks great. 

Before I merge this in, could you please increase unit test coverage - looks like quite a lot of things are [not covered](https://github.com/pytorch/botorch/pull/2754/checks?check_run_id=39657261673) by tests, including some important parts such as `BLLPosterior.rsample()`.

The tutorial failure is unrelated, we'll fix this on our end.

botorch_community/acquisition/bll_thompson_sampling.py
brunzema: I've now added an error raise in case none of the initializations result in a successful optimization--just to be sure.

botorch_community/models/vbll_helper.py
brunzema: yeah, as you said it does exactly the same so I removed it. This was probably from earlier iterations of the vbll repo.

brunzema:
hey @Balandat , again took a while--sorry about that! I tried to push the test coverage to 100%. Only places that I am unsure about are the following:

1) After how small discussion, I now added an error to the numerical optimization of the posterior samples paths (See https://github.com/pytorch/botorch/pull/2754#discussion_r2020170659):

```python
def _optimize_sample_path():
    ...
    
    optimization_successful = False
    for j in range(num_restarts):
        # map to bounds
        x0 = lb + (ub - lb) * x0s[j]

        # optimize sample path
        res = scipy.optimize.minimize(
            func, x0, jac=grad_func, bounds=bounds, method="L-BFGS-B"
        )

        # check if optimization was successful
        if res.success:
            optimization_successful = True
        if not res.success:
            logger.warning(f"Optimization failed with message: {res.message}")

        # store the candidate
        X_cand[j, :] = torch.from_numpy(res.x).to(dtype=torch.float64)
        Y_cand[j] = torch.tensor([-res.fun], dtype=torch.float64)

    if not optimization_successful:
        raise RuntimeError("All optimization attempts on the sample path failed.")
        
 ```
Here I am unsure how to actually test this as this never happened to be before. I am inclined to put a `# pragma: no cover` here. What do you think?

2) Should I write a separate test for the vbll helpers? The not covered lines are essentially all properties and the important parts (different parameterization of the VBLL head) are now all covered in the VBLL test. What is your opinion? Best place would be in `utils`?

Balandat:
Awesome, thanks for pushing this along. Once we cover the last few lines with tests this can go in. 

> Here I am unsure how to actually test this as this never happened to be before. I am inclined to put a # pragma: no cover here. What do you think?

I'd recommend having a lightweight mocked test that mocks out `scipy.optimize.minimize` to return a mocked `OptimizeResult` where `success = False`.

> Should I write a separate test for the vbll helpers? The not covered lines are essentially all properties and the important parts (different parameterization of the VBLL head) are now all covered in the VBLL test. What is your opinion? Best place would be in utils?

Yes, having those tests separately would be great. Looks like the misses are mainly in `botorch_community/models/vbll_helper.py`, so I would just test these in a new `test_community/models/test_vbll_helper.py` module.

Balandat:
Thanks @brunzema for the awesome contribution!

eytan:
Thanks Paul for the thorough contribution!  Congrats on landing the paper at ICLR and glad to see we could get it into BoTorch before the conference :)

brunzema:
Thank you guys also for the effort and the thorough reviews—I really appreciate it! Now let’s see what people build with it. 👍  Looking forward to seeing you (and/or others from the team) at ICLR!

Balandat:
> Now let’s see what people build with it. 👍

Indeed, looking forward to it! 

> Looking forward to seeing you (and/or others from the team) at ICLR!

I don't think anyone from the team is attending ICLR this time around - but hope to see you at a future conference!

PR #2761, opened 2025-03-06
Export NumericsWarning and UserInputWarning from botorch.exceptions

chrisyeh96:
## Motivation

Using `botorch.acquisition.ExpectedImprovement` (instead of LogEI) now triggers a `botorch.exceptions.warnings.NumericsWarning`. I'd like to be able to suppress this warning in cases that I actually intend to use EI instead of LogEI, such as when I am comparing the two.

I'd like to be able to write the following:

```python
import warnings

import botorch

with warnings.catch_warnings():
    warnings.filterwarnings('ignore', category=botorch.exceptions.NumericsWarning)
    acqf = botorch.acquisition.ExpectedImprovement(...)
```

However, `botorch.exceptions` does not export `NumericsWarning` as a symbol, so code editors (_e.g._, Visual Studio Code) complain that the symbol `botorch.exceptions.NumericsWarning` is not found. This pull request updates `botorch/exceptions/__init__.py` to export `NumericsWarning` and `UserInputWarning`, the only two warnings from `botorch/exceptions/warnings.py` that were previously unexported.

Yes.

## Test Plan

N/A

## Related PRs

I believe it may have been a simple oversight of commit 6892be9 (from pull request #2429) to add `NumericsWarning` to the list of exported symbols in `botorch/exceptions/__init__.py`.

### DIFF
--- a/botorch/exceptions/__init__.py
+++ b/botorch/exceptions/__init__.py
@@ -22,7 +22,7 @@ from botorch.exceptions.warnings import (
     NumericsWarning,
     OptimizationWarning,
     SamplingWarning,
-    UserInputWarning
+    UserInputWarning,
 )
 
 
--- a/botorch/exceptions/__init__.py
+++ b/botorch/exceptions/__init__.py
@@ -19,8 +19,10 @@ from botorch.exceptions.warnings import (
     BotorchWarning,
     CostAwareWarning,
     InputDataWarning,
+    NumericsWarning,
     OptimizationWarning,
     SamplingWarning,
+    UserInputWarning
 )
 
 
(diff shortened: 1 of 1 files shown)


chrisyeh96:
Tagging @SebastianAment

Balandat:
> However, botorch.exceptions does not export NumericsWarning as a symbol

As an alternative, could you instead just write the following?
```
from botorch.exceptions.warnings import NumericsWarning

with warnings.catch_warnings():
    warnings.filterwarnings('ignore', category=botorch.exceptions.NumericsWarning)
```

Not saying we shouldn't land this PR (seems fine), just curious if the above would work.

chrisyeh96:
@Balandat: Yeah, what you suggested works. I just found it odd that all but two of the warnings could be used via `botorch.exceptions.***Warning`, whereas only `NumericsWarning` and `UserInputWarning` could only be imported as `botorch.exceptions.warnings.***Warning`

botorch/exceptions/__init__.py
```diff
@@ -19,8 +19,10 @@
     BotorchWarning,
     CostAwareWarning,
     InputDataWarning,
+    NumericsWarning,
     OptimizationWarning,
     SamplingWarning,
+    UserInputWarning
```
Balandat, line 25:
```suggestion
    UserInputWarning,
```

PR #2766, opened 2025-03-11
Do not use deprecated last_dim_is_batch=True in OAK kernel

saitcakmak:
Summary:
This was producing a lot of `DeprecationWarning('The last_dim_is_batch argument is deprecated, and will be removed in GPyTorch 2.0. If you are using it as part of AdditiveStructureKernel or ProductStructureKernel, please update your code according to the "Kernels with Additive or Product Structure" tutorial in the GPyTorch docs.')`.

Updated to manually apply the transform rather than relying on it. We can remove support for the kwarg from custom kernel definition once the argument is fully deprecated in GPyTorch.

Differential Revision: D70992085


saitcakmak:
abandoned

PR #2793, opened 2025-03-31
Fix `custom_botorch_model_in_ax` tutorial

Balandat:
Recent changes on the Ax side broke this, fixing the breakage with minimal effort for now to get CI back to a green state.

With some of the Ax API changes in flight, we should revamp this tutorial more generally (and potentially not do this within botorch at all but in Ax instead).


mpolson64:
We should rewrite this to use the Ax API (+ a custom GS), label it a developer guide, publish it on the *Ax* website, then make this page just a link to the Ax tutorial.

Lets figure out who has the bandwidth for this (might be me if we can wait until next week).

mpolson64:
Would also be good to use the new Adapter language in place of Modelbridge etc

PR #2797, opened 2025-04-02
Detach tensor in `gen_candidates_scipy` to avoid test failure due to new warning

Balandat:
Summary:
https://github.com/pytorch/pytorch/pull/143261 added a new warning to pytorch that gets raised when converting a tensor with gradients to a scalar without detaching it first.

This caused some test failues in botorch (which are likely a bit overzealous, but we can fix that separately). This makes a change in `gen_candidates_scipy` to fix this.

There are most likely other occurrences of this in botorch that we should fix, but here I'm quickly addressing the test failures first.

Reviewed By: saitcakmak

Differential Revision: D72286925


Balandat:
Resolved by da6ee3a6e9820e06c79116f71d0242c3db6f189d (unclear why this wasn't auto-closed).

PR #2806, opened 2025-04-04
Rever black upgrade

Balandat:
#2789 reverted a previous black upgrade. This brings the version in the pre-commit config back in sync to avoid breakages in our tests / CI.


saitcakmak:
> https://github.com/pytorch/botorch/pull/2789 reverted a previous black upgrade

https://github.com/pytorch/botorch/pull/2789 seems to be matching pre-commit to requirements-fmt. 

> This brings the version in the pre-commit config back in sync to avoid breakages in our tests / CI.

The two are currently in-sync and this diff is downgrading them both. Is this an internal downgrade that we're trying to match in OSS?

Balandat:
@saitcakmak Yes this one: D72411339

saitcakmak (approved):

PR #2813, opened 2025-04-09
Fix `test_optimize_acqf_mixed_binary_only`

Balandat:
I ran into some test failures locally with the following error:
```
/opt/miniconda3/envs/py313/lib/python3.13/unittest/mock.py:989: in assert_called_once_with
    return self.assert_called_with(*args, **kwargs)
/opt/miniconda3/envs/py313/lib/python3.13/unittest/mock.py:975: in assert_called_with
    if actual != expected:
/opt/miniconda3/envs/py313/lib/python3.13/unittest/mock.py:2645: in __eq__
    return (other_args, other_kwargs) == (self_args, self_kwargs)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = OptimizeAcqfInputs(acq_function=qLogNoisyExpectedImprovement(
  (model): QuadraticDeterministicModel()
  (objective): ...rue, ic_generator=None, timeout_sec=None, return_full_tree=False, retry_on_optimization_warning=True, ic_gen_kwargs={})
other = OptimizeAcqfInputs(acq_function=qLogNoisyExpectedImprovement(
  (model): QuadraticDeterministicModel()
  (objective): ...rue, ic_generator=None, timeout_sec=None, return_full_tree=False, retry_on_optimization_warning=True, ic_gen_kwargs={})

>   ???
E   RuntimeError: Boolean value of Tensor with more than one value is ambiguous
```

This change uses `torch.equal` for `Tensor` object comparisons instead. We may want to consider overwriting `assertEqual()` in the `BotorchTestCase` to use this more generally across the tests (similar to how we implement `assertAllClose` [here](https://github.com/pytorch/botorch/blob/main/botorch/utils/testing.py#L93C9-L93C23)).

PR #2814, opened 2025-04-09
Test against python 3.13

Balandat:
Python 3.13 is out and used, so we should test against it.

This also require updating the minimum supported torch version in the install script. 

Side note: We should probably aim to bump the minium required pytorch version to 2.2.0 in new gpytorch / linear_operator / botorch releases so that we don't have to deal with the numpy<2.0 issues going forward.


Balandat:
Somehow the test failure only shows up on py3.13, presumably due to some torch version difference? Either way, #2813 fixes the test.

PR #2815, opened 2025-04-09
`[q]LogProbabilityOfFeasibility`

SebastianAment:
Summary: This commit adds `LogProbabilityOfFeasibility` and the corresponding batch Monte-Carlo version `qLogProbabilityOfFeasibility`. These new acquisition functions compute a probability of feasibility acquisition value that is particularly relevant for optimization runs that start out without any feasible observations. After a feasible observation is found, the acquisition function likely becomes over-exploitative, and one is better served switching to an acquisition function that also takes into account the objective value, e.g. `qLogNEI`.

Differential Revision: D72411936

### DIFF
--- a/botorch/acquisition/analytic.py
+++ b/botorch/acquisition/analytic.py
@@ -13,7 +13,7 @@ from __future__ import annotations
 
 import math
 
-from abc import ABC
+from abc import ABC, abstractmethod
 from contextlib import nullcontext
 from copy import deepcopy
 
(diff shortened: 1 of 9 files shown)


jduerholt:
Hi @SebastianAment,

this looks very useful, especially in the case described above. 

Just one question/remark: Will you also add the acqf to `get_acquisition_function` in `factory.py`? Or do you plan to deprecate this method in favor of the acqf constructors in the future? If yes, do you have some kind of documentation on how to use this constructors properly? Internally we are currently heavily relying on `get_acquisition_function` as it is a nice convenience function ;)

Best,

Johannes

Issue #2819, opened 2025-04-10
[Bug]: MVNXPB doesn't support multiple batch dimensions


JackBuck:
### What happened?

The documentation for `MVNXPB` indicates that `covariance_matrix` and `bounds` accept a `batch_shape`. However, when called with a `batch_shape` of more than one dimension it blows up. I would expect it to give the same as calling the algorithm in a loop for each element of the batch. Alternatively, the docs could specify only that one batch dimension is allowed, and the code should raise an informative error if more than one is passed.
https://botorch.readthedocs.io/en/stable/utils.html#botorch.utils.probability.mvnxpb.MVNXPB

### Please provide a minimal, reproducible example of the unexpected behavior.

```python
import torch
from botorch.utils.probability import MVNXPB

# Create inputs with a batch_shape of (2,2)
covmat = torch.diag(torch.tensor([1.0, 2.0, 3.0])**2).expand(2, 2, -1, -1)
bounds = torch.tensor(
    [
        [
            [[0, torch.inf], [0, torch.inf], [0, torch.inf]],
            [[-torch.inf, 0], [-torch.inf, 0], [-torch.inf, 0]],
        ],
        [
            [[1, torch.inf], [1, torch.inf], [1, torch.inf]],
            [[-torch.inf, -1], [-torch.inf, -1], [-torch.inf, -1]],
        ],
    ]
)

# Create and call the solver
solver = MVNXPB(covmat, bounds)
log_probs = solver.solve()
```

### Please paste any relevant traceback/logs produced by the example provided.

```shell
IndexError                                Traceback (most recent call last)
Cell In[1], line 21
     19 # Create and call the solver
     20 solver = MVNXPB(covmat, bounds)
---> 21 log_probs = solver.solve()

File .../lib/python3.13/site-packages/botorch/utils/probability/mvnxpb.py:236, in MVNXPB.solve(self, num_steps, eps)
    234     self.step += 1
    235     if should_update_chol:
--> 236         piv_chol.update_(eps=eps)
    238 # Factor in univariate probability if final term fell outside of a block.
    239 if self.step % 2:

File .../lib/python3.13/site-packages/botorch/utils/probability/linalg.py:128, in PivotedCholesky.update_(self, eps)
    126 rank1 = (rank1 * rank1.transpose(-1, -2)).tril()
    127 L[..., i + 1 :, i + 1 :] = L[..., i + 1 :, i + 1 :].clone() - rank1
[...]

PR #2823, opened 2025-04-13
Fix batch computation in Pivoted Cholesky

ltiao:
Summary:
## Summary 

TODO: 

## Changes

Updates a line containing indexing logic that breaks when `len(batch_shape) > 1`

Differential Revision: D72906531


ltiao:
Closes #2819

Issue #2403, opened 2024-06-29
[Documentation] Inaccurate and incomplete documentation for `SingleTaskMultiFidelityGP`


esantorella:
* The docstring for `linear_truncated` and the class docstring describe what the behavior would be if the default were `linear_truncated=False`. However, the default is `True`. The model uses a `LinearTruncatedFidelityKernel`, which isn't described at all.
* It is not clear how data fidelities are to be interpreted. Does a higher number indicate higher fidelity? Can they lie outside [0, 1]? Do they have a cardinal or ordinal interpretation?

This came up in #2393 , and I spent a while myself trying to understand how fidelities should be specified and how they affect the kernel.

Muhammad-Rebaal:
Hi @esantorella !

Could you please assign this issue to me ? I'd like to work on this .

Thank You !

esantorella:
@Muhammad-Rebaal Done, thank you!

PR #2847, opened 2025-05-11
[Documentation] fix Inaccurate and incomplete documentation for `SingleTaskMultiFidelityGP`

Muhammad-Rebaal:
closes [#2403](https://github.com/pytorch/botorch/issues/2403)

Hi @esantorella !
I've Improved the documentation related to `SingleTaskMultiFidelityGP` and `LinearTruncatedFidelityKernel` to clearly express that higher numerical values represent higher fidelities. Add the details about kernel behavior with fidelity values, including internal clamping to [0,1] range and correlation structure. Could you please review my PR .
Thank You !

### DIFF
--- a/botorch/models/kernels/linear_truncated_fidelity.py
+++ b/botorch/models/kernels/linear_truncated_fidelity.py
@@ -40,6 +40,15 @@ class LinearTruncatedFidelityKernel(Kernel):
         polynomial kernel between `x_1[..., [f_1, f_2]]` and
         `x_2[..., [f_1, f_2]]`.
 
+    Note:
+        Fidelity interpretation:
+        - Higher numerical values represent higher fidelities (e.g., 3 > 2 > 1 > 0)
+        - The kernel implementation internally clamps fidelity values to the range [0,1]
+        - The kernel is designed so that higher fidelity points have higher correlation
+            with each other than with lower fidelity points
+        - When using with SingleTaskMultiFidelityGP, ensure your fidelity values
+            follow this convention (higher numeric value = higher fidelity)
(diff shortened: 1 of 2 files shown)


botorch/models/kernels/linear_truncated_fidelity.py
```diff
@@ -40,6 +40,15 @@ class LinearTruncatedFidelityKernel(Kernel):
         polynomial kernel between `x_1[..., [f_1, f_2]]` and
         `x_2[..., [f_1, f_2]]`.
 
+    Note:
+        Fidelity interpretation:
+        - Higher numerical values represent higher fidelities (e.g., 3 > 2 > 1 > 0)
```
Balandat, line 45: Not a good example if the fidelity values are clamped
```suggestion
        - Higher numerical values represent higher fidelities (e.g., 0.7 > 0.5 > 0)
```

botorch/models/kernels/linear_truncated_fidelity.py
```diff
@@ -40,6 +40,15 @@ class LinearTruncatedFidelityKernel(Kernel):
...
+        - The kernel implementation internally clamps fidelity values to the range [0,1]
+        - The kernel is designed so that higher fidelity points have higher correlation
+            with each other than with lower fidelity points
```
Balandat, line 48: This seems like it would be the case for most if not all kernels? The kernel behavior is stationary w.r..t the fidelity parameters, right? As in, the covariance between 0.1 and 0.2 will be the same as the covariance between 0.8 and 0.9?

PR #2854, opened 2025-05-21
add model to __all__

jduerholt:
## Motivation

`RobustRelevancePursuitSingleTaskGP` is not listed in `__all__` in `bofire/models/__init__.py`. This PR adds it.

Yes.

## Test Plan

Not needed.

### DIFF
--- a/botorch/models/__init__.py
+++ b/botorch/models/__init__.py
@@ -27,6 +27,9 @@ from botorch.models.model import ModelList
 from botorch.models.model_list_gp_regression import ModelListGP
 from botorch.models.multitask import KroneckerMultiTaskGP, MultiTaskGP
 from botorch.models.pairwise_gp import PairwiseGP, PairwiseLaplaceMarginalLogLikelihood
+from botorch.models.robust_relevance_pursuit_model import (
+    RobustRelevancePursuitSingleTaskGP,
+)
 
 __all__ = [
     "add_saas_prior",
@@ -49,4 +52,5 @@ __all__ = [
     "SingleTaskGP",
     "SingleTaskMultiFidelityGP",
     "SingleTaskVariationalGP",
+    "RobustRelevancePursuitSingleTaskGP",
 ]


jduerholt:
Hmm, one gets circular imports when one adds it, I assume this was the reason for not adding it in the first place ..., so I think, I should close this PR again, or what do you think?

Balandat:
Hmm, interesting. I though this would be fixable by changing the `from botorch.models import SingleTaskGP` in https://github.com/pytorch/botorch/blob/main/botorch/models/robust_relevance_pursuit_model.py#L40 to `from botorch.models.gp_regression import SingleTaskGP` but that in itself will cause a circular import. 

@SebastianAment since you wrote this code, do you have a better sense of what the issue here is?

jduerholt:
It is also not important at all. We just encountered  that the `RobustRelevancePursuitSingleTaskGP` has to be imported in a different way than the other models and I thought it should be easily fixable so I fired up this PR without even running the tests locally ..., but this has no priority for us at all, we import now directly from the source file ...

Balandat:
Yeah - though the fact that things break even if you don't do this but instead adjust the import as in https://github.com/pytorch/botorch/pull/2854#issuecomment-2897994321 is disconcerting - we should understand what causes this and fix it.

SebastianAment:
> It is also not important at all. We just encountered  that the `RobustRelevancePursuitSingleTaskGP` has to be imported in a different way than the other models and I thought it should be easily fixable so I fired up this PR without even running the tests locally ..., but this has no priority for us at all, we import now directly from the source file ...

I suspect this is because I added the relevance pursuit model to the multiple dispatch of fit_gpytorch_mll. That file likely introduces the import issue, but haven't checked, currently on vacation.

Balandat:
Hmm yeah that would make sense. Let's take a look and see if that's the case when you're back; we may want to think about separating out the registries in a way that would avoid these circular import issues.

Issue #2864, opened 2025-06-04
[FEATURE REQUEST]: Support for categoricals in `optimize_acqf_mixed_alternating`


TobyBoyne:
### Motivation

Currently, I don't believe BoTorch supports mixed optimization over mixed spaces with categorical features. I would like to extend `optimize_acqf_mixed_alternating` to also optimize over such spaces. This would essentially be closing the following TODO:

https://github.com/pytorch/botorch/blob/d247a33486c22bc5c6f16107db3e71f1e3bf3acd/botorch/optim/optimize_mixed.py#L576

### Describe the solution you'd like to see implemented in BoTorch.

During the discrete step, this function currently proposes small perturbations in the 1-Manhattan distance ball around the incumbent. We can similarly propose perturbations in the 1-Hamming distance ball around the incumbent for categorical features. That is, we consider two points to be neighbours if all-but-one of their features are the same, and the remaining (categorical) feature takes any two distinct values.

### Describe any alternatives you've considered to the above solution.

1. You could enumerate over all possible combinations of categorical variables, then solve the continuous-only problem for each combination. This guarantees the best performance, but is costly, and infeasible in high-dim settings. For example, the popular Pest Control benchmark has 5^25 possible combinations.
2. You could treat the categorical features as ordinal discrete features. However, this imposes order to the categories that could harm the optimization.

### Is this related to an existing issue in BoTorch or another repository? If so please include links to those Issues here.

I have a working implementation in my fork of BoTorch, at https://github.com/TobyBoyne/botorch/tree/optimize-cat-alternating. I would be happy to develop the tests, and create a pull request.

### Pull Request

Yes

### Code of Conduct

- [x] I agree to follow BoTorch's Code of Conduct

Balandat:
@TobyBoyne this sounds great, we'd love to accept a PR with this functionality. Just submit and we'll review - I peeked at your code and the high level design seems sensible, we can discuss implementation details in the review.

TobyBoyne:
Added in #2866

PR #2866, opened 2025-06-04
Support categoricals in alternating optimization

TobyBoyne:
## Motivation

See issue #2864 

Yes

## Test Plan

There are already tests in place for mixed integer feature spaces. I have created a new test case for mixed categorical feature spaces. Once the overall approach has been reviewed, I will also add tests for the new functions that have been introduced

## Related PRs

None yet - happy to update the docs.


TobyBoyne:
Some questions I have about my implementation:
1. What should be the expected behaviour if the search space is not mixed (eg. everything is continuous, or everything is categorical)? I have added in a few checks that were necessary for spaces that are categorical+continuous, but no discrete, but I'm not sure which cases need to be well defined.
2. Should I introduce any additional keys to `options` to allow for more fine-grained control?
3. What should be the fallback when there are too many categories? For discrete ordinal features, continuous relaxation is in place.
4. I found some issues when using this with the OneHotToNumeric transform (see the FIXME comment in `optimize_mixed`. I would expect `get_X_baseline` to return the untransformed (ie. one-hot) inputs, but it seems to return the transformed

botorch/optim/optimize_mixed.py
```diff
@@ -164,10 +164,60 @@ def get_nearest_neighbors(
     return unique_neighbors
 
 
+def get_categorical_neighbors(
+    current_x: Tensor,
+    bounds: Tensor,
+    cat_dims: Tensor,
+):
```
Balandat:
```suggestion
) -> Tensor:
```

botorch/optim/optimize_mixed.py
```diff
@@ -552,6 +648,7 @@ def optimize_acqf_mixed_alternating(
     acq_function: AcquisitionFunction,
     bounds: Tensor,
     discrete_dims: list[int],
```
Balandat: would we want to make `discrete_dims` `None`able as well if there are only categorical dims?

botorch/optim/optimize_mixed.py
```diff
@@ -676,22 +773,29 @@ def optimize_acqf_mixed_alternating(
     tkwargs: dict[str, Any] = {"device": bounds.device, "dtype": bounds.dtype}
     # Remove fixed features from dims, so they don't get optimized.
     discrete_dims = [dim for dim in discrete_dims if dim not in fixed_features]
-    if len(discrete_dims) == 0:
+    cat_dims = cat_dims or []
+    cat_dims = [dim for dim in cat_dims if dim not in fixed_features]
+    non_cont_dims = [*discrete_dims, *cat_dims]
```
Balandat, line 801: we should probably validate that `cat_dims` and `discrete_dims` are disjoint here

Balandat: nvm I guess we're doing that below

botorch/optim/optimize_mixed.py
```diff
@@ -676,22 +773,29 @@ def optimize_acqf_mixed_alternating(
...
-        and min(discrete_dims) >= 0
-        and max(discrete_dims) <= dim - 1
+        isinstance(non_cont_dims, list)
+        and len(set(non_cont_dims)) == len(non_cont_dims)
+        and min(non_cont_dims) >= 0
+        and max(non_cont_dims) <= dim - 1
     ):
         raise ValueError(
-            "`discrete_dims` must be a list with unique integers "
+            "`discrete_dims` and `cat_dims` must be lists with unique integers "
```
Balandat:
```suggestion
            "`discrete_dims` and `cat_dims` must be lists with unique, disjoint integers "
```

Balandat (commented):
> What should be the expected behaviour if the search space is not mixed (eg. everything is continuous, or everything is categorical)? I have added in a few checks that were necessary for spaces that are categorical+continuous, but no discrete, but I'm not sure which cases need to be well defined.

I think if either it's fully continuous or fully discrete we should just raise an error. We have other acquisition optimizers for that.

Balandat (commented):
> Should I introduce any additional keys to options to allow for more fine-grained control?

Which ones would those be? In general I think unless we have a need for them we can keep things simple and add them later if needed. 

> What should be the fallback when there are too many categories? For discrete ordinal features, continuous relaxation is in place.

Good question. If we can't enumerate them all, then a natural thing to do would be to sample a subset in each dimension in each step rather than computing all.

botorch/optim/optimize_mixed.py
```diff
@@ -403,10 +474,16 @@ def generate_starting_points(
         and (X_baseline := get_X_baseline(acq_function=opt_inputs.acq_function))
         is not None
     ):
+        # FIXME: get_X_baseline should return untransformed inputs, but when
+        # OneHotToNumeric is used, it returns the transformed inputs.
+        X_baseline = opt_inputs.acq_function.model.input_transform.untransform(
+            X_baseline
+        )
```
Balandat: Hmm interesting. I am not super familiar with this part of the code but it seems that this handling here is not working the way it should: https://github.com/pytorch/botorch/blame/d247a33486c22bc5c6f16107db3e71f1e3bf3acd/botorch/optim/utils/acquisition_utils.py#L126-L129

Looking through the code, it seems that `_has_transformed_inputs` and `_has_transformed_inputs` is used very sparingly. This requires some more thorough audit, I think. 

The solution you have here will not work if the model doesn't actually have an `input_transform` attribute.

Balandat:
cc @saitcakmak who may be interested in this feature

botorch/optim/optimize_mixed.py
TobyBoyne: My solution was very much a temporary band-aid for something I was trying to run. I'm happy to remove this line, as I agree that it won't work if there are no `input_transform`s (and I think it might apply some input transforms twice). I will leave a TODO in the code noting that this sometimes breaks with bad input transforms.

TobyBoyne:
> I think if either it's fully continuous or fully discrete we should just raise an error. We have other acquisition optimizers for that.

I agree, although making this change does break this existing test, and therefore might break expected behaviour for some BoTorch users?

https://github.com/pytorch/botorch/blob/d247a33486c22bc5c6f16107db3e71f1e3bf3acd/test/optim/test_optimize_mixed.py#L535-L547

botorch/optim/optimize_mixed.py
Balandat: You should check if `opt_inputs.acq_function.model` has an `input_transform` attribute and do nothing if it doesn't.

Have you tried this with some other input transforms? Is this only an issue with `OneHotToNumeric`?

Balandat:
> I agree, although making this change does break this existing test, and therefore might break expected behaviour for some BoTorch users?

Good point - let's just fall back to to the standard `_optimize_acqf` in the same way in this case: https://github.com/pytorch/botorch/blob/d247a33486c22bc5c6f16107db3e71f1e3bf3acd/botorch/optim/optimize_mixed.py#L679-L680

botorch/optim/optimize_mixed.py
TobyBoyne: I've investigated further and it seems to not be a problem on BoTorch's end - when I make models only in BoTorch, I can't recreate the error with any kind of transform (OneHotToNumeric included). It's only when I'm using external code that I seem to get these issues.

I'll keep looking into this separately, but the issue is not with this PR (or even with BoTorch).

botorch/optim/optimize_mixed.py
```diff
@@ -164,10 +164,75 @@ def get_nearest_neighbors(
...
+    bounds: Tensor,
+    cat_dims: Tensor,
+    max_num_cat_values: int = MAX_DISCRETE_VALUES,
+) -> Tensor:
+    r"""Generate all 1-Hamming distance neighbors of a given input. The neighbors
+    are generated for the categorical dimensions only.
+
+    We assume that all categorical values are equidistant. If the number of values
+    is greater than `max_num_cat_values`, we sample uniformly from the range of
+    possible values for that dimension.
```
Balandat:
```suggestion
    is greater than `max_num_cat_values`, we sample uniformly from the
    possible values for that dimension.
```

botorch/optim/optimize_mixed.py
```diff
@@ -164,10 +164,75 @@ def get_nearest_neighbors(
...
+    Args:
+        current_x: The design to find the neighbors of. A tensor of shape `d`.
+        bounds: A `2 x d` tensor of lower and upper bounds for each column of `X`.
+        cat_dims: A tensor of indices corresponding to categorical parameters.
+        max_num_cat_values: Maximum number of values for a categorical parameter,
+            beyond which values are uniformly sampled.
+
+    Returns:
+        A tensor of shape `num_neighbors x d`, denoting all unique 1-Hamming
+        distance neighbors.
```
Balandat: should adjust this in light of the `max_num_cat_values` arg above.

botorch/optim/optimize_mixed.py
```diff
@@ -164,10 +164,75 @@ def get_nearest_neighbors(
...
+    """
+
+    # Neighbors are generated by considering all possible values for each
+    # categorical dimension, one at a time.
+    def _get_cat_values_tensor(lb: Tensor, ub: Tensor) -> Tensor:
```
Balandat: Implementing this in torch seems a bit unnecessary given that these are all going to be scalars, right? Could just work with python primitives instead here and probably have lower overhead and less dealing with dtypes and devices

botorch/optim/optimize_mixed.py
```diff
@@ -164,10 +164,75 @@ def get_nearest_neighbors(
     return unique_neighbors
 
 
+def get_categorical_neighbors(
```
Balandat, line 169: very nice!

botorch/optim/optimize_mixed.py
```diff
@@ -573,12 +686,12 @@ def optimize_acqf_mixed_alternating(
...
     be optimized using continuous relaxation.
 
-    # TODO: Support categorical variables.
-
     Args:
         acq_function: BoTorch Acquisition function.
         bounds: A `2 x d` tensor of lower and upper bounds for each column of `X`.
         discrete_dims: A list of indices corresponding to integer and binary parameters.
+        cat_dims: A list of indices corresponding to categorical parameters.
+            If `None`, no categorical parameters are assumed.
```
Balandat: Seems unnecessary, and we don't have this in the docstring for `discrete_dims` either
 ```suggestion
```

File: botorch/optim/optimize_mixed.py [hunk 11]
```diff
@@ -551,7 +663,8 @@ def continuous_step(
 def optimize_acqf_mixed_alternating(
```
Balandat, line 666: Let's update the docstring of the function to reflect the broader support, this currently says
> Optimizes acquisition function over mixed binary and continuous input spaces.

test/optim/test_optimize_mixed.py
```diff
@@ -148,6 +149,31 @@ def test_get_nearest_neighbors(self) -> None:
...
+        self.assertTrue(
+            torch.equal(
+                expected_neighbors.sort(dim=0).values,
+                get_categorical_neighbors(
+                    current_x=current_x, bounds=bounds, cat_dims=cat_dims
+                )
+                .sort(dim=0)
+                .values,
+            )
+        )
```
Balandat: readability nit
```suggestion
        neighbors = get_categorical_neighbors(
            current_x=current_x, bounds=bounds, cat_dims=cat_dims
        )
        self.assertTrue(
            torch.equal(
                expected_neighbors.sort(dim=0).values,
                neighbors.sort(dim=0).values,
            )
        )
```

test/optim/test_optimize_mixed.py
```diff
@@ -721,6 +763,140 @@ def test_optimize_acqf_mixed_integer(self) -> None:
...
+            {"maxiter": 100, "batch_limit": 5, "init_batch_limit": 20},
+        )
+
+        # Testing that continuous perturbations lead to lower acquisition values.
+        perturbed_candidates = candidates.clone()
+        perturbed_candidates[..., cont_dims] += 1e-2 * torch.randn_like(
+            perturbed_candidates[..., cont_dims], device=self.device
+        )
+        perturbed_candidates[..., cont_dims].clamp_(0, 1)
+        self.assertLess((acqf(perturbed_candidates) - acqf(candidates)).max(), 1e-12)
```
Balandat, line 847: did you do some stress runs for this? Not locking a seed can cause flaky tests - I don't think this is a big concern here since you're actually adding some tolerance below, but who knows what kind of numerical weirdnesses can happen here.

botorch/optim/optimize_mixed.py
```diff
@@ -164,10 +164,75 @@ def get_nearest_neighbors(
...
+    # Neighbors are generated by considering all possible values for each
+    # categorical dimension, one at a time.
+    def _get_cat_values_tensor(lb: Tensor, ub: Tensor) -> Tensor:
+        r"""Get a tensor of all of the values a categorical feature may take."""
+        if ub - lb + 1 <= max_num_cat_values:
+            return torch.arange(
+                lb, ub + 1, dtype=current_x.dtype, device=current_x.device
+            )
+        else:
+            return (
```
Balandat: As codecov says, this is currently not being tested - can you please add a test for the case with too many categorical values as well?

Balandat (changes requested):
This looks great, thanks a lot for the high quality contribution! 

I left a few nits inline; the main gap is to add a test for the case of too many categories (where we sample instead).

test/optim/test_optimize_mixed.py
TobyBoyne: This test mostly just follows the structure of the other tests. Only `mixed_binary_only` sets the seed currently. I've added `torch.manual_seed(0)` to all of the relevant tests just in case it causes any issues in the future.

TobyBoyne:
I think that should be everything, @Balandat! Thanks for the quick and thorough review!

test/optim/test_optimize_mixed.py
```diff
@@ -148,6 +149,48 @@ def test_get_nearest_neighbors(self) -> None:
...
+
+        neighbors = get_categorical_neighbors(
+            current_x=current_x,
+            bounds=bounds,
+            cat_dims=cat_dims,
+            max_num_cat_values=MAX_DISCRETE_VALUES,
+        )
+        # We expect the maximum number of neighbors in the first dim, and 8
+        # neighbors in the second dim.
+        self.assertTrue(neighbors.shape == torch.Size([MAX_DISCRETE_VALUES + 8, 2]))
```
Balandat, line 192: The sampling here introduces non-determinism, which is why the unit test failed in a couple of the runs. Can you lock the seed so that this becomes deterministic?

TobyBoyne: Ah, this actually helped me find a bug in the logic. When we randomly subsample, there is a chance that we sample the value at `current_x`. If we do, then after filtering out any values that are equal to `current_x`, then we end up with `MAX_DISCRETE_VALUES - 1` values for that dimension. I'll fix this bug now!

TobyBoyne: Okay, having fixed this bug, the test now should always pass, independent of the random seed. I have locked the seed as well so that the test is deterministic.

PR #2882, opened 2025-06-16
Probabilistic Reparameterization Part 2

TobyBoyne:
## Motivation

Probabilistic Reparameterization is a technique for optimizing mixed feature spaces, proposed in [1]. There is an existing pull request (#1533), which requires resolving merge conflicts, and writing tests. This PR continues that work.

[1] Samuel Daulton et al. Bayesian Optimization over Discrete and Mixed Spaces via Probabilistic Reparameterization, NeurIPS '22

Yes

## Test Plan

I have currently written tests that optimize the acquisition function. I will also write more fine grained tests: specifically, testing the structure of the chained transforms generated in `factory.py`, and testing the new transforms.

## Related PRs

This continues the work in #1533. The tutorial is in #1534, which also would need updating.


TobyBoyne:
Currently, the test comparing the Analytic vs MC methods for PR do not give the same (or close to the same) results, which doesn't seem correct to me. I will keep investigating, I imagine it is an issue with my code since the results in the paper seem to be fairly consistent between Analytic and MC (it seems to be that the Analytic approach is giving 0 acqf value everywhere). If anyone does spot any mistakes, please let me know!

For reference, the MC method in this test does give similar results to completely enumerating the mixed features.

esantorella:
cc @sdaulton

eytan:
Great to see you picking this up @TobyBoyne—thanks so much for your contributions to the project!

It'll be cool to see PR working with BARK—seems like a natural way to get around using MIP and a good use of those MCMC samples!

TobyBoyne:
Some questions I've found while working on this:
1. I have included the tests for the `*PRInputTransform` with the other PR tests. Should they be moved to `test/models/transform/input.py`? Similarly, should these transforms even be with the other `InputTransform`s, since they are highly specific to PR?
2. Flexibility in dimension position - we currently assume/enforce that the dimensions are ordered continuous->discrete->categorical. Should this be relaxed, or is it okay to leave for the sake of cleaner code?
3. Flexibility in encoding - PR currently only supports one-hot encoding for categoricals. In #2866, I assumed that categoricals were encoded ordinally. Is there a canonical/preferred encoding for categoricals? Should both be supported?

TobyBoyne:
> Analytic vs MC methods for PR do not give the same (or close to the same) results

For anyone interested, this was due to PR implicitly assuming that integer idxs are in the rightmost columns (after continuous), whereas the MixedAckley benchmark places them in the leftmost columns.

sdaulton:
> I have included the tests for the *PRInputTransform with the other PR tests. Should they be moved to test/models/transform/input.py? 

It seems fine to leave them with the other PR tests.

> Similarly, should these transforms even be with the other InputTransforms, since they are highly specific to PR?

I think it makes sense to put them in a separate file (perhaps acquisition/probabilistic_reparameterization.py), since they are only used via `AnalyticProbabilisticReparameterization` and `MCProbabilisticReparameterization`.


> Flexibility in dimension position - we currently assume/enforce that the dimensions are ordered continuous->discrete->categorical. Should this be relaxed, or is it okay to leave for the sake of cleaner code?

Ideally this would be relaxed, but I think we can leave that as a TODO for now if it adds to much complexity. If we do leave the ordering assumption, let's make sure we raise exceptions if the ordering is violated.


> Flexibility in encoding - PR currently only supports one-hot encoding for categoricals. In https://github.com/pytorch/botorch/pull/2866, I assumed that categoricals were encoded ordinally. Is there a canonical/preferred encoding for categoricals? Should both be supported?

We should only be using one-hot encoding (or potentially embeddings) for categoricals. The ordinal probabilistic reparameterization proposed in the paper really only makes sense if there is a inherent ordering to the categorical values (even if the GP is using say a categorical kernel), since to move from theta=3->theta=1 you would have to pass through 2, if an ordinal representation is used. One-hot is much more intuitive and avoids that scenario.

botorch/acquisition/probabilistic_reparameterization.py
```diff
@@ -0,0 +1,1219 @@
...
+
+    This will typically be used in conjunction with normalization as
+    follows:
+
+    In eval() mode (i.e. after training), the inputs pass
+    would typically be normalized to the unit cube (e.g. during candidate
+    optimization).
+    1. These are unnormalized back to the raw input space.
+    2. The discrete values are created.
+    3. All values are normalized to the unit cube.
```
TobyBoyne, line 234: How strong is this assumption, that the inputs will already be normalized? Is it worth using `check_min_max_scaling` during the forward pass to validate this assumption, or is that too expensive to do every time the transform is called?

botorch/acquisition/probabilistic_reparameterization.py
```diff
@@ -0,0 +1,1219 @@
...
+    tfs = OrderedDict()
+    if integer_indices is not None and len(integer_indices) > 0:
+        # unnormalize to integer space
+        tfs["unnormalize"] = Normalize(
+            d=one_hot_bounds.shape[1],
+            bounds=one_hot_bounds,
+            indices=integer_indices,
+            transform_on_train=False,
+            transform_on_eval=True,
+            transform_on_fantasize=False,
```
TobyBoyne, line 874: Why is this set to `False`? Note that the PR input transform has `transform_on_fantasize==True`. (This is from the original PR, I don't think added this)

botorch/acquisition/probabilistic_reparameterization.py
```diff
@@ -0,0 +1,1219 @@
...
+            # compute probabilities for integers
+            X_int = X_prob[..., self.integer_indices]
+            X_int_abs = X_int.abs()
+            offset = X_int_abs.floor()
+            if self.tau is not None:
+                X_prob[..., self.integer_indices] = torch.sigmoid(
+                    (X_int_abs - offset - 0.5) / self.tau
+                )
+            else:
+                X_prob[..., self.integer_indices] = X_int_abs - offset
```
TobyBoyne, line 384: Should we allow `tau` to be `None`? What would be the use case for not applying the sigmoid? If we do want to allow this, then I will change the type hint in the constructor to `float | None`.

botorch/acquisition/probabilistic_reparameterization.py
```diff
@@ -0,0 +1,1219 @@
...
+        """
+        self._check_input_shape(X)
+
+        X_expanded = X.expand(*X.shape[:-3], self.mc_samples, *X.shape[-2:]).clone()
+        X_prob = self.get_rounding_prob(X=X)
+        if self.integer_indices is not None:
+            X_int = X[..., self.integer_indices].detach()
+            assert X.ndim > 1
+            if X.ndim == 2:
+                X.unsqueeze(-1)
```
TobyBoyne, line 717: If `X.ndim == 2`, then to me I would imagine the shape of `X` is `n x d`. In that case, we should really unsqueeze in the first dimension to obtain a `1 x n x d` tensor, if I'm understanding this correctly.

botorch/acquisition/probabilistic_reparameterization.py
```diff
@@ -0,0 +1,1219 @@
...
+            torch.zeros(1, dtype=one_hot_bounds.dtype, device=one_hot_bounds.device),
+        )
+        self.register_buffer(
+            "ma_baseline",
+            torch.zeros(1, dtype=one_hot_bounds.dtype, device=one_hot_bounds.device),
+        )
+
+    def sample_candidates(self, X: Tensor) -> Tensor:
+        if "unnormalize" in self.input_transform:
+            unnormalized_X = self.input_transform["unnormalize"](X)
```
TobyBoyne, line 1031: Here, I think the candidate `X` should be in the unnormalized space. This would mean the current behaviour is incorrect, and that we should not unnormalize/normalize.

Balandat:
@sdaulton looks like there are still some open questions here from earlier; would you mind having a look?

PR #2883, opened 2025-06-16
ROBOT tutorial notebook

nataliemaus:
## Motivation

This PR adds a tutorial notebook for the Rank Ordered Bayesian Optimization with Trust Regions (ROBOT) method proposed by the AISTATS 2023 paper titled Discovering Many Diverse Solutions with Bayesian Optimization. The motivation for the notebook is to provide a simple clean example of how the ROBOT method can be applied in practice to optimize a set of diverse solutions to a black-box objective function. 

Yes.

## Test Plan

The only change made is to add an isolated Jupyter notebook (robot.ipynb) to the tutorials folder. I checked that the notebook runs as it should and produces desired outputs. Since none of the BoTorch code was actually changed at all (all I did was add this notebook), no other tests are required as far as I am aware. 

## Related PRs


nataliemaus:
@dme65 @jacobrgardner :)

tutorials/robot/robot.ipynb
```diff
@@ -0,0 +1,1041 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 1,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "import warnings\n",
+    "warnings.filterwarnings('ignore')\n",
```
esantorella: What warnings are filtered out? Can you make the filtering more granular? I wouldn't want to filter out any relevant warnings in case something starts breaking in the future.

tutorials/robot/robot.ipynb
```diff
@@ -0,0 +1,1041 @@
...
+    "torch.manual_seed(42);"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Define model class for GP surrogate model\n",
+    "The GPModel class defined below is a standard approximate GP model defined using gpytorch. As in [1], we use an approximate GP model for increased speed when updating on large quantities of data. \n",
+    "Additionally we define below a method update_gp that can be called to update the GP surrogate model on new data collected during optimization using gradient descent with the standard Adam optimizer. "
```
esantorella:
```suggestion
    "Additionally we define below a function update_gp that can be called to update the GP surrogate model on new data collected during optimization using gradient descent with the standard Adam optimizer. "
```

tutorials/robot/robot.ipynb
```diff
@@ -0,0 +1,1041 @@
...
+    "            self, \n",
+    "            inducing_points, \n",
+    "            variational_distribution, \n",
+    "            learn_inducing_locations=True,\n",
+    "        )\n",
+    "        super(GPModel, self).__init__(variational_strategy)\n",
+    "        self.mean_module = gpytorch.means.ConstantMean()\n",
+    "        self.covar_module = gpytorch.kernels.ScaleKernel(gpytorch.kernels.RBFKernel())\n",
+    "        self.num_outputs = 1\n",
+    "        self.likelihood = likelihood \n",
```
esantorella, line 78:
```suggestion
    "        super(GPModel, self).__init__(model=variational_strategy, likelihood=likelihood, num_outputs=1)\n",
    "        self.mean_module = gpytorch.means.ConstantMean()\n",
    "        self.covar_module = gpytorch.kernels.ScaleKernel(gpytorch.kernels.RBFKernel())\n",
```

tutorials/robot/robot.ipynb
```diff
@@ -0,0 +1,1041 @@
...
+    "Additionally we define below a method update_gp that can be called to update the GP surrogate model on new data collected during optimization using gradient descent with the standard Adam optimizer. "
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 2,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "class GPModel(ApproximateGP):\n",
```
esantorella, line 65: Could you explain more why you used this model instead of `ApproximateGP`? And given that most of the functionality of `ApproximateGP` is overridden, what to we get from subclassing it?

tutorials/robot/robot.ipynb
```diff
@@ -0,0 +1,1041 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "@dataclass\n",
+    "class TrustRegionState:\n",
+    "    center_point: torch.Tensor =  None \n",
```
esantorella:
```suggestion
    "    center_point: torch.Tensor | None =  None \n",
```

tutorials/robot/robot.ipynb
```diff
@@ -0,0 +1,1041 @@
...
+    "    length_min: float = 0.5 ** 7\n",
+    "    length_max: float = 1.6\n",
+    "    failure_counter: int = 0\n",
+    "    failure_tolerance: int = 32 \n",
+    "    success_counter: int = 0\n",
+    "    success_tolerance: int = 10 \n",
+    "    restart_triggered: bool = False\n",
+    "\n",
+    "\n",
+    "def update_state(state, Y_next):\n",
```
esantorella:
```suggestion
    "def update_state(state: TrustRegionState, Y_next):\n",
```

A type annotation for `Y_next` would be helpful for readability

tutorials/robot/robot.ipynb
```diff
@@ -0,0 +1,1041 @@
...
+    "On each iteration of optimization, ROBOT [1] uses Thompson sampling to generate candidates from within each of the $M$ trust regions. Thompson sampling is straightforward to implement using BoTorch."
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 4,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "def generate_batch(\n",
```
esantorella, line 182: BoTorch has `acquisition.thompson_sampling.PathwiseThompsonSampling` -- not sure if that would be helpful

tutorials/robot/robot.ipynb
```diff
@@ -0,0 +1,1041 @@
...
+    "\n",
+    "    return X_next"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Set up the rover black-box optimization task \n",
+    "In ths notebook, we optimize the same rover trajectory optimization task considered in [1]. The task consists of finding a 60 dimensional policy that allows a rover to move along some trajectory from a designated start point, to the \"goal\" point, while avoiding a set of obstacles. This optimization problem is useful for demonstrating the usefullness of ROBOT as it allows us to directly visualize the diverse solutions (trajectories) found by ROBOT. "
```
esantorella:
```suggestion
    "In ths notebook, we optimize the same rover trajectory optimization task considered in [1]. The task consists of finding a 60 dimensional policy that allows a rover to move along some trajectory from a designated start point, to the \"goal\" point, while avoiding a set of obstacles. This optimization problem is useful for demonstrating the usefulness of ROBOT as it allows us to directly visualize the diverse solutions (trajectories) found by ROBOT. "
```

tutorials/robot/robot.ipynb
```diff
@@ -0,0 +1,1041 @@
...
+    "For the rover optimization task, as in the original paper [1], we define the diversity function as the \"one way distance\" between a pair of paths taken by the rover, and we use $\\tau=0.15$. Thus, the $M$ solutions found by ROBOT must be meaningfully diverse from one another, as the solutions are forced to take distinct routes around the obstacles. "
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 7,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "def delta(x1, x2 ):\n",
```
esantorella, line 375: type annotations would be helpful for readability

tutorials/robot/robot.ipynb
```diff
@@ -0,0 +1,1041 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 9,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# Initializaiton data: \n",
+    "n_initialization_points = 1024\n",
```
esantorella:
```suggestion
    "n_initialization_points = 1024\n",
```

tutorials/robot/robot.ipynb
```diff
@@ -0,0 +1,1041 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 11,
+   "metadata": {},
+   "outputs": [
+    {
+     "name": "stdout",
+     "output_type": "stream",
+     "text": [
```
esantorella, line 507: Can we print less frequently?

tutorials/robot/robot.ipynb
```diff
@@ -0,0 +1,1041 @@
...
+    "plt.legend()\n",
+    "plt.tight_layout()\n",
+    "plt.show();"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Visualize ROBOT's optimized solutions\n",
```
esantorella, line 963: Love this visualization!

esantorella (commented):
Nice, thank you for submitting this! About how long does this notebook take to run? I think this probably belongs in `notebooks_community` since I'm not confident that the BoTorch team can promise to maintain this or run it regularly in the CI.

Edit: But feel free to keep it where it is for now so that we can keep getting CI signals on this -- seems like it's not running successfully now.

tutorials/robot/robot.ipynb
nataliemaus: As far as I'm aware this is the suggested way to use ApproximateGP class (https://github.com/cornellius-gp/gpytorch/blob/main/gpytorch/models/approximate_gp.py)?

tutorials/robot/robot.ipynb
nataliemaus: Yeah we could for example only print when the average reward improves, although I think can also be helpful for people to see an update on every iteration? I.e. in other tutorials such as turbo tutorial (https://github.com/pytorch/botorch/blob/main/tutorials/turbo_1/turbo_1.ipynb) there is a progress update printed on every iteration.

nataliemaus:
@esantorella, thank you for reviewing this! Regarding your concern about it taking too long to run, I agree and added a SMOKE_TEST option so that in SMOKE_TEST mode it only iterates a couple of times and completes much faster. Given that update I think it should be okay to keep this in tutorials/ ?

tutorials/robot/robot.ipynb
nataliemaus: Agreed, thanks for adding those, I've committed your suggested changes.

tutorials/robot/robot.ipynb
esantorella: Oh! It's GPyTorch `ApproximateGP`. I was thinking of BoTorch `ApproximateGPyTorchModel`. Never mind.

tutorials/robot/robot.ipynb
esantorella: I don't feel too strongly about it -- I think if you're reading the tutorial it's nice for the output to be more minimal, but if you're running this yourself on your own data, it's nice to see verbose output

esantorella:
> Regarding your concern about it taking too long to run, I agree and added a SMOKE_TEST option so that in SMOKE_TEST mode it only iterates a couple of times and completes much faster. Given that update I think it should be okay to keep this in tutorials/ ?

Sounds reasonable to me since it isn't using any functionality from botorch_community.

tutorials/robot/robot.ipynb
nataliemaus: As a compromise, I modified to print every 10 iterations instead of every iteration 👍

esantorella (approved):
Looks great, thank you!

PR #2889, opened 2025-06-18
Changes to I-BNN tutorial

esantorella:
## Motivation

* Take out comparison between methods. This makes the tutorial too long, and text descriptions of results can become inaccurate over time.
* Simplify code
* Put the MLP model that generates data into `eval` model (this changes results substantially)

## Test Plan

Looked at the notebook to make sure it rendered correctly


saitcakmak (approved):
Thanks for simplifying this!

PR #2894, opened 2025-06-25
Fix docstring of `FullyBayesianSingleTaskGP`

jduerholt:
## Motivation

Docstring of `FullyBayesianSingleTaskGP` is wrong (looks like a copy-paste thingy). This PR fixes it.

Yes.

## Test Plan

No tests needed, only fix on docstrings.

### DIFF
--- a/botorch/models/fully_bayesian.py
+++ b/botorch/models/fully_bayesian.py
@@ -224,18 +224,19 @@ class PyroModel:
         The prior has a mean value of 1 for each concentration and is very
         concentrated around the mean.
         """
+        d = len(self.indices) if self.indices is not None else self.ard_num_dims
         c0 = pyro.sample(
             "c0",
             pyro.distributions.LogNormal(
-                torch.tensor([0.0] * self.ard_num_dims, **tkwargs),
-                torch.tensor([0.1**0.5] * self.ard_num_dims, **tkwargs),
+                torch.tensor([0.0] * d, **tkwargs),
+                torch.tensor([0.1**0.5] * d, **tkwargs),
             ),
         )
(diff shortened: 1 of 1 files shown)


hvarfner:
Thanks for the PR @jduerholt! Although I appreciate the eponym, I'll go ahead and re-name it to something slightly more descriptive!

botorch/models/fully_bayesian.py
```diff
@@ -904,7 +904,7 @@ def construct_inputs(
 
 
 class FullyBayesianSingleTaskGP(AbstractFullyBayesianSingleTaskGP):
-    r"""A fully Bayesian single-task GP model with the SAAS prior.
+    r"""A fully Bayesian single-task GP model with the Hvarfner prior.
```
Balandat: We don't really talk about priors in other high-level model descriptions. Let's also adjust the wording below (can't suggest on non-modified code directly) to make clear what the dimension-scaling refers to:

> A model with a Matern-5/2 kernel and dimension-scaled priors on the hyperparameters from  [Hvarfner2024vanilla]_  is used by default.

```suggestion
    r"""A fully Bayesian single-task GP model.
```

jduerholt: I will adjust it!

jduerholt:
@Balandat: I adjusted it according to your suggestion. Best, Johannes

jduerholt:
@Balandat:

There is also a bug in the fully bayesian models regarding input warping of only a subset. I fixed it here. Is it fine to have it fixed here, or do you want a seperate PR for this?

Best,

Johannes

Issue #2879, opened 2025-06-13
[FEATURE REQUEST]: NumericToCategoricalEnconding Input Transform


jduerholt:
### Motivation

@TobyBoyne recently added the support for categorical dimensions to `optimize_mixed_alternating`. This implementation assumes that the categorical dimensions are integer encoded. As consequence, it is directly usable with the MixedSingleTaskGP, but not with GPs that for example assume one-hot encoded categoricals. A solution for this case would be to have a `NumericToOneHot` input transform that encodes the categorical feature(s) with in the model. 

Currently, botorch features a `OneHotToNumeric` input transform including a `untransform` functionality. It should be relatively straight-forward to come up with a `NumericToOneHot` transform based on this. This would solve the issue of models which expect one-hot encoded features, but not other possible encodings. For example, in chemistry one often uses descriptor encodings, in which one transforms the categorical feature in some kind of descriptors space based on chemical descriptors. From a software engineering point of view, these descriptor encodings of categoricals are very much the same as one-hot encodings: one transforms a vector into a matrix, just the mapping is different. For this reason, I would like to implement a generic `NumericToCategoricalEnconding` input transform which takes upon instantiation information regarding the dimensionality of the encoding space and a (non-differentiable) callable that performs the transformation. This would allow to use `optimize_mixed_alternating` with any kind of categorical encoding.

What do you think? Does this sounds reasonable to you?

Best,

Johannes

### Describe the solution you'd like to see implemented in BoTorch.

see above

### Pull Request

Yes

### Code of Conduct

- [x] I agree to follow BoTorch's Code of Conduct

saitcakmak:
Hi @jduerholt. That seems reasonable overall but I am curious about one aspect of the approach.

Why model using one-hot encoded features, rather than using something like `CategoricalKernel`? If you're optimizing it using a single numerical feature, you could do the same in the model and let the kernel handle it, rather than having your kernel be defined over one hot encoded features. We have [`MixedSingleTaskGP`](https://github.com/pytorch/botorch/blob/main/botorch/models/gp_regression_mixed.py#L29) that uses `CategoricalKernel`. I am hoping to replace `OneHot` in Ax with `CategoricalKernel` in H2 as well (haven't benchmarked this extensively yet).

saitcakmak:
I guess one could also use the `IndexKernel` here, instead of `CategoricalKernel`. That one uses an embedding under the hood to compute distances, which seems like it'd work just as well for categorical features.

[...]

PR #2907, opened 2025-07-01
NumericToCategoricalEncoding Input Transform.

jduerholt:
## Motivation

This PR refers to https://github.com/pytorch/botorch/issues/2879. It adds a new input transform that transforms a categorical degree of freedom encoded a an integer into some kind of vector based description. This could be for example a one-hot encoding, but also a descriptor encoding as it is often used in chemistry. It adds the possibility to use the alternating acqf optimizer also with surrogates that do not treat categoricals as integer based values. For example one could then also use a SAAS GP with the mixed alternating acqf optimizer and treat the categoricals under the hood as one-hots.

Yes.

## Test Plan

Unit tests, most of them are implemented (also to demonstrate the functionality), the ones which check the equality between transforms and correct behavior of transform on train etc. are still missing. My plan is to add them after a first feedback after a first review.

### DIFF
--- a/botorch/models/transforms/input.py
+++ b/botorch/models/transforms/input.py
@@ -1625,6 +1625,122 @@ class InputPerturbation(InputTransform):
         return p.transpose(-3, -2)  # p is batch_shape x n_p x n x d
 
 
+class NumericToCategoricalEncoding(InputTransform):
+    """Transform categorical parameters from an integer representation
+    to a vector based representation like one-hot encoding or a descriptor
+    encoding.
+    """
+
+    def __init__(
+        self,
+        dim: int,
+        categorical_features: dict[int, int],
+        encoders: dict[int, Callable[[Tensor], Tensor]],
+        transform_on_train: bool = True,
+        transform_on_eval: bool = True,
+        transform_on_fantasize: bool = True,
+    ):
+        r"""Initialize.
+
+        Args:
+            dim: The dimension of the numerically encoded input.
+            categorical_features: A dictionary mapping the index of each
+                categorical feature to its cardinality. This assumes that categoricals
+                are integer encoded.
+            encoders: A dictionary mapping the index of each categorical feature to
+                a callable that encodes the categorical feature into a vector
+                representation.
+            transform_on_train: A boolean indicating whether to apply the
+                transforms in train() mode. Default: False.
+            transform_on_eval: A boolean indicating whether to apply the
+                transform in eval() mode. Default: True.
+            transform_on_fantasize: A boolean indicating whether to apply the
+                transform when called from within a `fantasize` call. Default: False.
+        """
+        super().__init__()
+        self.transform_on_train = transform_on_train
+        self.transform_on_eval = transform_on_eval
+        self.transform_on_fantasize = transform_on_fantasize
+
+        self.encoders = encoders
+        self.categorical_features = categorical_features
+
+        if len(self.categorical_features) > dim:
+            raise ValueError(
+                "The number of categorical features exceeds the provided dimension."
+            )
+
+        # check that the encoders match the categorical features
+        if set(self.encoders.keys()) != set(self.categorical_features.keys()):
+            raise ValueError(
+                "The keys of `encoders` must match the keys of `categorical_features`."
+            )
+
+        self.ordinal_idx = list(
+            self.categorical_features.keys()
+        )  # indices of categorical features before encoding
+
+        self.numerical_idx = list(
+            set(range(dim)) - set(self.ordinal_idx)
+        )  # indices of numerical features before encoding
+
+        self.new_numerical_idx = []  # indices of numerical features after encoding
+        self.encoded_idx = []  # indices of categorical features after encoding
+
+        offset = 0
+        for idx in range(dim):
+            if idx in self.numerical_idx:
+                self.new_numerical_idx.append(idx + offset)
+            else:
+                card = self.categorical_features[idx]
+                self.encoded_idx.append(
+                    np.arange(
+                        idx + offset, idx + offset + card
+                    ).tolist()  # indices of categorical features after encoding
+                )
+                offset += card - 1  # adjust offset for next categorical feature
+
+    def transform(self, X: Tensor) -> Tensor:
+        r"""Transform the categorical inputs into a vector representation.
+
+        Args:
+            X: A `batch_shape x n x d`-dim tensor of inputs.
+
+        Returns:
+            A `batch_shape x n x d'`-dim tensor of where the integer encoded
+            categoricals are transformed to a vector representation.
+        """
+        if len(self.categorical_features) > 0:
+            s = list(X.shape)
+            s[-1] = len(self.numerical_idx) + len(np.concatenate(self.encoded_idx))
+            X_encoded = torch.zeros(size=s).to(X)
+            X_encoded[..., self.new_numerical_idx] = X[..., self.numerical_idx]
+            for i, idx in enumerate(self.categorical_features.keys()):
+                X_encoded[..., self.encoded_idx[i]] = self.encoders[idx](
+                    X[..., idx].long(),
+                ).to(X_encoded)
+            return X_encoded
+        return X
+
+    def equals(self, other: InputTransform) -> bool:
+        r"""Check if another input transform is equivalent.
+
+        Args:
+            other: Another input transform.
+
+        Returns:
+            A boolean indicating if the other transform is equivalent.
+        """
+        return (
+            type(self) is type(other)
+            and (self.transform_on_train == other.transform_on_train)
+            and (self.transform_on_eval == other.transform_on_eval)
+            and (self.transform_on_fantasize == other.transform_on_fantasize)
+            and self.categorical_features == other.categorical_features
+            and self.encoders == other.encoders
+        )
+
+
 class OneHotToNumeric(InputTransform):
     r"""Transform categorical parameters from a one-hot to a numeric representation."""
 
@@ -1649,10 +1765,6 @@ class OneHotToNumeric(InputTransform):
                 transform in eval() mode. Default: True.
             transform_on_fantasize: A boolean indicating whether to apply the
                 transform when called from within a `fantasize` call. Default: False.
-
-        Returns:
-            A `batch_shape x n x d'`-dim tensor of where the one-hot encoded
-            categoricals are transformed to integer representation.
         """
         super().__init__()
         self.transform_on_train = transform_on_train
--- a/botorch/models/transforms/input.py
+++ b/botorch/models/transforms/input.py
@@ -1699,16 +1699,18 @@ class NumericToCategoricalEncoding(InputTransform):
             )
 
         for idx, card in self.categorical_features.items():
-            if card == 1:
+            if card <= 1:
                 raise ValueError(
-                    f"Categorical feature at index {idx} has cardinality 1. "
+                    f"Categorical feature at index {idx} has cardinality {card}. "
                     f"All categorial features must have cardinality greater than 1."
                 )
 
         # check that the encoders match the categorical features
-        if (enc_keys := set(self.encoders)) != (cf_keys := set(self.categorical_features)):
+        if (enc_keys := set(self.encoders)) != (
+            cf_keys := set(self.categorical_features)
+        ):
             raise ValueError(
-                "The keys of `encoders` ({enc_keys}) must match the keys of "
+                f"The keys of `encoders` ({enc_keys}) must match the keys of "
                 f"of `categorical_features`  ({cf_keys})."
             )
 
@@ -1743,8 +1745,9 @@ class NumericToCategoricalEncoding(InputTransform):
             X: A `batch_shape x n x d`-dim tensor of inputs.
 
         Returns:
-            A `batch_shape x n x d'`-dim tensor with `d' = d + sum(categorical_features.values())`
-            in which the integer-encoded categoricals are transformed to a vector representation.
+            A `batch_shape x n x d'`-dim tensor with
+            `d' = d + sum(categorical_features.values())` in which the
+            integer-encoded categoricals are transformed to a vector representation.
         """
         s = list(X.shape)
         s[-1] = len(self.numerical_idx) + len(np.concatenate(self.encoded_idx))
--- a/botorch/models/transforms/input.py
+++ b/botorch/models/transforms/input.py
@@ -1660,7 +1660,7 @@ class NumericToCategoricalEncoding(InputTransform):
         transform_on_train: bool = True,
         transform_on_eval: bool = True,
         transform_on_fantasize: bool = True,
-    ):
+    ) -> None:
         r"""Initialize.
 
         Args:
@@ -1692,22 +1692,24 @@ class NumericToCategoricalEncoding(InputTransform):
                 "At least one categorical feature for encoding must be provided."
             )
 
-        if len(self.categorical_features) > dim:
+        if (num_cat := len(self.categorical_features)) > dim:
             raise ValueError(
-                "The number of categorical features exceeds the provided dimension."
+                f"The number of categorical features ({num_cat}) exceeds the "
+                f"provided dimension ({dim})."
             )
 
         for idx, card in self.categorical_features.items():
             if card == 1:
                 raise ValueError(
-                    f"Categorical feature at index {idx} has cardinality 1, "
-                    f"which is not allowed."
+                    f"Categorical feature at index {idx} has cardinality 1. "
+                    f"All categorial features must have cardinality greater than 1."
                 )
 
         # check that the encoders match the categorical features
-        if set(self.encoders.keys()) != set(self.categorical_features.keys()):
+        if (enc_keys := set(self.encoders)) != (cf_keys := set(self.categorical_features)):
             raise ValueError(
-                "The keys of `encoders` must match the keys of `categorical_features`."
+                "The keys of `encoders` ({enc_keys}) must match the keys of "
+                f"of `categorical_features`  ({cf_keys})."
             )
 
         self.ordinal_idx = list(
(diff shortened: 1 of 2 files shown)


jduerholt:
@saitcakmak @Balandat any thoughts on this? It would be totally fine for me, if you say that you do not see this functionality directly in botorch. Then I would integrate it into our codebase, which is also totally fine ;)

botorch/models/transforms/input.py
```diff
@@ -1625,6 +1625,122 @@ def _expanded_perturbations(self, X: Tensor) -> Tensor:
...
+        r"""Transform the categorical inputs into a vector representation.
+
+        Args:
+            X: A `batch_shape x n x d`-dim tensor of inputs.
+
+        Returns:
+            A `batch_shape x n x d'`-dim tensor of where the integer encoded
+            categoricals are transformed to a vector representation.
+        """
+        if len(self.categorical_features) > 0:
```
Balandat: Since the whole transform would be a nullop if no `categorical_features` is provided, should you instead just not allow this and error out? Or do you want to be able to apply the transform universally even if there aren't any categorical features?

botorch/models/transforms/input.py
```diff
@@ -1625,6 +1625,122 @@ def _expanded_perturbations(self, X: Tensor) -> Tensor:
         return p.transpose(-3, -2)  # p is batch_shape x n_p x n x d
 
 
+class NumericToCategoricalEncoding(InputTransform):
+    """Transform categorical parameters from an integer representation
+    to a vector based representation like one-hot encoding or a descriptor
+    encoding.
```
Balandat: Would be good to have a description of how the columns in the output of the transform are organized. Ideally there would be a concrete example in the docstring.

test/models/transforms/test_input.py
```diff
@@ -1207,6 +1209,159 @@ def test_warp_mro(self) -> None:
...
+            NumericToCategoricalEncoding(
+                dim=4,
+                categorical_features=categorical_features,
+                encoders={
+                    0: partial(one_hot, num_classes=2),
+                    2: partial(one_hot, num_classes=3),
+                },
+            )
+
+        torch.manual_seed(randint(0, 1000))
```
Balandat: This seems to defy the point of locking the seed?

test/models/transforms/test_input.py
```diff
@@ -1207,6 +1209,159 @@ def test_warp_mro(self) -> None:
...
+                categorical_features=categorical_features,
+                encoders={0: partial(one_hot, num_classes=3)},
+            )
+            tf.eval()
+            cat_numeric = torch.randint(0, 3, (3,), device=self.device)
+            cat_one_hot = one_hot(cat_numeric, num_classes=3)
+            cont = torch.rand(3, 2, dtype=dtype, device=self.device)
+
+            X_numeric = torch.cat(
+                [cat_numeric.view(-1, 1).to(dtype=dtype, device=self.device), cont],
```
Balandat: No need for the `.to()`, can do that directly above: `cat_numeric = torch.randint(0, 3, (3,), device=self.device, dtype=dtype)`

botorch/models/transforms/input.py
```diff
@@ -1625,6 +1625,122 @@ def _expanded_perturbations(self, X: Tensor) -> Tensor:
...
+        encoders: dict[int, Callable[[Tensor], Tensor]],
+        transform_on_train: bool = True,
+        transform_on_eval: bool = True,
+        transform_on_fantasize: bool = True,
+    ):
+        r"""Initialize.
+
+        Args:
+            dim: The dimension of the numerically encoded input.
+            categorical_features: A dictionary mapping the index of each
```
Balandat, line 1668: this arg name could be more descriptive, e.g. `numeric_cardinality` or sth like that

botorch/models/transforms/input.py
```diff
@@ -1625,6 +1625,122 @@ def _expanded_perturbations(self, X: Tensor) -> Tensor:
         return p.transpose(-3, -2)  # p is batch_shape x n_p x n x d
 
 
+class NumericToCategoricalEncoding(InputTransform):
```
Balandat, line 1628: wait, should this not be `CategoricalToNumeric`?

test/models/transforms/test_input.py
```diff
@@ -1207,6 +1209,159 @@ def test_warp_mro(self) -> None:
...
+            cat_one_hot1 = one_hot(cat_numeric1, num_classes=3)
+            cat_numeric2 = torch.randint(0, 2, (3,), device=self.device)
+            cat_one_hot2 = one_hot(cat_numeric2, num_classes=2)
+            cont = torch.rand(3, 2, dtype=dtype, device=self.device)
+
+            X_numeric = torch.cat(
+                [
+                    cont,
+                    cat_numeric1.view(-1, 1).to(dtype=dtype, device=self.device),
+                    cat_numeric2.view(-1, 1).to(dtype=dtype, device=self.device),
```
Balandat: same comment as above about `to()`

jduerholt:
Thanks for the review, I will go over it and adapt accordingly. Best, Johannes

botorch/models/transforms/input.py
jduerholt: Yeah, I am not sure how to call this propely, my thought was that here the categorical variable is encoded in a numeric fashion (by an integer) and we transform it some kind of categorical encoding, but this is definitly not ideal. Should I just rename to `CategoricalToNumeric`?

test/models/transforms/test_input.py
jduerholt: You are right, I was basing my tests on the ones from `OneHotToNumeric` and there it also done: https://github.com/pytorch/botorch/blob/71f03ea20abf219034f01db0db528ad7a74167aa/test/models/transforms/test_input.py#L1224

But, I will change it.

botorch/models/transforms/input.py
jduerholt: But `CategoricalToNumeric` is kind of strange in light of `OneHotToNumeric`, as this transform is a generalized version of the inverese of `OneHotToNumeric`, this is why I was naming it `NumericToCategoricalEncoding` ...

test/models/transforms/test_input.py
jduerholt: Done.

test/models/transforms/test_input.py
jduerholt: Done.

test/models/transforms/test_input.py
jduerholt: Done.

botorch/models/transforms/input.py
jduerholt: Hmm, I was not really thinking about this, but maybe it is good to just error out. I will update it accordingly.

jduerholt: Done.

test/models/transforms/test_input.py
jduerholt: I need to revert it. To be able to apply the torch one-hot functionality it has to be of type long.

test/models/transforms/test_input.py
jduerholt: I need to revert it partially, see comment above.

botorch/models/transforms/input.py
Balandat: hmm why is this strange?

jduerholt: Maybe, I have an error of thought, but my thinking goes as following:

We have currently a transform which is called `OneHotToNumeric`, it takes one-hot encoded categorical features and transforms them into a numeric representation. Numeric is meaning here just a number starting from zero and going up to `n_categories-1`. And this transform here is in principle the generalized opposite of `OneHotToNumeric`. We start with a number between 0 and `n_categories-1` and transform it to some kind of categorical encoding, for example a one-hot encoding. Since it is the opposite operation it make for me no sense to call it `CategoricalToNumeric`, as the result is not the same as in `OneHotToNumeric`.

Does this thinking makes somehow sense for you?

Balandat: hmm I see yeah that makes sense, I was misreading this.

botorch/models/transforms/input.py
jduerholt: Hmm, I was naming it like this as it also called like this in the related `OneHotToNumeric` transform. What do you think about `categorical_features_cardinality`?

botorch/models/transforms/input.py
jduerholt: I added an example ;)

jduerholt:
Hi @Balandat,

I have now implemented all your suggestions besides the one regarding the name, what do you think about my proposal there?

Best,

Johannes

botorch/models/transforms/input.py
Balandat: hmm yeah we should keep things consistent with `OneHotToNumeric`. I don't think the arg name is ideal (cc @sdaulton) but I don't know if it warrants changing things in `OneHotToNumeric` and having to deal with backward compatibility issues. So this seems ok to me.

botorch/models/transforms/input.py
```diff
@@ -1625,6 +1625,153 @@ def _expanded_perturbations(self, X: Tensor) -> Tensor:
...
+
+    def __init__(
+        self,
+        dim: int,
+        categorical_features: dict[int, int],
+        encoders: dict[int, Callable[[Tensor], Tensor]],
+        transform_on_train: bool = True,
+        transform_on_eval: bool = True,
+        transform_on_fantasize: bool = True,
+    ):
```
Balandat:
```suggestion
    ) -> None:
```

Balandat: nit

botorch/models/transforms/input.py
```diff
@@ -1625,6 +1625,153 @@ def _expanded_perturbations(self, X: Tensor) -> Tensor:
...
+
+        if len(self.categorical_features) == 0:
+            raise ValueError(
+                "At least one categorical feature for encoding must be provided."
+            )
+
+        if len(self.categorical_features) > dim:
+            raise ValueError(
+                "The number of categorical features exceeds the provided dimension."
+            )
```
Balandat:
```suggestion
        if (num_cat := len(self.categorical_features)) > dim:
            raise ValueError(
                f"The number of categorical features ({num_cat}) exceeds the "
                f"provided dimension ({dim})."
            )
```

botorch/models/transforms/input.py
```diff
@@ -1625,6 +1625,153 @@ def _expanded_perturbations(self, X: Tensor) -> Tensor:
...
+
+        for idx, card in self.categorical_features.items():
+            if card == 1:
+                raise ValueError(
+                    f"Categorical feature at index {idx} has cardinality 1, "
+                    f"which is not allowed."
```
Balandat:
```suggestion
                    f"Categorical feature at index {idx} has cardinality 1. "
                    f"All categorial features must have cardinality greater than 1."
```

botorch/models/transforms/input.py
```diff
@@ -1625,6 +1625,153 @@ def _expanded_perturbations(self, X: Tensor) -> Tensor:
...
+                "At least one categorical feature for encoding must be provided."
+            )
+
+        if len(self.categorical_features) > dim:
+            raise ValueError(
+                "The number of categorical features exceeds the provided dimension."
+            )
+
+        for idx, card in self.categorical_features.items():
+            if card == 1:
```
Balandat: Should we consider catching the case where `card` is zero or not an integer, just so as to avoid hard-to-parse errors?

botorch/models/transforms/input.py
```diff
@@ -1625,6 +1625,153 @@ def _expanded_perturbations(self, X: Tensor) -> Tensor:
...
+                raise ValueError(
+                    f"Categorical feature at index {idx} has cardinality 1, "
+                    f"which is not allowed."
+                )
+
+        # check that the encoders match the categorical features
+        if set(self.encoders.keys()) != set(self.categorical_features.keys()):
+            raise ValueError(
+                "The keys of `encoders` must match the keys of `categorical_features`."
+            )
```
Balandat:
```suggestion
        if (enc_keys := set(self.encoders)) != (cf_keys := set(self.categorical_features)):
            raise ValueError(
                "The keys of `encoders` ({enc_keys}) must match the keys of "
                f"of `categorical_features`  ({cf_keys})."
            )
```

botorch/models/transforms/input.py
```diff
@@ -1625,6 +1625,153 @@ def _expanded_perturbations(self, X: Tensor) -> Tensor:
...
+
+    def transform(self, X: Tensor) -> Tensor:
+        r"""Transform the categorical inputs into a vector representation.
+
+        Args:
+            X: A `batch_shape x n x d`-dim tensor of inputs.
+
+        Returns:
+            A `batch_shape x n x d'`-dim tensor of where the integer encoded
+            categoricals are transformed to a vector representation.
```
Balandat:
```suggestion
            A `batch_shape x n x d'`-dim tensor with `d' = d + sum(categorical_features.values())`
            in which the integer-encoded categoricals are transformed to a vector representation.
```

botorch/models/transforms/input.py
```diff
@@ -1625,6 +1625,153 @@ def _expanded_perturbations(self, X: Tensor) -> Tensor:
...
+        """
+        s = list(X.shape)
+        s[-1] = len(self.numerical_idx) + len(np.concatenate(self.encoded_idx))
+        X_encoded = torch.zeros(size=s).to(X)
```
Balandat:
```suggestion
        X_encoded = torch.zeros(size=s, device=X.device, dtype=X.dtype)
```

botorch/models/transforms/input.py
```diff
@@ -1625,6 +1625,153 @@ def _expanded_perturbations(self, X: Tensor) -> Tensor:
...
+        X_encoded[..., self.new_numerical_idx] = X[..., self.numerical_idx]
+        for i, idx in enumerate(self.categorical_features.keys()):
+            X_encoded[..., self.encoded_idx[i]] = self.encoders[idx](
+                X[..., idx].long(),
+            ).to(X_encoded)
```
Balandat: slight readability tweak
```suggestion
            encoded_val = self.encoders[idx](X[..., idx].long()).to(X_encoded)
            X_encoded[..., self.encoded_idx[i]] = encoded_val
```

test/models/transforms/test_input.py
```diff
@@ -1207,6 +1209,239 @@ def test_warp_mro(self) -> None:
...
+                dim=-1,
+            ).to(dtype=dtype)
+
+            expected = torch.cat(
+                [cat_one_hot, cont],
+                dim=-1,
+            )
+            X_one_hot = tf(X_numeric)
+            self.assertTrue(torch.equal(X_one_hot, expected))
+            # two categoricals at end
```
Balandat, line 1295:
```suggestion
            
            # two categoricals at end
```

test/models/transforms/test_input.py
```diff
@@ -1207,6 +1209,239 @@ def test_warp_mro(self) -> None:
...
+                dim=-1,
+            ).to(dtype=dtype)
+
+            expected = torch.cat(
+                [cont, cat_one_hot1, cat_one_hot2],
+                dim=-1,
+            )
+            X_one_hot = tf(X_numeric)
+            self.assertTrue(torch.equal(X_one_hot, expected))
+            # two categoricals, one at start, one at end
```
Balandat, line 1329:
```suggestion
            
            # two categoricals, one at start, one at end
```

test/models/transforms/test_input.py
```diff
@@ -1207,6 +1209,239 @@ def test_warp_mro(self) -> None:
...
+                dim=-1,
+            )
+
+            expected = torch.cat(
+                [cat_one_hot1, cont, cat_one_hot2],
+                dim=-1,
+            )
+            X_one_hot = tf(X_numeric)
+            self.assertTrue(torch.equal(X_one_hot, expected))
+            # only categoricals
```
Balandat, line 1363:
```suggestion
            
            # only categoricals
```

test/models/transforms/test_input.py
```diff
@@ -1207,6 +1209,239 @@ def test_warp_mro(self) -> None:
...
+                dim=-1,
+            )
+
+            expected = torch.cat(
+                [cat_one_hot1, cat_one_hot2],
+                dim=-1,
+            )
+            X_one_hot = tf(X_numeric)
+            self.assertTrue(torch.equal(X_one_hot, expected))
+        # test no transform on eval
```
Balandat, line 1395:
```suggestion
        
        # test no transform on eval
```

test/models/transforms/test_input.py
```diff
@@ -1207,6 +1209,239 @@ def test_warp_mro(self) -> None:
...
+            dim=dim,
+            categorical_features=categorical_features,
+            encoders={
+                0: partial(one_hot, num_classes=3),
+                1: partial(one_hot, num_classes=2),
+            },
+            transform_on_train=False,
+        )
+        self.assertTrue(tf.equals(tf2))
+        # test different transform_on_train
```
Balandat, line 1437:
```suggestion
        
        # test different transform_on_train
```

test/models/transforms/test_input.py
```diff
@@ -1207,6 +1209,239 @@ def test_warp_mro(self) -> None:
...
+            dim=dim,
+            categorical_features=categorical_features,
+            encoders={
+                0: partial(one_hot, num_classes=3),
+                1: partial(one_hot, num_classes=2),
+            },
+            transform_on_train=True,
+        )
+        self.assertFalse(tf3.equals(tf2))
+        # test categorical features
```
Balandat, line 1449:
```suggestion
        
        # test categorical features
```

Balandat (commented):
Thanks, this looks great. A number of minor nits, happy to merge once updated.

botorch/models/transforms/input.py
jduerholt: Done.

jduerholt:
@Balandat: Should be done now ;)

PR #2923, opened 2025-07-11
Non-equidistant discrete dimensions for `optimize_acqf_mixed_alternating`

jduerholt:
## Motivation

As dicussed in https://github.com/pytorch/botorch/discussions/2904, this is the first PR in planned series of PRs regarding adding additional functionality to `optimize_acqf_mixed_alternating`. In this one, non-equidistant discrete dimensions are introduced.

CC: @TobyBoyne 

Yes.

## Test Plan

Unit tests.


botorch/optim/optimize_mixed.py
```diff
@@ -84,21 +84,31 @@ def _setup_continuous_relaxation(
     `discrete_dims` and `post_processing_func` is updated to round
     them to the nearest integer.
     """
-    discrete_dims_t = torch.tensor(discrete_dims, dtype=torch.long)
-    num_discrete_values = (
-        bounds[1, discrete_dims_t] - bounds[0, discrete_dims_t]
+    discrete_dims_t = torch.tensor(list(discrete_dims.keys()), dtype=torch.long)
```
Balandat: I think you can simplify this function a lot and get rid of all the tensor logic

```
    dims_to_relax, dims_to_keep = {}, {}
    for index, values in discrete_dims.items():
        if len(values) > max_discrete_values:
            dims_to_relax[index] = values
        else:
            dims_to_keep[index] = values

    ...
```

botorch/optim/optimize_mixed.py
```diff
@@ -148,26 +158,56 @@ def get_nearest_neighbors(
...
+
+    t_discrete_dims = torch.tensor(
+        list(discrete_dims.keys()), dtype=torch.long, device=current_x.device
+    )
+
+    for dim, values in discrete_dims.items():
+        val = current_x[dim].item()
+        # Find the index of the closest value in values
+        idx = min(range(len(values)), key=lambda i: abs(values[i] - val))
+        discrete_indices.append(idx)
```
Balandat, line 172: I am a bit worried about this loopiness here being slow if there are many discrete dims and many values per dim. Have you profiled this? I wonder if this can/should be done in a vectorized fashion (doesn't seem obvious how without using "ragged" tensors though).

Same thing for the "untransform" below

botorch/optim/optimize_mixed.py
```diff
@@ -815,14 +885,31 @@ def optimize_acqf_mixed_alternating(
...
+        if not float(values[0]) == bounds[0, dim].item():
+            raise ValueError(
+                f"Discrete dimension {dim} must start at the lower bound "
+                f"{bounds[0, dim].item()} but starts at {values[0]}."
+            )
+        if not float(values[-1]) == bounds[1, dim].item():
+            raise ValueError(
+                f"Discrete dimension {dim} must end at the upper bound "
+                f"{bounds[1, dim].item()} but end at {values[-1]}."
+            )
```
Balandat:
```suggestion
    for dim, values in discrete_dims.items():
        lower_bnd, upper_bnd = bounds[:, dim].tolist()
        lower, upper = values[0], values[-1]
        if lower != lower_bnd:
            raise ValueError(
                f"Discrete dimension {dim} must start at the lower bound "
                f"{lower_bnd} but starts at {lower}."
            )
        if upper != upper_bnd:
            raise ValueError(
                f"Discrete dimension {dim} must end at the upper bound "
                f"{upper_bnd} but end at {upper}."
            )
```

Balandat (commented):
I didn't go through in full detail; at a high level this makes sense, the main thing I'm worried is that if there are many dimensions and many values this could end up becoming quite slow. Do you have any timing / profiling data on this?

jduerholt:
> the main thing I'm worried is that if there are many dimensions

In principle this should be of the same speed as the original implementation, main difference is the rounding operation, are you interested in profiling of this operation? Or do you see issues with speed also in other parts of the code?

jduerholt:
Ah, now I see, you see the main problem in `get_nearest_neighbors`, or? Are interested in profiling data on this? I can generate it, but have not looked into it, as I wanted to first get things started and working again. I will have a look into this.

botorch/optim/optimize_mixed.py
jduerholt: What do you mean with "ragged" tensors?

jduerholt:
Here is some code for profiling for the mentioned operation:



``` python
from botorch.optim.optimize_mixed import get_nearest_neighbors
import time
import torch

d = 100
nd = 90
n_discrete_values = 300

current_x = torch.rand(d)

discrete_dims = {i: list(range(n_discrete_values+1)) for i in range(nd)}

for i in range(nd):
    current_x[i] = torch.randint(0, n_discrete_values, (1,)).item()

bounds = torch.tensor([[0.0] * d, [n_discrete_values] * d])

start = time.time()
neighbors = get_nearest_neighbors(current_x=current_x, discrete_dims=discrete_dims, bounds=bounds)
end = time.time()
print(f"Time taken to find nearest neighbors: {end - start:.4f} seconds")
```

For this rather extreme case, it takes around 0.006 seconds on my macbook with a M1 processor. The original one takes for this case around 0.001 seconds, the new one is defintely slower, but I think that it will be not the rate limiting step (speeking as a chemist :D), or?

botorch/optim/optimize_mixed.py
Balandat: Basically the fact that the length of the possible values is heterogeneous, see https://docs.pytorch.org/docs/stable/nested.html

> Nested tensors allow for ragged-shaped data to be contained within and operated upon as a single tensor. Such data is stored underneath in an efficient packed representation, while exposing a standard PyTorch tensor interface for applying operations.

Balandat:
Yeah, that seems like it should be fine.

botorch/optim/optimize_mixed.py
jduerholt: Nice, I did not know this feature of pytorch. Thanks for pointing it out.

botorch/optim/optimize_mixed.py
jduerholt: Thanks for the hint, I updated accordingly.

botorch/optim/optimize_mixed.py
jduerholt: Good catch, I updated it accordingly.

jduerholt:
Hi @Balandat,

I implemented your comments, so it should be ready for rereview.

Best,

Johannes

botorch/optim/optimize_mixed.py
```diff
@@ -415,11 +471,14 @@ def generate_starting_points(
...
     bounds = opt_inputs.bounds
     binary_dims = []
-    for dim in discrete_dims:
-        if bounds[0, dim] == 0 and bounds[1, dim] == 1:
+    for dim, values in discrete_dims.items():
+        if len(values) == 2:
             binary_dims.append(dim)
     num_binary = len(binary_dims)
-    num_integer = len(discrete_dims) - num_binary
+    num_integer = len(discrete_dims) - num_binary  # maybe rename to num_discrete???
```
Balandat: `num_discrete` makes sense as a name

botorch/optim/optimize_mixed.py
```diff
@@ -415,11 +471,14 @@ def generate_starting_points(
...
+    # Remove binary dims from discrete_dims
+    # WHY?
+    # discrete_dims = {dim: values for dim, values in discrete_dims.items()}
```
Balandat: what's going on here?

botorch/optim/optimize_mixed.py
```diff
@@ -815,14 +876,33 @@ def optimize_acqf_mixed_alternating(
...
         raise NotImplementedError(
             "`optimize_acqf_mixed_alternating` only supports "
             "sequential optimization."
         )
 
     cat_dims = cat_dims or []
-    discrete_dims = discrete_dims or []
+    discrete_dims = discrete_dims or {}
+
+    discrete_dims = OrderedDict(sorted(discrete_dims.items(), key=lambda x: x[0]))
```
Balandat: why is this needed? Seems unnecessary to order the dict by dim key?

botorch/optim/optimize_mixed.py
jduerholt: This can be completely ignored, I was confusing myself :D I will remove the comments.

botorch/optim/optimize_mixed.py
jduerholt: I renamed it accordingly ;)

botorch/optim/optimize_mixed.py
jduerholt: It is removed ;)

botorch/optim/optimize_mixed.py
jduerholt: Could be ;) But by adding this one line, one is on the safe side, as I tested the code only for the case that it is ordered. Does it hurt to keep it?

Balandat: Hmm it's a bit confusing to explicitly order something that doesn't need to be ordered, I think it makes understanding the rest of the code harder (it confused me, for one). So I'd prefer to not do this if not necessary (from skimming the code it seems like it should not need to be ordered and it should be easy to test the behavior without doing this).

botorch/optim/optimize_mixed.py
```diff
@@ -148,26 +149,56 @@ def get_nearest_neighbors(
...
+    unique_neighbors = unique_neighbors[
+        ~(unique_neighbors == current_x_int).all(dim=-1)
+    ]
+
+    # Transform unique_neighbors back to the original non-integer space
+    for dim, values in discrete_dims.items():
+        t_values = torch.tensor(
+            values, device=unique_neighbors.device, dtype=unique_neighbors.dtype
+        )
+        idx = unique_neighbors[:, dim].long().clamp(0, len(values) - 1)
```
saitcakmak: Why do we need to clamp here? Both `plus_neighbors` and `minus_neighbors` have been clamped, right? If we actually need to clamp here, I think that may produce duplicate neighbors.

botorch/optim/optimize_mixed.py
```diff
@@ -148,26 +149,56 @@ def get_nearest_neighbors(
...
+        discrete_indices, dtype=current_x.dtype, device=current_x.device
+    )
     num_discrete = len(discrete_dims)
     diag_ones = torch.eye(num_discrete, dtype=current_x.dtype, device=current_x.device)
     # Neighbors obtained by increasing a discrete dimension by one.
-    plus_neighbors = current_x.repeat(num_discrete, 1)
-    plus_neighbors[:, discrete_dims] += diag_ones
+    plus_neighbors = current_x_int.repeat(num_discrete, 1)
+    plus_neighbors[:, t_discrete_dims] += diag_ones
     plus_neighbors.clamp_(max=bounds[1])
```
saitcakmak: Does it make sense to clamp to bounds anymore, since the neighbors now have index values that are between 0 & len(values) - 1?

botorch/optim/optimize_mixed.py
```diff
@@ -148,26 +149,56 @@ def get_nearest_neighbors(
...
     # Neighbors obtained by decreasing a discrete dimension by one.
-    minus_neighbors = current_x.repeat(num_discrete, 1)
-    minus_neighbors[:, discrete_dims] -= diag_ones
+    minus_neighbors = current_x_int.repeat(num_discrete, 1)
+    minus_neighbors[:, t_discrete_dims] -= diag_ones
     minus_neighbors.clamp_(min=bounds[0])
```
saitcakmak: Same question about clamping here. I think this one is particularly problematic since the bounds[0] may be some large number even though the index value has a fixed range. I guess the same argument would apply to bounds[1] as well since the bounds can be negative as well.

botorch/optim/optimize_mixed.py
```diff
@@ -515,8 +571,8 @@ def generate_starting_points(
         )
         x_init_candts = torch.cat([x_init_candts, perturb_nbors], dim=0)
 
-    # Process the candidates to make sure they are all feasible.
-    x_init_candts[..., discrete_dims] = x_init_candts[..., discrete_dims].round()
+    # For each discrete dimension, map to the nearest allowed value
+    x_init_candts = round_discrete_dims(X=x_init_candts, discrete_dims=discrete_dims)
```
saitcakmak, line 586: Don't we need to round the categorical dims as well?

botorch/optim/optimize_mixed.py
```diff
@@ -767,7 +824,8 @@ def optimize_acqf_mixed_alternating(
     Args:
         acq_function: BoTorch Acquisition function.
         bounds: A `2 x d` tensor of lower and upper bounds for each column of `X`.
-        discrete_dims: A list of indices corresponding to integer and binary parameters.
+        discrete_dims: A dictionary mapping indices of discrete and binary
+            dimensions to a list of allowed values for that dimension.
         cat_dims: A list of indices corresponding to categorical parameters.
```
saitcakmak, line 840: So, the categoricals are still required to be [0, 1, 2, ...] but discrete dimensions are relaxed. I think it'd be good to treat them equally here if it's not too much extra work.

botorch/optim/optimize_mixed.py
saitcakmak: I agree with Max here. If we don't need to order it, adding this operation just confuses the reader. It takes a few seconds of careful reading to really parse what this is doing.

botorch/optim/optimize_mixed.py
```diff
@@ -881,9 +957,13 @@ def optimize_acqf_mixed_alternating(
...
     # Remove fixed features from dims, so they don't get optimized.
-    discrete_dims = [dim for dim in discrete_dims if dim not in fixed_features]
+    discrete_dims = {
+        dim: values
+        for dim, values in discrete_dims.items()
+        if dim not in fixed_features
+    }
     cat_dims = [dim for dim in cat_dims if dim not in fixed_features]
-    non_cont_dims = [*discrete_dims, *cat_dims]
+    non_cont_dims = [*sorted(discrete_dims.keys()), *cat_dims]
```
saitcakmak: This sorting also seems unnecessary

botorch/optim/optimize_mixed.py
```diff
@@ -903,7 +983,7 @@ def optimize_acqf_mixed_alternating(
             "integers between 0 and num_dims - 1."
         )
     discrete_dims_t = torch.tensor(
-        discrete_dims, dtype=torch.long, device=tkwargs["device"]
+        list(discrete_dims.keys()), dtype=torch.long, device=tkwargs["device"]
     )
```
saitcakmak, line 997: This seems unused after your changes

saitcakmak (changes requested):
This seems reasonable overall. My main concern is with clamping operations, which may be introducing a bug. Curious to hear your take on extending this to categoricals as well

botorch/optim/optimize_mixed.py
Balandat: I thought so as well at first but it's used in l 1020

botorch/optim/optimize_mixed.py
jduerholt: You are right, I remember now why I added it, it was nonsense. I will remove it.

botorch/optim/optimize_mixed.py
jduerholt: Yeah, this PR is only dealing with the discrete ones. My plan is to file a consecutive PR in which I introduce the same syntax for categoricals. My plan was to get first the feedback on the discrete ones to have then less work with the categoricals ;) Do you want to have the categoricals also in this PR?

botorch/optim/optimize_mixed.py
jduerholt: See answer below.

botorch/optim/optimize_mixed.py
jduerholt: Good catch, we need to clamp, but not with the bounds ;) I overlooked this. I will change it.

botorch/optim/optimize_mixed.py
jduerholt: Totally correct, we need to clamp with zero here.

botorch/optim/optimize_mixed.py
jduerholt: I will investigate this, after fixing the clamping from above.

botorch/optim/optimize_mixed.py
jduerholt: I agree.

botorch/optim/optimize_mixed.py
Balandat: I think it's fine to work incrementally and deal with the categoricals in a separate PR, so long as this PR clearly states hat limitation in the docstring and errors our if the input doesn't comply with that requirement.

botorch/optim/optimize_mixed.py
Balandat: @jduerholt this still needs to be changed it seems

botorch/optim/optimize_mixed.py
Balandat: same here

Balandat (commented):
Looks like a couple of items are still open, o/w this is getting there!

jduerholt:
> Looks like a couple of items are still open, o/w this is getting there!

I will work on it, at latest next week ;) Best, Johannes

botorch/optim/optimize_mixed.py
jduerholt: Good catch, it was not necessary and is removed now.

botorch/optim/optimize_mixed.py
jduerholt: Done.

botorch/optim/optimize_mixed.py
jduerholt: Done.

jduerholt:
@saitcakmak @Balandat I fixed the raised point, it should be ready for re-review now ;) Best, Johannes

saitcakmak (approved):
Thanks!

saitcakmak:
Looks like line 694 is missing in codecov. I think it should be ok to mark it with #pragma: no cover

Update: looks like I have write access, fixed.

jduerholt:
> Looks like line 694 is missing in codecov. I think it should be ok to mark it with #pragma: no cover
> 
> Update: looks like I have write access, fixed.

Thanks! I plan to implement the same syntax for categoricals next week.

jduerholt:
> > Looks like line 694 is missing in codecov. I think it should be ok to mark it with #pragma: no cover
> > Update: looks like I have write access, fixed.
> 
> Thanks! I plan to implement the same syntax for categoricals next week in a seperate PR.

PR #2925, opened 2025-07-12
Fix edge case with NaNs in `is_non_dominated`

### LINKED ISSUE #2924: [Bug]: is_non_dominated returns True for Nan values at idx 0
### What happened?

I just came across a small bug in _botorch.utils.multi_objective.pareto.is_non_dominated_, when dealing with `NaN` values.

When the first value of a batch is `NaN` (at index n=0), the method unexpectedly returns `True` at index 0 instead of the expected `False`. This issue consistently occurs across all batches at n=0, also with a batch size of 1. Other indices with `NaN` values return `False` as expected. If the first value is not `NaN`, the output is as expected.

**Expected Behavior**:
Always return `False` if the value is `NaN`.

**Actual Behavior**:
Returns `True` for `NaN` values at index 0.

```
import torch

nans = torch.full((2, 2, 3), torch.nan)
rands = torch.rand((2, 5, 3))
Y = torch.hstack([nans, rands])
non_dominated = is_non_dominated(Y)
```

```shell
>>> non_dominated
tensor([[ True, False, False, False,  True,  True, False],
        [ True, False,  True,  True,  True,  True,  True]])
```

### BoTorch Version

0.14.0

### Python Version

3.13.5

### Operating System

macOS

_No response_

### Pull Request

None

### Code of Conduct

Balandat:
Addresses an issue of elements being marked as non-dominated when all elements of a tensor are `NaN`.

Fixes #2924

### DIFF
--- a/botorch/utils/multi_objective/pareto.py
+++ b/botorch/utils/multi_objective/pareto.py
@@ -56,13 +56,14 @@ def is_non_dominated(
     if n > 1000 or n**2 * Y.shape[:-2].numel() * el_size / 8 > MAX_BYTES:
         return _is_non_dominated_loop(Y, maximize=maximize, deduplicate=deduplicate)
 
+    is_all_nan = Y.isnan().all(dim=-1)  # edge case: all elements are NaN
     Y1 = Y.unsqueeze(-3)
     Y2 = Y.unsqueeze(-2)
     if maximize:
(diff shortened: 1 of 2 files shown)


Balandat:
Test failures are unrelated.

saitcakmak (approved):

PR #2931, opened 2025-07-18
optimize sequence of acquisition functions

bletham:
Summary: Enables sequential q-batch optimization using a sequence of acquisition functions rather than the same acquisition function for each point in the batch. Right now just for when using optimize_acqf, thus focused on continuous search spaces.

Differential Revision: D78560867

### DIFF
--- a/botorch/optim/optimize.py
+++ b/botorch/optim/optimize.py
@@ -65,7 +65,7 @@ class OptimizeAcqfInputs:
     See docstring for `optimize_acqf` for explanation of parameters.
     """
 
-    acq_function: AcquisitionFunction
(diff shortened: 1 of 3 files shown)


jduerholt:
Very interesting PR, one question: how does it relate to `optimize_acqf_list`? https://github.com/pytorch/botorch/blob/e7fef3bfa235b468af7ad43a3bd38aa690a161f9/botorch/optim/optimize.py#L855

bletham:
Essentially this folds the functionality of optimize_acqf_list up into the sequential_q optimizer. I can see optimize_acqf_list being deprecated at some point in the future after this goes in.

PR #2942, opened 2025-07-28
Update syntax for categoricals in `optimize_acqf_mixed_alternating`

jduerholt:
## Motivation

Following https://github.com/pytorch/botorch/pull/2923, this PR updates the syntax for the categorical dimensions in order to match the syntax for the discrete ones.

Yes.

## Test Plan

Unit tests.

### DIFF
--- a/test/optim/test_optimize_mixed.py
+++ b/test/optim/test_optimize_mixed.py
@@ -276,7 +276,6 @@ class TestOptimizeAcqfMixed(BotorchTestCase):
         # where we fall back to randomly sampling a subset.
         random.seed(0)
         current_x = torch.tensor([50.0, 5.0], device=self.device)
-        # cat_dims = torch.tensor([0, 1], device=self.device, dtype=torch.long)
         cat_dims = {0: list(range(101)), 1: list(range(9))}
 
         neighbors = get_categorical_neighbors(


botorch/optim/optimize_mixed.py
```diff
@@ -246,29 +245,31 @@ def get_categorical_neighbors(
...
+        dtype=torch.long,
+    )
     num_cat_values.clamp_(max=max_num_cat_values)
     new_cat_idcs = torch.cat(
         tuple(
             torch.full((num_cat_values[dim].item(),), dim, device=current_x.device)
-            for dim in cat_dims
+            for dim in range(len(cat_dims))
         )
     )
```
esantorella: We could do something like

```python
new_cat_values_dict = {dim: _get_cat_values(dim=dim) for dim in cat_dims.keys()}
new_cat_values_lst = list(itertools.chain.from_iterable(new_cat_values_dict.values()))
new_cat_values = torch.tensor(
    new_cat_values_lst, device=current_x.device, dtype=current_x.dtype
)
new_cat_idcs = torch.cat(
    tuple(
        torch.full((len(values),), dim, device=current_x.device)
        for dim, values in new_cat_values_dict.items()
    )
)
```

botorch/optim/optimize_mixed.py
```diff
@@ -807,7 +810,7 @@ def optimize_acqf_mixed_alternating(
     acq_function: AcquisitionFunction,
     bounds: Tensor,
     discrete_dims: dict[int, list[float]] | None = None,
-    cat_dims: list[int] | None = None,
+    cat_dims: dict[int, list[float]] | None = None,
```
esantorella, line 811: This is going to create a backwards compatibility issue since this is a function other libraries or users may call (Ax uses `optimize_acqf_mixed_alternating`). Could you keep `list[int]` allowed and deprecate that usage?

jduerholt: Hmm, not sure about this, I did the same for `discrete_dims` in https://github.com/pytorch/botorch/pull/2923 and I discussed with @Balandat and @saitcakmak that we want do do the same for `cat_dims`. We also did not keep `list[int]` allowed for the `discrete_dims`. Furthermore we discussed in https://github.com/pytorch/botorch/discussions/2904#discussioncomment-13631993 that it is fine to introduce these breaking changes.

But I could try to keep the backward compatibility by reading the categorical values from the bounds. But then one should do the same also for the `discrete_dims`, or?

saitcakmak: I think this should be fine. For the `discrete_dims` change, I landed it together with a diff that updated the Ax usage

esantorella: Sounds good, I can put in an Ax change that we'll land at the same time.

botorch/optim/optimize_mixed.py
jduerholt: I updated it accordingly and added the clamping operation.

botorch/optim/optimize_mixed.py
jduerholt: Thanks!

test/optim/test_optimize_mixed.py
```diff
@@ -259,12 +276,11 @@ def test_get_categorical_neighbors(self) -> None:
         # where we fall back to randomly sampling a subset.
         random.seed(0)
         current_x = torch.tensor([50.0, 5.0], device=self.device)
-        bounds = torch.tensor([[0.0, 0.0], [100.0, 8.0]], device=self.device)
-        cat_dims = torch.tensor([0, 1], device=self.device, dtype=torch.long)
+        # cat_dims = torch.tensor([0, 1], device=self.device, dtype=torch.long)
+        cat_dims = {0: list(range(101)), 1: list(range(9))}
```
esantorella:
```suggestion
        cat_dims = {0: list(range(101)), 1: list(range(9))}
```

test/optim/test_optimize_mixed.py
```diff
@@ -303,15 +319,29 @@ def test_sample_feasible_points(self, with_constraints: bool = False) -> None:
...
+            num_points=10,
+        )
+        self.assertTrue(
+            torch.all(torch.isin(X[..., 1], torch.tensor([2.0, 3.0, 5.0]).to(X)))
+        )
+        self.assertEqual(X.shape, torch.Size([10, 3]))
+        self.assertTrue(torch.all(X[..., 0] == 0.5))
+        if with_constraints:
+            self.assertTrue(torch.all(X[..., 1] >= 4.0))
+        self.assertAllClose(X[..., 1], X[..., 1].round())
```
esantorella, line 338: Thanks for these checks!

jduerholt:
Thanks for the quick review. Next PR will deal with linear equality constraints and nonlinear inequality constraints for the continuous degrees of freedom in `optimize_acqf_mixed` ;) Best, Johannes

PR #2944, opened 2025-07-30
Equality constraints for `optimize_acqf_mixed_alternating`

jduerholt:
## Motivation

Following https://github.com/pytorch/botorch/pull/2942 and https://github.com/pytorch/botorch/pull/2923, this PR adds support for linear equality constraints (only for continuous degrees of freedom) for `optimize_acqf_mixed_alternating`.

I am thinking about adding also support for non-linear inequality constraints here. I would do it as for `optimize_acqf`: If nonlinear constraints are present, the user has to provide a sampler (`ic_generator`). What do you think?

Yes.

## Test Plan

Unit tests.

### DIFF
--- a/botorch/optim/optimize_mixed.py
+++ b/botorch/optim/optimize_mixed.py
@@ -108,7 +108,9 @@ def _setup_continuous_relaxation(
 
 
 def _filter_infeasible(
-    X: Tensor, inequality_constraints: list[tuple[Tensor, Tensor, float]] | None
+    X: Tensor,
+    inequality_constraints: list[tuple[Tensor, Tensor, float]] | None,
+    equality_constraints: list[tuple[Tensor, Tensor, float]] | None,
 ) -> Tensor:
     r"""Filters infeasible points from a set of points.
 
@@ -123,15 +125,26 @@ def _filter_infeasible(
             `\sum_i (X[indices[i]] * coefficients[i]) >= rhs`. `indices` and
             `coefficients` should be torch tensors. See the docstring of
             `make_scipy_linear_constraints` for an example.
+        equality_constraints: A list of tuples (indices, coefficients, rhs),
+            with each tuple encoding an equality constraint of the form
+            `\sum_i (X[indices[i]] * coefficients[i]) == rhs`. `indices` and
+            `coefficients` should be torch tensors. See the docstring of
+            `make_scipy_linear_constraints` for an example.
 
     Returns:
         The tensor `X` with infeasible points removed.
     """
-    if inequality_constraints is None:
+    if inequality_constraints is None and equality_constraints is None:
         return X
     is_feasible = torch.ones(X.shape[:-1], device=X.device, dtype=torch.bool)
-    for idx, coef, rhs in inequality_constraints:
-        is_feasible &= (X[..., idx] * coef).sum(dim=-1) >= rhs
+    if inequality_constraints is not None:
+        for idx, coef, rhs in inequality_constraints:
+            is_feasible &= (X[..., idx] * coef).sum(dim=-1) >= rhs
+    if equality_constraints is not None:
+        for idx, coef, rhs in equality_constraints:
+            is_feasible &= torch.isclose(
+                (X[..., idx] * coef).sum(dim=-1), torch.tensor(rhs).to(coef)
+            )
     return X[is_feasible]
 
 
@@ -385,8 +398,10 @@ def sample_feasible_points(
     all_points = torch.empty(
         0, bounds.shape[-1], device=bounds.device, dtype=bounds.dtype
     )
-    constraints = opt_inputs.inequality_constraints
-    if constraints is None:
+    if (
+        opt_inputs.inequality_constraints is None
+        and opt_inputs.equality_constraints is None
+    ):
         # Generate base points using Sobol.
         sobol_engine = SobolEngine(dimension=bounds.shape[-1], scramble=True)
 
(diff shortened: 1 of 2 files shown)


test/optim/test_optimize_mixed.py
```diff
@@ -1013,6 +1065,53 @@ def test_optimize_acqf_mixed_integer(self) -> None:
...
+            q=3,
+            raw_samples=32,
+            num_restarts=4,
+            options={"batch_limit": 5, "init_batch_limit": 20},
+            equality_constraints=[constraint],
+        )
+        self.assertAllClose(
+            candidates[:, [1, 2]].sum(dim=-1),
+            torch.ones(3, device=self.device),
+        )
```
esantorella: Thanks for confirming that it works!

botorch/optim/optimize_mixed.py
```diff
@@ -123,15 +125,26 @@ def _filter_infeasible(
...
-    for idx, coef, rhs in inequality_constraints:
-        is_feasible &= (X[..., idx] * coef).sum(dim=-1) >= rhs
+    if inequality_constraints is not None:
+        for idx, coef, rhs in inequality_constraints:
+            is_feasible &= (X[..., idx] * coef).sum(dim=-1) >= rhs
+    if equality_constraints is not None:
+        for idx, coef, rhs in equality_constraints:
+            is_feasible &= torch.isclose(
+                (X[..., idx] * coef).sum(dim=-1), torch.tensor(rhs).to(coef)
+            )
```
esantorella, line 147: Could you use `evaluate_feasibility` from `botorch.optim.parameter_constraints` to increase code-sharing? If we need to change a tolerance, it would be better to not have to do that in multiple places.

botorch/optim/optimize_mixed.py
```diff
@@ -123,15 +125,26 @@ def _filter_infeasible(
             `\sum_i (X[indices[i]] * coefficients[i]) >= rhs`. `indices` and
             `coefficients` should be torch tensors. See the docstring of
             `make_scipy_linear_constraints` for an example.
+        equality_constraints: A list of tuples (indices, coefficients, rhs),
+            with each tuple encoding an equality constraint of the form
+            `\sum_i (X[indices[i]] * coefficients[i]) == rhs`. `indices` and
+            `coefficients` should be torch tensors. See the docstring of
+            `make_scipy_linear_constraints` for an example.
```
esantorella:
```suggestion
            `coefficients` should be torch tensors. Example:
            `[(torch.tensor([1, 3]), torch.tensor([1.0, 0.5]), -0.1)]`
```

botorch/optim/optimize_mixed.py
```diff
@@ -882,6 +917,12 @@ def optimize_acqf_mixed_alternating(
             `\sum_i (X[indices[i]] * coefficients[i]) >= rhs`. `indices` and
             `coefficients` should be torch tensors. See the docstring of
             `make_scipy_linear_constraints` for an example.
+        equality_constraints: A list of tuples (indices, coefficients, rhs),
+            with each tuple encoding an equality constraint of the form
+            `\sum_i (X[indices[i]] * coefficients[i]) == rhs`. `indices` and
+            `coefficients` should be torch tensors. See the docstring of
+            `make_scipy_linear_constraints` for an example. Equality
+            constraints can only be used with continuous degrees of freedom.
```
esantorella:
```suggestion
            `coefficients` should be torch tensors. Example:
            `[(torch.tensor([1, 3]), torch.tensor([1.0, 0.5]), -0.1)]` Equality
            constraints can only be used with continuous degrees of freedom.
```

test/optim/test_optimize_mixed.py
```diff
@@ -677,6 +695,40 @@ def test_continuous_step(self):
...
+                X_new[0, [8, 9]].sum(),
+                torch.tensor(
+                    [
+                        1.0,
+                    ],
+                    device=self.device,
+                ),
+            )
+        )
+
```
esantorella: Nit
```suggestion
        self.assertAllClose(
            X_new[0, [8, 9]].sum(),
            torch.tensor([1.0], device=self.device),
        )

```

test/optim/test_optimize_mixed.py
```diff
@@ -1013,6 +1065,53 @@ def test_optimize_acqf_mixed_integer(self) -> None:
             torch.tensor([0.5, 0.5, 2.0], device=self.device).repeat(3, 1),
         )
 
+        # Test with equality constraints.
+        constraint = (  # X[..., 0] + X[..., 1] >= 1.
```
esantorella:
```suggestion
        constraint = (  # X[..., 1] + X[..., 2] >= 1.
```

test/optim/test_optimize_mixed.py
```diff
@@ -1013,6 +1065,53 @@ def test_optimize_acqf_mixed_integer(self) -> None:
...
+            options={"batch_limit": 5, "init_batch_limit": 20},
+            equality_constraints=[constraint],
+        )
+        self.assertAllClose(
+            candidates[:, [1, 2]].sum(dim=-1),
+            torch.ones(3, device=self.device),
+        )
+
+        # Test with equality constraint and fixed feature in constraint.
+        constraint = (  # X[..., 0] + X[..., 1] >= 1.
```
esantorella:
```suggestion
        constraint = (  # X[..., 1] + X[..., 2] >= 1.
```

esantorella (changes requested):

esantorella:
Thanks so much for putting this in! This looks pretty good to me. Just a few minor comments.

botorch/optim/optimize_mixed.py
jduerholt: Good point, will do so!

jduerholt: Hi @esantorella,

I had a deeper look into `evaluate_feasibility` https://github.com/pytorch/botorch/blob/17a066bf7bba95400925a826212c52bf93ab1ad2/botorch/optim/parameter_constraints.py#L613

It is not directly applicable here as it always checks feasibility on the q-batch level and not on the row level as needed here. The reason why the checks are perfomed on the q level is that it also checks for interpoint constraints. One could modify the behavior of the method by introducing a flag if it should perform the check on the row/candidate level, and if the flag is set ignore the interpoint constraints. 

This could make sense in the spirit of code sharing as I also want to introduce nonlinear constraints (as mentioned above).

What do you think regarding the flag and the possible inclusion of interpoint constraints?

Best,

Johannes

esantorella: Oh I see. Yes, I think it would be great to give that function an option to evaluate feasibility either interpoint or not. Another option would be to call that function separately for each element of the batch. Making it easier to incorporate nonlinear constraints sounds great. I'll leave it up to you.

esantorella (approved):

botorch/optim/optimize_mixed.py
jduerholt: Hi @esantorella , I propose the following: Let us merge this PR as is, at is already reviewed etc. I will then file a consecutive one which introduces nonlinear constraints and tidies up the feasibility evaluation. Then one has it step by step which is also easier for review. Fine for you? Best, Johannes

esantorella: Sounds great, thank you for making these easy to review!

PR #2952, opened 2025-08-05
Tidy-up feasibility evaluation in `optimize_mixed_alternating`

jduerholt:
## Motivation

As discussed with @esantorella in https://github.com/pytorch/botorch/pull/2944#discussion_r2250503884, this PR tidies-up the feasibility evaluation in `optimize_mixed_alternating.` It is now refactored in a way that it uses `optim.parameter_constraints.evaluate_feasibility`, which improves code sharing.

Regarding nonlinear inequality constraints that were also discussed in the thread above, I faced the problem that nonlinear constraints only work with a batch limit of 1, which would render `optimize_mixed_alternating` highly inefficient as it makes use of batch optimization. So, I think, we have to first get rid of the limitation of `batch_limit==1` for nonlinear constraints. Should I create a separate issue for this?

Yes.

## Test Plan

Unit tests.

### DIFF
--- a/botorch/optim/optimize_mixed.py
+++ b/botorch/optim/optimize_mixed.py
@@ -138,7 +138,7 @@ def _filter_infeasible(
     # X is reshaped to [n, 1, d] in order to be able to apply
     # `evaluate_feasibility` which operates on the batch level
     is_feasible = evaluate_feasibility(
-        X=X.reshape(X.shape[0], 1, X.shape[-1]),
+        X=X.unsqueeze(-2),
         inequality_constraints=inequality_constraints,
         equality_constraints=equality_constraints,
         nonlinear_inequality_constraints=None,


botorch/optim/optimize_mixed.py
```diff
@@ -134,17 +135,14 @@ def _filter_infeasible(
...
-            is_feasible &= (X[..., idx] * coef).sum(dim=-1) >= rhs
-    if equality_constraints is not None:
-        for idx, coef, rhs in equality_constraints:
-            is_feasible &= torch.isclose(
-                (X[..., idx] * coef).sum(dim=-1), torch.tensor(rhs).to(coef)
-            )
+    # X is reshaped to [n, 1, d] in order to be able to apply
+    # `evaluate_feasibility` which operates on the batch level
+    is_feasible = evaluate_feasibility(
+        X=X.reshape(X.shape[0], 1, X.shape[-1]),
```
saitcakmak:
```suggestion
        X=X.unsqueeze(-2),
```

saitcakmak (approved):
Thanks! 

Re non-linear: I think the alternating optimizer is overall slower than the continuous optimizers. I think it'd be great to keep it batched to retain the recent speed ups.

PR #2972, opened 2025-08-13
Fixed tutorial link in "getting_started.mdx"

SpencerHalle:
The link was broken, so I updated it, it works now.

## Motivation

The link didn't work for the tutorial section, so I fixed it

Yes

## Test Plan

If you click the link, it goes to the page

## Related PRs

### DIFF
--- a/docs/getting_started.mdx
+++ b/docs/getting_started.mdx
@@ -88,7 +88,7 @@ Here's a quick run down of the main components of a Bayesian Optimization loop.
 ## Tutorials
 
 Our Jupyter notebook tutorials help you get off the ground with BoTorch.
-View and download them [here](https://botorch.org/docs/tutorials/).
+View and download them [here](tutorials/index.mdx).
 
 
 ## API Reference


Balandat:
Hmm so these docs are mainly used on the website where docusaurus (which powers our website) automatically resolves these links. @SpencerHalle could you verify that the link works for you on the rendered docs on https://botorch.org/docs/getting_started ?

esantorella:
This is weird. I went to botorch.org/docs/getting_started and hovered over the link, so you can see it in the lower left of the screenshot. I sometimes see this link point to the correct place, botorch.org/docs/tutorials:
image

But after refreshing the page, the link is incorrect, pointing to botorch.org/docs/getting_started/tutorials:

image

Balandat:
cc @CristianLara

CristianLara:
Thank you so much for the contribution @SpencerHalle!

You're right there's a problem here. Docusaurus is resolving that link as a url path relative from the "getting_started" page, and that behavior is different depending whether the url contains a trailing slash or not (which github pages will sometimes but not always add 😓). We can see this in @esantorella's screenshots above where the only difference is the presence of a trailing slash and that causes the link to be resolved differently.

The change in this PR fixes the issue by writing out the full URL and circumventing link resolution, but that creates another issue: the full URL link will ignore our website versioning which is specified using a `/<version>/` prefix in the url. Links in our docs should always take you to the linked doc within the version you are currently viewing. Docusaurus handles this for us automatically unless the link is already a fully constructed url.

## Proposed Solution

Our preferred solution here is pretty simple, if we define the link as a relative **file** path instead of a **url** path then docusaurus will resolve this correctly whether there is a trailing slash or not.

```diff
- View and download them [here](tutorials).
+ View and download them [here](tutorials/index.mdx).
```

**File** paths have the benefit of always being resolved against the current file's directory rather than the path in the url (which can be affected by trailing slashes). See the [Docusaurus docs](https://docusaurus.io/docs/markdown-features/links) for more details

docs/getting_started.mdx
```diff
@@ -88,7 +88,7 @@ Here's a quick run down of the main components of a Bayesian Optimization loop.
 ## Tutorials
 
 Our Jupyter notebook tutorials help you get off the ground with BoTorch.
-View and download them [here](tutorials).
+View and download them [here](https://botorch.org/docs/tutorials/).
```
Balandat: As per @CristianLara's suggestion:
```suggestion
View and download them [here](tutorials/index.mdx).
```

PR #2982, opened 2025-08-18
lint: add py.typed

CompRhys:
https://github.com/facebook/Ax/pull/4139 repeated for botorch.


Balandat:
@CompRhys note that while BoTorch uses type annotations, we actually (at least currently) don't use a typechecker on it. This is largely b/c in the past this has caused substantial headaches with pytorch code. Do we want this file in light of that?

CompRhys:
I still think it makes sense to add because it allows type checkers above to know that the outputs have types here. If they need to be ignored in some cases that's fine

Issue #2996, opened 2025-08-29
[Bug]: Interpoint constraints cannot be used with `optimize_acqf_mixed`


AVHopp:
### What happened?

When using interpoint constraints in `optimize_acqf_mixed`, an index error is happening internally - however, I am not sure if this is actually a bug or if interpoint constraints cannot be used with this function in general (see description of fix).

### Please provide a minimal, reproducible example of the unexpected behavior.

from __future__ import annotations

import torch
from botorch.acquisition import qLogExpectedImprovement
from botorch.fit import fit_gpytorch_mll
from botorch.models import SingleTaskGP
from botorch.optim import optimize_acqf_mixed
from gpytorch.mlls import ExactMarginalLogLikelihood

# 1 discrete dim, 2 continuous dims
BATCH_SIZE = 3  # q
D_DISC = 1
D_CONT = 2

bounds = torch.stack(
    [
        torch.zeros(D_DISC + D_CONT),
        torch.ones(D_DISC + D_CONT),
    ]
)

fixed_features_list = [
    {0: 0.0},
    {0: 0.5},
    {0: 1.0},
]

indices_interpoint = torch.tensor([(b, 1) for b in range(BATCH_SIZE)])
coefficients = torch.ones(BATCH_SIZE)
rhs = 1.0

interpoint_eq_constraint = (indices_interpoint, coefficients, rhs)

if __name__ == "__main__":
    train_X = torch.rand(8, D_DISC + D_CONT)
    train_Y = train_X.pow(2).sum(dim=-1, keepdim=True)

    model = SingleTaskGP(train_X, train_Y)
    mll = ExactMarginalLogLikelihood(model.likelihood, model)
    fit_gpytorch_mll(mll)

    best_f = train_Y.max().item()
    acqf = qLogExpectedImprovement(model=model, best_f=best_f)

    optimize_acqf_mixed(
        acq_function=acqf,
        bounds=bounds,
        q=BATCH_SIZE,
        num_restarts=5,
        raw_samples=32,
        fixed_features_list=fixed_features_list,
        equality_constraints=[interpoint_eq_constraint],
        inequality_constraints=None,
    )

### Please paste any relevant traceback/logs produced by the example provided.

```shell
Traceback (most recent call last):
  File "/Users/m290244/Desktop/Programming/BayesianBE/baybe/examples/misc/botorch_mixed_interpoint_failure.py", line 45, in <module>
    optimize_acqf_mixed(
[...]

PR #3003, opened 2025-09-05
Update documentation and error handling for inter-point constraints

AVHopp:
## Motivation

This PR improves the documentation and error handling when attempting to use inter-point constraints within `optimize_acqf_mixed` which is not supported. It implements the solution to #2996 discussed in this issue.

Yes.

## Test Plan

I verified my code by ensuring that the example posted in #2996 raises an error message. I did not add any tests, but would be happy to turn that example into a dedicated test if being pointed to the correct place for doing so.

## Related PRs

None.

### DIFF
--- a/botorch/optim/optimize.py
+++ b/botorch/optim/optimize.py
@@ -1055,6 +1055,9 @@ def optimize_acqf_mixed(
     For q > 1 this function always performs sequential greedy optimization (with
     proper conditioning on generated candidates).
 
+    NOTE: This method does not support the kind of "inter-point constraints" that
+    are supported by `optimize_acqf()`.
+
     Args:
         acq_function: An AcquisitionFunction
         bounds: A `2 x d` tensor of lower and upper bounds for each column of `X`
@@ -1076,19 +1079,8 @@ def optimize_acqf_mixed(
             `\sum_i (X[indices[i]] * coefficients[i]) = rhs`
         nonlinear_inequality_constraints: A list of tuples representing the nonlinear
             inequality constraints. The first element in the tuple is a callable
-            representing a constraint of the form `callable(x) >= 0`. In case of an
-            intra-point constraint, `callable()`takes in an one-dimensional tensor of
-            shape `d` and returns a scalar. In case of an inter-point constraint,
-            `callable()` takes a two dimensional tensor of shape `q x d` and again
-            returns a scalar. The second element is a boolean, indicating if it is an
-            intra-point or inter-point constraint (`True` for intra-point. `False` for
-            inter-point). For more information on intra-point vs inter-point
-            constraints, see the docstring of the `inequality_constraints` argument to
-            `optimize_acqf()`. The constraints will later be passed to the scipy
-            solver. You need to pass in `batch_initial_conditions` in this case.
-            Using non-linear inequality constraints also requires that `batch_limit`
-            is set to 1, which will be done automatically if not specified in
-            `options`.
+            representing a constraint of the form `callable(x) >= 0`. The `callable()`
+            takes in an one-dimensional tensor of shape `d` and returns a scalar.
         post_processing_func: A function that post-processes an optimization
             result appropriately (i.e., according to `round-trip`
             transformations).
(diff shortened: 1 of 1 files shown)


botorch/optim/optimize.py
```diff
@@ -1283,8 +1296,7 @@ def optimize_acqf_discrete(
     """
     if isinstance(acq_function, OneShotAcquisitionFunction):
         raise UnsupportedError(
-            "Discrete optimization is not supported for"
-            "one-shot acquisition functions."
+            "Discrete optimization is not supported forone-shot acquisition functions."
```
Balandat:
```suggestion
            "Discrete optimization is not supported for one-shot acquisition functions."
```

botorch/optim/optimize.py
```diff
@@ -1123,6 +1115,27 @@ def optimize_acqf_mixed(
...
+    # Check for existence of inter-point constraints
+    # Code adapted from _validate_sequential_inputs
+    if inequality_constraints is not None:
+        for i, constraint in enumerate(inequality_constraints):
+            if len(constraint[0].shape) > 1:
+                raise UnsupportedError(const_err_message.format(i, "linear inequality"))
+    if equality_constraints is not None:
+        for i, constraint in enumerate(equality_constraints):
+            if len(constraint[0].shape) > 1:
+                raise UnsupportedError(const_err_message.format(i, "linear equality"))
```
Balandat:
```suggestion
        for i, (indices, _, _) in enumerate(inequality_constraints):
            if indices.ndim > 1:
                raise UnsupportedError(const_err_message.format(i, "linear inequality"))
    if equality_constraints is not None:
        for i, (indices, _, _) in enumerate(equality_constraints):
            if indices.ndim > 1:
                raise UnsupportedError(const_err_message.format(i, "linear equality"))
```

Balandat (changes requested):
Thanks for adding this!

> I did not add any tests, but would be happy to turn that example into a dedicated test if being pointed to the correct place for doing so.

It would be great if you could add a lightweight `test_error_on_interpoint_constraints()` test to `TestOptimizeAcqfMixed` here https://github.com/pytorch/botorch/blob/main/test/optim/test_optimize.py#L1765 to validate that the error gets raised when passing in inter-point constraints.

AVHopp:
@Balandat I've applied your suggestions and added a small test here: 92140e9843ceb899f78e6f15fd4b4aea78c6fdfe

test/optim/test_optimize.py
```diff
@@ -1934,6 +1934,29 @@ def test_optimize_acqf_mixed_empty_ff(self):
                 raw_samples=10,
             )
 
+    def test_optimize_acqf_mixed_inter_point_constraints(self):
+        mock_acq_function = MockAcquisitionFunction()
+        with self.assertRaisesRegex(
+            UnsupportedError,
+            expected_regex="Inter-point constraints are not supported for sequential optimization. "
```
Balandat: @AVHopp this line is a bit too long (failed the lint), mind updating this? You can run the pre-commit check locally to make sure lint is ok as described here: https://github.com/pytorch/botorch/blob/main/CONTRIBUTING.md#pre-commit-hooks

test/optim/test_optimize.py
```diff
@@ -1934,6 +1934,29 @@ def test_optimize_acqf_mixed_empty_ff(self):
                 raw_samples=10,
             )
 
+    def test_optimize_acqf_mixed_inter_point_constraints(self):
```
esantorella: Would you be able to add tests for equality constraints and nonlinear inequality constraints too? That would remedy the codecov failure.

AVHopp: Added in c6e00f11ddc53474e70ad044fdffe0dff61571d7

test/optim/test_optimize.py
AVHopp: FIxed in 5bb83c215e8993908e1bdbdd1b6f276e57cde211

botorch/optim/optimize.py
```diff
@@ -1076,19 +1079,12 @@ def optimize_acqf_mixed(
             `\sum_i (X[indices[i]] * coefficients[i]) = rhs`
         nonlinear_inequality_constraints: A list of tuples representing the nonlinear
             inequality constraints. The first element in the tuple is a callable
-            representing a constraint of the form `callable(x) >= 0`. In case of an
-            intra-point constraint, `callable()`takes in an one-dimensional tensor of
-            shape `d` and returns a scalar. In case of an inter-point constraint,
-            `callable()` takes a two dimensional tensor of shape `q x d` and again
-            returns a scalar. The second element is a boolean, indicating if it is an
+            representing a constraint of the form `callable(x) >= 0`. The `callable()`
```
AVHopp, line 1082: Note: I have realized that the `nonlinear_inequality_constraints` typically a `list[tuple[Callable, bool]]` type, where the cool indicates inter vs. intra-point constraint. With the change here, the bool is now basically redundant. However, for consistency, I decided to keep it but adjusted the docstring accordingly. This comment is this to make you aware of this change, in particular if you'd prefer to handle this bool differently :)

Balandat: I think this is great for now; we probably don't want to update the signature to make sure we don't have to do a bunch of case distinctions to format the function inputs.

Issue #3009, opened 2025-09-12
[Bug]: Poor performance of `compute_best_feasible_objective` with no feasible points


TobyBoyne:
### What happened?

I am performing an optimization with qLogNEI and an outcome constraint. When generating candidates with no feasible values in the training data, the acqf is zero almost everywhere, leading to candidates that ignore the constraint. This leads to degenerate behaviour, where the optimizer will keep proposing the same (infeasible) point throughout the BO loop.

I think that the acqf is zero everywhere because the `compute_best_feasible_objective` does not give a good lower bound for infeasible values. I have found some issues with the current `compute_best_feasible_objective`, each of which contributes some part to this problem. Sorry for the very long ticket!!

**1 Normalization.** The convex weights are normalized in the wrong dimension below - the weights are `(32, N)` and should be normalized along `N` so that the weights assigned to each point sum to 1. The current implementation essentially scales down all points to search near the origin. Solution is to just fix the dimension here.
https://github.com/pytorch/botorch/blob/1518b304f47f5cdbaf9c175e808c90b3a0a6b86d/botorch/acquisition/utils.py#L192

**2 Convex search excludes boundaries.** By using convex weights, we don't explore the edges of the space which is where points satisfying the threshold are likely to lie. I don't think there is a good solution to this, since we don't have access to `bounds` when the class is initialized.

**3 Random convex weights favour middle.** By using random convex weights, we tend to explore very close to the middle. I think this is due to the Central Limit Theorem, but I haven't thought about this too deeply. Try simulating:
```python
w = torch.rand(32, N)
(w / w.sum(dim=1, keepdim=True)) @ torch.linspace(0, 1, N)
```
As `N` increases, the above samples tend closer and closer to `0.5`, so we don't actually get to explore the space very much. A solution would be to pass `torch.cat((X, convex_weights @ X))` to `get_infeasible_cost`, to at least ensure that we evaluate the minima at the points on the convex hull.
https://github.com/pytorch/botorch/blob/1518b304f47f5cdbaf9c175e808c90b3a0a6b86d/botorch/acquisition/utils.py#L198

**4 Prune baseline makes this worse** When we pass `prune_baseline=True` (the default), we only keep the infeasible point that maximizes the objective. However, we should also keep the point that is closest to being feasible, as we otherwise only evaluate the lower bound at an area of the search space that is very far away from being feasible. Solution is to either change this default, or be less aggressive in pruning the baseline.

[...]

PR #3011, opened 2025-09-13
Improve best feasible objective

TobyBoyne:
## Motivation

See #3009.

Yes

## Test Plan

After feedback on the initial draft of the PR, I will implement tests similar to the problem structure in #3009, confirming that the behaviour of the modified code behaves as desired.

## Related PRs

None

### DIFF
--- a/botorch/acquisition/utils.py
+++ b/botorch/acquisition/utils.py
@@ -198,11 +198,13 @@ def _estimate_objective_lower_bound(
     """
     # we do not have access to `bounds` here, so we infer the bounding box
     # from data, expanding by 10% in each direction
-    X_lb = X.min(dim=-2).values
-    X_ub = X.max(dim=-2).values
+    X_lb = X.min(dim=-2, keepdim=True).values
+    X_ub = X.max(dim=-2, keepdim=True).values
     X_range = X_ub - X_lb
     X_padding = 0.1 * X_range
-    uniform_samples = torch.rand(32, X.shape[-1], dtype=X.dtype, device=X.device)
+    uniform_samples = torch.rand(
+        *X.shape[:-2], 32, X.shape[-1], dtype=X.dtype, device=X.device
+    )
     X_samples = X_lb - X_padding + uniform_samples * (X_range + 2 * X_padding)
     # infeasible cost M is such that -M < min_x f(x), thus
     # 0 < min_x f(x) - (-M), so we should take -M as a lower


TobyBoyne:
If reviewers are happy with these changes, I will also implement them for the non-log version, as well as investigating the multi objective acqfs to see if similar changes are needed. I will also write tests in this PR, once the ideas in this PR have been approved.

botorch/acquisition/logei.py
```diff
@@ -282,6 +282,7 @@ def __init__(
         tau_relu: float = TAU_RELU,
         marginalize_dim: int | None = None,
         incremental: bool = True,
+        infeasible_obj: Tensor | float | None = None,
```
renzph: Should this also be added to `qNoisyExpectedImprovement`?

TobyBoyne: Yes - if the changes look good to whoever reviews, then I will also apply them to qNEI, as well as investigating q(Log)NEHVI too. Just wanted to get confirmation that the changes were good before making them elsewhere :)

botorch/acquisition/utils.py
```diff
@@ -183,19 +187,19 @@ def _estimate_objective_lower_bound(
...
-    weights_sum = convex_weights.sum(dim=0, keepdim=True)
-    convex_weights = convex_weights / weights_sum
+    # we do not have access to `bounds` here, so we infer the bounding box
+    # from data, expanding by 10% in each direction
+    X_lb = X.min(dim=-2)
+    X_ub = X.max(dim=-2)
+    X_range = X_ub - X_lb
+    X_padding = 0.1 * X_range
+    uniform_samples = torch.rand(32, X.shape[-1], dtype=X.dtype, device=X.device)
+    X_samples = X_lb - X_padding + uniform_samples * (X_range + X_padding)
```
Balandat: Should this be as follows?
```suggestion
    X_samples = X_lb - X_padding + uniform_samples * (X_range + 2 * X_padding)
```

botorch/acquisition/utils.py
```diff
@@ -235,8 +239,18 @@ def objective(Y: Tensor, X: Tensor | None = None):
...
+    # We check both the upper and lower bound of the posterior, since the objective
+    # may be increasing or decreasing. For objectives that are neither (eg. absolute
+    # distance from a target), this should still provide a good bound.
+    lb = torch.stack(
+        [
+            objective(posterior.mean - 6 * posterior.variance.clamp_min(0).sqrt(), X=X),
+            objective(posterior.mean + 6 * posterior.variance.clamp_min(0).sqrt(), X=X),
+        ],
+        dim=0,
+    )
```
Balandat, line 263: nit 
```suggestion
    six_stdv = 6 * posterior.variance.clamp_min(0).sqrt()
    lb = torch.stack(
        [
            objective(posterior.mean - six_stdv, X=X),
            objective(posterior.mean + six_stdv, X=X),
        ],
        dim=0,
    )
```

botorch/acquisition/utils.py
```diff
@@ -235,8 +239,18 @@ def objective(Y: Tensor, X: Tensor | None = None):
...
+
+    if lb.ndim - 1 < posterior.mean.ndim:
```
Balandat: why this change?

TobyBoyne: We add an extra dimension in the line above, by stacking the `mean - 6*std` and `mean + 6*std`. This then changes the check below. Maybe it would be better for me to be explicit above, taking the minimum of `mean +- 6*std` instead of stacking, and then leaving this line unchanged?

TobyBoyne:
Hi @Balandat, thanks for the review! I don't have any results showing the effect yet. Would you be interested in seeing the effect of these changes on `_estimate_objective_lower_bound`, or results from downstream BO tasks? I will be able to continue working on this PR in the coming week :)

Balandat:
@TobyBoyne ideally both - it seems that the behavior in this setting is degenerate enough that I'd expect to see pretty clear effects on the downstream BO tasks even without having to run many replications. 

And thanks a lot for your effort on improving this!

botorch/acquisition/logei.py
TobyBoyne: Decided to remove this argument, since the issue is best resolved by using ProbabilityOfFeasibility.

botorch/acquisition/utils.py
TobyBoyne: I have changed this to be more explicit.

TobyBoyne:
Hi @Balandat, I've carried out some investigation on the issue (#3009). Unfortunately not as cut-and-dry as I'd like, since the old implementation works for the "wrong reasons", ie. it evaluates the objective near 0, which can still lead to good estimation of a lower bound of the objective.

I've also had a look into the multi objective case, but I actually don't think any changes are required here: the current code explicitly handles all-infeasible points [here](https://github.com/meta-pytorch/botorch/blob/eba2dceb009b079ecaed794ecdb9c6cc80fb29df/botorch/acquisition/multi_objective/utils.py#L128) and [here](https://github.com/meta-pytorch/botorch/blob/eba2dceb009b079ecaed794ecdb9c6cc80fb29df/botorch/utils/multi_objective/hypervolume.py#L692), allowing `X_baseline` to be empty. I tried recreating the issue for the multiobjective case, and was unable to.

I think that's everything for this PR - please let me know if I've missed anything!

PR #3026, opened 2025-09-26
Enable evaluating PFNs trained with pytorch/PFNs to be evaluated in Ax

SamuelGabriel:
Summary: This PR enables PFNs generally to work in our MAST evaluation, as the registry didn't quite work before, and it additionally allows to use training checkpoints from pytorch/PFNs to be used to do evaluations.

Reviewed By: Balandat

Differential Revision: D80944578


Balandat:
These changes have landed as part of another / other commits

PR #3034, opened 2025-10-02
Fix bug in `condition_on_observation`

jduerholt:
## Motivation

Fix the bug outlined here: https://github.com/meta-pytorch/botorch/issues/3033

Tests are not yet adapted,  first, I want to get your opinion on the fix ;)

Yes.

## Test Plan

Unit tests, not yet adapted.

### DIFF
--- a/test/models/test_gpytorch.py
+++ b/test/models/test_gpytorch.py
@@ -732,9 +732,6 @@ class TestModelListGPyTorchModel(BotorchTestCase):
             conditioned_model = model.condition_on_observations(
                 condition_X, condition_Y
             )
-
-            # why is `original_train_inputs` a tensor and `train_inputs` a list?
-            # is this due to this helper model `SimpleGPyTorchModel`?
             self.assertAllClose(conditioned_model._original_train_inputs, X)
             self.assertAllClose(conditioned_model.train_inputs[0], input_transform(X))
 


hvarfner:
Hi @jduerholt , 

Thanks for spotting and fixing this. I haven't looked into it yet, but I'll start right away!

saitcakmak (approved):
Lgtm. I'll leave it to @hvarfner to give it a final look before merging.

hvarfner (commented):
This looks like a great solution to me! Did you have a test or two for this in the works, too?

jduerholt:
I will file some tests over the weekend ;)

jduerholt:
Hi @hvarfner, I added a test that checks for the desired behavior. I was only wondering regarding one data structure issue, see my comment in the code ;) Best, Johannes

test/models/test_gpytorch.py
```diff
@@ -691,6 +698,46 @@ def test_condition_on_observations_model_list(self):
...
+                train_X, train_Y, input_transform=input_transform
+            )
+            model.eval()
+            _ = model.posterior(train_X)
+
+            conditioned_model = model.condition_on_observations(
+                condition_X, condition_Y
+            )
+
+            # why is `original_train_inputs` a tensor and `train_inputs` a list?
```
hvarfner: @jduerholt I honestly don't know. Regardless, it's fine for the purpose of the test.

I assume the reason that train_inputs is a list is due to some `ModelList` dynamic. @saitcakmak Do you know the reason for _original_train_inputs being a tensor, and why train_input is a list of tensors?

saitcakmak: This is a GPyTorch thing. `train_inputs` is an attribute of `ExactGP` (a tuple, but I don't know when it has more than one element -- never seen it). `_original_train_inputs` is something I had added to support input transforms.

jduerholt: Thanks for the explanation, I removed the comment!

PR #3042, opened 2025-10-07
Correct handling of input transforms for `AdditiveMapSaasSingleTaskGP`

jduerholt:
## Motivation

In `AdditiveMapSaasSingleTaskGP`, input transforms that alter the shape of the training data are not handled correctly. This PR fixes it. 

A few questions to the Map Saas complex in general:
- We have now the ensemble version and the non-ensemble version, which one would you recommend in general?
- We have getter methods that generate fitted or unfitted versions of the MAP models and we have the option to just initialise the respective class and fit it in the default MAP way. What do you recommend here? Currently, I am using the `AdditiveMapSaasSingleTaskGP` via initialising the class and fitting it like any SingleTaskGP.

Yes.

## Test Plan

Unit tests.

### DIFF
--- a/test/models/test_map_saas.py
+++ b/test/models/test_map_saas.py
@@ -7,6 +7,7 @@
 import itertools
 import math
 import pickle
+from functools import partial
 from itertools import product
 from typing import Any
 from unittest import mock
@@ -27,7 +28,12 @@ from botorch.models.map_saas import (
     get_gaussian_likelihood_with_gamma_prior,
     get_mean_module_with_normal_prior,
 )
-from botorch.models.transforms.input import AppendFeatures, FilterFeatures, Normalize
+from botorch.models.transforms.input import (
+    AppendFeatures,
+    FilterFeatures,
+    Normalize,
+    NumericToCategoricalEncoding,
+)
 from botorch.models.transforms.outcome import Standardize
 from botorch.optim.utils import get_parameters_and_bounds, sample_all_priors
 from botorch.posteriors.fully_bayesian import GaussianMixturePosterior
@@ -43,6 +49,7 @@ from gpytorch.means import ConstantMean
 from gpytorch.mlls.exact_marginal_log_likelihood import ExactMarginalLogLikelihood
 from gpytorch.priors import GammaPrior, HalfCauchyPrior, NormalPrior
 from torch import Tensor
+from torch.nn.functional import one_hot
 
 
 class TestMapSaas(BotorchTestCase):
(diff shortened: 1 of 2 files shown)


saitcakmak:
Hi @jduerholt, great questions! I'll start by saying that the current stack around MAP SAAS is a bit of a mess and we need to do some clean up. The story goes something like this:
- We have fully Bayesian SAAS model, it works great but it is expensive.
- @dme65, looking for a way to make it faster, invents the MAP SAAS model.
- The single model doesn't work as well, so he creates a helper that individually fits multiple MAP SAAS models and loads them into a fully Bayesian model to produce an ensemble. 
- These helpers that get you a fitted model are designed around how the legacy Ax generators fitted models -- using a function call that returns a fitted model. This is not how a typical BoTorch model works and they're not compatible with Ax's Modular BoTorch Generator (MBG).
- I was working on migrating everything to MBG and we (mostly David) came up with `AdditiveMapSaas` that worked just like a typical BoTorch model. It tended to overfit on occasion (due to over-parameterized kernel & a single shared likelihood) but the BO performance was comparable, so we kept it around for a while and punted on the ensemble variant.
- We really wanted to get rid of the remaining legacy Ax generators, of which MAP SAAS was the primary one that was still in use. I had some changes sitting around that made the individually fitted, later ensembled setup work with MBG that I didn't like. While thinking through how to improve that, I realized that the ensemble model was just a multi-output model that had a different posterior and then came the new `EnsembleMapSaasSingleTaskGP`.

So, what's the future of this? 
- `EnsembleMapSaasSingleTaskGP` is here to stay. We're updating Ax / internal usage to use this instead of the old ensemble or the additive model. 
- `get_fitted_....` will be eventually deprecated. BoTorch models should be fit using `fit_gpytorch_mll` (or the nuts variant) and shouldn't need a separate helper.
- `AdditiveMapSaas` models will be deprecated as well (after adding MT variant of the ensemble etc). Having too many variants causes confusion. These perform good enough but the ensemble variants tend to be more robust to overfitting.
- `get_map_saas_model` may stay around or get replaced with a simple `MapSaasSingleTaskGP`. It has a nested prior over the sparsity (rather than fixed tau, tau has a half-cauchy prior -- which works better than fixed tau), so it is not equivalent to the ensemble model with a single tau. We don't use this one internally. I haven't benchmarked it recently but it does work well (not as good as the ensemble) and it is cheaper.

test/models/test_map_saas.py
```diff
@@ -647,6 +654,31 @@ def _get_data_and_model(
...
+                encoders={2: partial(one_hot, num_classes=3)},
+            )
+
+            model = AdditiveMapSaasSingleTaskGP(
+                train_X=X,
+                train_Y=Y,
+                input_transform=input_transform,
+            )
+            print(model.covar_module)
+            assert model.covar_module.kernels[0].base_kernel.ard_num_dims == 5
```
saitcakmak:
```suggestion
            self.assertEqual(model.covar_module.kernels[0].base_kernel.ard_num_dims, 5)
```

saitcakmak (approved):
thanks for the fix!

jduerholt:
Hi @saitcakmak,

thanks for the explanation, internally I was experimenting in recent times a lot with `AdditiveMapSaas` and it was really working fine, I fitted it always in the usual way. I am looking forward to use the Ensemble one in the future. 

Another idea, on which was thinking in recent weeks:

Have you thought to extend the SAAS model family (which is really great) to mixed models like the MixedSingleTaskGP? This could be really powerful. What do you think?

Best,

Johannes

saitcakmak:
> Have you thought to extend the SAAS model family (which is really great) to mixed models like the MixedSingleTaskGP?

We've been using one hot encoding in Ax as far as I've known, but I want to change that. One option is to rely on mixed models like `MixedSingleTaskGP` for this, but that requires adding a mixed variant of each model we're interested in using, which I am not a fan of. In addition, having separate model classes for different search spaces just adds to the complexity of dispatching to the right model.
So, I do want to better support categorical variables in models, but I want to do this without adding more model classes.
- One option could be to give each model a `categorical_dims` input and have them call some helper that adds a categorical kernel to the base kernel of the model. There is some research / benchmarking to do to determine the best way to combine a base kernel with a categorical kernel though.
- Another option I've thought of is to have one base model that we construct using components built in Ax. We'd look at the properties of the search space and other problem structure (for multi-task etc) and construct the appropriate kernel and use that to construct the model. This idea works if we always want to use the auto generated kernels but it can get complicated with giving people options to customize things.

Let me know if you have any other ideas for this! We generally discourage our users from using categorical variables, so we don't see a lot of high cardinality categoricals. This has been on my wish list for a while though, so it'd be good to address this soon.

PR #3048, opened 2025-10-17
Tutorial addition: Feasibility-driven trust Region Bayesian Optimization

paoloascia:
## Motivation

Hi! This tutorial explains how to implement the Feasibility-driven trust Region Bayesian Optimization (FuRBO) method presented last September at AutoML2025. We believe it is a relevant contribution, as it builds upon Scalable Constrained BO to enhance the treatment of constraints and accelerate convergence to a feasible region.

Yes

## Test Plan

We did not change the previous code. The tutorial has been tested prior to upload.

## Related PRs

No documentation needs to be added.


hvarfner:
Thanks @paoloascia for the contribution! Overall I think this looks good and about ready to be merged. I'll get back later today with a few suggestions for improvements.

paoloascia:
Hi @hvarfner ! Thank you for the comment. I'm looking forward to your feedback :).

notebooks_community/FuRBO/FuRBO.ipynb
```diff
@@ -0,0 +1,1011 @@
...
+    "from botorch.fit import fit_gpytorch_mll\n",
+    "from botorch.models import SingleTaskGP\n",
+    "from botorch.models.transforms.outcome import Standardize\n",
+    "\n",
+    "from gpytorch.constraints import Interval\n",
+    "from gpytorch.kernels import MaternKernel, ScaleKernel\n",
+    "from gpytorch.likelihoods import GaussianLikelihood\n",
+    "from gpytorch.mlls import ExactMarginalLogLikelihood\n",
+    "\n",
+    "from scipy.stats import invgauss\n",
```
hvarfner: Unused imports

notebooks_community/FuRBO/FuRBO.ipynb
```diff
@@ -0,0 +1,1011 @@
...
+    "from scipy.stats import ecdf\n",
```
hvarfner: Unused imports

notebooks_community/FuRBO/FuRBO.ipynb
```diff
@@ -0,0 +1,1011 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 172,
+   "id": "1e0f4e8d-657f-49d8-bb59-eb8c2ff4a9f5",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "class norm_():\n",
```
hvarfner: It would be good to see the utilities we have in botorch used here. The class Normalize(d=dim, bounds=bounds) should do the trick to normalize the inputs to [0, 1] for the models. Similarly, creating a different class `ack` that wraps Ackley can be avoided by using the same transforms.

notebooks_community/FuRBO/FuRBO.ipynb
```diff
@@ -0,0 +1,1011 @@
...
+    "    with gpytorch.settings.max_cholesky_size(max_cholesky_size):\n",
+    "        fit_gpytorch_mll(mll, \n",
+    "                         optimizer_kwargs={'method': 'L-BFGS-B'})\n",
+    "\n",
+    "    return model\n",
+    "\n",
+    "from botorch.models.model_list_gp_regression import ModelListGP\n",
+    "from torch.quasirandom import SobolEngine\n",
+    "\n",
+    "class Furbo_state():\n",
```
hvarfner: Please use pythonic class naming (i.e. `FurboState`). Ideally, use type hints too, so that users understan what is going on.

notebooks_community/FuRBO/FuRBO.ipynb
```diff
@@ -0,0 +1,1011 @@
...
+    "    '''Class to track optimization status with restart'''\n",
+    "    # Initialization of the status\n",
+    "    def __init__(self,              \n",
```
hvarfner: Thank for commenting what each variable does. However, the conventional way to do this would be via a docstring. This is also cleaner, and should be a quick fix.

notebooks_community/FuRBO/FuRBO.ipynb
```diff
@@ -0,0 +1,1011 @@
...
+    "        self.sobol = SobolEngine(dimension=self.dim, scramble=True, seed=1)\n",
+    "        \n",
+    "    # Update the status\n",
+    "    def update(self,\n",
+    "               X_next,          # Samples X (input values) to update the status\n",
+    "               Y_next,          # Samples Y (objective value) to update the status\n",
+    "               C_next,          # Samples C (constraints values) to update the status\n",
+    "               **tkwargs):\n",
+    "        \n",
+    "        '''Function to update optimization status'''\n",
```
hvarfner: Good that there is a docstring here. Please move the comments up at the function definition here instead.

notebooks_community/FuRBO/FuRBO.ipynb
```diff
@@ -0,0 +1,1011 @@
...
+    "            self.X = X_next\n",
+    "            self.Y = Y_next\n",
+    "            self.C = C_next\n",
+    "        else:\n",
+    "            # Else, concatenate the new batch to the previous samples\n",
+    "            self.X = torch.cat((self.X, X_next), dim=0)\n",
+    "            self.Y = torch.cat((self.Y, Y_next), dim=0)\n",
+    "            self.C = torch.cat((self.C, C_next), dim=0)\n",
+    "\n",
+    "        # update GPR surrogates\n",
```
hvarfner, line 288: Why is there a try/catch here? Does the model fitting fail, and if so, do we have a sense on why that is?

notebooks_community/FuRBO/FuRBO.ipynb
```diff
@@ -0,0 +1,1011 @@
...
+    "    model = SingleTaskGP(\n",
+    "        X,\n",
+    "        Y,\n",
+    "        covar_module=covar_module,\n",
+    "        likelihood=likelihood,\n",
+    "        outcome_transform=Standardize(m=1),\n",
+    "    )\n",
+    "    mll = ExactMarginalLogLikelihood(model.likelihood, model)\n",
+    "\n",
+    "    with gpytorch.settings.max_cholesky_size(max_cholesky_size):\n",
```
hvarfner: Is the `max_cholesky_size` argument ever changed during optimization? If not, please remove it here and elsewhere to clean up the code.

For context, botorch's max_cholesky_size is 4096 by default: https://github.com/meta-pytorch/botorch/blob/07ce3763b3d4336ad6ba9ba02195a53d607fb59b/botorch/__init__.py#L36-L49

If you want to change it, you can change it once at the top of the file and still remove it everywhere else!

notebooks_community/FuRBO/FuRBO.ipynb
```diff
@@ -0,0 +1,1011 @@
...
+    "        fit_gpytorch_mll(mll, \n",
+    "                         optimizer_kwargs={'method': 'L-BFGS-B'})\n",
+    "\n",
+    "    return model\n",
+    "\n",
+    "from botorch.models.model_list_gp_regression import ModelListGP\n",
```
hvarfner: Imports at the top of each cell please

notebooks_community/FuRBO/FuRBO.ipynb
```diff
@@ -0,0 +1,1011 @@
...
+    "        else:\n",
+    "            # Else, concatenate the new batch to the previous samples\n",
+    "            self.X = torch.cat((self.X, X_next), dim=0)\n",
+    "            self.Y = torch.cat((self.Y, Y_next), dim=0)\n",
+    "            self.C = torch.cat((self.C, C_next), dim=0)\n",
+    "\n",
+    "        # update GPR surrogates\n",
+    "        try:\n",
+    "            self.Y_model = get_fitted_model(self.X, self.Y, self.dim, max_cholesky_size = float(\"inf\"))\n",
+    "            self.C_model = ModelListGP(*[get_fitted_model(self.X, C.reshape([C.shape[0],1]), self.dim, max_cholesky_size = float(\"inf\")) for C in self.C.t()])\n",
```
hvarfner: `C.reshape([C.shape[0],1])` --> `C.reshape(-1,1)`

hvarfner: ...and remove `max_cholesky_size` since it shouldn't be needed here.

notebooks_community/FuRBO/FuRBO.ipynb
```diff
@@ -0,0 +1,1011 @@
...
+   "id": "b5831947-283e-4682-aae4-bd19bcce03e0",
+   "metadata": {},
+   "source": [
+    "# Feasibility-driven trust Region Bayesian Optimization (FuRBO)\n",
+    "\n",
+    "- Contributors: paoloascia, elenaraponi\n",
+    "- Last update 17 October 2025\n",
+    "- BoTorch version: 0.12.0\n",
+    "\n",
+    "This tutorial shows how to implement Feasibility-driven trust Region Bayesian Optimization (FuRBO) with restarts in a closed loop [1].\n",
```
hvarfner: In general, we strive to explain the method, how and why it works to users in the tutorials. In this regard, the notebook is currently lacking a bit. I think an overarching goal should be to have a potential user understand when the method should be used, and ideally see the intuition behind it. Here is one example where there are both visualizations and some explanations of the method accompanying the implementation:


https://github.com/meta-pytorch/botorch/blob/main/tutorials/relevance_pursuit_robust_regression/relevance_pursuit_robust_regression.ipynb

This is admittedly an extreme example, but hopefully it demonstrates the point. As an example, what does FurBO accomplish that TurBO/SCBO does not in this particular case?

notebooks_community/FuRBO/FuRBO.ipynb
```diff
@@ -0,0 +1,1011 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 179,
+   "id": "81f9bd26",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "def restart_criterion(state, radius_min):\n",
```
hvarfner:
```
def restart_criterion(state, radius_min):
  return state.radius < radius_min
```

notebooks_community/FuRBO/FuRBO.ipynb
```diff
@@ -0,0 +1,1011 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 178,
+   "id": "f0d2b342-601b-4d07-9fe3-459f28ecead2",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "def GP_restart_criterion(state):\n",
```
hvarfner: This does not seem like a good thing to include to me, especially when we only run for 100 iterations. Are you sure there is not a bug here?

notebooks_community/FuRBO/FuRBO.ipynb
```diff
@@ -0,0 +1,1011 @@
...
+    "obj = ack(dim = 2,\n",
+    "          negate=True,\n",
+    "          **tkwargs)\n",
+    "cons = list([sum_(threshold = 0,\n",
+    "                  lb = obj.lb,\n",
+    "                  ub = obj.ub), \n",
+    "             norm_(threshold = 0.5, \n",
+    "                   lb = obj.lb, \n",
+    "                   ub = obj.ub)])\n",
+    "batch_size = int(1)#3 * obj.dim)\n",
```
hvarfner: Remove commented code

notebooks_community/FuRBO/FuRBO.ipynb
```diff
@@ -0,0 +1,1011 @@
...
+      "98) Best value: -2.16e+00, MND radius: 1.0\n",
+      "99) Best value: -2.16e+00, MND radius: 1.0\n",
+      "100) Best value: -4.88e-01, MND radius: 0.5\n"
+     ]
+    }
+   ],
+   "source": [
+    "import warnings\n",
+    "\n",
+    "warnings.filterwarnings(\"ignore\")\n",
```
hvarfner: What warnings are suppresed?

notebooks_community/FuRBO/FuRBO.ipynb
```diff
@@ -0,0 +1,1011 @@
...
+     "metadata": {},
+     "output_type": "display_data"
+    }
+   ],
+   "source": [
+    "import matplotlib.pyplot as plt\n",
+    "import numpy as np\n",
+    "from matplotlib import rc\n",
+    "\n",
+    "def ensure_tensor(x, tkwargs, dim=0):\n",
```
hvarfner: Isn't `torch.cat(x).to(**tkwargs)` enough here?

notebooks_community/FuRBO/FuRBO.ipynb
```diff
@@ -0,0 +1,1011 @@
...
+    "        state.radius /= 2.0\n",
+    "        state.failure_counter = 0\n",
+    "    \n",
+    "    for ind, x_candidate in enumerate(state.best_X):\n",
+    "        # Generate the samples to evaluathe the feasible area on\n",
+    "        radius = state.radius\n",
+    "        samples = multivariate_circular(x_candidate, radius, n_samples, lb=lb, ub=ub, **tkwargs)\n",
+    "    \n",
+    "        # Evaluate samples on the models of the objective -> yy Tensor\n",
+    "        state.Y_model.eval()\n",
```
hvarfner: `posterior()` sets the model in eval mode, so these can be removed!

notebooks_community/FuRBO/FuRBO.ipynb
```diff
@@ -0,0 +1,1011 @@
...
+    "        with torch.no_grad():\n",
+    "            posterior = state.Y_model.posterior(samples)\n",
+    "            samples_yy = posterior.mean.squeeze()\n",
+    "        \n",
+    "        # Evaluate samples on the models of the constraints -> yy Tensor\n",
+    "        state.C_model.eval()\n",
```
hvarfner: and here

notebooks_community/FuRBO/FuRBO.ipynb
```diff
@@ -0,0 +1,1011 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 174,
+   "id": "3d36cf01-c5be-44ee-96bb-12564225bd7b",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "def multivariate_circular(centre,       # Centre of the multivariate distribution\n",
```
hvarfner: This to me seems like the secret sauce for the method, right? If so, it could benefit from some additional explanation in the markdown cells!

notebooks_community/FuRBO/FuRBO.ipynb
```diff
@@ -0,0 +1,1011 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 177,
+   "id": "189aae72-f033-49db-bcc8-0e791774bd76",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "def stopping_criterion(state, n_iteration):\n",
+    "    '''Function to evaluate if the maximum number of allowed iterations is reached.'''\n",
```
hvarfner, line 638: `return state.it_counter <= n_iteration`

hvarfner (changes requested):
Looks good implementation-wise, but there are some changes that would be good to see.
1. Please make the notebook more informative: The user should get a sense of what the method does from the markdown cells and plots. Now, they have to go through the code. Since the community is pretty well-informed on what TurBO does by now (and probably even SCBO to some extent), try to focus on what distinguishes FurBO from those two. This may entail some extra work, such as adding an informative plot, some math and maybe even a comparison to SCBO.
2. Please use a linter to clean up the code. This includes unused imports, having comments in the right places and using docstrings instead of comments where appropriate.
3. Get rid of arguments that are not actually used. Max_cholesky_size and "method": L-BFGS-B are covered by current defaults, so they shouldn't be needed.
4. Warnings & try/catches: It's odd to me that GP fitting fails - I really don't think it should with this few observations. Please have a look at this.

paoloascia:
Hi @hvarfner!

Sorry for taking so long to get back to your comments. We reworked the entire tutorial to better address your comments. We divided the notebook into two parts: 1. the workflow of FuRBO, 2. two examples to compare with SCBO.

Let us know what you think about it :).
Cheers!

hvarfner:
Hi @paoloascia,

This looks really nice! Love the effort of explaining the method in greater detail. I think that will come in handy for people who want to understand and use the method for themselves.

I'll go through the code one more time (I would guess tomorrow), and then we should be good to merge.

notebooks_community/FuRBO/FuRBO.ipynb
```diff
@@ -0,0 +1,1907 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 17,
+   "id": "5a56a1e4",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "class speedReducer():\n",
```
hvarfner: Please use https://github.com/meta-pytorch/botorch/blob/main/botorch/test_functions/synthetic.py#L1315

notebooks_community/FuRBO/FuRBO.ipynb
```diff
@@ -0,0 +1,1907 @@
...
+    "warnings.filterwarnings(\"ignore\")\n",
+    "\n",
+    "# Defining objective function\n",
+    "fun = Ackley(dim=10, negate=True).to(**tkwargs)\n",
+    "fun.bounds[0, :].fill_(-5)\n",
+    "fun.bounds[1, :].fill_(10)\n",
+    "\n",
+    "def eval_objective(x):\n",
+    "    \"\"\"This is a helper function we use to unnormalize and evalaute a point\"\"\"\n",
+    "    return fun(unnormalize(x - 0.3, fun.bounds))\n"
```
hvarfner: The boundary issue likely stems from this offset term

notebooks_community/FuRBO/FuRBO.ipynb
```diff
@@ -0,0 +1,1907 @@
...
+    "tkwargs = {\"device\": device, \"dtype\": dtype}\n",
+    "\n",
+    "warnings.filterwarnings(\"ignore\")\n",
+    "\n",
+    "# Defining objective function\n",
+    "fun = Ackley(dim=10, negate=True).to(**tkwargs)\n",
+    "fun.bounds[0, :].fill_(-5)\n",
+    "fun.bounds[1, :].fill_(10)\n",
+    "\n",
+    "def eval_objective(x):\n",
```
hvarfner, line 83: You can use `input_transform=Normalize(d=objective.dim, bounds=objective.bounds)` when creating the model. Then, the untransforms and helpers are not needed.

notebooks_community/FuRBO/FuRBO.ipynb
hvarfner: @paoloascia Did you check this?

paoloascia: @hvarfner Yes, I did. In simple cases like these, the GPR does not fail. When the function has big jumps in value between closely positioned samples, the GPR generally fails. I noticed this happening when the optimization converges and places samples close to one another, but the restart trigger is still not fulfilled.

notebooks_community/FuRBO/FuRBO.ipynb
paoloascia: @hvarfner Thank you for the input. However, I'm not sure this approach would work with the current definition of the trust region. To make it compatible, I would guess that the trust region should be scaled by the bounds of the problem. I can try out the changes for both algorithms, but before doing so, I'd like to ask if there are any reasons why Normalize wasn't used in the SCBO tutorial. Given the necessary scaling, I think that using Normalize could increase the overall complexity.

hvarfner:
@paoloascia Nice job, the notebook runs without issue and looks great. Let's ship it!

Issue #3050, opened 2025-10-20
[Bug]: Error raised when comparing input transforms with differently shaped tensors


TobyBoyne:
### What happened?

When comparing two input transforms, if any elements of their `state_dict`s are Tensors that have different (non-broadcastable) shapes, then comparison raises a RuntimeError.

The simplest example is when comparing two `Normalize`s, which act on different number of dimensions, as below.

### Please provide a minimal, reproducible example of the unexpected behavior.

```python
from botorch.models.transforms.input import Normalize
norm1 = Normalize(d=5, indices=[0, 1])
norm2 = Normalize(d=5, indices=[1, 2, 3])
print(norm1.equals(norm2))
# expect "False", get a RuntimeError
```

### Please paste any relevant traceback/logs produced by the example provided.

```shell
File "botorch\botorch\models\transforms\input.py", line 127, in <genexpr>
    torch.allclose(v, other_state_dict[k].to(v))
RuntimeError: The size of tensor a (2) must match the size of tensor b (3) at non-singleton dimension 0
```

### BoTorch Version

Latest commit on main (as of Oct 20th)

### Python Version

3.12

### Operating System

WSL

### (Optional) Describe any potential fixes you've considered to the issue outlined above.

The bug is caused by the following lines:

https://github.com/meta-pytorch/botorch/blob/b0d492d349f506210cd64c809d611180bddcb3a9/botorch/models/transforms/input.py#L126-L129

If the two arguments to `torch.allclose` are different shapes, an error is raised. A simple fix would be to just add a `_allclose` helper function that looks something like below. There may also be a built-in method that I'm not familiar with, in the way that `torch.equal` checks for size first and doesn't throw an error.

```python
def _allclose(a, b):
    if isinstance(a, Tensor) and isinstance(b, Tensor):
        if a.shape != b.shape: return False
        return torch.allclose(a, b)
    return a == b
```

### Pull Request

Yes

### Code of Conduct

- [x] I agree to follow BoTorch's Code of Conduct

Balandat:
That makes a lot of sense. I'm not aware of a built in method for this but your suggestion is straightforward so I think we can just go with that. Would you mind putting up a PR for this?

PR #3051, opened 2025-10-21
Check shape of state dict when comparing input transforms

TobyBoyne:
## Motivation

See #3050 

Y

## Test Plan

I have added an additional test case to `test/models/transforms/test_input.py`. Without the changes in this PR, the new test case would fail with a runtime error, as described in #3050. The test now correctly passes.


Balandat:
Thanks, this looks great

PR #3053, opened 2025-10-22
Adding Kaiming/He initialization for the VBLL mean

brunzema:
Hi everyone :)

I have a minor update to the VBLLs. This update does not change the default behavior of the current implementation.

## Motivation

In the VBLL repo, we observed improved performance on regression tasks, when applying Kaiming/He initialization to the VBLL mean (https://github.com/VectorInstitute/vbll/commit/fdc6ad0a9b41df3af15aa56d1aff4b7d4ea5fe93).

This has not yet been tested extensively in BO tasks, but we still wanted to migrate this option here to the community branch. The default initialization remains unchanged but one can choose to switch to the new init which is now also in the VBLL repo.

Yes

## Test Plan

Added tests to maintain coverage.

## Related PRs

- Original VBLL PR: #2754 
- Commit with Kaiming/He initialization in VBLL repo: https://github.com/VectorInstitute/vbll/commit/fdc6ad0a9b41df3af15aa56d1aff4b7d4ea5fe93

### DIFF
--- a/botorch_community/models/vbll_helper.py
+++ b/botorch_community/models/vbll_helper.py
@@ -381,10 +381,18 @@ class Regression(nn.Module):
         parameterization : str
             Parameterization of covariance matrix.
             Currently supports {'dense', 'diagonal', 'lowrank', 'dense_precision'}
+        mean_initialization : str or None
+            Initialization method for the mean of the weights.
+            Supports {'kaiming', None}. If None, weights are initialized from
+            a standard normal distribution. Defaults to None.
         prior_scale : float
             Scale of prior covariance matrix
         wishart_scale : float
             Scale of Wishart prior on noise covariance
+        cov_rank : int or None
+            For 'lowrank' parameterization, the rank of the covariance matrix.
+        clamp_noise_init : bool
+            Whether to clamp the noise initialization to be positive.
         dof : float
             Degrees of freedom of Wishart prior on noise covariance
         """
@@ -415,7 +423,7 @@ class Regression(nn.Module):
 
         if mean_initialization is None:
             self.W_mean = nn.Parameter(
-                torch.zeros(out_features, in_features, dtype=self.dtype)
+                torch.randn(out_features, in_features, dtype=self.dtype)
             )
         elif mean_initialization == "kaiming":
             self.W_mean = nn.Parameter(
--- a/test_community/models/test_vblls.py
+++ b/test_community/models/test_vblls.py
@@ -5,7 +5,9 @@
 # LICENSE file in the root directory of this source tree.
 
 import copy
+from unittest.mock import patch
 
+import numpy as np
 import torch
 
 from botorch.utils.testing import BotorchTestCase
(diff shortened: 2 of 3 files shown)


botorch_community/models/vbll_helper.py
```diff
@@ -412,9 +412,26 @@ def __init__(
...
         # last layer distribution
         self.W_dist = get_parameterization(parameterization)
-        self.W_mean = nn.Parameter(
-            torch.randn(out_features, in_features, dtype=self.dtype)
-        )
+
+        if mean_initialization is None:
+            self.W_mean = nn.Parameter(
+                torch.zeros(out_features, in_features, dtype=self.dtype)
+            )
```
Balandat, line 427: As this changes the default from `torch.randn` to `torch.zeros` this does change the default behavior - is this intended?

botorch_community/models/vbll_helper.py
```diff
@@ -361,6 +360,7 @@ def __init__(
         out_features,
         regularization_weight,
         parameterization="dense",
+        mean_initialization=None,
```
Balandat, line 363: Let's add that option to the docstring as well?

test_community/models/test_vblls.py
```diff
@@ -82,6 +83,54 @@ def test_initialization(self) -> None:
...
+        """Test different mean_initialization options."""
+        d, num_hidden, num_outputs, num_layers = 2, 3, 1, 4
+
+        # Test None initialization (default to zeros)
+        model = VBLLModel(
+            in_features=d,
+            hidden_features=num_hidden,
+            num_layers=num_layers,
+            out_features=num_outputs,
+            mean_initialization=None,
```
Balandat, line 98: Maybe also test not passing in `mean_initialization` explicitly here to test the default?

test_community/models/test_vblls.py
```diff
@@ -82,6 +83,54 @@ def test_initialization(self) -> None:
...
+        )
+        print(model.head.W_mean)
+        expected_zeros = torch.zeros(num_outputs, num_hidden, dtype=torch.float64)
```
Balandat:
```suggestion
        expected_zeros = torch.zeros(num_outputs, num_hidden, dtype=torch.float64)
```

test_community/models/test_vblls.py
```diff
@@ -82,6 +83,54 @@ def test_initialization(self) -> None:
...
+        self.assertTrue(torch.allclose(model.head.W_mean, expected_zeros))
+
+        # Test Kaiming initialization
```
Balandat: this is not testing too much; you could do a mock assert called on `np.sqrt` to check that this was actually used

botorch_community/models/vbll_helper.py
brunzema: oh wow, no it is not---thank you for the catch... Was copying it from a different branch and should have realized at latest when doing the tests. I fixed this and I adjusted the test accordingly.

test_community/models/test_vblls.py
brunzema: nice idea, added this in the test

brunzema:
Thank you for the catch! Reverted back to the actual default, added the docstring (+docstrings that were missing), and included the mock assert.

PR #3056, opened 2025-10-23
Use python 3.10 in tutorial tests against stable

Balandat:
Without this, we run into the following error when testing the notebooks against the stable versions:
```
  × No solution found when resolving dependencies:
  ╰─▶ Because torch==2.0.1 has no wheels with a matching Python ABI tag
      (e.g., `cp313`) and you require torch==2.0.1, we can conclude that your
      requirements are unsatisfiable.

      hint: You require CPython 3.13 (`cp313`), but we only found wheels for
      `torch` (v2.0.1) with the following Python ABI tags: `cp38`, `cp39`,
      `cp310`, `cp311`
```

### DIFF
--- a/.github/workflows/reusable_tutorials.yml
+++ b/.github/workflows/reusable_tutorials.yml
@@ -31,10 +31,16 @@ jobs:
     - uses: actions/checkout@v4
     - name: Install uv
       uses: astral-sh/setup-uv@v5
-    - name: Set up Python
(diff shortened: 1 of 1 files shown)


Balandat:
We should probably think about bumping the minimum pytorch requirement at some point - that should help resolve some of this complexity with the installation setup. 2.0.1 is also quite old so this shouldn't cause many compatibility issues for folks.

saitcakmak (approved):

PR #3062, opened 2025-10-24
Add missing int cast

### LINKED ISSUE #3061: [Bug]: Inconsistent random seed type
### What happened?

Now that the package is typed, I noticed the following small type inconsistency in `MCSampler`, where the fallback type is `int | float` due to a missing cast on the return value of the `item()` call. I think adding the cast makes sense as it would then have a type that is compatible with that of the input argument.

```python
from botorch.posteriors import Posterior
from botorch.sampling.base import MCSampler
from torch import Tensor
from typing_extensions import override


class CustomMCSampler(MCSampler):
    """Customer sampler for beta posterior."""

    @override
    def forward(self, posterior: Posterior) -> Tensor:
        reveal_type(self.seed)
```

```shell
Revealed type is "builtins.int | builtins.float"
```

### BoTorch Version

0.16.0

### Python Version

3.11

### Operating System

macOS

A simple `int` call would resolve the problem.

### Pull Request

Yes

### Code of Conduct

AdrianSosic:
Fixes #3061

### DIFF
--- a/botorch/sampling/base.py
+++ b/botorch/sampling/base.py
@@ -60,7 +60,9 @@ class MCSampler(Module, ABC):
                 f"got {sample_shape}."
             )
         self.sample_shape = sample_shape
-        self.seed = seed if seed is not None else torch.randint(0, 1000000, (1,)).item()
+        self.seed = (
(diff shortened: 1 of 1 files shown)


Balandat:
Thanks!

PR #3082, opened 2025-11-18
Classification Models

sunnyshen321:
Summary:
X-link: https://github.com/facebook/Ax/pull/4555

Implement RandomForest, SVC, and XGBoost models as BoTorch models for modeling deterministic failures.

Differential Revision: D87278626


jduerholt:
Nice PR, just two questions: 

- how would you integrate them into ACQF optimization? 
- Regarding the RF, why a generict deterministic model as base? You could you use an Ensemble model and get the uncertainity from the forest. Have a look here: https://github.com/experimental-design/bofire/blob/main/bofire/surrogates/random_forest.py

PR #3106, opened 2025-12-08
Add a note about batching for random restarts

nabenabe0928:
## Motivation

This PR adds [a paper](https://arxiv.org/abs/2511.13625) related to the following PR:
- https://github.com/meta-pytorch/botorch/pull/2870 

(Write your answer here.)

### DIFF
--- a/docs/batching.md
+++ b/docs/batching.md
@@ -151,17 +151,13 @@ starting points. To efficiently optimize an acquisition function for a $q$-batch
 of candidate points using $r$ random restarts, BoTorch uses batched
 evaluation on a $r \times q \times d$ set of candidate points to independently
 evaluate and optimize each random restart in parallel.
-In order to optimize the $r$ acquisition functions using gradient information,
-the acquisition values of the $r$ random restarts are summed before
-back-propagating.
-Note that convergence slows down at each random restart when sharing
-(quasi Newton-based) acquisition function optimizer states, as claimed by
-[^Irie2026], so the optimizer states are decoupled across random restarts.
+By default, optimization states are decoupled across the restarts to avoid
+slowdowns in convergence (see e.g., [^Irie2026]).
 
 [^Irie2026]:
-    I. Couckuyt, D. Deschrijver and T. Dhaene. Towards Efficient Multiobjective
-    Optimization: Multiobjective statistical criterions. Workshop on AI to
-    Accelerate Science and Engineering at AAAI, Singapore, 2026.
+    K. Irie, S. Watanabe and M. Onishi. Batch Acquisition Function Evaluations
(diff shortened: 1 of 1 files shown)


nabenabe0928:
Hi @Balandat , I published a PR as mentioned in the email:)

docs/optimization.md
```diff
@@ -71,7 +71,16 @@ uses
...
+parallelism, while maintaining high quality solutions. Note, however, that
+convergence slows down at each random restart when sharing `scipy` optimizer
+states, as claimed by [^Irie2026], so the optimizer states are decoupled across
+random restarts.
+
+[^Irie2026]:
+    I. Couckuyt, D. Deschrijver and T. Dhaene. Towards Efficient Multiobjective
+    Optimization: Multiobjective statistical criterions. Workshop on AI to
+    Accelerate Science and Engineering at AAAI, Singapore, 2026.
+    [paper](https://arxiv.org/abs/2511.13625)
```
Balandat: Actually, since we decouple the optimizations let's maybe not talk about the issues that don't occur here (since it'll be confusing for the reader). If you like you can reference the paper here to provide detail on the setup (note though that the reference is wrong, this is not the Cockuyt paper!). 
```suggestion
found over the random restarts. 

By default, BoTorch will exploit hardware parallelism by batching the acquisition 
function evaluation across multiple random restarts.
```

docs/batching.md
```diff
@@ -154,6 +154,15 @@ evaluate and optimize each random restart in parallel.
...
 back-propagating.
+Note that convergence slows down at each random restart when sharing
+(quasi Newton-based) acquisition function optimizer states, as claimed by
+[^Irie2026], so the optimizer states are decoupled across random restarts.
+
+[^Irie2026]:
+    I. Couckuyt, D. Deschrijver and T. Dhaene. Towards Efficient Multiobjective
+    Optimization: Multiobjective statistical criterions. Workshop on AI to
+    Accelerate Science and Engineering at AAAI, Singapore, 2026.
+    [paper](https://arxiv.org/abs/2511.13625)
```
Balandat: This seems to be the wrong reference here as well? 

```suggestion
By default, optimization states are decoupled across the restarts to avoid 
slowdowns in convergence (see e.g. [^Irie2026]). 
```

docs/optimization.md
nabenabe0928: Thank you for the suggestion, I addressed your comment!

docs/batching.md
nabenabe0928: Thank you for the catch, I applied your suggestion!

Balandat (approved):

PR #3109, opened 2025-12-10
Clamp to 0 to avoid sqrt of negative numbers in mvn_hellinger_distance

Jack-Sandberg:
## Motivation

When trying out ScoreBO, I noticed that it tended to explore excessively. I found that the acquisition function occasionally contained NaNs that originated from the mvn_hellinger_distance function. When computing 1 - x.exp() for small x, the result is sometimes negative due to floating point inaccuriacies which causes the sqrt to output NaN.

Yes.

## Test Plan

The change is rather minor but I've added a test to `test_community/utils/test_stat_dist.py`.

### DIFF
--- a/test_community/utils/test_stat_dist.py
+++ b/test_community/utils/test_stat_dist.py
@@ -79,13 +79,11 @@ class TestStatDist(BotorchTestCase):
 
     def test_mvn_hellinger_distance_numerical_stability(self):
         # Two almost equal distributions. Distance is approximatly 0.
-        dist1_mean = torch.tensor([[-0.4615126826879162]], dtype = torch.float64)
-        dist2_mean = torch.tensor([[-0.46151268268791173]], dtype = torch.float64)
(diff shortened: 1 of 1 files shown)


hvarfner:
@Jack-Sandberg Pease make sure to go through the [Contributor License Agreement](https://code.facebook.com/cla) so that the PR can be merged!

Jack-Sandberg:
Hi @hvarfner, 

The CLA should have been signed now. 

I was adapting ScoreBO to my own setting where I use discrete arms and discrete lengthscale values. I did not use standardize input or standardize output because my data is from a GP that matches the prior assumptions and should be reasonably well-behaved. The GP has a variance of 1 and the noise has a variance of 0.25^2.

Currently, the NaNs show up in the second iteration at the location of the first data point. With discrete arms, the acquisition function must be evaluated at the _exact_ same location as the data has been collected (and potentially where the sampled optimas $x^*$ lie) which might explain why the parameters in the test case are almost identical. I have not investigated the behavior in continuous settings.

PR #3123, opened 2026-01-06
Tutorial update workflow

saitcakmak:
Modifies the existing run tutorials script to have it save outputs on demand. A new workflow is added, which uses the modified script with save output options to run all tutorials on a monthly cadence and creates a PR to update their outputs.

Also includes a tutorial update which was generated using the updated script. The workflow itself is not yet tested (it does not show up in the actions tab until the PR is committed).

### DIFF
--- a/.github/workflows/update_tutorials.yml
+++ b/.github/workflows/update_tutorials.yml
@@ -51,11 +51,11 @@ jobs:
 
     - name: Run tutorials and update outputs
       run: |
+        name_option=""
         if [ -n "${{ inputs.tutorial_name }}" ]; then
-          python scripts/run_tutorials.py -p "$(pwd)" -n "${{ inputs.tutorial_name }}" --save-outputs
-        else
-          python scripts/run_tutorials.py -p "$(pwd)" --save-outputs
+          name_option="-n ${{ inputs.tutorial_name }}"
         fi
+        python scripts/run_tutorials.py -p "$(pwd)" $name_option --save-outputs
 
     - name: Check for changes
       id: check_changes
--- a/scripts/run_tutorials.py
+++ b/scripts/run_tutorials.py
@@ -49,6 +49,11 @@ def run_tutorial(
     Runs the tutorial in a subprocess, catches any raised errors and returns
     them as a string, and returns runtime and memory information as a dict.
     """
+    if smoke_test and save_outputs:
+        raise ValueError(
+            "Cannot use both smoke_test and save_outputs. "
+            "Saving outputs from smoke tests is not supported."
+        )
     timeout_minutes = 5 if smoke_test else 30
     tic = time.monotonic()
     print(f"Running tutorial {tutorial.name}.")


.github/workflows/update_tutorials.yml
```diff
@@ -0,0 +1,102 @@
...
+        python-version: "3.13"
+
+    - name: Fetch all history for all tags and branches
+      run: git fetch --prune --unshallow || true
+
+    - name: Install latest PyTorch & GPyTorch
+      run: |
+        uv pip install --pre torch torchvision --index-url https://download.pytorch.org/whl/nightly/cpu
+        uv pip install git+https://github.com/cornellius-gp/linear_operator.git
+        uv pip install git+https://github.com/cornellius-gp/gpytorch.git
```
Balandat, line 46: Should this be using the latest, or should this use the latest stable?

.github/workflows/update_tutorials.yml
```diff
@@ -0,0 +1,102 @@
...
+      run: |
+        uv pip install .[tutorials]
+
+    - name: Run tutorials and update outputs
+      run: |
+        if [ -n "${{ inputs.tutorial_name }}" ]; then
+          python scripts/run_tutorials.py -p "$(pwd)" -n "${{ inputs.tutorial_name }}" --save-outputs
+        else
+          python scripts/run_tutorials.py -p "$(pwd)" --save-outputs
+        fi
```
Balandat, line 57: something like this maybe?

```suggestion
        name_option = ""
        if [ -n "${{ inputs.tutorial_name }}" ]; then
          name_option = "-n ${{ inputs.tutorial_name }}" 
        fi
        python scripts/run_tutorials.py -p "$(pwd)" $(name_option) --save-outputs
```

.github/workflows/update_tutorials.yml
```diff
@@ -0,0 +1,102 @@
...
+        BRANCH_NAME="update-tutorial-outputs"
+        git checkout -b "$BRANCH_NAME"
+        git add tutorials/
+        git commit -m "Update tutorial outputs"
+
+        # Push branch (force push to update if branch exists)
+        git push -f origin "$BRANCH_NAME"
+
+        # Check if PR already exists
+        EXISTING_PR=$(gh pr list --head "$BRANCH_NAME" --json number --jq '.[0].number')
```
Balandat, line 89: Do we have access to the `gh` command line utility here?

scripts/run_tutorials.py
```diff
@@ -37,7 +43,7 @@ def run_script(
 
 
 def run_tutorial(
-    tutorial: Path, smoke_test: bool = False
+    tutorial: Path, smoke_test: bool = False, save_outputs: bool = False
```
Balandat, line 46: Should we error out here if both `smoke_test` and `save_outputs` are true?

.github/workflows/update_tutorials.yml
saitcakmak: We sometimes add tutorials that rely on latest linear operator / gpytorch, so using latest seems like a safer option. Latest stable would be closer to what a user would get if they were to run the tutorial themselves though.

.github/workflows/update_tutorials.yml
saitcakmak: Sure, there is an if block and same number of lines in either case.

.github/workflows/update_tutorials.yml
saitcakmak: yes, it is pre-installed https://docs.github.com/en/actions/how-tos/write-workflows/choose-what-workflows-do/use-github-cli

scripts/run_tutorials.py
saitcakmak: it could be useful for testing the save outputs functionality, but I doubt it would get much use. We can error out

.github/workflows/update_tutorials.yml
Balandat: make sense; output should hopefully not be too different for the ones not relying on the newest changes

Balandat (approved):

PR #3124, opened 2026-01-07
KroneckerMultiTaskGP doesn't clear caches correctly

slishak-PX:
## Motivation

When training a `KroneckerMultiTaskGP`, adding a validation callback to `fit_gpytorch_mll` (calling `model.eval(); model.posterior(test_x); model.train()`) results in nonsensical results due to the caches not being correctly cleared. In particular, the analytical mean is incorrect.

It would probably be even safer to use `gpytorch.utils.memoize.clear_cache_hook`, as I assume the root cause of this issue is that `data_data_roots` and `task_root` were removed and `predictive_mean_cache` was added, but this method was never updated.

Yes.

## Test Plan

No unit test added yet; maybe could check that the cache is empty after calling `model.eval()`?

```python
import torch
from botorch.models.multitask import KroneckerMultiTaskGP
from botorch.models.transforms import Normalize, Standardize
from plotly.subplots import make_subplots

n_inputs = 1
n_outputs = 2
n_train = 128
n_test = 1
device = torch.device("cuda")

torch.manual_seed(42)
train_x = torch.rand(n_train, n_inputs, dtype=torch.float64, device=device)
train_y = torch.hstack([train_x, -train_x])
test_x = torch.ones(n_test, 1, n_inputs, dtype=torch.float64, device=device)

model = KroneckerMultiTaskGP(
    train_x, 
    train_y, 
    input_transform=Normalize(n_inputs),
    outcome_transform=Standardize(n_outputs),
)

def callback(*args, **kwargs):
    with torch.no_grad():
        model.eval()
        _ = model.posterior(test_x).mean.detach()
        model.train()

fit_gpytorch_mll(
    mll, 
    options={"maxiter": 5},
    optimizer_kwargs={"callback": callback},
)

posterior_q = model.posterior(test_x)
mean = posterior_q.mean.detach().cpu().numpy()
samples = posterior_q.sample(torch.Size([256])).detach().cpu().numpy()

fig = make_subplots()
fig.add_histogram(
    x=samples[:, 0, 0, 0],
    name="Samples",
    showlegend=True,
)
fig.add_vline(
    x=mean[0, 0, 0], 
    line_color="red", 
    showlegend=True, 
    name="Analytic mean",
)
```

Before:
image

After:
image

## Related PRs

### DIFF
--- a/botorch/models/multitask.py
+++ b/botorch/models/multitask.py
@@ -883,7 +883,7 @@ class KroneckerMultiTaskGP(ExactGP, GPyTorchModel, FantasizeMixin):
 
     def train(self, val=True, *args, **kwargs):
         if val:
-            fixed_cache_names = ["data_data_roots", "train_full_covar", "task_root"]
+            fixed_cache_names = ["predictive_mean_cache", "train_full_covar"]
             for name in fixed_cache_names:
                 try:
                     pop_from_cache(self, name)


saitcakmak:
I don't have context on the changes related to removal of `data_data_roots` and `task_root`, but the change here seems reasonable to me. I also agree that it'd be better to use `gpytorch.utils.memoize.clear_cache_hook` to future proof the behavior here. I am curious why GPyTorch itself doesn't call `clear_cache_hook` in `ExactGP` or `Module` `train` method. `ExactGP` only clears its `prediction_strategy`. I guess there can be train time caches that a custom model may want to retain and they don't want to mess with those. cc @Balandat

PR #3126, opened 2026-01-07
Revert D90273915

saitcakmak:
Summary:
This diff reverts D90273915
The original diff that made the linter changes has been reverted. We will need to re-do this after they re-upgrade the linter.

Depends on D90273915

Reviewed By: saitcakmak

Differential Revision: D90283320


saitcakmak:
committed

PR #3157, opened 2026-01-28
Add previous winner support to qEUBO

mytkymt:
## Motivation
I want to use EUBO with a previous_winner while also applying a posterior_transform (e.g., a weighted sum). 
There were two options: 
- add posterior_transform to AnalyticEUBO
- add previous_winner to qEUBO. 

AnalyticEUBO doesn’t naturally support posterior_transform because it assumes a single‑output preference model, while qEUBO already accepts posterior_transform and can follow the same previous_winner handling as in AnalyticEUBO. So I added previous_winner support in qExpectedUtilityOfBestOption to enable that workflow.

Yes. 

## Test Plan
I have conducted tests here:
test_preference.py
test_input_constructors.py

## Related PRs
N/A

### DIFF
--- a/botorch/acquisition/input_constructors.py
+++ b/botorch/acquisition/input_constructors.py
@@ -1677,6 +1677,7 @@ def construct_inputs_qeubo(
     sampler: MCSampler | None = None,
     objective: MCAcquisitionObjective | None = None,
     posterior_transform: PosteriorTransform | None = None,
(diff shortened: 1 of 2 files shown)


saitcakmak (changes requested):
Hi @mytkymt. Sorry for the late review here. I do not see any tests for the changes in the PR. Can you add some unit tests that use `previous_winner` argument?

PR #3164, opened 2026-02-05
Links to Active Learning Papers

jduerholt:
## Motivation

I was looking up papers referenced in the active learning acqfs and thought it could be useful to have the links directly in, then the next does not need to search for them again. 

Yes.

## Test Plan

Just docs, I hope they build and look correct as I am not 100% sure how they are displayed in the API references if put in as I did ... let's see ;)

### DIFF
--- a/botorch/acquisition/active_learning.py
+++ b/botorch/acquisition/active_learning.py
@@ -8,20 +8,20 @@ r"""
 Active learning acquisition functions.
 
 .. [Seo2014activedata]
-    S. Seo, M. Wallat, T. Graepel, and K. Obermayer. [Gaussian process regression:
-    Active data selection and test point rejection]
-    (https://ieeexplore.ieee.org/document/861310).
+    S. Seo, M. Wallat, T. Graepel, and K. Obermayer. `Gaussian process regression:
+    Active data selection and test point rejection.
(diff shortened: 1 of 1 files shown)


Balandat:
It's pretty easy to check locally how things will render, see the instructions [here](https://github.com/meta-pytorch/botorch/blob/main/CONTRIBUTING.md#documentation).

jduerholt:
> It's pretty easy to check locally how things will render, see the instructions [here](https://github.com/meta-pytorch/botorch/blob/main/CONTRIBUTING.md#documentation).

Thanks for the hint, just one comment here: If you build the docs locally with the command from your link, it opens up a locally served website, but in this locally served website, it is still pointed to the readthedocs global website for the API reference. I only had a quick look but could not figure out where this is hardcoded. So I build the api references just directly via sphinx (`make html`). It is rendering now correctly, so from my side this is good to go. 

image

Balandat:
Thanks - yes good point I guess I hadn't really tested this with the new website linking to readthedocs (we got rid of our own custom setup b/c it was too brittle). 

@CristianLara any best practices for how to do this kind of doc testing with readthedocs? Should we update the CONTRIBUTING.md file accordingly?

Issue #3158, opened 2026-01-29
[FEATURE REQUEST]: Allow custom kernels in `SingleTaskMultiFidelityGP`


TobyBoyne:
### Motivation

I have a BO problem that is multifidelity, and requires the use of a custom kernel. I therefore cannot easily use a [`SingleTaskMultiFidelityGP`](https://github.com/meta-pytorch/botorch/blob/b77e81a4959be191f01817f682603effa6357105/botorch/models/gp_regression_fidelity.py#L51), since it only supports using the default `get_covar_module_with_dim_scaled_prior` for the non-fidelity dims.

### Describe the solution you'd like to see implemented in BoTorch.

I would like to add a `cont_kernel_factory` argument to `SingleTaskMultiFidelityGP`, similar to the same argument in `SingleTaskMixedGP`.

As an aside, I also wonder if we should remove the `ScaleKernel`, since this has also been removed across most other  models? Although maybe it's required since the `DownsamplingKernel` is not bounded to [0, 1] compared to, for example, the `RBFKernel`.

### Describe any alternatives you've considered to the above solution.

One could simply construct the kernel first, and pass it as the `covar_module` argument into `SingleTaskGP`. This works, but would require duplicating a lot of the logic in  `_setup_multifidelity_covar_module`, as well as missing the setup logic in `SingleTaskMultiFidelityGP.__init__`

### Pull Request

Yes

### Code of Conduct

- [x] I agree to follow BoTorch's Code of Conduct

Balandat:
Hi @TobyBoyne, this makes a lot of sense to me; we'd be more than happy to accept a PR for this. 

Regarding the question about the `ScaleKernel` - should be fine to remove this since in either case we have a product kernel with either an RBF or Matern kernel so having a `ScaleKernel` would overparameterize the scale (assuming of course that `cont_kernel_factory` will generate a similar such kernel with a scaling parameter). cc @hvarfner who may have thoughts on this.

hvarfner:
No concerns here, sounds good to me!

TobyBoyne:
Hi @Balandat, thanks for the reply! I've created a PR (#3167), and left a couple of small open questions.

I just wanted to clarify your comment - I was under the impression that `cont_kernel_factory` will **not** have a scaling parameter, as discussed in #2451? I have also left in the `ScaleKernel` for the linear truncated kernel, since it is the sum of kernels and likely needs to be rescaled to fit the standardized data.

Thanks!

[...]

PR #3167, opened 2026-02-09
Add cont_kernel_factory argument to SingleTaskMultifidelityGP

TobyBoyne:
## Motivation

See #3158

Yes

## Test Plan

Written tests in `test_gp_regression_fidelity` (specifically `test_cont_kernel_factory`) to verify behaviour

## Related PRs

N/A

### DIFF
--- a/botorch/models/gp_regression_fidelity.py
+++ b/botorch/models/gp_regression_fidelity.py
@@ -197,7 +197,7 @@ def _setup_multifidelity_covar_module(
             of the default kernel.
         nu: The smoothness parameter for the Matern kernel: either 1/2, 3/2, or
             5/2. Only used when ``linear_truncated=True``.
-        covar_module: The module for computing the covariance matrix between
+        data_covar_module: The module for computing the covariance matrix between
                 the non-fidelity features. Defaults to ``RBFKernel``.
     Returns:
         The covariance module and subset_batch_dict.


botorch/models/gp_regression_fidelity.py
```diff
@@ -255,11 +278,15 @@ def _setup_multifidelity_covar_module(
...
-        kernel, batch_shape=aug_batch_shape, outputscale_prior=GammaPrior(2.0, 0.15)
-    )
+    if linear_truncated:
+        covar_module = ScaleKernel(
+            kernel, batch_shape=aug_batch_shape, outputscale_prior=GammaPrior(2.0, 0.15)
+        )
+        key_prefix = "covar_module.base_kernel.kernels"
+    else:
+        covar_module = kernel
+        key_prefix = "covar_module.kernels"
```
TobyBoyne, line 284: Since the linear truncated model is the sum of many kernels, I think we only need a scale kernel for the case that `linear_truncated==True`? We don't need a scale kernel for the full model (where there will be a fixed outputscale of 1, as per https://github.com/meta-pytorch/botorch/discussions/2451).

botorch/models/gp_regression_fidelity.py
```diff
@@ -205,6 +219,15 @@ def _setup_multifidelity_covar_module(
 
     kernels = []
 
+    if linear_truncated and cont_kernel_factory is not None:
+        raise ValueError(
+            "Non-fidelity kernel factory cannot be specified when using a linear "
+            "truncated kernel."
+        )
```
TobyBoyne: Is this necessary? There is no similar error in the case, for example, `not linear_truncated and nu is not None`.

botorch/models/gp_regression_fidelity.py
```diff
@@ -72,6 +72,8 @@ def __init__(
         data_fidelities: Sequence[int] | None = None,
         linear_truncated: bool = True,
```
TobyBoyne, line 74: Should `linear_truncated=True` be the default here? The docstring for the class makes it sounds like `Downsampling/ExponentialDecayKernel` are the default kernels, but actually they aren't used by default here. 

Specifically, the `__init__` docstring reads: `linear_truncated: If True, use a ``LinearTruncatedFidelityKernel`` instead of the default kernel.`

hvarfner:
Hi @TobyBoyne ,

Thanks for the PR! I'm familiarizing myself with the code on my end, so I'll have answers for you in a bit!

botorch/models/gp_regression_fidelity.py
hvarfner: Good question. Let's keep the default for now and possibly change that later.

botorch/models/gp_regression_fidelity.py
hvarfner: Yes, this is perfect. We should clean up the `linear_truncated` logic regardless, but I think it's a good idea to keep the `cont_kernel_factory` self-contained.

botorch/models/gp_regression_fidelity.py
hvarfner: I think this snippet is enough, since cont_kernel_factory would be a non-default argument. `nu` has been silently ignored previously when `linear_truncated` was false, and this would be in line with that behavior IIUC.

hvarfner:
@TobyBoyne Overall, this looks great. I've added one inline change. Otherwise, two changes:

1. Can you please add a few rows to test the newly introduced `subset_batch_dict` changes
2. The tests contain `assert`. Please change these to `self.assert*Something*`.

botorch/models/gp_regression_fidelity.py
hvarfner, line 152:
```suggestion
self._subset_batch_dict = {
            "mean_module.raw_constant": -1,
            **subset_batch_dict,
        }
if linear_truncated:
    self._subset_batch_dict["covar_module.raw_outputscale"] = -1
```

hvarfner (changes requested):

test/models/test_gp_regression_fidelity.py
```diff
@@ -427,6 +432,54 @@ def test_construct_inputs(self):
...
+            (None, custom_kernel_factory), expected_kernels
+        ):
+            model = SingleTaskMultiFidelityGP(
+                **model_kwargs, cont_kernel_factory=cont_kernel_factory
+            )
+            non_fid_kernel = model.covar_module.kernels[0]
+            assert isinstance(non_fid_kernel, expected_kernel)
+            assert non_fid_kernel.ard_num_dims == d
+            assert non_fid_kernel.batch_shape == aug_batch_shape
+            assert non_fid_kernel.active_dims == torch.as_tensor(list(range(d)))
```
hvarfner: This test fails on cuda:

`line 475, in test_cont_kernel_factory
    assert non_fid_kernel.active_dims == torch.as_tensor(list(range(d)))
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
RuntimeError: Expected all tensors to be on the same device, but found at least two devices, cuda:0 and cpu!`


`torch.as_tensor(list(range(d)), device=self.device)` should do the trick.

TobyBoyne:
Thanks for the review, @hvarfner! This should be everything :)

hvarfner:
@TobyBoyne thanks once again for the PR! Sorry for the inconvenience, but would you be okay with designing this like https://github.com/meta-pytorch/botorch/blob/main/botorch/models/multitask.py instead, i.e. passing a covar_module and creating the `ProductKernel` with the `ExponentialDecayKernel` thereafter. This would help with serialization and allow for greater customizability for the user. The `linear_truncated` option can be dealt with in exactly the same way as before.

@saitcakmak

TobyBoyne:
@hvarfner that makes sense, I hope this lines up with what you were thinking! Happy to make more changes if needed.

botorch/models/gp_regression_fidelity.py
```diff
@@ -205,6 +213,12 @@ def _setup_multifidelity_covar_module(
 
     kernels = []
 
+    if linear_truncated and data_covar_module is not None:
+        raise ValueError(
+            "Non-fidelity kernel factory cannot be specified when using a linear "
+            "truncated kernel."
```
saitcakmak:
```suggestion
            "Non-fidelity covariance module cannot be specified when using a linear "
            "truncated kernel."
```

botorch/models/gp_regression_fidelity.py
```diff
@@ -190,7 +197,8 @@ def _setup_multifidelity_covar_module(
             of the default kernel.
         nu: The smoothness parameter for the Matern kernel: either 1/2, 3/2, or
             5/2. Only used when ``linear_truncated=True``.
-
+        covar_module: The module for computing the covariance matrix between
```
saitcakmak:
```suggestion
        data_covar_module: The module for computing the covariance matrix between
```

saitcakmak (approved):
Thanks @TobyBoyne! This will be much easier to integrate with Ax in the future.

Curious if you have tried the linear truncated version here. From an API perspective, I am wondering if we should just clean up that option (not in this PR) since we now have an explicit way for the user to pass in a custom data kernel. The origins of this model are rather old research projects, so we don't have any recent experiments comparing various options it supports.

TobyBoyne:
Hi @saitcakmak, no I haven't tried the linear truncated kernel yet, since my use case only has one fidelity dimension, and my understanding is that the linear truncated kernel is useful to reduce the number of interaction terms between fidelities. 

Some more reasons that I'm not using the linear truncated kernel include not being able to find any references for this kernel, and the default priors seeming outdated to me - e.g. they use a Matern kernel with Gamma(3,6) on the lengthscale, whereas the BoTorch default today is the dim-scaling RBF kernel. There's also an open discussion on whether this kernel is even a valid kernel, #2706.

> I am wondering if we should just clean up that option (not in this PR) since we now have an explicit way for the user to pass in a custom data kernel

I don't think the API introduced in this PR is sufficient to define a LinearTruncatedFidelityKernel, since you can't choose the fidelity kernel and they are slightly different. As a user, it's not clear to me whether $c + (1-f_1)^p \cdot (1 - f_2)^p$ is a better/worse fidelity kernel than $(1-f_1)\cdot(1-f_2)\cdot(1+f_1 \cdot f_2)^p$, so it would be nice to not have to make the choice!

Issue #3168, opened 2026-02-10
[FEATURE REQUEST]: add `return_acq_val` parameter to `optimize_acqf`


atinary-lbrey:
### Motivation

Dear BoTorch community, 

I realized that when optimizing the acquisition function with any `optimize_acqf` (continuous, discrete, local search), a final evaluation of the acqf is performed in order to return the acq values. 

From some memory profiling I have been running recently, it seems like this call is pretty memory intensive especially for the `optimize_discrete_local_search` since performed jointly (see [here](https://github.com/meta-pytorch/botorch/blob/747107733cf721d28356116281b826bf68337446/botorch/optim/optimize.py#L1672C5-L1677C33)). In my specific case, such call is useless as I am only interested in returning the optimized candidates and I am not using the acq_values. 

In order to avoid this unnecessary computation, I would like to make this computation optional.

### Describe the solution you'd like to see implemented in BoTorch.

The easiest way to do it in my opinion would be to add a flag `return_acq_value` for example to the `optimize_acqf`s args. If true, then the joint evaluation is performed. Otherwise it returns only the optimized candidates. 
For the implementation, I could imagine maybe the acqf returning a tuple(candidates, None) when the bool is False and tuple(candidates, acq_values) otherwise. Not sure if it is the cleanest way to do it tho. 

### Describe any alternatives you've considered to the above solution.

For now, since in my code I don't need the acq_values. I just copied the methods in my code and removed the computation of the acq_values at the end of the optimization. My function is thus returning only the candidates. 

### Pull Request

Yes

### Code of Conduct

- [x] I agree to follow BoTorch's Code of Conduct

hvarfner:
Hi @atinary-lbrey ,

I think this is sensible solution, and the optional `return (candidates, None)` sounds minimally invasive as far as I can tell. We strive to have all `optimize_acqf` variants share an interface as much as possible, so it would be good if all of them had this option. 

With that said, it doesn't quite make sense to me that the last the forward pass you linked would be any more memory intensive than the acquisition function optimization of the `q`-th candidate. In fact, it should be `1 / batch_size` times as memory intensive, IIUC. What is the setting in which this is occurs?

atinary-lbrey:
Dear Carl, 

Thanks for your quick response. I realized the hyperlink I used is a bit broken.
[...]

PR #3209, opened 2026-03-03
Optionally return acq_values 

atinary-lbrey:
## Motivation

The idea is to optionally returns the acq_values only if needed, else returns None. Also computes them only when necessary to avoid useless forward pass of the acqf. See #3168 for more background. 

Yes I did.

## Test Plan

I have updated the API for the `optimize_acqf` and the different flavors of it. I added some test to try out if setting the flag `return_acq_values=False` does indeed return `None` and that by default it returns the `acq_values`.

## Related PRs

No related PRs, only this issue #3168


Balandat:
Thanks for the PR, @atinary-lbrey. Would you mind updating the unit tests to bump up coverage for the new functionality to make sure this is tested across all functions? Thanks!

PR #3279, opened 2026-04-20
Fix DeprecationWarning in Standardize.untransform_posterior

AdrianSosic:
## Motivation

Fixes a deprecation warning by wrapping `scale_tril` in a `TriangularLinearOperator` before passing it to `linear_operator.operators.CholLinearOperator`, which deprecated the use of dense tensors.

### Minimal Reproducing Example
```python
import warnings

import torch
from botorch.posteriors.gpytorch import GPyTorchPosterior
from gpytorch.distributions import MultivariateNormal

warnings.simplefilter("error")

mean = torch.tensor([1.0, 2.0])
covar = torch.eye(2) * 0.5
mvn = MultivariateNormal(mean, covar)
posterior = GPyTorchPosterior(mvn)
transform = Standardize(m=1)
transform(torch.tensor([[1.0], [2.0]]))
transform.untransform_posterior(posterior)
```

```python
DeprecationWarning: chol argument to CholLinearOperator should be a TriangularLinearOperator. Passing a dense tensor will cause errors in future versions.
```

## Comments

There might be other places where a similar fix is necessary – I guess the BoTorch team has a better overview 🙃

### DIFF
--- a/botorch/models/transforms/outcome.py
+++ b/botorch/models/transforms/outcome.py
@@ -36,7 +36,11 @@ from botorch.models.transforms.utils import (
 from botorch.models.utils.assorted import get_task_value_remapping
 from botorch.posteriors import GPyTorchPosterior, Posterior, TransformedPosterior
 from botorch.utils.transforms import normalize_indices
-from linear_operator.operators import CholLinearOperator, DiagLinearOperator
+from linear_operator.operators import (
+    CholLinearOperator,
+    DiagLinearOperator,
+    TriangularLinearOperator,
+)
 from torch import Tensor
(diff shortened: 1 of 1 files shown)


esantorella:
Pending the CI passing, this looks good to me. This should be a no-op: The tensor previously passed to `CholLinearOperator` must be lower-triangular. If it's dense, then previously, `CholLinearOperator` would previously convert it to `TriangularLinearOperator(chol, upper=False)`, which is what this PR does (since upper=False) is default. Thanks for putting this in!

saitcakmak (approved):

PR #3287, opened 2026-04-23
docs: force no trailing slash

shimanek:
## Motivation

I noticed on the documentation site that some links appeared to be broken. Looking into it, I think it's due to the trailing slash of addresses being rendered in some cases and not in others.

For example, going to the main site > Docs > Basic Concepts > Acquisition Functions

The address is then 
`https://botorch.org/docs/acquisition`
but when I refresh (on Firefox), the address becomes 
`https://botorch.org/docs/acquisition/`
so that the relative link ("... counterpart qExpectedImprovement can be found in **this tutorial**.") goes to 
`https://botorch.org/docs/acquisition/tutorials/compare_mc_analytic_acquisition`
instead of 
`https://botorch.org/docs/tutorials/compare_mc_analytic_acquisition`

I reproduced this locally by adding a [setting](https://docusaurus.io/docs/api/docusaurus-config#trailingSlash) of
`"trailingSlash": true`

To keep using relative links, I think the easiest fix is to specify 
`"trailingSlash": false`

But just thought I'd point it out.

Yes.

## Test Plan

Reproduced site issue locally and tested fix locally.

## Related PRs

None found.

### DIFF
--- a/website/docusaurus.config.js
+++ b/website/docusaurus.config.js
@@ -42,6 +42,7 @@ module.exports={
     "wrapPagesHTML": true
   },
   "onBrokenLinks": "throw",
+  "trailingSlash": false,
   "presets": [
     [
       "@docusaurus/preset-classic",


esantorella:
Thanks so much for the fix! This has been an issue in the past that we haven't resolved comprehensively (e.g. #2972 ). 

@CristianLara Does this seem like the right fix to you?

Although this sounds like the right fix to me, I'm a bit confused, because botorch.org is hosted by GitHub Pages, and [this](https://docusaurus.io/docs/deployment) Docusaurus page claims that "GitHub Pages adds a trailing slash to Docusaurus URLs by default." But based on what we've seen and Cris's past comments, maybe that is not true 100% of the time? Also, I want to make sure we don't have URLs that will be broken by the addition of a trailing slash. I tried to reproduce this locally on Chrome but couldn't generate the trailing slash issue.

PR #3301, opened 2026-05-04
fix: Numerical tolerance for filtering of X_avoid

atinary-lbrey:
## Motivation
Realized there was some numerical instabilities with hard equality check when filtering X_avoid points from choices in `optimize_discrete*`

Yes

## Test Plan

Added a test case in the TestOptimizeDiscrete by adding noise up to hard tolerance (`1e-8`) and checking that the behavior was the same as in the exact case. 

## Related PRs


saitcakmak:
Hi @atinary-lbrey. Do you have examples of cases where the exact equality checks cause issues? It is a bit surprising that exact equality checks against candidates (presumably) produced by the discrete optimizer would fail. Do you provide `X_avoid` separately from an external source?

Overall, this seems like a reasonable solution at first but I have my concerns. These broadly apply to any discrete optimizer. `isclose` uses `atol=1e-8, rtol=1e-5` by default. Both of these are problematic in two extremes, which are both unlikely to happen. 
- `atol=1e-8`: If the parameter magnitudes are extremely small, they may get equal by default. Such small values are likely to cause numerical issues elsewhere as well, so they're not recommended.
- `rtol=1e-5`: If a parameter spans a small range with a large common factor, like `[1_000_001, 1_000_002, ...]`, these will also count as equal. I think this is also very unlikely.
If we're using `isequal` with these tolerances, it'd be good to flag these edge cases in the docstring, so that the underlying assumptions are clearly communicated to the user. One step further would be to validate that the provided search spaces do not violate these assumptions when the optimizer is called, and error out if an unsupported search space is provided.

PR #3302, opened 2026-05-05
Local entropy search as community contribution

avrohr:
## Motivation

This PR adds a BoTorch community implementation of Local Entropy Search (LES) [1], an information-theoretic local Bayesian optimization method.

LES targets information gain about the local descent sequence started from a current incumbent, rather than information gain about a global optimizer. In our paper we show that this is useful for high-complexity problems.

This PR adds a BoTorch re-implementation of our methods. The original code is available [here](https://github.com/Data-Science-in-Mechanical-Engineering/local-entropy-search).

This PR also adds a community notebook demonstrating LES on a Branin example.

Yes.

## Test Plan

Tests are in ``test_local_entropy_search.py``. Please let me know if additional tests are required.

## Related PRs

This PR is self contained and does not change existing functionality.

## References

[1] David Stenger, Armin Lindicke, Alexander von Rohr, Sebastian Trimpe.
*Local Entropy Search over Descent Sequences for Bayesian Optimization*. ICLR 2026.
[https://openreview.net/forum?id=cPxmLZmFa7](https://openreview.net/forum?id=cPxmLZmFa7)


Balandat:
This is great, @avrohr, thanks for the contribution. Let me take a closer look at this later today, but overall I don't see any issues and this would be great to have.

botorch_community/acquisition/local_entropy_search.py
```diff
@@ -0,0 +1,445 @@
...
+
+            for _ in range(self.num_descent_steps):
+                optimizer.zero_grad(set_to_none=True)
+                path_values = matheron_paths(X_var).squeeze(-1).squeeze(-1)
+                objective = (
+                    path_values.sum() if not self.maximize else -path_values.sum()
+                )
+                objective.backward()
+
+                optimizer.step()
```
Balandat, line 298: @avrohr The pytorch optimizer makes it easy to batch things here. But this is typically a low-dimensional problem, so using a second order or quasi-second order optimizer should work well here. Have you considered doing that? We do have a batched version of L-BFGS-B in botorch that we could try out here. Certainly not blocking, but an interesting question.

Balandat (approved):
looks great, thanks!

botorch_community/acquisition/local_entropy_search.py
avrohr: @Balandat Thanks for feedback and fast merge. A nice thing about LES is that it can work with any iterative optimizer to generate the distribution over sequences that the entropy search acquisition uses. In the paper we test LES with simple gradient descent, ADAM, and CMAES. I tried to keep the API simple and limit the support to pytorch optimizer but if there is interest I could expand the API to support a broader range of underlying optimizer. Generally, it would make sense to also support optimizer that directly respect the bounds of the search space.

I had a quick look at botorch's L-BFGS-B and it seems like it doesn't expose the sequence of estimates and only returns the final value. While LES can work with distributions of local minima it doesn't perform as well as it does on sequences.

Balandat: Makes sense. It would be great to have a proper version of batched L-BFGS-B in scipy that also exposes the optimization trajectories for each batch. Alas, that will require a bit of work and time.

PR #3320, opened 2026-06-08
Linear constraints as sparse constraints

jduerholt:
## Motivation

`make_scipy_linear_constraints` currently returns a `list[ScipyConstraintDict]` — one dense dict per `(restart, q-row)` block, where each `jac(x)` closure builds an `np.zeros(n)` row with a few entries set. That works for SLSQP/COBYLA (which only accept the dict form), but it throws away the constraint Jacobian's sparsity structure for *any* solver that could exploit it.

Motivating use case:

**Sparse-aware custom solvers.** When a user passes a callable via `options["method"]` (scipy's `_custom` dispatch), the callable receives the constraints as-is. With dicts, the solver has no way to recover sparsity short of fragile probing. With a `LinearConstraint(sparse_A, lb, ub)` object, it can read `A.row`/`A.col`/`A.data` directly and hand Ipopt-style solvers a sparse `jacobianstructure` / `jacobian`. This can help for aggregated BoTorch problems with b·q·d up into the hundred or even thousands.

This PR switches `make_scipy_linear_constraints` to return up to **two** `scipy.optimize.LinearConstraint` objects (a `list` of 0, 1, or 2 entries): one stacking all inequality rows, one stacking all equality rows. Each carries a single `scipy.sparse.coo_array` for `A` plus the corresponding `lb` / `ub` vectors (equalities: `lb=ub=rhs`; inequalities: `lb=rhs, ub=+inf`, matching BoTorch's existing `A @ x >= rhs` convention). Keeping eq and ineq in separate objects avoids the `OptimizeWarning` scipy emits during SLSQP conversion when both row types share an `A`.

**Backward compatibility** is preserved end-to-end:

- SLSQP and COBYLA still work via scipy's `standardize_constraints`, which auto-converts each `LinearConstraint` back into per-row dicts via `new_constraint_to_old` before dispatch.
- Mixed lists like `[LinearConstraint, LinearConstraint, dict, dict]` (linear + nonlinear) are natively supported by `scipy.optimize.minimize`.
- The only internal consumers (`botorch/generation/gen.py:345` and `botorch/optim/parameter_constraints.py:project_to_feasible_space_via_slsqp`) are migrated to the new return type with a trivial `list(...)` wrap so the result composes with downstream nonlinear-constraint dicts.

Yes.

## Test Plan

Existing tests for `_make_linear_constraints` and `make_scipy_linear_constraints` are rewritten to assert on the new structure (sparse `A`, `lb`/`ub` vectors, list-of-LCs return type) while preserving the semantic checks. New tests added in `test/optim/test_parameter_constraints.py`:

- `test_make_scipy_linear_constraints` now uses `scipy.optimize._minimize.standardize_constraints` to round-trip the list of `LinearConstraint`s back into the legacy dict form and verifies the per-row `fun(x)` / `jac(x)` math is unchanged. Catches both incorrect COO emission *and* any future regression in scipy's converter. The round-trip is run under `warnings.simplefilter("error", OptimizeWarning)` to pin the eq/ineq split (any future re-merge into a single `LinearConstraint` will fail the test).
- `test_make_scipy_linear_constraints_sparsity_intra_point` — for `shapeX=(b=4, q=3, d=5)` intra-point mixture, asserts the per-row column sets are pairwise disjoint (block-diagonal) and `nnz == b * q * len(indices)`.
- `test_make_scipy_linear_constraints_inter_point_pattern` — for an inter-point constraint `[(0,0), (1,0), (2,0)]`, asserts the resulting row has non-zeros at the strided positions `[0, d, 2d]`.
- `test_make_scipy_linear_constraints_mixed_intra_inter` — stacks one intra-point and one inter-point equality constraint in the same call and verifies the COO row-offset arithmetic, the `lb`/`ub` slicing across blocks, per-row math at five row positions, and the `standardize_constraints` round-trip. This case was never covered before this PR.

Separately, a small dispatch smoke test in `test/generation/test_gen.py`:

- `test_gen_candidates_scipy_custom_method_dispatch` — passes a callable that raises a sentinel exception as `options["method"]` and asserts the exception bubbles up through `gen_candidates_scipy → minimize_with_timeout → scipy → _custom`. Pins the integration point that custom solvers (pounce, ipopt-via-cyipopt, …) rely on; if any layer ever drops the callable on the floor, this test catches it.

## Related PRs

https://github.com/meta-pytorch/botorch/pull/2368 is related, but it can be closed after this PR as the option to pass in a seperate callable to scipy.optimize.minimize makes changes in botorch (despite expression of the sparsity structure obsolete).

### DIFF
--- a/botorch/generation/gen.py
+++ b/botorch/generation/gen.py
@@ -342,11 +342,12 @@ def gen_candidates_scipy(
 
             x0 = candidates_.flatten()
 
-            constraints = make_scipy_linear_constraints(
+            linear_lc = make_scipy_linear_constraints(
                 shapeX=candidates_.shape,
                 inequality_constraints=_no_fixed_features.inequality_constraints,
                 equality_constraints=_no_fixed_features.equality_constraints,
             )
+            constraints = [linear_lc] if linear_lc is not None else []
             if _no_fixed_features.nonlinear_inequality_constraints:
                 # Make sure ``batch_limit`` is 1 for now.
                 if not (len(candidates_.shape) == 3 and candidates_.shape[0] == 1):
--- a/botorch/generation/gen.py
+++ b/botorch/generation/gen.py
@@ -342,12 +342,13 @@ def gen_candidates_scipy(
 
             x0 = candidates_.flatten()
 
-            linear_lc = make_scipy_linear_constraints(
-                shapeX=candidates_.shape,
-                inequality_constraints=_no_fixed_features.inequality_constraints,
-                equality_constraints=_no_fixed_features.equality_constraints,
+            constraints = list(
+                make_scipy_linear_constraints(
+                    shapeX=candidates_.shape,
+                    inequality_constraints=_no_fixed_features.inequality_constraints,
+                    equality_constraints=_no_fixed_features.equality_constraints,
+                )
             )
-            constraints = [linear_lc] if linear_lc is not None else []
             if _no_fixed_features.nonlinear_inequality_constraints:
                 # Make sure ``batch_limit`` is 1 for now.
                 if not (len(candidates_.shape) == 3 and candidates_.shape[0] == 1):
--- a/botorch/generation/gen.py
+++ b/botorch/generation/gen.py
@@ -342,13 +342,12 @@ def gen_candidates_scipy(
 
             x0 = candidates_.flatten()
 
-            constraints = list(
-                make_scipy_linear_constraints(
-                    shapeX=candidates_.shape,
-                    inequality_constraints=_no_fixed_features.inequality_constraints,
-                    equality_constraints=_no_fixed_features.equality_constraints,
-                )
+            constraints = make_scipy_linear_constraints(
+                shapeX=candidates_.shape,
+                inequality_constraints=_no_fixed_features.inequality_constraints,
+                equality_constraints=_no_fixed_features.equality_constraints,
             )
+
             if _no_fixed_features.nonlinear_inequality_constraints:
                 # Make sure ``batch_limit`` is 1 for now.
                 if not (len(candidates_.shape) == 3 and candidates_.shape[0] == 1):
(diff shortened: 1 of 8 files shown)


jduerholt:
@saitcakmak, any thoughts on this?

botorch/generation/gen.py
```diff
@@ -342,10 +342,12 @@ def f(x):
 
             x0 = candidates_.flatten()
 
-            constraints = make_scipy_linear_constraints(
-                shapeX=candidates_.shape,
-                inequality_constraints=_no_fixed_features.inequality_constraints,
-                equality_constraints=_no_fixed_features.equality_constraints,
+            constraints = list(
```
jduerholt: I think this list is not needed anymore, i need to get rid of it.

botorch/optim/parameter_constraints.py
```diff
@@ -88,63 +107,131 @@ def make_scipy_linear_constraints(
...
+    ineq_lc = _stack(ineq_blocks)
+    if ineq_lc is not None:
+        result.append(ineq_lc)
+    eq_lc = _stack(eq_blocks)
+    if eq_lc is not None:
+        result.append(eq_lc)
+    return result
 
 
 def eval_lin_constraint(
```
saitcakmak: Looks like eval_lin_constraint and lin_constraint_jac are now dead code

saitcakmak (approved):
Sorry about the late review. Lgtm!

botorch/generation/gen.py
jduerholt: Done.

botorch/optim/parameter_constraints.py
jduerholt: Good catch, I overlooked this, and you are right, this is not anymore needed as it is now handled directly in scipy.

jduerholt:
Hi @saitcakmak,

thanks for the review. I just fixed the code that you flagged.
Best,

Johannes

jduerholt:
Hi @saitcakmak, anything missing here to get this merged? Best, Johannes

saitcakmak:
This is all good, it should be landing soon. We have some additional restrictions around landing changes around big shopping events (currently amazon prime day) since I am now in the Ads org.

PR #3323, opened 2026-06-12
Added input constructor for LogRegionalExpectedImprovement and qLogRegionalExpectedImprovement

Sahil-kirangi:
Registers `construct_inputs_rei` for both `LogRegionalExpectedImprovement`
  and `qLogRegionalExpectedImprovement` in `botorch_community/acquisition/input_constructors.py`.
- Infers `best_f` from training data when not provided, consistent with
  the pattern used by `construct_inputs_best_f`.
- Passes through `X_dev`, `posterior_transform`, `length`, and `bounds`.

## Motivation

Without this constructor, calling `get_acquisition_function` with either
REI class raises a KeyError. This fills the gap left when the original
REI implementation was merged.

Yes, I filled out the application for these as well.

## Test Plan

- Added `test_construct_inputs_rei` to `test_community/acquisition/test_input_constructors.py`.
- Covers: auto-inferred `best_f`, explicit `best_f`, explicit `length`
  and `bounds`, and successful acqf instantiation from returned kwargs.
- Tests pass for both analytic and MC variants.

## Related PRs

### DIFF
--- a/botorch_community/acquisition/input_constructors.py
+++ b/botorch_community/acquisition/input_constructors.py
@@ -8,7 +8,10 @@ r"""
 A registry of helpers for generating inputs to acquisition function
 constructors programmatically from a consistent input format.
 
-Contributor: hvarfner (bayesian_active_learning, scorebo)
+Contributors:
+    hvarfner (bayesian_active_learning, scorebo)
+    SaiAakash (rei)
+    sahilKirangi (113878) (rei input constructor)
 """
 
 from __future__ import annotations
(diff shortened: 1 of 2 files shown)


saitcakmak (changes requested):
Hi @113878. Thanks for the PR. 

The input constructors are typically used for compatibility with Ax's modular BoTorch generator, which only passes a certain list of possible constructor arguments to the input constructors ([see here](https://github.com/facebook/Ax/blob/main/ax/generators/torch/botorch_modular/acquisition.py#L417)). In this case, the input constructor requires `X_dev`, which is not an argument Ax supplies. So, I don't think this constructor will work. Could you instead sample `X_dev` using `bounds` or some other available input?

Have you tested this constructor with Ax? What does your intended usage look like?

PR #3324, opened 2026-06-14
Add AddTree GP for tree-structured / conditional parameter spaces

maxc01:
Implements the AddTree covariance kernel and Bayesian-optimisation helpers from Ma & Blaschko, AISTATS 2020 (https://arxiv.org/abs/2006.11771), contributed under botorch_community.

Many real-world BO problems have *conditional* parameter spaces: a top-level categorical decision selects which sub-parameters apply, with arbitrary nesting. Examples include neural architecture search, algorithm-configuration tuning, and any pipeline whose downstream hyperparameters depend on an earlier choice. The AddTree kernel encodes the tree structure into the GP prior so the model shares information across paths whose continuous parameters live in distinct subspaces.

User-facing API
===============

A space is described declaratively, either as a Python literal::

    from botorch_community.models.addtree import (
        AddTreeSpace, AddTreeGP, encode, decode,
    )

    space = AddTreeSpace.from_dict({
        'name': 'root',
        'choices': [{
            'name': 'L1',
            'options': {
                'left':  {'name': 'L_left',
                          'continuous': [{'name': 'v', 'lo': 0., 'hi': 1.}]},
                'right': {'name': 'L_right',
                          'continuous': [{'name': 'v', 'lo': 0., 'hi': 1.}]},
            },
        }],
    })

or from a YAML file via AddTreeSpace.from_yaml(path), or from Node / Choice / Continuous dataclasses for programmatic construction.

The objective and BO loop work in a flat user-facing dict::

    p = decode(space, x)
    # -> {'L1': 'left', 'L_left.v': 0.42}

Inputs to the GP are BFS-encoded tensors produced by encode():

    train_X = torch.stack([encode(space, p) for p in observed_params])
    model = AddTreeGP(space, train_X, train_Y)
    fit_gpytorch_mll(ExactMarginalLogLikelihood(model.likelihood, model))

Acquisition optimisation uses optimize_acqf_mixed under the hood, so both q=1 (UpperConfidenceBound) and q>1 (qUpperConfidenceBound) are supported with no user-side branching::

    from botorch_community.acquisition.addtree import optimize_addtree_acqf
    candidates, vals = optimize_addtree_acqf(model, beta=3.0, q=1)

What is added
=============

  - botorch_community/models/addtree/space.py       declarative spec
  - botorch_community/models/addtree/encoding.py    encode/decode
  - botorch_community/models/addtree/kernels.py     AddTreeDeltaKernel +
                                                     build_addtree_kernel
  - botorch_community/models/addtree/model.py       AddTreeGP
  - botorch_community/acquisition/addtree.py        optimize_addtree_acqf
                                                     (thin wrapper over
                                                     optimize_acqf_mixed)
  - examples/jenatton_small.py        two-level binary tree, target 0.1
  - examples/jenatton3_synthetic.py   three-level binary, target 0.05
  - test_community/models/test_addtree_*.py         36 tests

Validation
==========

  - All 36 unit tests pass.
  - ufmt + flake8 clean.
  - examples/jenatton_small.py reaches the global optimum (0.1) on seed=0 with the default n_init=10, n_iter=100, beta=3.
  - examples/jenatton3_synthetic.py reaches gap 0.0003 to the global optimum 0.05 on seed=0 with the default n_init=20, n_iter=60, beta=3 (BFS dim 29, 8 root-to-leaf paths).
  - On 10 random seeds of jenatton_small (n_iter=30, beta=1) the algorithm reaches the global optimum on 7/10 seeds, matching the upstream reference implementation's success rate.

Contributor: maxc01

## Motivation

(Write your motivation here.)

(Write your answer here.)

## Test Plan

## Related PRs

### DIFF
--- a/botorch_community/acquisition/addtree.py
+++ b/botorch_community/acquisition/addtree.py
@@ -1,0 +1,95 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+#
+# This source code is licensed under the MIT license found in the
+# LICENSE file in the root directory of this source tree.
+
+r"""Acquisition-function optimisation for the AddTree conditional space.
+
+This is a **thin** wrapper around BoTorch's
+:func:`~botorch.optim.optimize.optimize_acqf_mixed`: the AddTree spec
+already exposes its ``fixed_features_list``, so optimisation is a single
+call.
+
+Sign convention:
+    BoTorch maximises; AddTree's original acquisition (LCB) minimises.
+    Negate observed objective values when constructing the
+    :class:`~botorch_community.models.addtree.model.AddTreeGP` and pass a
+    maximisation acquisition function (the default
+    :class:`~botorch.acquisition.UpperConfidenceBound` here).
+
+Contributor: maxc01
+"""
+
+from __future__ import annotations
+
+from typing import Any
+
+from botorch.acquisition import (
+    AcquisitionFunction,
+    qUpperConfidenceBound,
+    UpperConfidenceBound,
+)
+from botorch.optim import optimize_acqf_mixed
+
+from botorch_community.models.addtree.model import AddTreeGP
+from torch import Tensor
+
+
+__all__ = ["optimize_addtree_acqf"]
+
+
+def optimize_addtree_acqf(
+    model: AddTreeGP,
+    *,
+    acqf: AcquisitionFunction | None = None,
+    beta: float = 1.0,
+    q: int = 1,
+    num_restarts: int = 2,
+    raw_samples: int = 128,
+    **optimize_acqf_mixed_kwargs: Any,
+) -> tuple[Tensor, Tensor]:
+    r"""Optimise an acquisition function over the AddTree space.
+
+    Args:
+        model: A fitted :class:`AddTreeGP`. The space is read from
+            ``model.addtree_space``.
+        acqf: Optional acquisition function. If ``None``, defaults to
+            :class:`~botorch.acquisition.UpperConfidenceBound` with the
+            supplied ``beta``.
+        beta: ``beta`` for the default UCB acquisition. Ignored if
+            ``acqf`` is provided.
+        q: Batch size (number of candidates). q > 1 is supported for
+            q-batch acquisitions; sequential greedy is used internally.
+        num_restarts: Multistart count for L-BFGS-B (forwarded).
+        raw_samples: Sobol initialisation samples per fixed-features
+            entry (forwarded).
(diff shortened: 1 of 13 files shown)


hvarfner:
Hi @maxc01 ,


Thanks for putting this PR together! The citation link is probably not the right one, so you may want to update that.

Before going line-by-line, I think the PR could be quite a bit leaner — a lot of it overlaps with things botorch and Ax already provide. A few high-level thoughts:

1. How does this differ from the existing HierarchicalConditionalKernelGP? botorch already has the hierarchical conditional-kernel GP in botorch/models/hierarchical/, which is the same additive tree-structured kernel idea over conditional search spaces. As far as I can tell the two are barely different in practice, so I'd love to understand what AddTree adds that the existing model can't. If there's a genuinely new piece, I think a thin kernel/model on top of the existing components is enough — the per-node Delta * RBF sum can largely be assembled from existing kernels (CategoricalKernel/DeltaKernel for the equality gate + per-branch RBFKernel) on a plain SingleTaskGP, so I'd rather not add a parallel model/kernel stack if we can avoid it.

2. Search-space scaffolding belongs in Ax. AddTreeSpace and its dict/yaml DSL is a parameter-space modeling layer, and that's really Ax's job — botorch models just take tensors. The Jenatton objective should be in test_functions or in a notebook example.

3. Let the caller build the acquisition function. optimize_addtree_acqf constructs the acqf internally; the usual botorch pattern is for the caller to build it and pass it to optimize_acqf. As written it's mostly a passthrough to optimize_acqf_mixed.

Examples should be notebooks rather than scripts. The example .py files under examples/ would fit better as tutorial notebooks.

A general cleanup pass would help to remove the excessive verbosity in some places. 

Happy to dig into specifics once the overall shape settles. Thanks again!

hvarfner:
@saitcakmak

saitcakmak:
I agree with @hvarfner  here. We've tried out AddTree kernels but we did not find them to work well, unless the test functions had very specific additive structure as assumed by the kernel. I would like to avoid adding a lot of bespoke code unless there is a clear value add for the users.

maxc01:
Thanks for the careful review. After looking at `HierarchicalConditionalKernel` in detail, I agree --  the existing hierarchical-search-space stack already covers this problem class well enough that a parallel kernel/model here doesn't make sense, so I'll close this PR. The two approaches share the broader idea of additive active-feature-gated kernels but differ in encoding, default partition, and hyperparameter priors -- related but not the same kernel.

Issue #3325, opened 2026-06-15
[FEATURE REQUEST]: `SingleTaskMultiFidelityGP` does not have a `mean_module` in its `__init__`


AVHopp:
### Motivation

The `SingleTaskMultiFidelityGP` does not offer the possibility to provide a `mean_module` in its `__init__`:

```python
def __init__(
        self,
        train_X: Tensor,
        train_Y: Tensor,
        train_Yvar: Tensor | None = None,
        iteration_fidelity: int | None = None,
        data_fidelities: Sequence[int] | None = None,
        linear_truncated: bool = True,
        nu: float = 2.5,
        covar_module: Module | None = None,
        likelihood: Likelihood | None = None,
        outcome_transform: OutcomeTransform | _DefaultType | None = DEFAULT,
        input_transform: InputTransform | None = None,
    ) -> None:
```

Consequently, it always uses the default mean of its parent class, the `SingleTaskGP`. Is this an intentional design/decision or is this indeed an error and one should be able to hand over the `mean_module`? In this case, I'd volunteer to draft a small PR fixing this :)

### Describe the solution you'd like to see implemented in BoTorch.

Rather easy, just add the `mean_module` to the `__init__` and hand it over in the `suprt.__init__()` call

### Pull Request

Yes

### Code of Conduct

- [x] I agree to follow BoTorch's Code of Conduct

hvarfner:
Hi @AVHopp ,

This is most likely an oversight on our end. Feel free to draft a PR and I'll try and be quick to review!

AVHopp:
@hvarfner created a PR at #3326 :)

PR #3326, opened 2026-06-17
Add `mean_module` as `__init__` argument for `SingleTaskMultiFidelityGP`

AVHopp:
## Motivation

This PR implements #3325 by adding a `mean_module` argument to the `__init__` of the `SingleTaskMultiFidelityGP`, as well as corresponding tests. As mentioned in that issue, this argument seems to have simply been overlooked and "forgotten" so far. As a consequence, the fallback of a `ConstantMean` was used. The changes in this PR now allow to use different mean modules.

In addition to the addition of the argument, the following parts were changed:
- The code only adds `"mean_module.raw_constant": -1` to the `_subset_batch_dict` if the provided `mean_module` actually has a parameter of that name and its shape is equal to `_aug_batch_shape`.
- The code overrides `subset_outputs`. This is necessary due to users now potentially hading over custom mean modules.

Yes.

## Test Plan

Three new tests were being added:
- Providing a custom mean module works
- Explicit test for the `subset_output` override, once with a mean with and once with a mean without parameters
- Test for verifying that an error is thrown when a custom mean module with output-specific parameters is used for which the model does not know how to subset them safely.

In addition, pre-commit hooks were run.

## Related PRs

None.

## Further comments

The argument is added as the last argument to not break anything relying on the positions of the arguments: If `mean_module` would be added at another point of the signature (e.g. next to the `cover_module`), this would be a breaking change. One could even add it as a dedicated Keyword-only argument to be even safer.

For transparency: AI was used in the creation of this code, but everything was checked and verified by myself.

### DIFF
--- a/botorch/models/gp_regression_fidelity.py
+++ b/botorch/models/gp_regression_fidelity.py
@@ -26,7 +26,7 @@ without having to do too many expensive high-fidelity evaluations.
 from __future__ import annotations
 
 from collections.abc import Sequence
-from typing import Any
+from typing import Any, cast
 
 import torch
 from botorch.exceptions.errors import UnsupportedError
@@ -44,6 +44,7 @@ from botorch.utils.types import _DefaultType, DEFAULT
 from gpytorch.kernels.kernel import ProductKernel
 from gpytorch.kernels.scale_kernel import ScaleKernel
 from gpytorch.likelihoods.likelihood import Likelihood
+from gpytorch.means.mean import Mean
 from gpytorch.module import Module
 from gpytorch.priors.torch_priors import GammaPrior
 from torch import Tensor
@@ -77,6 +78,7 @@ class SingleTaskMultiFidelityGP(SingleTaskGP):
         likelihood: Likelihood | None = None,
         outcome_transform: OutcomeTransform | _DefaultType | None = DEFAULT,
         input_transform: InputTransform | None = None,
+        mean_module: Mean | None = None,
     ) -> None:
         r"""
         Args:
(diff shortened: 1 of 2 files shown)


botorch/models/gp_regression_fidelity.py
```diff
@@ -172,6 +185,21 @@ def construct_inputs(
         inputs["data_fidelities"] = fidelity_features
         return inputs
 
+    def subset_output(self, idcs: list[int]) -> SingleTaskMultiFidelityGP:
```
saitcakmak, line 188: Do we need the custom `subset_output` implementation here? I am generally suspicious of the `subset_output` functionality across botorch models. I don't think it generally works correctly with custom mean / covar modules (this suggest to me it doesn't: https://github.com/AVHopp/botorch/blob/b5a8768d343bd33266a04dbaaec01f7f35e2da0e/botorch/models/gp_regression.py#L202).

Is `subset_output` something you utilize? I am wondering if we should deprecate it to simplify the model code.

AVHopp: I personally do not use it. To be fully honest and transparent, this change of the `subset_output` functionality is something that an AI agent suggested me should also be fixed and changed alongside my change to the `__init__`, and the presented argumentation for it seemed reasonable to me, hence I included it.
That being said, I'm happy to change this part as you see fit - from keeping it as it was, changing it to raise errors/warnings in case somebody uses custom modules from doing a full deprecation (which would however probably be outside of the scope of this PR I assume?), I am open to everything 😄

PR #3337, opened 2026-07-10
Fix qLogEHVI IndexError on empty box decompositions

### LINKED ISSUE #3335: [Bug]: qLogEHVI raises IndexError when the box decomposition has no cells
### What happened?

While moving optuna's BoTorch integration from qEHVI to qLogEHVI (optuna/optuna-integration#302), CI hit an IndexError inside qLogEHVI on a case that qEHVI handles fine.

With several objectives and only a few observed points, `NondominatedPartitioning` with `alpha > 0` can prune every cell, so `get_hypercell_bounds()` returns a tensor with `num_cells == 0`. The legacy `qExpectedHypervolumeImprovement` evaluates to 0 for that input, which matches the math (no cells means no measurable improvement region). `qLogExpectedHypervolumeImprovement` instead raises: `_compute_log_qehvi` ends with `logmeanexp(logsumexp(log_areas_per_segment, dim=-1), dim=0)`, and the `logsumexp` reduction over the empty cells dim hits the `amax` call in `safe_math._inf_max_helper`, which does not accept zero size dims.

Interestingly, plain `torch.logsumexp` returns `-inf` for a zero size reduction dim, which is the mathematically expected value here (log of an empty sum). So `safe_math.logsumexp` diverges from the torch function it wraps only because of the `amax` pre step.

`qLogNoisyExpectedHypervolumeImprovement` is not affected, its incremental path handles this case.

Reproduces on 0.9.5, 0.18.1, and current main (0.18.2.dev10+g46cc96bc8).

```python
import torch

import botorch
from botorch.acquisition.multi_objective import monte_carlo
from botorch.acquisition.multi_objective.logei import qLogExpectedHypervolumeImprovement
from botorch.utils.multi_objective.box_decompositions import NondominatedPartitioning

print("botorch:", botorch.version.version)
print("torch:", torch.__version__)

M = 7
torch.manual_seed(87)
train_Y = -torch.rand(3, M, dtype=torch.double)
train_X = torch.rand(3, M, dtype=torch.double)

ref_point = train_Y.min(dim=0).values - 1e-8
partitioning = NondominatedPartitioning(ref_point=ref_point, Y=train_Y, alpha=0.1)
print("num_cells:", partitioning.get_hypercell_bounds().shape[-2])  # 0

model = SingleTaskGP(train_X, train_Y, outcome_transform=Standardize(m=M))
X_test = torch.rand(1, 1, M, dtype=torch.double)

legacy = monte_carlo.qExpectedHypervolumeImprovement(
    model=model, ref_point=ref_point.tolist(), partitioning=partitioning
)
print("qEHVI:", legacy(X_test))  # tensor([0.]) as expected

qlog = qLogExpectedHypervolumeImprovement(
    model=model, ref_point=ref_point.tolist(), partitioning=partitioning
)
print("qLogEHVI:", qlog(X_test))  # IndexError
```

```shell
  File "repro.py", line 33, in <module>
    print("qLogEHVI:", qlog(X_test))  # IndexError
                       ~~~~^^^^^^^^
  File ".../site-packages/torch/nn/modules/module.py", line 1778, in _wrapped_call_impl
    return self._call_impl(*args, **kwargs)
           ~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^
  File ".../site-packages/torch/nn/modules/module.py", line 1789, in _call_impl
    return forward_call(*args, **kwargs)
  File ".../site-packages/botorch/utils/transforms.py", line 400, in decorated
    return method(cls, X, **kwargs)
  File ".../site-packages/botorch/utils/transforms.py", line 300, in decorated
    output = method(acqf, X, *args, **kwargs)
  File ".../site-packages/botorch/utils/transforms.py", line 366, in decorated
    output = method(acqf, X, *args, **kwargs)
  File ".../site-packages/botorch/acquisition/multi_objective/logei.py", line 483, in forward
    return self._compute_log_qehvi(samples=samples, X=X)
           ~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^
  File ".../site-packages/botorch/acquisition/multi_objective/logei.py", line 432, in _compute_log_qehvi
    return logmeanexp(logsumexp(log_areas_per_segment, dim=-1), dim=0)
                      ~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File ".../site-packages/botorch/utils/safe_math.py", line 143, in logsumexp
    return _inf_max_helper(torch.logsumexp, x=x, dim=dim, keepdim=keepdim)
  File ".../site-packages/botorch/utils/safe_math.py", line 167, in _inf_max_helper
    M = x.amax(dim=dim, keepdim=True)
IndexError: amax(): Expected reduction dim -1 to have non-zero size.
```

### BoTorch Version

0.18.2.dev10+g46cc96bc8 (main); also reproduces on 0.18.1 and 0.9.5

### Python Version

3.13

### Operating System

macOS (Apple Silicon); also hit on ubuntu-latest in optuna-integration CI

Three directions, depending on where you want the invariant to live:

1. Make `safe_math._inf_max_helper` handle zero size reduction dims, for example by deferring to the wrapped function directly, so `safe_math.logsumexp` keeps the same semantics as `torch.logsumexp` (returns `-inf`). This fixes every fat max style op at once.
2. Guard in `_compute_log_qehvi`: if the decomposition has no cells, return a suitably small constant with the right batch shape, consistent with how the fat approximations bound other log zero cases.
3. Have `NondominatedPartitioning` with `alpha > 0` never prune down to zero cells (always keep the largest cell), since an empty approximate decomposition defeats the purpose of the approximation.

Happy to open a PR for whichever direction you prefer.

### Pull Request

Yes

### Code of Conduct

adrianhtt:
## Motivation

Fixes #3335.

qLogExpectedHypervolumeImprovement raises `IndexError: amax(): Expected reduction dim -1 to have non-zero size` when the box decomposition it receives has zero hypercells. This happens with an approximate decomposition (alpha > 0) when there are many objectives and few observed points, so every cell overlapping the pareto front gets pruned. The legacy qExpectedHypervolumeImprovement returns 0 for the same input.

The root cause is in `safe_math._inf_max_helper`, which backs `logsumexp`, `fatmax` and `smooth_amax`: the amax pre step does not accept zero size reduction dims, while plain `torch.logsumexp` returns -inf for them.

## Changes

- `safe_math._inf_max_helper` returns -inf for zero size reduction dims, keeping the result connected to the autograd graph so gradients stay defined. This restores parity with `torch.logsumexp` and is the correct limit for the smooth max approximations, since the maximum over an empty set is -inf.
- `initialize_q_batch` falls back to random initial conditions with `BadInitialCandidatesWarning` when acquisition values are non finite, instead of crashing in `torch.multinomial` on nan sampling weights. Without this, `optimize_acqf` still failed downstream whenever the acquisition function evaluates to -inf for all raw samples.
- The fused C++ kernel dispatch in `_compute_log_qehvi` skips the zero cell case. There is nothing to fuse and the Python path handles it directly.

With these three pieces the full `optimize_acqf` pipeline degrades gracefully on an empty decomposition: qLogEHVI evaluates to -inf, the initializer warns and picks random initial conditions, and finite candidates come back, matching what users get from the legacy qEHVI today.

## Test plan

- New `TestEmptyReductionDims` in test_safe_math.py: logsumexp, fatmax and smooth_amax on zero size dims, int and tuple dims, keepdim on and off, exact agreement with `torch.logsumexp`, and finite gradients.
- New case in `test_initialize_q_batch`: all -inf acquisition values warn and return random initial conditions instead of raising.
- New regression test in test_logei.py: a deterministic 7 objective, 3 point input whose approximate decomposition (alpha=0.1) prunes every cell; qLogEHVI evaluates to -inf for both fat=True and fat=False.
- ufmt and flake8 clean on all changed files. `test_gen_value_function_initial_conditions` fails for me on current main with and without this change when running the whole file, so it appears unrelated.

### DIFF
--- a/botorch/acquisition/multi_objective/logei.py
+++ b/botorch/acquisition/multi_objective/logei.py
@@ -322,6 +322,9 @@ class qLogExpectedHypervolumeImprovement(
             and self.fat
             and obj.device.type == "cpu"
             and q <= _FUSED_MAX_I
+            # An approximate box decomposition can have zero cells; there is
+            # nothing to fuse then and the Python path handles it directly.
+            and self.cell_lower_bounds.shape[-2] > 0
         )
 
         if use_fused and cell_batch_ndim > 0:
--- a/botorch/optim/initializers.py
+++ b/botorch/optim/initializers.py
@@ -1004,10 +1004,13 @@ def initialize_q_batch(
         return X, acq_vals
 
     Ystd = acq_vals.std(dim=0)
-    if torch.any(Ystd == 0):
+    if torch.any(Ystd == 0) or not torch.isfinite(Ystd).all():
+        # A non-finite std indicates non-finite acquisition values, for example when
+        # the acquisition function evaluates to -inf for all raw samples. Sampling
+        # probabilities proportional to exp(eta * Z) are not defined in either case.
         warnings.warn(
-            "All acquisition values for raw samples points are the same for "
-            "at least one batch. Choosing initial conditions at random.",
+            "All acquisition values for raw samples points are the same or not "
+            "finite for at least one batch. Choosing initial conditions at random.",
             BadInitialCandidatesWarning,
             stacklevel=3,
         )
--- a/botorch/utils/safe_math.py
+++ b/botorch/utils/safe_math.py
@@ -162,8 +162,15 @@ def _inf_max_helper(
 
     Returns:
         The Tensor representing the smooth approximation to the maximum over the
-        specified dimensions.
+        specified dimensions. If the reduced dimensions of ``x`` have zero size,
+        the result is ``-inf``, consistent with the maximum over an empty set.
     """
+    dims = (dim,) if isinstance(dim, int) else tuple(dim)
+    if any(x.shape[d] == 0 for d in dims):
+        # The maximum over an empty set is -inf, which also matches the behavior of
+        # torch.logsumexp over a zero-size dimension. The sum keeps the result
+        # connected to the autograd graph, so gradients w.r.t. x remain defined.
+        return x.sum(dim=dim, keepdim=keepdim) - torch.inf
     M = x.amax(dim=dim, keepdim=True)
     is_inf_max = torch.logical_and(*torch.broadcast_tensors(M.isinf(), x == M))
     has_inf_max = _any(is_inf_max, dim=dim, keepdim=True)
--- a/test/acquisition/multi_objective/test_logei.py
+++ b/test/acquisition/multi_objective/test_logei.py
@@ -153,6 +153,43 @@ class TestLogQExpectedHypervolumeImprovement(BotorchTestCase):
         else:  # This is an interesting difference between the exp and the fat tail.
             self.assertEqual(0, exp_log_res)
 
+    def test_q_log_expected_hypervolume_improvement_empty_partitioning(self):
+        # An approximate box decomposition (alpha > 0) can prune all hypercells
+        # when there are many objectives and few observations. The hypervolume
+        # improvement over an empty cell set is zero, so the log acquisition
+        # value is -inf rather than an IndexError, see
+        # https://github.com/meta-pytorch/botorch/issues/3335.
+        tkwargs = {"device": self.device, "dtype": torch.double}
+        Y = torch.tensor(
+            [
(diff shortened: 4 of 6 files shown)


saitcakmak (approved):
Thanks, lgtm!

PR #3338, opened 2026-07-14
Add Alpha Entropy Search (AES) and a corresponding tutorial

fernandezdaniel:
## Motivation

This pull request adds an implementation of Alpha Entropy Search (AES) [1], an
information-based acquisition function for Bayesian Optimization, along with a
test file and a tutorial notebook that illustrates its use and compares it against 
standard baselines.

AES generalizes Joint Entropy Search (JES) [2]. JES measures the dependency
between `{x*, y*}` and `y` using the KL-divergence between their joint
distribution and the corresponding factorized distribution. AES replaces this
KL-divergence with Amari's alpha-divergence, which has a free parameter
`alpha` that controls the sensitivity of the divergence to different regions
of the two distributions. Values of `alpha` close to `0` and close to `1`
recover the reversed and direct KL-divergence limits, respectively, and
different values of `alpha` give rise to different acquisition functions. All
of them are evaluated in closed form, using the same Gaussian approximations
as JES.

[1] D. Fernández-Sánchez, E. C. Garrido-Merchán, D. Hernández-Lobato.
Alpha entropy search for new information-based Bayesian optimization.
Knowledge-Based Systems, 322, 113612, 2025.

[2] C. Hvarfner, F. Hutter, N. Nardi. Joint Entropy Search for
Maximally-informed Bayesian Optimization. NeurIPS, 2022.

Yes, I have read the Contributing Guidelines on pull requests. This is a
community contribution under `botorch_community`, `test_community`, and
`notebooks_community`, per the "Community Contributions (Beta)" section of
`CONTRIBUTING.md`.

## Test Plan

Unit tests are added in `test_community/acquisition/test_alpha_entropy_search.py`,
modeled after the existing `test/acquisition/test_joint_entropy_search.py`
test suite and extended with coverage specific to AES.

`test_singleobj_alpha_entropy_search` mirrors the JES test and checks:

- Shapes, dtypes (`float`/`double`), single-model vs. `ModelListGP`, and
  standardized vs. non-standardized outcomes, across `condition_noiseless`
  True/False.
- `posterior_transform` correctly negates/scalarizes `optimal_output_values`.
- An invalid `estimation_type` raises `ValueError`.
- A `SaasFullyBayesianSingleTaskGP` model raises `NotImplementedError` (AES
  does not yet support fully Bayesian models).

Three more tests cover behavior specific to AES:

- `test_alpha_edge_values` checks that `alpha=0.0` and `alpha=1.0`, which are
  undefined for Amari's alpha-divergence and are clamped internally via `eps`,
  produce finite acquisition values instead of dividing by zero.
- `test_q_greater_than_one_error` checks that `q > 1` raises an
  `AssertionError`, since AES only supports `q=1`.
- `test_x_pending_incompatible_with_q1` checks that a non-empty `X_pending`
  raises an `AssertionError`, since pending points are concatenated onto the
  `q` dimension and AES only supports `q=1`.

All 4 tests pass locally with `pytest test_community/acquisition/test_alpha_entropy_search.py`.

The tutorial notebook (`notebooks_community/alpha_entropy_search/alpha_entropy_search.ipynb`)
has two parts:

1. A 1D visualization reproducing the qualitative behavior of the paper's
   figure comparing JES, AES for several values of `alpha`, and an ensemble
   over `alpha` values, on a fixed-hyperparameter `SingleTaskGP`. It reports
   the number of local maxima and the global maximizer location for each
   acquisition function.
2. A closed-loop BO comparison on the noiseless 3D Hartmann function against
   `EI`, `MES`, and `JES`, following the paper's benchmark protocol (Matérn
   5/2 ARD kernel, fixed small likelihood noise, `S=32` samples of `{x*, y*}`
   per iteration, L-BFGS-B with 1 restart and 200 candidate points, 10 initial
   observations), averaged over 10 random seeds and 100 BO iterations. AES
   (for `alpha` in `{0.1, 0.3, 0.5, 0.7}`) converges to a noticeably lower
   regret than EI, MES, and JES.

The notebook supports `SMOKE_TEST=1` to run a reduced version (fewer seeds,
iterations, and Monte Carlo samples) for CI, and was executed end-to-end both
in smoke-test mode and at full scale without errors. The committed notebook
includes the full-scale outputs.

## Related PRs

Not applicable.


botorch_community/acquisition/alpha_entropy_search.py
```diff
@@ -0,0 +1,306 @@
...
+        check_no_nans(noiseless_var)
+
+        # get stdv of noiseless variance
+        noiseless_stdv = noiseless_var.sqrt()
+        # batch_shape x 1
+        normal = Normal(
+            torch.zeros(1, device=X.device, dtype=X.dtype),
+            torch.ones(1, device=X.device, dtype=X.dtype),
+        )
+        noiseless_normalized_mvs = (self.optimal_outputs - mean_cond) / noiseless_stdv
```
saitcakmak: To make sure posterior transform is consistently applied to both values
```suggestion
        noiseless_normalized_mvs = (self.optimal_output_values - mean_cond) / noiseless_stdv
```

botorch_community/acquisition/alpha_entropy_search.py
```diff
@@ -0,0 +1,306 @@
...
+        self,
+        model: Model,
+        optimal_inputs: Tensor,
+        optimal_outputs: Tensor,
+        condition_noiseless: bool = True,
+        posterior_transform: PosteriorTransform | None = None,
+        X_pending: Tensor | None = None,
+        estimation_type: str = "LB",
+        num_samples: int = 64,
+        alpha: int = 0.5,
```
saitcakmak:
```suggestion
        alpha: float = 0.5,
```

botorch_community/acquisition/alpha_entropy_search.py
```diff
@@ -0,0 +1,306 @@
...
+        noiseless_cdf_mvs = normal.cdf(noiseless_normalized_mvs).clamp_min(CLAMP_LB)
+        noiseless_pdf_mvs = torch.exp(normal.log_prob(noiseless_normalized_mvs))
+
+        noiseless_ratio = noiseless_pdf_mvs / noiseless_cdf_mvs
+
+        noiseless_var_cond_trunc = noiseless_var * (
+            1 - (noiseless_normalized_mvs + noiseless_ratio) * noiseless_ratio
+        ).clamp_min(CLAMP_LB)
+
+        var_cond_trunc = noiseless_var_cond_trunc + (var_cond - noiseless_var) - 1e-8
```
saitcakmak, line 273: Why do we subtract 1e-8? If the variance is tiny, this could end up negative.

botorch_community/acquisition/alpha_entropy_search.py
```diff
@@ -0,0 +1,306 @@
...
+)
+
+
+class qAlphaEntropySearch(AcquisitionFunction, MCSamplerMixin):
+    r"""The batch mode for the acquisition function Alpha Entropy Search,
+    it is not supported.
+
+    Alpha entropy search (AES) acquisition function is a generalization of
+    joint entropy search (JES). Instead of computing the Kullback-Leibler
+    divergence between the joint distribution between the joint distribution
```
saitcakmak: duplicate
```suggestion
    divergence between the joint distribution
```

botorch_community/acquisition/alpha_entropy_search.py
```diff
@@ -0,0 +1,306 @@
...
+        self.num_samples = optimal_inputs.shape[0]
+        self.condition_noiseless = condition_noiseless
+        self.initial_model = model
+
+        # Here, the optimal inputs have shapes num_optima x [num_models if FB] x 1 x D
+        # and the optimal outputs have shapes num_optima x [num_models if FB] x 1 x 1
+        # The third dimension equaling 1 is required to get one optimum per model,
+        # which raises a BotorchTensorDimensionWarning.
+        if isinstance(model, FullyBayesianSingleTaskGP):
+            raise NotImplementedError(FULLY_BAYESIAN_ERROR_MSG)
```
saitcakmak, line 136: How about the multi-task or any other variants? Is this specific to fully bayesian models or a general restriction on all batched / ensemble models? Depending on the answer, let's use one of these helpers:
`from botorch.utils.transforms import is_fully_bayesian, is_ensemble`

botorch_community/acquisition/alpha_entropy_search.py
```diff
@@ -0,0 +1,306 @@
...
+
+        Args:
+            X: A ``batch_shape x q x d``-dim Tensor of ``batch_shape``
+                t-batches with ``q`` ``d``-dim design points each.
+
+        Returns:
+            A ``batch_shape``-dim Tensor of acquisition values at the given design
+            points ``X``.
+        """
+        if self.estimation_type == "LB":
```
saitcakmak, line 192: This is already validated in `__init__`. Could we just remove `forward` implementation and inherit it from the parent class?

saitcakmak (changes requested):
Thanks, I left a few minor comments. Let's also make sure we have full test coverage.

botorch_community/acquisition/alpha_entropy_search.py
fernandezdaniel: Thanks for pointing this out. We intentionally added `-1e-8` to ensure that `var_cond_trunc` remains strictly below `var_pred`, as we rely on this when computing `v3`, where we evaluate `(1.0 / var_cond_trunc - 1.0 / var_pred)`. You're right that, in principle, it could become negative if `var_cond_trunc` were already close to its lower bound. To guard against that, we can add a `.clamp_min(CLAMP_LB)` immediately after the subtraction, following the same pattern already used a few lines above for `noiseless_var_cond_trunc`.

botorch_community/acquisition/alpha_entropy_search.py
fernandezdaniel: We've believed it's better to remove the check from `__init__` and keep it in `forward`, following the same pattern used by `qJointEntropySearch`. This should also address the issue Codecov pointed out, namely that there was an unreachable line. We've already made this change in [620438c](https://github.com/meta-pytorch/botorch/pull/3338/commits/620438c04).

botorch_community/acquisition/alpha_entropy_search.py
fernandezdaniel: Although it is possible to implement a multi-task variant, we have neither developed nor tested it. A variant supporting batched evaluation could also be possible, for example by following the approach used in JES, but we have not tested that either. In this implementation, we focus on the method presented in this paper, where we only consider the simplest setting, without multi-output GPs or batched evaluation.

Regarding the restriction on using fully Bayesian models, we inherited this limitation from the structure of the JES code. I think `is_ensemble(model)` would actually be the more appropriate check in this case, because when we have a collection of models (whether from MCMC or any other ensemble) and draw a sample of `num_optima` optima, each model in the ensemble could, in principle, have its own distinct optimum. Correctly conditioning each ensemble member on its own optimum is, to the best of my understanding, future work.

test_community/acquisition/test_alpha_entropy_search.py
```diff
@@ -0,0 +1,175 @@
...
+
+from itertools import product
+
+import torch
+from botorch.acquisition.objective import ScalarizedPosteriorTransform
+from botorch.models.fully_bayesian import SaasFullyBayesianSingleTaskGP
+from botorch.sampling.normal import SobolQMCNormalSampler
+from botorch.utils.test_helpers import get_model
+from botorch.utils.testing import BotorchTestCase
+
```
saitcakmak:
```suggestion
```

PR #3342, opened 2026-08-05
Compute CategoricalKernel diagonal without materializing the pairwise matrix

atinary-dpacheco:
## Motivation

`CategoricalKernel.forward` builds the full `n1 x n2 x d` comparison tensor and only
reduces it to the diagonal at the end, so asking for `n` diagonal values costs `n^2 * d`
elements. `diag=True` is exactly what `LazyEvaluatedKernelTensor._diagonal` requests, so
this sits on the hot path for posterior variances on models with categorical features.

Comparing the rows elementwise gives the same values from an `n x d` intermediate. Peak
allocation drops by a factor of `n`:

| n (d=3, float64) | peak before | peak after | time before | time after |
|---|---|---|---|---|
| 1000 | 22.89 MB | 0.023 MB | 8.22 ms | 0.019 ms |
| 5000 | 572.20 MB | 0.114 MB | 247.01 ms | 0.132 ms |

The dense (`diag=False`) branch is untouched — it needs every entry.

`diag=True` with differing row counts now raises `BotorchTensorDimensionError` rather
than silently returning the `min(n1, n2)` entries `torch.diagonal` used to produce
(per review). That path was already broken: GPyTorch's
`LazyEvaluatedKernelTensor._diagonal` expects `n1` entries back, so the old result would
fail its debug shape check anyway.

Yes. `ufmt` and `flake8` are clean.

## Test Plan

`pytest test/models/` — 422 passed (the one failure, `test_with_different_batch_shapes`,
is a missing-JAX `ImportError` that reproduces on an unmodified checkout).

Three tests added:

- `test_diag_matches_dense_diagonal` — compares `diag=True` against `torch.diagonal` of
  the dense result across batched/ARD kernels and both `last_dim_is_batch` values.
  Lengthscales are randomized so a misaligned lengthscale axis cannot cancel out.
- `test_diag_raises_on_mismatched_shapes` — covers the new error, and checks the dense
  path still accepts uneven inputs.
- `test_diag_does_not_materialize_pairwise_matrix` — bounds the largest tensor allocated
  during a diagonal request. Fails on the current implementation
  (`786432 not less than or equal to 6144`), so it guards against regressing to the
  dense path.

I also fuzzed the new implementation against the old one over 724 configurations (batch
shapes, row counts, `d`, float32/float64, ARD, and both flags): outputs are bitwise
identical (`torch.equal`) in value and shape.

## Related PRs

None.

### DIFF
--- a/botorch/models/kernels/categorical.py
+++ b/botorch/models/kernels/categorical.py
@@ -5,6 +5,7 @@
 # LICENSE file in the root directory of this source tree.
 
 import torch
+from botorch.exceptions.errors import BotorchTensorDimensionError
 from gpytorch.kernels.kernel import Kernel
 from torch import Tensor
 
@@ -30,11 +31,14 @@ class CategoricalKernel(Kernel):
         last_dim_is_batch: bool = False,
     ) -> Tensor:
         if diag:
+            if x1.shape[-2] != x2.shape[-2]:
+                raise BotorchTensorDimensionError(
+                    "diag=True requires `x1` and `x2` to have the same number of "
+                    f"points, but got {x1.shape[-2]} and {x2.shape[-2]}."
+                )
             # Comparing rows directly avoids the `n1 x n2 x d` cross-product that would
-            # otherwise be built and then discarded. `torch.diagonal` yields
-            # `min(n1, n2)` entries for uneven inputs, so truncate to match.
-            n = min(x1.shape[-2], x2.shape[-2])
-            dists = (x1[..., :n, :] != x2[..., :n, :]) / self.lengthscale
+            # otherwise be built and then discarded.
+            dists = (x1 != x2) / self.lengthscale
             if last_dim_is_batch:
                 return torch.exp(-dists.transpose(-1, -2))
             return torch.exp(-dists.mean(-1))
--- a/botorch/models/kernels/categorical.py
+++ b/botorch/models/kernels/categorical.py
@@ -29,13 +29,20 @@ class CategoricalKernel(Kernel):
         diag: bool = False,
         last_dim_is_batch: bool = False,
     ) -> Tensor:
+        if diag:
+            # Comparing rows directly avoids the `n1 x n2 x d` cross-product that would
+            # otherwise be built and then discarded. `torch.diagonal` yields
+            # `min(n1, n2)` entries for uneven inputs, so truncate to match.
+            n = min(x1.shape[-2], x2.shape[-2])
+            dists = (x1[..., :n, :] != x2[..., :n, :]) / self.lengthscale
+            if last_dim_is_batch:
+                return torch.exp(-dists.transpose(-1, -2))
+            return torch.exp(-dists.mean(-1))
+
         delta = x1.unsqueeze(-2) != x2.unsqueeze(-3)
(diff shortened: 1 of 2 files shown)


botorch/models/kernels/categorical.py
```diff
@@ -29,13 +29,25 @@ def forward(
...
     ) -> Tensor:
+        if diag:
+            # Comparing `x1` and `x2` row-wise yields the diagonal directly. Forming
+            # the full cross-product of comparisons and then discarding everything
+            # off the diagonal would allocate an `n1 x n2 x d` intermediate just to
+            # return `n` values, which dominates memory whenever only the diagonal is
+            # needed -- most notably when evaluating posterior variances at a large
+            # set of test points.
+            # `torch.diagonal` returns `min(n1, n2)` entries when the inputs have
+            # differing numbers of rows, so truncate to match that behavior.
```
atinary-dpacheco:
```suggestion
            # Comparing `x1` and `x2` row-wise yields the diagonal directly. 
```

botorch/models/kernels/categorical.py
```diff
@@ -29,13 +29,20 @@ def forward(
         diag: bool = False,
         last_dim_is_batch: bool = False,
     ) -> Tensor:
+        if diag:
+            # Comparing rows directly avoids the `n1 x n2 x d` cross-product that would
+            # otherwise be built and then discarded. `torch.diagonal` yields
+            # `min(n1, n2)` entries for uneven inputs, so truncate to match.
+            n = min(x1.shape[-2], x2.shape[-2])
+            dists = (x1[..., :n, :] != x2[..., :n, :]) / self.lengthscale
```
esantorella: Hm, I think it would be better to error if `x1.shape[-2] != x2.shape[-2]` and `diag=True`. I'm not sure why someone would want that behavior. This looks like a pre-existing bug to me, so thanks for flagging.

esantorella (changes requested):
Good idea, and thanks for getting this in! Left a comment on the issue you flagged.

botorch/models/kernels/categorical.py
atinary-dpacheco: Thanks for the review. That's a good point. Updated 👍🏼 https://github.com/meta-pytorch/botorch/pull/3342/changes/f3a082c3b84a08c57edf4897be704e4620d43c99

atinary-dpacheco: Let me know if the rest looks good

esantorella (approved):
Looks good to me!

botorch/models/kernels/categorical.py
atinary-dpacheco: Is there anything else I should do in order to merge this?

atinary-dpacheco:
Let me know if this can be merged 👍🏼 Thanks for the review

atinary-dpacheco:
Hi @esantorella thanks for the review. I re-requested a review by mistake. I think the PR should be ready to be merged whenever possible. Thanks a lot 👍🏼

atinary-dpacheco:
Hello, any news from this?

esantorella:
Sorry for the delay. Landing this now.

PR #3353, opened 2026-08-09
Fix Gumbel max-value sampling for multiple batch dimensions

### LINKED ISSUE #3352: [Bug]: Gumbel max-value sampling fails with multiple batch dimensions
## What happened?

`_sample_max_value_Gumbel` fails when the posterior has more than one batch dimension. A batched candidate set with shape `(2, 3, n, d)` is accepted by the model and by the Thompson max-value sampler, but the Gumbel sampler passes non-scalar bounds to SciPy and raises `RuntimeError: Unable to parse arguments`.

The expected result is a finite tensor with shape `(num_samples, 2, 3)`, matching `_sample_max_value_Thompson` and preserving the posterior batch shape.

The implementation currently transposes `mu` and `sigma` only when they are exactly rank 3, then assumes `mu.shape[1]` is the only batch dimension. For rank-4 and higher posteriors, the leading dimensions remain in place and each root-finding call receives an array-valued bound.

## Minimal reproduction

```python
import torch
from botorch.acquisition.max_value_entropy_search import (
    _sample_max_value_Gumbel,
    _sample_max_value_Thompson,
)


torch.manual_seed(0)
model = SingleTaskGP(
    torch.rand(10, 2, dtype=torch.double),
    torch.rand(10, 1, dtype=torch.double),
)
candidate_set = torch.rand(2, 3, 10, 2, dtype=torch.double)

with torch.no_grad():
    print(model.posterior(candidate_set).mean.shape)
    print(_sample_max_value_Thompson(model, candidate_set, 5).shape)
    print(_sample_max_value_Gumbel(model, candidate_set, 5).shape)
```

Output on current `main`:

```text
torch.Size([2, 3, 10, 1])
torch.Size([5, 2, 3])
RuntimeError: Unable to parse arguments
```

## Environment

- BoTorch: `0.18.2.dev13+g60d0e565e` (`main` at `60d0e565`)
- PyTorch: `2.11.0+cu130`, CPU execution
- Python: `3.13`
- OS: Windows 11

## Suggested fix

Flatten all posterior batch dimensions before the scalar quantile searches, then reshape the sampled maximum values back to `num_samples x batch_shape`. I have a focused regression and am willing to open a pull request.

aswanth-07:
## Motivation

Fixes #3352.

The Gumbel max-value sampler only transposed posterior means and variances when they were exactly rank 3, then treated `mu.shape[1]` as the sole batch dimension. With two or more posterior batch dimensions, SciPy's scalar root finder received array-valued bounds and raised `RuntimeError: Unable to parse arguments`.

This change flattens all posterior batch dimensions for the scalar quantile searches and restores the original batch shape on the sampled values. It also extends the shared sampler regression to unbatched, one-dimensional-batch, and two-dimensional-batch candidate sets. Thompson and Gumbel sampling now expose the same output-shape behavior.

Yes.

## Test Plan

- `python -m pytest test/acquisition/test_max_value_entropy_search.py -q` ? 8 passed
- `python -m ufmt check botorch/acquisition/max_value_entropy_search.py test/acquisition/test_max_value_entropy_search.py` ? 2 files already formatted
- `python -m flake8 botorch/acquisition/max_value_entropy_search.py test/acquisition/test_max_value_entropy_search.py` ? passed
- `git diff --check` ? passed
- Manual CPU comparison: a posterior with batch shape `(2, 3)` now returns finite Gumbel and Thompson samples with shape `(5, 2, 3)`; each Gumbel batch slice matches an independent call under the same seed.

The pristine `main` reproduction fails in `_sample_max_value_Gumbel` with `RuntimeError: Unable to parse arguments` while the Thompson sampler returns `(5, 2, 3)`.

## Related PRs

None. Targeted issue, pull request, and commit searches found no existing report or fix for multiple Gumbel posterior batch dimensions. The original implementation came from #298 and the later sampler refactor was #734; neither discusses this case.

## AI assistance

This pull request was prepared with OpenAI Codex assistance. I reviewed the reproduction, implementation, relevant project history, and reported test results before submission.

### DIFF
--- a/botorch/acquisition/max_value_entropy_search.py
+++ b/botorch/acquisition/max_value_entropy_search.py
@@ -996,8 +996,8 @@ def _sample_max_value_Gumbel(
     # Flatten all posterior batch dimensions for the scalar quantile searches.
     # mu, sigma are now n X num_batches, where num_batches is one for an
     # unbatched posterior.
-    mu = mu.squeeze(-1).reshape(-1, n).transpose(0, 1)
-    sigma = sigma.squeeze(-1).reshape(-1, n).transpose(0, 1)
+    mu = mu.reshape(-1, n).transpose(0, 1)
+    sigma = sigma.reshape(-1, n).transpose(0, 1)
 
     # bisect search to find the quantiles 25, 50, 75
     lo = (mu - 3 * sigma).min(dim=0).values
--- a/botorch/acquisition/max_value_entropy_search.py
+++ b/botorch/acquisition/max_value_entropy_search.py
@@ -938,8 +938,8 @@ def _sample_max_value_Thompson(
 
     Args:
         model: A fitted single-outcome model.
-        candidate_set: A ``n x d`` Tensor including ``n`` candidate points to
-            discretize the design space.
+        candidate_set: A ``batch_shape x n x d`` Tensor including ``n`` candidate
+            points per batch to discretize the design space.
         num_samples: Number of max value samples.
         posterior_transform: A PosteriorTransform. If using a multi-output model,
             a PosteriorTransform that transforms the multi-output posterior into a
@@ -947,7 +947,8 @@ def _sample_max_value_Thompson(
         maximize: If True, consider the problem a maximization problem.
 
     Returns:
-        A ``num_samples x 1`` Tensor of posterior max value samples.
+        A ``num_samples x batch_shape`` Tensor of posterior max value samples,
+        where ``batch_shape`` is ``(1,)`` for an unbatched posterior.
     """
     posterior = model.posterior(candidate_set, posterior_transform=posterior_transform)
     weight = 1.0 if maximize else -1.0
@@ -973,8 +974,8 @@ def _sample_max_value_Gumbel(
 
     Args:
         model: A fitted single-outcome model.
-        candidate_set: A ``n x d`` Tensor including ``n`` candidate points to
-            discretize the design space.
+        candidate_set: A ``batch_shape x n x d`` Tensor including ``n`` candidate
+            points per batch to discretize the design space.
         num_samples: Number of max value samples.
         posterior_transform: A PosteriorTransform. If using a multi-output model,
             a PosteriorTransform that transforms the multi-output posterior into a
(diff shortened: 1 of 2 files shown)


botorch/acquisition/max_value_entropy_search.py
```diff
@@ -973,36 +974,39 @@ def _sample_max_value_Gumbel(
...
-        mu = mu.squeeze(-1).T
-        sigma = sigma.squeeze(-1).T
-    # mu, sigma is now n X num_fantasies or n X 1
+    batch_shape = mu.shape[:-2]
+    n = mu.shape[-2]
+    # Flatten all posterior batch dimensions for the scalar quantile searches.
+    # mu, sigma are now n X num_batches, where num_batches is one for an
+    # unbatched posterior.
+    mu = mu.squeeze(-1).reshape(-1, n).transpose(0, 1)
+    sigma = sigma.squeeze(-1).reshape(-1, n).transpose(0, 1)
```
saitcakmak: Squeeze seems redundant since you're calling reshape right after

saitcakmak (commented):
Lgtm. Happy to re-review once you publish the draft

aswanth-07:
Thanks! I removed the redundant squeeze() as you pointed out and reran the tests all are passing. The PR is now ready for review : )

test/acquisition/test_max_value_entropy_search.py
```diff
@@ -398,9 +398,15 @@ def _test_max_value_sampler_base(self, sampler) -> None:
...
-            self.assertEqual(samples.shape, torch.Size([5, 3]))
+            for batch_shape in ((), (3,), (2, 3)):
+                candidate_set = torch.rand(
+                    *batch_shape, 10, 2, device=self.device, dtype=dtype
+                )
+                samples = sampler(
+                    model=model, candidate_set=candidate_set, num_samples=5
+                )
+                expected_batch_shape = batch_shape or (1,)
+                self.assertEqual(samples.shape, torch.Size([5, *expected_batch_shape]))
```
saitcakmak, line 409: Can we extend this to verify that we get correct values when using batched inputs (so that the values are not scrambled)? You can compare the batched output to the same inputs evaluated in a loop and verify that the output is as expected. This should help catch any correctness bugs in the future.

saitcakmak (changes requested):

test/acquisition/test_max_value_entropy_search.py
aswanth-07: Thanks, good point. Added a regression that compares the `(2, 3)` batched Gumbel output against the same six candidate-set slices evaluated individually under identical sampling noise. It covers both float32 and float64; the full test module passes (8 tests).

saitcakmak (approved):
Thanks!

aswanth-07:
:)

